PR #737, opened 2020-04-22
Include evolutionary-keras

RoyStegeman:
This branch contains the same commits as `n3fit-nnfit_in_n3fit`, but is instead rebased to the current point of master, while we want to preserve `n3fit-nnfit_in_n3fit` for future reference as this is the branch used to do the test with the NGA and CMA-ES.  

Once n3fit has been optimized and TF2.2 released, this branch can be merged.


scarlehoff:
I'll close this one*. I think at this point if anyone really wants to come back here they would need to either reimplement most of it

PR #983, opened 2020-10-21
Add __cause__ to produce_data_input

wilsonmr:
After #937 I thought it would be good if we also got a cause when trying to produce `data_input` so I extracted the underlying function which raises the inner error from the outer one and then call it in both `fitinputcontext` and `data_input` production rules.

This is slightly gross but I tried to add decent amount of docstring so that it's at least vaguely clear to other people

### DIFF
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -232,7 +232,7 @@ class CoreConfig(configparser.Config):
         _, theory = self.parse_from_("fit", "theory", write=False)
         thid = theory["theoryid"]
 
-        data_input = self.parse_data_input_from_(
+        data_input = self._parse_data_input_from_(
             "fit", {"theoryid": thid}
         )
         return {"theoryid": thid, "data_input": data_input}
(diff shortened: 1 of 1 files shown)


validphys2/src/validphys/config.py
```diff
@@ -232,22 +232,9 @@ def produce_fitinputcontext(self, fit):
...
-            try:
-                with self.set_context(ns=self._curr_ns.new_child({"theoryid": thid})):
-                    _, data_val = self.parse_from_("fit", data_key, write=False)
-            except ConfigError as inner_error:
-                raise e from inner_error
-        # now fill in a unique key "data_input" regardless of new or old fit
-        # for uniformity
-        data_input = self.produce_data_input(**{data_key: data_val})
+        data_input = self._parse_data_input_from_(
+            "fit", {"theoryid": thid}
```
RoyStegeman, line 236: @Zaharid  do you know why ` {"theoryid": thid}` is passed here? It seems unnecessary to me.

Zaharid: It may have had something to do with how `experiments` work(ed). But seems unneeded for getting the right `dataset_inputs`.

PR #1091, opened 2021-02-10
Parquet regression tests

siranipour:
Relies on: https://github.com/NNPDF/reportengine/pull/40

This doesn't work, but I do need help. So [it turns out](https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html#io-parquet) parquet can't save columns that are not `dtype=str` (which is problematic e.g when saving covmats ). This is not an issue, I've just made it change to `str` before writing to disk on the report engine side..

The issue comes on the vp side of testing. When it makes the comparison between dataframes, it loads the parquet file which has columns of type `Int64Index` while the loaded df has columns of type `Index`.

The assertion then fails because it compares columns. Any ideas?


Zaharid:
I think serde and afaict bincode are in the same category as pickle, so way too tied to the (current version of the) implementation and difficult to interface with external codes, so good for network exchange but not so much for storing things permanently. Ideally a person would be able to write a read pinnaple grids without the rust library and the format would contain as much information as  possible.

scarlehoff:
Since we have moved from the C++ reading (and new fktables are now pineappl) I think we should close this.

(also, I believe having the test regression in human readable / github diff-able `csv` is actually very useful)

PR #1316, opened 2021-07-02
Add fit-based targets to hyperparameter optimization

scarlehoff:
This is a continuation for #1313 and #1314

After this PR, targets for the hyperoptimizaton based on the fits coming from the folds can be used. As a first example I've added the distance between folds. I will also add the future tests as a metric _in some way_ (still not sure what the metric should be).

```yaml
kfold:
  target: fit_distance 
```

Obviously by default you have one single replica to the CV, but you can add now the right flags to the runcard:

```yaml
genrep: true
parallel_models: true
same_trvl_per_replica: true
```

And off you go:
```bash
n3fit runcards/Basic_hyperopt.yml 1 -r 5 --hyperopt 10
```

### DIFF
--- a/n3fit/src/n3fit/hyper_optimization/penalties.py
+++ b/n3fit/src/n3fit/hyper_optimization/penalties.py
@@ -59,6 +59,7 @@ def saturation(pdf_models=None, n=100, min_x=1e-6, max_x=1e-4, flavors=None, **_
         xpdf = y[0, :, flavors]
         slope = np.diff(xpdf) / np.diff(np.log10(x))
         pen = abs(np.mean(slope, axis=1)) + np.std(slope, axis=1)
+        # Add a small offset to avoid ZeroDivisionError
         extra_loss += np.sum(1.0 / (1e-7 + pen))
     return extra_loss
 


scarlehoff:
This is now ready for review (and most importantly, it doesn't crash :P)

n3fit/src/n3fit/model_trainer.py
```diff
@@ -533,8 +532,8 @@ def _generate_observables(
             force_set_smallest = input_arr.min() > 1e-9
             if force_set_smallest:
                 new_xgrid = np.linspace(
-                        start=1/input_arr_size, stop=1.0, endpoint=False, num=input_arr_size
-                        )
+                    start=1 / input_arr_size, stop=1.0, endpoint=False, num=input_arr_size
```
RoyStegeman, line 535:
```suggestion
                    start=1/input_arr_size, stop=1.0, endpoint=False, num=input_arr_size
```
just seems a bit odd

n3fit/src/n3fit/model_trainer.py
```diff
@@ -690,8 +691,8 @@ def _train_and_fit(self, training_model, stopping_object, epochs=100):
         return self.failed_status
 
     def _hyperopt_override(self, params):
-        """ Unrolls complicated hyperopt structures into very simple dictionaries"""
-        # If the input contains all parameters, then that's your dictionary of hyperparameters
+        """Unrolls complicated hyperopt structures into very simple dictionaries"""
+        # If the input contains all parameters, then that's your dicttionary of hyperparameters
```
RoyStegeman, line 695:
```suggestion
        # If the input contains all parameters, then that's your dictionary of hyperparameters
```

n3fit/src/n3fit/model_trainer.py
```diff
@@ -795,14 +796,17 @@ def hyperparametrizable(self, params):
...
+        exp_models = []
 
         ### Training loop
         for k, partition in enumerate(self.kpartitions):
             # Each partition of the kfolding needs to have its own separate model
             # and the seed needs to be updated accordingly
             seeds = self._nn_seeds
             if k > 0:
-                seeds = [np.random.randint(0, pow(2, 31))]
+                seeds = [np.random.randint(0, pow(2, 31)) for _ in seeds]
```
RoyStegeman, line 809: Maybe this is completely irrelevant for hyperopt, but you can for example get different seeds for replica 2 depending on wheter you ran `n3fit 2` or `n3fit 1 -r 2`.

validphys2/src/validphys/hyperoptplot.py
```diff
@@ -76,8 +91,12 @@ def params(self):
         apl = hyperparameters.get("activation_per_layer")
         if apl is not None and isinstance(apl, str) and "nodes_per_layer" in hyperparameters:
             # If apl is a string, it can only bring information if there is `nodes_per_layer`
-            apl = [apl]*(len(hyperparameters["nodes_per_layer"])-1) + ["linear"]
+            apl = [apl] * (len(hyperparameters["nodes_per_layer"]) - 1) + ["linear"]
             hyperparameters["activation_per_layer"] = apl
+        # 2. If there _was_ originally a reward included in this parameter (because it comes from and
+        #   previous hyperoptimization) mark it as such
```
RoyStegeman, line 97:
```suggestion
        # 2. If there _was_ originally a reward included in this parameter (because it comes from a
        #   previous hyperoptimization) mark it as such
```

validphys2/src/validphys/hyperoptplot.py
```diff
@@ -76,8 +91,12 @@ def params(self):
...
+        #         if (reward:=hyperparameters.pop("reward", None)) is not None:
+        #             hyperparameters["old_reward"] = reward
```
RoyStegeman, line 99: Might as well skip running this hyperparameter setup in that case, though I imagine this happens so rarely that it's really worth the effort to not do the simplest thing and just overwrite it. You might want to remove this comment though.

n3fit/src/n3fit/model_trainer.py
scarlehoff: When doing hyperopt these are two different things so it's fine (the first command will populate only `replica_2` while the second populates only `replica_1`)

n3fit/src/n3fit/hyper_optimization/rewards.py
```diff
@@ -14,26 +14,125 @@
...
+        # Get the full input and the total chi2
+        full_input = exp_model.input
+        exp_chi2 = exp_model.compute_losses()["loss"][0]
+
+        # Now update the loss with the PDF covmat
+        for layer in exp_model.get_layer_re(".*_exp$"):
+            # Get the input to the loss layer
+            model_output = MetaModel(full_input, layer.input)
+            # Get the full predictions and generate the PDF covmat
+            y = model_output.predict()[0]  # The first is a dummy-dim
```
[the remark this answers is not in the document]
scarlehoff: This is an "old-TF" bug, since (from the discussion in #1405) we are keeping compatibility until the paper is published I will add a workaround so it can be removed once we drop 2.3

(although now I'll have to find out up to which version exactly the workaround is needed...)

doc/sphinx/source/n3fit/hyperopt.rst
```diff
@@ -315,3 +315,57 @@ The group of datasets that were left out followed the algorithm :ref:`mentioned
...
+
+   n3fit runcard.yml 1 -r 10 --hyperopt 20
+
+The above runcard will, for a sample of 20 trials in ``210508-hyperopt_for_paper`` (according to their rewards),
+run two fits of 10 replicas each.
+The first fit will hide the data from HERA and the second one (an empty fold) will take into consideration all data.
+The figure of metric will be the difference between the :math:`\chi2` of the second fit to the folded data and the :math:`\chi2` of the first fit to the folded data *including* pdf errors.
+
+.. math::
+   L_{\rm hyperopt} = \chi^{2}_{(1) \rm pdferr} - \chi^{2}_{(2)}
```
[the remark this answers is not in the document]
scarlehoff: The future test is referred to at the beginning of the page, but I'll link it here as well.

n3fit/src/n3fit/hyper_optimization/rewards.py
```diff
@@ -14,26 +14,125 @@
...
+
+    # Loop over all models but the last (our reference!)
+    total_loss = 0.0
+    for n3pdf, exp_model in zip(n3pdfs[:-1], experimental_models[:-1]):
+
+        _set_central_value(n3pdf, exp_model)
+
+        # Get the full input and the total chi2
+        full_input = exp_model.input
+        exp_chi2 = exp_model.compute_losses()["loss"][0]
```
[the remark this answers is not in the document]
scarlehoff: Sorry, good catch. Both this and the one above were indeed for debugging and forgot to remove them!

Zaharid:
@siranipour if you could have a look at this, once you are back, that would be great.

n3fit/src/n3fit/hyper_optimization/rewards.py
```diff
@@ -14,26 +17,137 @@
...
+        raise ValueError("fit_future_test needs n3pdf models to act upon")
+    if experimental_models is None:
+        raise ValueError("fit_future_test needs experimental_models to compute chi2")
+
+    from n3fit.backends import MetaModel
+
+    compatibility_mode = False
+    try:
+        import tensorflow as tf
+        from n3fit.backends import set_eager
```
RoyStegeman, line 103: pylint complains about importing outside top-level. Also in another function where you import backeds.operations.

n3fit/src/n3fit/hyper_optimization/penalties.py
```diff
@@ -44,24 +45,25 @@ def saturation(pdf_model, stopping_object, n=100, min_x=1e-6, max_x=1e-4, flavor
...
+        y = pdf_model.predict([xin])
+        xpdf = y[0, :, flavors]
+        slope = np.diff(xpdf) / np.diff(np.log10(x))
+        pen = abs(np.mean(slope, axis=1)) + np.std(slope, axis=1)
+        extra_loss += np.sum(1.0 / (1e-7 + pen))
     return extra_loss
 
 
-def patience(pdf_model, stopping_object, alpha=1e-4):
+def patience(pdf_models, stopping_object, alpha=1e-4):
```
RoyStegeman, line 66: Could you do the same here as you did in rewards.py? Namely, using kwargs to stop pylint complaining about unused  arguments.

RoyStegeman (commented):
Looks okay to me, though I guess we should still give @siranipour a chance to have a look before merging.

n3fit/src/n3fit/hyper_optimization/rewards.py
RoyStegeman: I understand why, but currently pylints use is limited due to the huge number of warnings, so if we want to improve this we should prevent adding more of them. To stop pylint from complaining you don't need to move this to top-level, instead you can add `# pylint: disable=import-outside-toplevel`, maybe even in the `.pylintrc` if you prefer.

scarlehoff: In this file pylint only gives 5  warnings. I don't really think that limits its use.

scarlehoff:
Merged as per Code Meeting

PR #1324, opened 2021-07-13
Add metadata groups to closure tables

wilsonmr:
Without spending more time on this, this was the best I could do. There seems to be a fair amount of redundancy but I think there isn't a better way to do it without getting rid of backwards compat.

Example report: https://vp.nnpdf.science/6oHHKWGkSw-d-LxFX59Fig==

the groups are much smaller so there are bigger fluctuations. The worst one is probably jets which actually has a fair amount of data. Anyway people asked for it, so here it is.

### DIFF
--- a/validphys2/src/validphys/closuretest/multiclosure.py
+++ b/validphys2/src/validphys/closuretest/multiclosure.py
@@ -650,7 +650,25 @@ experiments_bootstrap_bias_variance = collect(
 )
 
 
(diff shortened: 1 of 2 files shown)


RoyStegeman:
@tgiani  can this be merged?

tgiani:
I still have to review this, will do after I m done with the ct paper

tgiani (approved):

PR #1335, opened 2021-07-22
Build with Python 3.9

Zaharid:


scarlehoff:
I think we should close this and open a new one with the github action (since travis is no more)

PR #1339, opened 2021-07-26
Run eagerly in debug mode

siranipour:
Thought it may be useful to have this flag so you can have `embed`s and so on inside the model when in debug mode 

note: I've not tested this because tf is a nightmare on M1 and the linux machines I have access to are down at the minute

### DIFF
--- a/n3fit/src/n3fit/backends/keras_backend/MetaModel.py
+++ b/n3fit/src/n3fit/backends/keras_backend/MetaModel.py
@@ -245,6 +245,7 @@ class MetaModel(Model):
         loss=None,
         target_output=None,
         clipnorm=None,
+        run_eagerly=None,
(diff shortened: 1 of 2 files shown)


scarlehoff:
Hi @siranipour, do you want to have this or should we close it?

I was thinking maybe an option (since this is a developers only option) is to add some docs on how to do it? I'm not sure we want to have this option in the runcard at all (unless it also starts the shell automatically) because otherwise an clueless user could start `n3fit` in this mode and just be very confused about why is it so slow (and they won't ever be able to use it because it's only useful for debugging).

(doing some bookkeeping)

siranipour:
Yeah let's close it, after discussing this with you I realised it was more involved than I had previously thought

Issue #1337, opened 2021-07-23
nf=4 PDFs


Zaharid:
We should make sure that we can evolve PDFs with nf=4 in a way that is correct and automtic.

PR #1346, opened 2021-08-03
adding --theory_id flag to evolven3fit

scarrazza:
Provides an extra optional flag to `evolven3fit` to select a custom evolution theory.
Should help in #1337 but may require some extra management of the output files.


RoyStegeman:
I prepared the NLO and NNLO pdfs in nf=4: 
- NLO: https://vp.nnpdf.science/88VZ9ZY7TtCJl9YsGJl6oQ==/
- NNLO: https://vp.nnpdf.science/DuJBij4-SHOgJsEs-gSwXQ==/
- NNLO NNPDF3.1: https://vp.nnpdf.science/1ZMzOBGbQAuSZkESO8nJ1g==/
 
Where as a sensibility check, it is clear by looking at the lumi plots the what we do for NNPDF4.0 is the same as what was done for NNPDF3.1. 

I suppose that technically I could complete the list of PDFs in chapter 10 at this point. However, the theories used in the evolution are from NNPDF3.1 (theory ID's are 120 and 121). Do these need to be updated?

scarrazza:
@RoyStegeman thank you so much.
@enocera shall we create new theories for these nf variations?

enocera:
@RoyStegeman @scarrazza I think that I've updated `theory.db` with the theories required for the nf variation fits that we list in the paper. The theories relevant to each PDF set are also summarised here https://www.wiki.ed.ac.uk/display/nnpdfwiki/NNPDF4.0+final+fits. Thanks.

RoyStegeman:
That's very helpful. Thanks!

RoyStegeman:
I produced those PDFs by just doing `evolven3fit` on the original fit, meaning the 1000 replica fit in case of the baseline, and then postfit afterwards. As a result, the replicas in the nf versions are not necessarily exactly the same as those in the 'regular' version of the PDF. I guess this isn't much of a problem, but this is what it says in the paper: 
Screenshot_20210915_200322

While e.g. by looking at the NNLO PDFs at Q=1.65GeV it is clear that they are not perfectly identical, is that acceptable?

Zaharid:
@RoyStegeman  note that the metadata is the trickier bit here (and why we wanted a script ideally...). What changes have been made compared to the default 4.0?

enocera:
> I produced those PDFs by just doing `evolven3fit` on the original fit, meaning the 1000 replica fit in case of the baseline, and then postfit afterwards. 

Question: shouldn't we start from the compressed set of 100 replicas and re-evolve this one instead of the fit with 1000 replicas?

RoyStegeman:
> Question: shouldn't we start from the compressed set of 100 replicas and re-evolve this one instead of the fit with 1000 replicas?

To be honest, I am not familiar enough with `eveolven3fit` to answer that. Essentially `evolven3fit` takes as input `.exportgrid` files and outputs `.dat` files in LHAPDF format, and I don't know if there's an easy way to do take a PDF as input instead of a fit?

RoyStegeman:
> @RoyStegeman note that the metadata is the trickier bit here (and why we wanted a script ideally...). What changes have been made compared to the default 4.0?

Re. metadata, the `.info` file is generated by `evolven3fit` with parameters depending on the theory settings (such as the `NumFlavors` you mentioned earlier). Is there any metadata that is not automatically taken care of that should be done by a script instead?

enocera:
> > Question: shouldn't we start from the compressed set of 100 replicas and re-evolve this one instead of the fit with 1000 replicas?
> 
> To be honest, I am not familiar enough with `eveolven3fit` to answer that. Essentially `evolven3fit` takes as input `.exportgrid` files and outputs `.dat` files in LHAPDF format, and I don't know if there's an easy way to do take a PDF as input instead of a fit?

Ah. This answers my question. If the input is `.exportgrid`, I'd say that the compressed set with 100 replicas cannot be used because `.exportgrid` does not exist for that set.

Zaharid:
> > @RoyStegeman note that the metadata is the trickier bit here (and why we wanted a script ideally...). What changes have been made compared to the default 4.0?
> 
> Re. metadata, the `.info` file is generated by `evolven3fit` with parameters depending on the theory settings (such as the `NumFlavors` you mentioned earlier). Is there any metadata that is not automatically taken care of that should be done by a script instead?

@RoyStegeman @scarrazza see https://github.com/NNPDF/nnpdf/pull/1346#issuecomment-892114938

RoyStegeman:
@Zaharid you're linking:
> These keys should be completed correctly if we select an appropriate theory id. The AlphaS_MZ is the most tricky and arbitrary object to deal with, for the previous sets we do some backevolution with full flavours and then up evolution with nf flavours.

The way I understand that message is that almost all metadata is filled in correctly by evolven3fit, except for AlphaS_MZ. AlphaS_MZ is usually taken from the theory, but perhaps now we need to do "some backevolution" followed by "up evolution" instead? Essentially, I just don't know exactly what steps to take in order to determine the correct AlphaS_MZ.

scarrazza:
@RoyStegeman could you please share the link with the info file? In principle we don't have to touch anything, but let me check explicitly.

RoyStegeman:
NNPDF40_nlo_as_01180_nf_4.info
```
SetDesc: "NNPDF x.x"
SetIndex: 
Authors: NNPDF Collaboration.
Reference: arXiv:xxxx.xxxxxx
Format: lhagrid1
DataVersion: 1
NumMembers: 101
Particle: 2212
Flavors: [-4, -3, -2, -1, 21, 1, 2, 3, 4, ]
OrderQCD: 1
FlavorScheme: variable
NumFlavors: 4
ErrorType: replicas
XMin: 1.0000000e-09
XMax: 1.0000000e+00
QMin: 1.6500000e+00
QMax: 1.0000000e+05
MZ: 9.1187600e+01
MUp: 0
MDown: 0
MStrange: 0
MCharm: 1.5100000e+00
MBottom: 4.9200000e+00
MTop: 1.7250000e+02
AlphaS_MZ: 0.1126178
AlphaS_OrderQCD: 1
AlphaS_Type: ipol
AlphaS_Qs: [1.6500000e+00, 1.7793910e+00, 1.9247332e+00, 2.0884973e+00, 2.2736105e+00, 2.4835548e+00, 2.7224883e+00, 2.9953960e+00, 3.3082803e+00, 3.6683990e+00, 4.0845679e+00, 4.5675444e+00, 5.1305182e+00, 5.7897416e+00, 6.5653429e+00, 7.4823833e+00, 8.5722373e+00, 9.8744064e+00, 1.1438917e+01, 1.3329511e+01, 1.5627919e+01, 1.8439624e+01, 2.1901683e+01, 2.6193436e+01, 3.1551250e+01, 3.8289016e+01, 4.6826827e+01, 5.7731478e+01, 7.1774136e+01, 9.0013186e+01, 1.1391437e+02, 1.4552663e+02, 1.8774214e+02, 2.4468456e+02, 3.2229530e+02, 4.2922805e+02, 5.7822977e+02, 7.8829758e+02, 1.0880891e+03, 1.5213834e+03, 2.1559410e+03, 3.0980757e+03, 4.5169626e+03, 6.6857830e+03, 1.0052439e+04, 1.5363022e+04, 2.3880935e+04, 3.7782467e+04, 6.0883588e+04, 1.0000000e+05]
AlphaS_Vals: [3.2443514e-01, 3.1293255e-01, 3.0182345e-01, 2.9109647e-01, 2.8074028e-01, 2.7074365e-01, 2.6109544e-01, 2.5178466e-01, 2.4280049e-01, 2.3413228e-01, 2.2576960e-01, 2.1770223e-01, 2.0992020e-01, 2.0241375e-01, 1.9517341e-01, 1.8818995e-01, 1.8145439e-01, 1.7495804e-01, 1.6869247e-01, 1.6264950e-01, 1.5682123e-01, 1.5120001e-01, 1.4577845e-01, 1.4054942e-01, 1.3550604e-01, 1.3064166e-01, 1.2594988e-01, 1.2142452e-01, 1.1705963e-01, 1.1284947e-01, 1.0878852e-01, 1.0487147e-01, 1.0109317e-01, 9.7448708e-02, 9.3933320e-02, 9.0542431e-02, 8.7271634e-02, 8.4116683e-02, 8.1073490e-02, 7.8138115e-02, 7.5306766e-02, 7.2575786e-02, 6.9941653e-02, 6.7400973e-02, 6.4950475e-02, 6.2587003e-02, 6.0307519e-02, 5.8109088e-02, 5.5988885e-02, 5.3944181e-02]
AlphaS_Lambda4: 0.342207
AlphaS_Lambda5: 0.239
```

scarrazza:
Theory 120, right?

RoyStegeman:
Technically 220, but they're equivalent: NLO fitted charm nf=4

scarrazza:
Right, so we have to override the fields `AlphaS_MZ` and `MZ` with the values of `alphas` and `Qref` from the theory.db

RoyStegeman:
Is this something we'll always want to do if evolving with a theory that's different from the one used during fitting (or maybe even if they are the same theory)?

scarrazza:
@RoyStegeman not really, in fact I find this choice of reusing `MZ` as `MB` and `AlphaS_MZ` as `AlphaS_MB` (from users of nf sets) rather misleading. Now, if we want to automatize this behaviour for nf sets, we can touch the evolvegrid.cc and update both lines involving these variables and use the theory.db values instead of asking APFEL the reference values.

RoyStegeman:
Yes that's what I just changed: taking `MZ` from `Qref` in the db and `Alphas_MZ` from `alphas` in the db instead of APFEL. I was just wondering if this was behaviour we'd always want under certain conditions (e.g. when manually entering theory_id) so that it should be added to this PR, or if I should jut keep it local. But I guess it's the second?

scarrazza:
Yes, the second, in fact the nf variation sets assume that some metadata fields have another meaning.

RoyStegeman:
@enocera  Can you also add nf=4 theories for as=0.117 and as=0.119 for both NLO and NNLO? I think this is needed for the `_pdfas` PDFs.

enocera:
@RoyStegeman OK, but let me determine the value of alphas in these cases - we don't have these theories from NNPDF3.1.

enocera:
@RoyStegeman I've added the theories as you requested:
- 224 -> corresponds to what we call nf=4, alphas=0.117, NLO;
- 225 -> as 224 but at NNLO;
- 226 -> corresponds to what we call nf=4, alphas=0.119, NLO;
- 227 -> as 226 but at NNLO.

RoyStegeman:
That's great. Thanks!

RoyStegeman:
@Zaharid I added a python script that does the same thing that my adjustment local adjustment to `evolven3fit` of this branch did, and that I used to produce the nf variations fit. 

I still don't really see the benefit of a separate python script, but oh well :man_shrugging:. Is this anything like what you had in mind? The --theory_id flag takes care of everything else for the nf variations.

Zaharid:
Without having looked at it in detail, yes it is. I would offer the fact that this PR was opened on August 3, and was the last result for the paper as evidence that something like it was in fact needed :).

RoyStegeman:
Good, I'll still add some docs on the nf variations before this can be closed.

Well personally I think I would've preferred a simple flag in evolven3fit, but this works as well.

PR #1352, opened 2021-08-11
Collect over groups to arrange groups_chi2 table

### LINKED ISSUE #1351: `custom_group` and chi2 table doesn't arrange correctly
The following runcard 
```yaml
metadata_group: custom_group

fit: NNPDF31_nnlo_as_0118_DISonly

pdf:
  from_: fit

use_cuts: internal

dataset_inputs:
  - {dataset: NMCPD, frac: 0.5, custom_group: foo}
  - {dataset: NMC, frac: 0.5, custom_group: bar}
  - {dataset: SLACP, frac: 0.5, custom_group: foo}
  - {dataset: SLACD, frac: 0.5, custom_group: foo}

theoryid: 53

template_text: |
  {@groups_chi2_table@}

actions_:
  - report(main=True)
```
doesn't work as expected: output is here
https://vp.nnpdf.science/vNBX1kDXSgGoKgKLZhZInw==

Note that, for example, the `npoints` within each custom_group don't add up. It is possible to replicate by having a `dataset_input` with one custom group sandwiched between other `dataset_input`s having a different grouping

siranipour:
closes #1351

The [weird](https://stackoverflow.com/a/716489/7708866) sum(..., []) is to concatenate a list of lists which is as a result of a collect.

### DIFF
--- a/validphys2/src/validphys/results.py
+++ b/validphys2/src/validphys/results.py
@@ -815,13 +815,12 @@ groups_each_dataset_chi2 = collect("each_dataset_chi2", ("group_dataset_inputs_b
 def groups_chi2_table(groups_data, pdf, groups_chi2, groups_each_dataset_chi2):
     """Return a table with the chi² to the groups and each dataset in
     the groups."""
-    dschi2 = iter(sum(groups_each_dataset_chi2, []))
     records = []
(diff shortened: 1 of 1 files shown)


siranipour:
The above advice has now been implemented

Zaharid (approved):

PR #1366, opened 2021-08-24
Paperchi2plot

enocera:
This PR amends the PLOTTING files to allow one to group datasets as requested for the NNPDF4.0 paper. Maybe a better name for the grouping may be used.


siranipour:
Closing now because #1398 supersedes this

PR #1395, opened 2021-09-01
Add compression tools to docs

scarlehoff:
For completeness.

### DIFF
--- a/doc/sphinx/source/external-code/pdf-codes.md
+++ b/doc/sphinx/source/external-code/pdf-codes.md
@@ -26,8 +26,11 @@ general information about PDF evolution and the DGLAP equations, you can go to t
 section](dglap.md).
 
 ## PDF compression
-In order to maintain the faithfulness of released NNPDF grids, while keeping a reasonable number of replicas,
-a PDF set with 1000 replicas is generated and then compressed to a set of 100 replicas.
+PDF compression seeks to maintain the statistical accuracy of a large sample of replicas
+produced by a fit when using a PDF set with a smaller number of replicas (and thus fewer 
+convolutions required to compute cross sections with PDF uncertainties). For example the 
+main published PDFs are typically based on a 1000 replica fit, which can then be compressed to 
+around a 100 replicas PDF set while maintaining good accuracy of most relevant statistical estimators.
 This is done with the [pyCompressor](https://n3pdf.github.io/pycompressor/) library,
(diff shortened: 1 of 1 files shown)


RoyStegeman:
> This is fine by me! The only thing that I am not sure about is whether it is worth having this in a separate section, say Reduced Sets, and also include some mention about the Hessian.

I wouldn't, `pycompressor` is exteral code and has its own docs, so making a special "reduced sets" section seems excessive to me.  If you want, I suppose it might be nice to add a tutorial on "how to compress mc pdf sets", similar to this tutorial for mc2hessian: https://docs.nnpdf.science/tutorials/mc2hessian.html

Radonirinaunimi:
> I wouldn't, `pycompressor` is exteral code and has its own docs, so making a special "reduced sets" section seems excessive to me. If you want, I suppose it might be nice to add a tutorial on "how to compress mc pdf sets".
> 
> Btw, since mc2hessian is no longer external, something like that already exists for mc2hessian: https://docs.nnpdf.science/tutorials/mc2hessian.html

That makes sense! I suggested that while forgetting that there has already been a documentation on the vp-mc2hessian. I wouldn't add a tutorial on the NNPDF doc though. This will just repeat the same information/tutorials that are already present in the N3PDF documentation.

scarlehoff:
Rebased in order to have the test before merging.

doc/sphinx/source/external-code/pdf-codes.md
```diff
@@ -25,6 +25,15 @@ written to replace APFEL, which is currently dubbed EKO ('Evolution Kernel Opera
 general information about PDF evolution and the DGLAP equations, you can go to the [Theory
 section](dglap.md).
 
+## PDF compression
+In order to maintain the faithfulness of released NNPDF grids, while keeping a reasonable number of replicas,
+a PDF set with 1000 replicas is generated and then compressed to a set of 100 replicas.
```
Zaharid, line 30: The wording here could be more general

```suggestion
PDF compression seeks to maintain the statistical accuracy of a large sample of replicas
produced by a fit when using a PDF set with a smaller number of replicas (and thus fewer 
convolutions required to compute cross sections with PDF uncertainties). For example the 
main published PDFs are typically based on a 1000 replica fit, which can then be compressed to 
around a 100 replicas while maintaining good accuracy of most relevant statistical estimators.
```

PR #1401, opened 2021-09-03
Temporarily remove nf=4 comment

Zaharid:
This page is now public and linked from the main readme prominently. The dots are not a good look at the moment.


Zaharid:
Incidentally we do need to do something for #133... (here the ellipsis is justified).

RoyStegeman:
> Incidentally we do need to do something for #133... (here the ellipsis is justified).

Do you mean the HERA data issue and nf=4? I'm not really involved with either of those. Looking at the theory database there exist theories with nf=3,4,6 so maybe that's what was done for NNPDF3.1 as opposed to evolution?

RoyStegeman (approved):

Zaharid:
I mean the script or whatever that reevolves the grids and fills in the metadata.

PR #1411, opened 2021-09-23
Add modified_tanh activation function

RoyStegeman:
Does as the title suggests.

### DIFF
--- a/n3fit/src/n3fit/backends/keras_backend/base_layers.py
+++ b/n3fit/src/n3fit/backends/keras_backend/base_layers.py
@@ -22,7 +22,7 @@ from tensorflow.keras.layers import concatenate, Input # pylint: disable=unused-
 from tensorflow.keras.layers import Dense as KerasDense
 from tensorflow import expand_dims
 from tensorflow.keras.regularizers import l1_l2
-from tensorflow import nn
+from tensorflow import nn, math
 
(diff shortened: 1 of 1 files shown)


n3fit/src/n3fit/backends/keras_backend/base_layers.py
```diff
@@ -30,14 +31,18 @@ def square_activation(x):
     """ Squares the input """
     return x*x
 
+def modified_tanh(x):
+    return (nn.relu(x)+nn.relu(-x))*nn.tanh(x)
```
scarlehoff, line 35: Please add some docstring.
Also, why `relu(x) + relu(-x)` instead of `abs(x)`? Wouldn't it be the same?

RoyStegeman: I think it had something to do with how the gradient was defined in x=0, but let me see if that is correct.

RoyStegeman: I thought I saw somewhere that the gradient of abs(x) was not well defined at x=0 while that of relu just goes to 0. Not sure why I thought that: I just explicitly tested and the gradient of abs(x) at x=0 is also 0.

scarlehoff (approved):

PR #1415, opened 2021-09-23
Fix commodata loader bug

### LINKED ISSUE #1409: Error commondata when running a closure test
While trying to run a closure test I got an error related to the line in commondata parser below. It removes the DATA_ prefix from commondata filenames, but instead, while running the closure test a commondata filename starting with FILTER_ was passed, so the prefix was not fully removed and as a result the name of the dataset is not recognised. 

https://github.com/NNPDF/nnpdf/blob/f020b0bf2926fc85196441825896c8f955f1e69b/validphys2/src/validphys/commondataparser.py#L23 

I also noticed these lines below. Does this mean DATA_ is the old format, and if I get a name with FILTER_ in the data parser above, the FILTER_ prefix should simply be removed or is there a more important difference between the two? 

https://github.com/NNPDF/nnpdf/blob/f020b0bf2926fc85196441825896c8f955f1e69b/validphys2/src/validphys/loader.py#L238-L243

Maybe @siranipour knows more about this?

RoyStegeman:
Fixes #1409.

### DIFF
--- a/validphys2/src/validphys/commondataparser.py
+++ b/validphys2/src/validphys/commondataparser.py
@@ -19,8 +19,7 @@ def load_commondata(spec):
     Returns an instance of CommonData
     """
     commondatafile = spec.datafile
-    # Getting set name from commondata file name
-    setname = commondatafile.name[5:-4] # DATA prefix and .dat suffix
+    setname = spec.name
     systypefile = spec.sysfile
 
     commondata = parse_commondata(commondatafile, systypefile, setname)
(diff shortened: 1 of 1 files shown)


siranipour (approved):
Looks good, let's merge once the tests pass

RoyStegeman:
I'm still doubting a bit between doing this and adding another option (if-statement) to remove the FILTER_ prefix when necessary. Not sure there is a reason why spec.name could be different from the name in DATA_[name].dat, but that they are the same is explicitly being checked in the `parse_commondata` function.

Issue #1414, opened 2021-09-23
Fitenvironment is broken for nnpdfcpp fits


Zaharid:
https://github.com/NNPDF/nnpdf/pull/1328 added code like https://github.com/NNPDF/nnpdf/blame/f2ad3ea60cec9c93b958b5ec77387d82142e8575/validphys2/src/validphys/config.py#L274 which is not valid for nnpdfcpp fits since they don't have the corresponding keys. While we have #1405 for now it would be good to be on top of what works and what doesn't (but not sure about what to do here).

Zaharid:
Closed in #1418.

Issue #1413, opened 2021-09-23
Remove old fit format compatibility


Zaharid:
We have had a json format for a while now and in particular for everything in 4.0. We should only use that and remove legacy formats both in terms of producing them and reading them in vp.

This is part of #1405 and should be coordinated with it.

scarlehoff:
https://github.com/NNPDF/nnpdf/releases/tag/4.0.10

PR #1418, opened 2021-09-24
Collecting over cuts context for pseudodata

siranipour:
The reason the bug in https://github.com/NNPDF/nnpdf/issues/1414 happened was because we were collecting over an environment that was over specified. This does beg the question whether we still want to fix the point raised in #1414 in light of #1413

### DIFF
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -253,6 +253,18 @@ class CoreConfig(configparser.Config):
         replicas = fitted_replica_indexes(pdf)
         return NSList(replicas, nskey='replica')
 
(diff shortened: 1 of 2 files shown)


Zaharid:
I think we want these fixes until we say we don't want them any longer :)

Zaharid:
I don't think this is working though. It seems to be getting stuck in a loop.

Zaharid (approved):

Zaharid (changes requested):
Hold on

siranipour:
Old style fits are now supported too

Zaharid (approved):

PR #1422, opened 2021-09-30
Move varflavours to a more inconspicuous location

Zaharid:
This seems a bit too specialized to be sitting at top level.


Zaharid:
@RoyStegeman  FWIW  was thinking that the script should be doing all the calling of evolvefit and PDF renaming. I agree it is a bit bare bones as it is. But I guess it is too late for that.

RoyStegeman:
I agree, I just put it here since I didn't really think it fitted anywhere else either but I was surprised that SC didn't ask me to move it elsewhere.

Just leave this PR open, if it doesn't cause some unforeseen difficulties I can add a call to `vp-fitrename` and  `evolven3fit` to the script as well.

RoyStegeman:
@Zaharid I updated the script a bit and I think it's ready to be merged.

n3fit/evolven3fit/varflavors.py
```diff
@@ -0,0 +1,99 @@
...
+            shutil.copytree(path_input_fit, copied_fit, symlinks=True)
+            newpath = change_name(copied_fit, args.new_name)
+            newpath.rename(path_output_fit)
+        log.info("Renaming with copy completed")
+
+    # 2. Run evolven3fit
+    evolven3fit_command = (
+        f"evolven3fit {path_output_fit} {args.max_replicas} --theory_id {args.theory_id} "
+    )
+    subprocess.call(evolven3fit_command, shell=True)
```
Zaharid, line 77: It would be somewhat better to use `subprocess.run` here, particularly because then you can do `check=True`, which is a good idea.

Zaharid (commented):
While we are at it, we could as well place this under `n3fit/scripts` and arrange `setup.py` so it gets installed.

RoyStegeman:
Thanks, that indeed makes more sense

n3fit/src/n3fit/scripts/varflavors.py
```diff
@@ -0,0 +1,99 @@
...
+    # the theory db, repsectively
+    theory_dict = fetch_theory(path_db, args.theory_id)
+    Qref = theory_dict["Qref"]
+    alphas = theory_dict["alphas"]
+
+    path_temp_info = path_info_file.parent / "temp.info"
+    with open(path_info_file, "r+") as f:
+        with open(path_temp_info, "w") as tempfile:
+            for line in f:
+                if line.startswith("AlphaS_MZ:"):
```
Zaharid, line 90: I think we should actually parse the header with YAML here. See examples in e.g. the rename scripts. This `startswith` thing is a bit of an anachronism from when we didn't have a roundtrip YAML parser.

n3fit/src/n3fit/scripts/varflavors.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+    args = parser.parse_args()
+
+    path_input_fit = Path(args.fit_folder)
+    input_fit_name = path_input_fit.name
+    path_info_file = path_input_fit / f"nnfit/{input_fit_name}.info"
+
+    # 1. If the new_name flag is used, create a copy of the input fit
+    if args.new_name:
+        import tempfile
```
Zaharid, line 60: @siranipour is there a shortcut for all this?

n3fit/src/n3fit/scripts/varflavors.py
RoyStegeman: I tried, but the roundtrip does change some of the formatting. Maybe this can be fixed by manually entering settings for the roundtrip parser, but I don't trust it as much as simply doing as it is now.

Zaharid: well the converse problem is that one could potentially have 

```
In [9]: yaml.safe_load(r"""
   ...: "AlphaS_MZ": 
   ...:     0.118
   ...: """)
Out[9]: {'AlphaS_MZ': 0.118}
```
(note the quotes and the line break).

RoyStegeman: Not really, the .info file is really written line-by-line during evolution, it's not a dictionary dumped to a yaml file. So the .info file will never be formatted in any other way.

Zaharid:
@siranipour please have a look at this and approve eventually.

PR #1428, opened 2021-10-08
Swap the libnnpdf profile for a python dictionary holding all relevant information

scarlehoff:
This is a very small change, I've simply moved the `get_\w+_path` methods to a python class. This simplifies the code in various places and removes yet another dependency in libNNPDF. I tried to make it more or less similar to how it used to work.

This is quite small and self-contained. The choice of having the "profile-class" inside `loader.py` was due to being the only file that really uses it (`uploadutil` being the other one) but I'm open to any other options, I'm don't have any particular opinions about this, just wanted to removed the libNNPDF import.

### DIFF
--- a/validphys2/src/validphys/app.py
+++ b/validphys2/src/validphys/app.py
@@ -23,47 +23,47 @@ from validphys import mplstyles
 
 
 providers = [
-        'validphys.results',
-        'validphys.commondata',
-        'validphys.pdfgrids',
-        'validphys.pdfplots',
-        'validphys.dataplots',
-        'validphys.fitdata',
-        'validphys.arclength',
-        'validphys.sumrules',
-        'validphys.reweighting',
-        'validphys.kinematics',
-        'validphys.correlations',
-        'validphys.chi2grids',
-        'validphys.eff_exponents',
-        'validphys.paramfits.dataops',
-        'validphys.paramfits.plots',
-        'validphys.theorycovariance.construction',
-        'validphys.theorycovariance.output',
-        'validphys.theorycovariance.tests',
-        'validphys.replica_selector',
-        'validphys.closuretest',
-        'validphys.mc_gen',
-        'validphys.theoryinfo',
-        'validphys.pseudodata',
-        'validphys.renametools',
-        'validphys.covmats',
-        'validphys.hyperoptplot',
-        'validphys.deltachi2',
-        'validphys.n3fit_data',
-        'validphys.mc2hessian',
-        'reportengine.report',
-        ]
+    "validphys.results",
+    "validphys.commondata",
+    "validphys.pdfgrids",
+    "validphys.pdfplots",
+    "validphys.dataplots",
+    "validphys.fitdata",
+    "validphys.arclength",
+    "validphys.sumrules",
+    "validphys.reweighting",
+    "validphys.kinematics",
+    "validphys.correlations",
+    "validphys.chi2grids",
+    "validphys.eff_exponents",
+    "validphys.paramfits.dataops",
+    "validphys.paramfits.plots",
+    "validphys.theorycovariance.construction",
+    "validphys.theorycovariance.output",
+    "validphys.theorycovariance.tests",
+    "validphys.replica_selector",
+    "validphys.closuretest",
+    "validphys.mc_gen",
+    "validphys.theoryinfo",
+    "validphys.pseudodata",
+    "validphys.renametools",
+    "validphys.covmats",
+    "validphys.hyperoptplot",
+    "validphys.deltachi2",
+    "validphys.n3fit_data",
+    "validphys.mc2hessian",
+    "reportengine.report",
+]
 
 log = logging.getLogger(__name__)
+
 
 class App(app.App):
 
     environment_class = Environment
     config_class = Config
 
-    critical_message = (
-            """A critical error ocurred. This is likely due to one of the following reasons:
+    critical_message = """A critical error ocurred. This is likely due to one of the following reasons:
 
  - A bug in validphys.
  - Corruption of the provided resources (e.g. incorrect plotting files).
@@ -76,13 +76,12 @@ including the contents of the following file:
 
 %s
 """
-)
 
     @property
     def default_style(self):
         return os.fspath(mplstyles.smallstyle)
 
-    def __init__(self, name='validphys', providers=providers):
+    def __init__(self, name="validphys", providers=providers):
         super().__init__(name, providers)
 
     @property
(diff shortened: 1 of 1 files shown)


Zaharid:
I find the choices of what is named with underscores here a bit confusing. Both regarding the class itself and e.g. `_profile` which is used somewhere else and yet has an underscore.

More generally, there is the question of why we have a Loader class in the first place: I think the original idea was to be able to give it different profile settings, which originally were the folders, but then settings started getting added to the profile. Now the arguments to Loader are mostly useless because it is fully controlled by the profile, for which there is no explicit configuration.  I think the correct interface would be a profile dictionary (with the current profile as a default). On the other hand nobody has ever wanted customized loader settings so we could as well remove all input settings...

scarlehoff:
I'll have a look tomorrow. The reason it's not like that is that I didn't want to change how it is accessed in various places.

If it looks simple enough to make the changes I'll change it, otherwise we merge this (it's not like this one makes anything more complicated and probably facilitates things for the future)

validphys2/src/validphys/app.py
```diff
@@ -23,37 +23,37 @@
...
+        'validphys.theoryinfo',
+        'validphys.pseudodata',
+        'validphys.renametools',
+        'validphys.covmats',
+        'validphys.hyperoptplot',
+        'validphys.deltachi2',
+        'validphys.n3fit_data',
+        'validphys.mc2hessian',
+        'reportengine.report',
+        ]
```
Zaharid, line 56: This edit is both adding to the diff and still leaving the formatting in a bit of a peculiar state: either revert these altogether or run e.g.  black (or this thing that I haven't used https://pypi.org/project/darker/) to not have this square braked dangling (and use four spaces).

Same for the formatting changes below.

validphys2/src/validphys/loader.py
```diff
@@ -62,46 +62,69 @@ class CutsNotFound(LoadFailedError): pass
...
+    """
+    profile_path = os.environ.get("NNPDF_PROFILE_PATH", profile_path)
+    if profile_path is None:
+        # Check both sys paths
+        prefix_paths = [sys.prefix, sys.base_prefix]
+        for prefix in prefix_paths:
+            check = pathlib.Path(prefix) / "share/NNPDF/nnprofile.yaml"
+            if check.exists():
+                profile_path = check
+                break
```
Zaharid, line 89: The corresponding one liner is
```python
next((p for prefix in  (sys.prefix, sys.base_prefix) if (p:=(pathlib.Path(prefix)/"share/NNPDF/nnprofile.yaml")).exists() ),None)
```
but seeing as I spent 10 minutes searching for the appropriate syntax for the walrus (which I find backwards) I am not really recommending it.

validphys2/src/validphys/loader.py
```diff
@@ -62,46 +62,69 @@ class CutsNotFound(LoadFailedError): pass
...
+
+    If no profile is found a LoaderError will be thrown
+    """
+    profile_path = os.environ.get("NNPDF_PROFILE_PATH", profile_path)
+    if profile_path is None:
+        # Check both sys paths
+        prefix_paths = [sys.prefix, sys.base_prefix]
+        for prefix in prefix_paths:
+            check = pathlib.Path(prefix) / "share/NNPDF/nnprofile.yaml"
+            if check.exists():
```
Zaharid, line 87:
```suggestion
            if check.is_file():
```

validphys2/src/validphys/loader.py
```diff
@@ -62,46 +62,69 @@ class CutsNotFound(LoadFailedError): pass
...
+                profile_path = check
+                break
+    if profile_path is None:
+        raise LoaderError("Missing an NNPDF profile file")
+
+    mpath = pathlib.Path(profile_path)
+    try:
+        profile_dict = yaml.safe_load(mpath.open("r"))
```
Zaharid, line 95: It is generally recommended to use `open` as a context manager.
```suggestion
        with mpath.open() as f:
            profile_dict = yaml.safe_load(f)
```

validphys2/src/validphys/loader.py
```diff
@@ -62,46 +62,69 @@ class CutsNotFound(LoadFailedError): pass
...
+    except yaml.YAMLError as e:
```
Zaharid, line 96:
```suggestion
    except (OSError, yaml.YAMLError) as e:
```

validphys2/src/validphys/loader.py
```diff
@@ -1020,3 +1042,5 @@ def __getattribute__(self, attr):
             if hasattr(RemoteLoader, 'download_' + resname):
                 return super().__getattribute__('make_checker')(resname)
         return super().__getattribute__(attr)
+
+
```
Zaharid, line 1046:
```suggestion
```

Zaharid (commented):
Looks good other than the minor stylistic quibbles.

validphys2/src/validphys/loader.py
scarlehoff: I don't really one liners like that. Looks cool but it's not like one gains anything.
I like the walrus though, since we are 3.8 now I might start using it

validphys2/src/validphys/app.py
scarlehoff: I think the problem is I did use black

Zaharid: For me black formats that list to

```python
providers = [
    'validphys.results',
    'validphys.commondata',
    'validphys.pdfgrids',
    'validphys.pdfplots',
    'validphys.dataplots',
    'validphys.fitdata',
    'validphys.arclength',
    'validphys.sumrules',
    'validphys.reweighting',
    'validphys.kinematics',
    'validphys.correlations',
    'validphys.chi2grids',
    'validphys.eff_exponents',
    'validphys.paramfits.dataops',
    'validphys.paramfits.plots',
    'validphys.theorycovariance.construction',
    'validphys.theorycovariance.output',
    'validphys.theorycovariance.tests',
    'validphys.replica_selector',
    'validphys.closuretest',
    'validphys.mc_gen',
    'validphys.theoryinfo',
    'validphys.pseudodata',
    'validphys.renametools',
    'validphys.covmats',
    'validphys.hyperoptplot',
    'validphys.deltachi2',
    'validphys.n3fit_data',
    'validphys.mc2hessian',
    'reportengine.report',
]
```

which I find okay.

validphys2/src/validphys/loader.py
Zaharid: FWIW every time I have thought about using it I either got wrong the operator precedence or had do search the precise scoping rules, so not much of a gain overall...

validphys2/src/validphys/app.py
scarlehoff: Funny. I wonder whether I have a funny default set or I opened it with the wrong program or smt.
Anyway, I'll revert it when I'm in front of the computer, I just wanted to remove those lines.

validphys2/src/validphys/loader.py
scarlehoff: I've avoided until now because python 3.6 is still everywhere by means of old computers amd clusters so I don't have much experience with it.
But I like the fact that it's a walrus :P

PR #1430, opened 2021-10-11
Change ThPredictions to python predictions

### LINKED ISSUE #1347: `ThPredictionsResult` should not use C++
At the moment when computing theory predictions we make calls to `C++`

https://github.com/NNPDF/nnpdf/blob/70ca11df5c42079ef4d3daec774a4222473372a7/validphys2/src/validphys/results.py#L131

afaict this can be replaced in favour of `validphys.convolution`. In particular the stats no longer need to be done using C++ too. However, before fixing this I just want to make sure of the following:

When we have

https://github.com/NNPDF/nnpdf/blob/70ca11df5c42079ef4d3daec774a4222473372a7/validphys2/src/validphys/results.py#L108

Is this equivalent to:

```python
np.std(self._rawdata, axis=1)
```

because at the moment, (for NMC for example)

```python
np.allclose(self.std_error, np.std(self._rawdata, axis=1)
# False
```
cc @scarrazza

scarlehoff:
Closes #1347. In practice (for the user) nothing changes afaik, one can continue using ThPredictionResult.from_convolution but now it will use the python convolutions module instead. Everything works just the same, only a few regression test change but these changes are in the third digit (the annoying part is that the picture in test_plot changes also in a few pixels which is enough for the test to fail, so I had to redo that as well).

I added a few checks to convolutions.py but on second thought I don't know if they are very useful or at all.

Note: I'm not touching https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/chi2grids.py since that module gets completely changed in #1424 and these two PR don't overlap for any other action.


scarlehoff:
Regarding the fact of doing central predictions and the envelope separately: this can be easily implemented but it needs to be done in #1433 since I can add a flag to `grid_values` to include or not central values (instead in this PR I would need to either modify `convolution` or `libNNPDF` to be then remove in #1433)

scarlehoff:
The `central_value` problem will be solved two PR after this one so this one can be accepted and merged if people is pleased with the rest (if people is not pleased then such information would also be appreciated)

validphys2/src/validphys/results.py
```diff
@@ -35,6 +34,7 @@
     calc_phi,
     bootstrap_values,
 )
+from validphys.convolution import predictions, central_predictions, PredictionsRequireCutsError
```
Zaharid, line 37:
```suggestion
from validphys.convolution import (
    predictions,
    central_predictions,
    PredictionsRequireCutsError,
)

```

validphys2/src/validphys/results.py
```diff
@@ -96,12 +102,20 @@ def sqrtcovmat(self):
...
+    """
+    def __init__(self, dataobj, stats_class, label=None, central_value=None):
         self.stats_class = stats_class
         self.label = label
-        self._std_error = dataobj.get_error()
-        self._rawdata = dataobj.get_data()
-        super().__init__(dataobj)
+        # Ducktype the input into numpy arrays
+        try:
+            self._std_error = dataobj.std(axis=1).to_numpy()
```
Zaharid, line 113: I don't believe this is correct: If the constructor is accepting  a stats_class then that should be used to compute statistics. As discussed, ideally we would get rid of these objects and use `StatsResult` instead.

validphys2/src/validphys/results.py
```diff
@@ -123,16 +137,28 @@ def make_label(pdf, dataset):
...
+            datasets = dataset.datasets
+        except AttributeError:
+            datasets = (dataset,)
+
+        try:
+            all_preds = []
+            all_centrals = []
+            for d in datasets:
+                all_preds.append(predictions(d, pdf))
+                all_centrals.append(central_predictions(d, pdf))
```
Zaharid, line 152: Just a note to mostly myself that we are supposed to be  rid of this in some subsequent PR.

Zaharid (changes requested):
I think we need to look into how hessian sets are supposed to work here.

validphys2/src/validphys/results.py
scarlehoff: The stats class was not being used here at all. Also, the plans for the `StatsResult` would probably need changes to PDF so that should wait for the changes to the PDF.

validphys2/src/validphys/results.py
scarlehoff: All I can do is to promise I will work on this as soon as the dust is settled on the rest so that it doesn't get forgotten.

validphys2/src/validphys/results.py
Zaharid: But that line is incorrect for Hessian sets, no?

scarlehoff: because of replica 0?

scarlehoff: ohh, I see what you mean now. 
Because they are not necessarily symmetric and on top of that they don't necessarily represent the 68cl version. I'll fix that with the Stats Class which I guess is bringing all the ifs in the libNNPDF side

scarlehoff:
> I think we need to look into how hessian sets are supposed to work here.

Maybe we should have hessian regression tests as well...

scarlehoff:
Mm, by using stats_class the MC pdfs are broken instead.

scarlehoff:
In the last commit I've had to update the regression test but it passes for `rtol=1e-4` so I think it is ok.

Zaharid:
FWIW this is still using the cpp commondata, for e.g. `results`. I suppose we would want to change that later.

Zaharid (approved):

Zaharid:
I would want to make sure that nothing catastrophic happens with the vp report, but other than that it looks fine (given the various future discussions).

Zaharid:
@scarlehoff the problem looks real.

validphys2/src/validphys/results.py
```diff
@@ -45,10 +45,16 @@ class Result:
...
 
-    def __init__(self, dataobj):
-        self._central_value = dataobj.get_cv()
+    def __init__(self, dataobj, central_value=None):
+        if central_value is None:
+            try:
+                central_value = dataobj.mean(axis=1).to_numpy()
+            except AttributeError:
+                central_value = dataobj.get_cv()
+        self._central_value = np.array(central_value).squeeze()
```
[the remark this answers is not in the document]
Zaharid: Are we sure that the `squeeze` here is the right thing to do? Looks like it is is gong to cause problems for one point datasets.

Zaharid (changes requested):
This seems to fail.

Zaharid:
Btw, didn't we have a test with one datapoint at some point?

scarlehoff:
> Btw, didn't we have a test with one datapoint at some point?

Apparently not.
This PR is revealing a lot of corner cases we don't test for. I'll fix this and add a test that fails for this.

siranipour:
We have it for the covmat pipeline
https://github.com/NNPDF/nnpdf/blob/9467c2527e48795abaab288a0aec7763d8695029/validphys2/src/validphys/tests/test_covmats.py#L179

scarlehoff:
> We have it for the covmat pipeline
> 
> https://github.com/NNPDF/nnpdf/blob/9467c2527e48795abaab288a0aec7763d8695029/validphys2/src/validphys/tests/test_covmats.py#L179

Precisely in that module a few test break if I add an extra dataset to the test list... I'm not entirely sure why.

scarlehoff:
(making this into draft until it is rebased on top of #1473)

scarlehoff:
The test after the rebase (so, commit 1489276) will fail because of the `squeeze` and a small change in the regression tests (`1e-3`, they pass using `rtol` / `atol`).

Zaharid:
@scarlehoff this error https://github.com/NNPDF/nnpdf/runs/4324437054?check_suite_focus=true  seems genuine (and not obvious to be what it is by looking quickly at it).

scarlehoff:
Is the random NaN problem that I am unable to reproduce. I have a computer with Ubuntu looping on the test but it never fails...

scarlehoff:
Rebased. After a tortuous journey this should all pass and work fine now.

Zaharid (approved):
I have ran this with a few options, and also looked at it many times by now. Seems fine to me.

PR #1460, opened 2021-11-02
Testing the conda 3.9 build

scarlehoff:
I'm able to compile this in my computer (we had too many pins) but I prefer to let the CI test.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: [3.8]
+        python-version: [3.9]
         include:
           - os: ubuntu-latest
             CONDA_OS: linux-64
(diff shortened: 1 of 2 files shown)


RoyStegeman:
Why not prepend the conda-forge channel to the .condarc?

scarlehoff:
I think the best option is to force gcc > 10, that way it works even if people has the wrong channel in their `condarc`.

scarlehoff:
Pinning openssl > 3.0 forces everything relevant (including python) to come from conda-forge.
Just pinning the compiler was not enough. But this doesn't solve the problem because there are way too many conflicts that arise in the `run` section by doing that.
The problem is I don't know which package is the one that needs to come from conda-forge so I haven't been able to isolate the issue. I'll close this for now.

Zaharid:
I think we should indeed swap the channels and even try to make all of them conda-forge (i.e. remove defaults as an option).

PR #1461, opened 2021-11-02
Changed reserve -> resize in FutureColliders.cc in buildmaster

J-M-Moore:
Fixing runtime error in FutureColliders.cc.


Zaharid (approved):

Zaharid:
This looks like it is fixing that problem.

Issue #1458, opened 2021-11-01
Tensorflow package for python 3.9


Zaharid:
I finally got around trying to build the package with python 3.9 and `boa`. Instead of the horrible error messages in #1335 I now see something that looks like explaining the cause of the problem. I set the environment variable `CONDA_PY=39` and run `conda mambabuild conda-recipe` I get the error

```
Mamba failed to solve:
 - coverage
 - swig 3.0.10.*
 - pytest-mpl
 - sysroot_linux-64 2.17.*
 - pytest
 - hypothesis
 - cmake
 - gcc_linux-64
 - gxx_linux-64
 - opt_einsum >=2.3.2
 - gsl >=2.4,<2.5.0a0
 - gsl >=2.4,<2.5.0a0
 - sphinxcontrib-bibtex
 - libstdcxx-ng >=9.3.0
 - sphinx_rtd_theme >0.5
 - yaml-cpp >=0.6.3,<0.6.4.0a0
 - yaml-cpp >=0.6.3,<0.6.4.0a0
 - psutil
 - gast 0.3.3.*
 - pandas
 - python >=3.9,<3.10.0a0
 - blessings >=1.7
 - tensorflow >=2 *eigen*
 - requests
 - pkg-config
 - sphinx >=4.0.2
 - docutils 0.16.*
 - seaborn
 - prompt_toolkit
 - recommonmark
 - validobj
 - matplotlib >=3.3.0
 - libarchive >=3.4,<3.5.0a0
 - libarchive >=3.4,<3.5.0a0
 - hyperopt
 - numpy
 - scipy >=0.19.1
 - sqlite >=3.36.0,<4.0a0
 - libgcc-ng >=9.3.0
 - reportengine >=0.30.25
 - lhapdf >=6.4.0,<6.4.1.0a0
 - apfel >=3.0.5.0,<3.0.5.1.0a0

with channels:

The reported errors are:
   Encountered problems while solving:
     - package tensorflow-2.0.0-eigen_py36ha83d16c_0 requires tensorflow-base 2.0.0 eigen_py36h4ed9498_0, but none of the providers can be installed
```

which looks like it is saying that there is no "eigen" version that works.  Is there some other package that would work with python 3.9 without memory leaks and other known problems?

Zaharid:
Incidentally I have the impression that #1284 would be an easy and quick win e.g. since it gave me this error in seconds rather than hours.

RoyStegeman:
[...]

PR #1463, opened 2021-11-03
Enable python 3.9 and mamba build

scarlehoff:
As per https://github.com/NNPDF/nnpdf/issues/1458#issuecomment-958807144

If the workflow works I'll enable also 3.8 so that we have both for the time being.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -8,7 +8,7 @@ on:
 
 # some general variables
 env:
-  CONDA_PY: 38
+  CONDA_PY: 39
   IMAGE_NAME: nnpdf
 
 jobs:
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8]
+        python-version: [3.9]
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v1
--- a/.github/workflows/fitbot.yml
+++ b/.github/workflows/fitbot.yml
@@ -111,7 +111,7 @@ jobs:
         url=$(vp-upload output/ 2>&1 | grep https)
         echo "REPORT_URL=$url" >> $GITHUB_ENV
     - name: Save the package if something fails
-      if: ${{ failure }}
+      if: failure()
       shell: bash -l ${0}
       run: |
         scp /usr/share/miniconda/envs/test/conda-bld/linux-64/nnpdf*.tar.bz2  nnpdf.fisica.unimi.it:debug_pkgs/
--- a/.github/workflows/fitbot.yml
+++ b/.github/workflows/fitbot.yml
@@ -11,7 +11,7 @@ env:
   N3FIT_MAXNREP: 20 # total number of replicas to fit
   POSTFIT_NREP: 16 # requested replicas for postfit
   REFERENCE_SET: NNBOT-d3034a7cc-2021-06-11 # reference set for vp
-  CONDA_PY: 38
+  CONDA_PY: 39
 
 jobs:
   build:
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8]
+        python-version: [3.9]
     runs-on: ${{ matrix.os }}
     env:
       NETRC_FILE: ${{ secrets.NETRC_FILE }}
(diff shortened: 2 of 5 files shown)


scarlehoff:
I'm waiting for the 3.8 fitbot and then I'll send the 3.9 fitbot to see whether there are any differences.

scarlehoff:
Mm, the fit itself is ok and I seem to be able to run comparefit with python 3.9 so this cannot be merged yet. There's some package that it's not working correctly...

Zaharid:
I am

> Mm, the fit itself is ok and I seem to be able to run comparefit with python 3.9 so this cannot be merged yet. There's some package that it's not working correctly...

Not sure I understand this.

scarlehoff:
> Not sure I understand this.

I shouldn't write complex sentences too early in the morning. It's missing "but for the fit fails, so we cannot merge this yet".
I've downloaded the fit and I am able to run comparefits in my computer (and the fit looks ok, which makes sense because the test pass) so I need to compare my packages with the bot packages to see what created the problem.

Zaharid:
FWIW the error is

```
   File "/usr/share/miniconda/envs/nnpdfenv/lib/python3.9/site-packages/validphys/covmats.py", line 391, in sqrt_covmat

    decomp = la.cholesky(correlation_matrix)
  File "/usr/share/miniconda/envs/nnpdfenv/lib/python3.9/site-packages/scipy/linalg/decomp_cholesky.py", line 88, in cholesky
    c, lower = _cholesky(a, lower=lower, overwrite_a=overwrite_a, clean=True,
  File "/usr/share/miniconda/envs/nnpdfenv/lib/python3.9/site-packages/scipy/linalg/decomp_cholesky.py", line 17, in _cholesky
    a1 = asarray_chkfinite(a) if check_finite else asarray(a)
  File "/usr/share/miniconda/envs/nnpdfenv/lib/python3.9/site-packages/numpy/lib/function_base.py", line 488, in asarray_chkfinite
    raise ValueError(
ValueError: array must not contain infs or NaNs
```

It might be something with numpy or rather pandas.

scarlehoff:
Doesn't seem to be the case though.

RoyStegeman:
Is it a problem with python 3.9 or with boa? Since you changed both at the same time.

scarlehoff:
I have no idea. I'm inclined to blame the cosmic rays though because the test passed with both 3.9 and boa.

Zaharid:
@scarlehoff just to be clear, are you able to run comparefits with the equivalent package built locally? Also what is the difference with the fit above that passed?

scarlehoff:
> @scarlehoff just to be clear, are you able to run comparefits with the equivalent package built locally? Also what is the difference with the fit above that passed?

I was able to run comparefits with a (dirty that I used for testing yesterday) 3.9 conda environment, but I checked the usual suspects (`numpy`, `pandas`, `scipy`) and the versions where those of the bot. So that's all I can say at the moment.

I have to create a fresh environment, built the package, redownload resources and test to be equivalent to what the bot did.

scarlehoff:
> @scarlehoff just to be clear, are you able to run comparefits with the equivalent package built locally? Also what is the difference with the fit above that passed?

Now the answer is yes: https://vp.nnpdf.science/ILCPf1olSs-GWwDP_vDjdw==/

Zaharid:
Uh.

Is there a difference between the fit that failed and the one that succeeded?

scarlehoff:
> Is there a difference between the fit that failed and the one that succeeded?

No. Which should be the case (all packages should be the same, the only difference being whether the box is running in 3.8 or 3.9).

scarlehoff:
So it was really due to cosmic rays...

RoyStegeman:
I guess this is done now, right? Just remove the lines that save the package upon test failure. 

And do you want to run the tests for both 3.8 and 3.9, or just 3.9?

scarlehoff:
> Just remove the lines that save the package upon test failure.

It might be useful in the future.

> And do you want to run the tests for both 3.8 and 3.9, or just 3.9?

Both, it's "free" (maybe not that free in Mac? We could skip that...) and we want to have packages for both.

RoyStegeman:
> Both, it's "free" (maybe not that free in Mac? We could skip that...) and we want to have packages for both.

I think MacOS is free as well. At least it's not mentioned that it's the excepoiton to the first sentence: 
https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#about-billing-for-github-actions

scarlehoff:
Not in terms of money but I think (I might be wrong) that while you can run many Linux actions in parallel, for Mac OS that's limited to something like 5 (and then if both 3.8 and 3.9 are active at the same time only two commits can be running at the same time instead of 5 of them).

RoyStegeman:
> and then if both 3.8 and 3.9 are active at the same time only two commits can be running at the same time instead of 5 of them

This doesn't really seem very limiting for us.

scarlehoff:
@Zaharid have a look at this and if you're ok with it merge, like and subscribe.

Zaharid (approved):

Zaharid:
Thanks for this work!

As discussed I think we should move to conda forge in another PR and see how disruptive that is.

PR #1466, opened 2021-11-11
Fix pseudodata indexing mismatch

siranipour:
This caused the indexing of the pseudodata to be messed up

### DIFF
--- a/validphys2/src/validphys/commondata.py
+++ b/validphys2/src/validphys/commondata.py
@@ -29,6 +29,10 @@ def loaded_commondata_with_cuts(commondata, cuts):
     return lcd.with_cuts(cuts)
 
 dataset_inputs_loaded_cd_with_cuts = collect(
+    "loaded_commondata_with_cuts", ("data_input",)
+)
+
+groups_dataset_inputs_loaded_cd_with_cuts = collect(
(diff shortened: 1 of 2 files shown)


Zaharid:
@siranipour Actually I am not sure if it is free of trouble to change this: `dataset_inputs_loaded_cd_with_cuts` is used in the covmats module, where instead it is expected that it is ordered as  `data_input`. Maybe we should do one of:

  - Define a different provider as done in this pr, as done in this PR.
  - Have an index that doesn't reorder the inputs.

siranipour:
Okay should be an easy fix

Zaharid:
Also, would it be easy to add a test for this?

Zaharid:
@scarlehoff I am going to  go ahead and say that this isn't failing twice in the same way because of cosmic rays.

scarlehoff:
Interesting. It does seem to be random though.

scarlehoff:
Sadly my line to copy the package also failed :__

scarlehoff:
I still cannot reproduce it so here's your report: https://vp.nnpdf.science/Ki5hG87QQvusLVERrPUlFw==

Zaharid (approved):

PR #1467, opened 2021-11-11
Disable spinner in non interactive mode

Zaharid:
Print something to the terminal instead. This avoids needlessly
cluttering CI outputs.

Example

```python
import time
import sys
import subprocess

from validphys.renametools import Spinner

with Spinner():
    time.sleep(5)

if sys.argv[-1] != "no":
    print(
        subprocess.run(
            [sys.executable, __file__, "no"], capture_output=True, text=True
        ).stdout
    )
```

### DIFF
--- a/validphys2/src/validphys/renametools.py
+++ b/validphys2/src/validphys/renametools.py
@@ -12,6 +12,10 @@ class Spinner:
     """ Context manager to provide a spinning cursor
     while validphys performs some other task silently.
 
+    When exececuted in a TTY, it shows a spinning cursor for the duration of
+    the context manager. In non interactive prompts, it prints to stdout at the
+    beginning and end.
+
     Example
     -------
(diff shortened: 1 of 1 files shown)


validphys2/src/validphys/renametools.py
```diff
@@ -12,6 +12,10 @@ class Spinner:
     """ Context manager to provide a spinning cursor
     while validphys performs some other task silently.
 
+    When exececuted in a TTY, it shows a spinning cursor for the duration of
```
siranipour, line 15: What's TTY?

siranipour (approved):

validphys2/src/validphys/renametools.py
Zaharid: Uh https://www.howtogeek.com/428174/what-is-a-tty-on-linux-and-how-to-use-the-tty-command/

PR #1468, opened 2021-11-12
[WIP] Fixing the bot

scarlehoff:
Given that we have a case that is supposed to fail, let's see whether we can actually reproduce it.

### DIFF
--- a/.github/workflows/fitbot.yml
+++ b/.github/workflows/fitbot.yml
@@ -74,36 +74,36 @@ jobs:
       run: |
         conda activate nnpdfenv
         cd $RUNFOLDER
-        for ((i=1; i<=$N3FIT_MAXNREP; i+=2)); do n3fit $RUNCARD.yml $i -r $((i+1)); done
+        #for ((i=1; i<=$N3FIT_MAXNREP; i+=2)); do n3fit $RUNCARD.yml $i -r $((i+1)); done
     # performing DGLAP
     - name: Running dglap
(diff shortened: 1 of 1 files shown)


Zaharid:
I wonder if I can do something to make the error message better for that problem.

scarlehoff:
Yes, something like "it seems we are in a period of high cosmic activity"

.github/workflows/fitbot.yml
```diff
@@ -114,7 +114,7 @@ jobs:
       if: failure()
       shell: bash -l ${0}
       run: |
-        scp /usr/share/miniconda/envs/test/conda-bld/linux-64/nnpdf*.tar.bz2  nnpdf.fisica.unimi.it:debug_pkgs/
+        scp -o StrictHostKeyChecking=no $CONDA_PREFIX/conda-bld/${{matrix.CONDA_OS}}/*.tar.bz2 nnpdf@fisica.unimi.it:debug_pkgs/
```
scarlehoff, line 117:
```suggestion
        ls $CONDA_PREFIX/conda-bld/*/*.tar.bz2
        scp -v -o StrictHostKeyChecking=no $CONDA_PREFIX/conda-bld/*/*.tar.bz2 nnpdf@fisica.unimi.it:debug_pkgs/
```

scarlehoff:
Well, it consistently fails even with master so I should be able to reproduce that in my computer if I do enough loops.

Issue #1288, opened 2021-06-10
Fix unknown process type warnings


Zaharid:
Running comparefits we get a bunch of

```
[WARNING]: Unknown process type for dataset CHORUSNUPb_dw_ite
```

These should not be there.

PR #1472, opened 2021-11-19
In process_function take nnpdf31_process from PLOTTING_*.yaml

RoyStegeman:
addresses #1288

### DIFF
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -52,7 +52,6 @@ from validphys.gridvalues import LUMI_CHANNELS
 
 from validphys.paramfits.config import ParamfitsConfig
 
-from validphys.theorycovariance.theorycovarianceutils import process_lookup
 from validphys.plotoptions import get_info
 
 import validphys.scalevariations
@@ -763,9 +762,13 @@ class CoreConfig(configparser.Config):
 
             with self.set_context(ns=self._curr_ns.new_child(spec)):
                 _, data_input = self.parse_from_(None, "data_input", write=False)
-                names = {
-                    (process_lookup(dsin.name), dsin.name): dsin for dsin in data_input
-                }
+
+                names = {}
+                for dsin in data_input:
+                    cd = self.produce_commondata(dataset_input=dsin)
+                    proc = get_info(cd).nnpdf31_process
+                    ds = dsin.name
+                    names[(proc, ds)] = dsin
 
                 all_names.append(names)
         used_set = set.intersection(*(set(d) for d in all_names))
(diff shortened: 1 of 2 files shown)


Zaharid:
A couple of things need to be done here:

- We have higher level functions for recovering the plotting file information, specifically `validphys.plotoptios.get_info`. See e.g. https://github.com/NNPDF/nnpdf/blob/9467c2527e48795abaab288a0aec7763d8695029/validphys2/src/validphys/config.py#L1466 Also ideally this would not touch a Loader and instead recover a commondata object, but looking at the uses of this function that might be involved.

- In the code triggering the  warnings in comparefits, specifically https://github.com/NNPDF/nnpdf/blob/9467c2527e48795abaab288a0aec7763d8695029/validphys2/src/validphys/config.py#L767 we should definitively use the get_info lookup (also  I was not too happy about adding anything related to processes there but not looking at what that breaks).

RoyStegeman:
> we should definitively use the get_info lookup

Sure, in the config Class everything is already there to do that instead of the process_lookup, but that's not the case for the other instances of process_lookup. 

> Also ideally this would not touch a Loader and instead recover a commondata object

Not really sure what you mean here. A commondata object is alsways obtained through a Loader at some stage, right? Both the lines you point at are inside CoreConfig, while process_lookup is called outside that class in most cases so the commondata object is not as readily available there.

RoyStegeman:
@Zaharid could you clarify your points above?

Zaharid:
> > we should definitively use the get_info lookup
> 
> Sure, in the config Class everything is already there to do that instead of the process_lookup, but that's not the case for the other instances of process_lookup.
> 
> > Also ideally this would not touch a Loader and instead recover a commondata object

I meant that line specifically, which is used in a lot of code paths.

> 
> Not really sure what you mean here. A commondata object is alsways obtained through a Loader at some stage, right? Both the lines you point at are inside CoreConfig, while process_lookup is called outside that class in most cases so the commondata object is not as readily available there.

I like the idea of having some separation of concerns between what happens before and after the graph runs. Stuff like funding appropriate paths to commonda happens before as of now. On top of the (arguable) theoretical purity, it does enable certain features, like for example allowing to [override](https://docs.nnpdf.science/modules/validphys/validphys.html?highlight=use_fitcommondata#validphys.config.CoreConfig.parse_use_fitcommondata) the commondata lookup for the purposes of  closure tests. That works because there is a single point that knows how to get the commondata(spec) and knows how to handle that keyword. I suppose it is mostly inconsequential for the remaining uses, but would be good if it wasn't needed.

Zaharid (approved):

PR #1473, opened 2021-11-19
Add a single data point dataset to the tests

scarlehoff:
This would fail with #1430 and adds two nice features:

1) A single data point dataset is used in a variety of tests
2) The group of data now contains two datasets from the same experiment which is another "corner case" which was not tested for.

The problem is that 2) breaks the tests for the covmat. So a few commits are needed to fix that.


siranipour:
Seems good, let's see what the tests do

scarlehoff:
There's also the problem that many test use libNNPDF... I might just remove the libNNPDF part for the regression test.

scarlehoff:
Found the problem, these two calls produce the same results if and only if the dataset from the experiments come together in the list...

```python
covmat = API.dataset_inputs_t0_covmat_from_systematics(**config)
cpp_covmat = API.groups_covmat(**config)
```

Zaharid:
See also #1466

scarlehoff:
Suggestions welcome for the test that I've added in the last commit. Not sure what to test other than the fact that it works. I don't think this function was used in this way in any test.

siranipour (approved):

Issue #1476, opened 2021-11-25
np.exp() makes libNNPDF fail down the line


scarlehoff:
The `test_rebuilddata.py` test fails for python 3.9 in github actions.

- It doesn't fail in my own installation
- I've tried using a PC with Ubuntu to be a bit closer to the github action and it also works
- It used to be inconsistent but for #1430 has failed 4 times in a row

https://github.com/NNPDF/nnpdf/runs/4324994519?check_suite_focus=true#step:6:2498

The only way of recreating the issue I can think of is using a docker image. I'll do it when I have some free time, if anyone has any clues or want to debug this feel free. (I would do it directly with #1430)

Zaharid:
I wonder if that particular bug happens because the commondata files in the fit used for testing contain `"nan"` strings which are parsed in an undefined way, similar to #1465. That doesn't really explain why it happens intermittently or with the new code, but it may be worth checking.

Additionally, it would be nice if we could get better error messages out of the failure, e.g. by placing some explicit test here https://github.com/NNPDF/nnpdf/blob/eeae19b15c960b4f8f974c3ac2ba2af36ed845bc/validphys2/src/validphys/covmats.py#L391 that tries to print out the index of the offending data on failure. @RoyStegeman mind having a look?

RoyStegeman:
Sure, I can try

Zaharid:
Trying to run that test locally I also see nothing resembling the issue on github.

scarlehoff:
> Trying to run that test locally I also see nothing resembling the issue on github.

The github server are clearly in the upper atmosphere

scarlehoff:
Ok, with the docker img I can get it to fail.

Turns out, `test_rebuildata.py` fails only when it is run after `test_plost.py`. Looking inside I can trace the problem down to 

https://github.com/NNPDF/nnpdf/blob/eeae19b15c960b4f8f974c3ac2ba2af36ed845bc/validphys2/src/validphys/plotoptions/core.py#L352

And more specifically, the test fails due to this line here

https://github.com/NNPDF/nnpdf/blob/eeae19b15c960b4f8f974c3ac2ba2af36ed845bc/validphys2/src/validphys/plotoptions/kintransforms.py#L97 

The test running that line is `test_plot_xq2`. I've checked and there's no infinite or anything weird there.  And anyway the test fails _after_ that. That is, the next test fails.

RoyStegeman:
> And more specifically, the test fails due to this line here

[...]

Issue #1475, opened 2021-11-25
Proposed changes to the behaviour of the PDF class.


scarlehoff:
This is a collection of issues with the current implementation of the PDF class.

This issue will be organized as:
- Problem
- Examples
- Solution

# Error Type

Half of the issues with the PDF class is the management of the error type.

The `nnpdf_error` attribute:

https://github.com/NNPDF/nnpdf/blob/eeae19b15c960b4f8f974c3ac2ba2af36ed845bc/validphys2/src/validphys/core.py#L180

I will remove this attribute and all references to it. If something is accessing this attribute then it is a libNNPDF thing and should be removed. Instead (as was suggested) any behavior that depends on the error type of the PDF should be interfaced through its associated `stats_class`.

# Actual PDF

Currently `load` loads a libNNPDF PDF set.

https://github.com/NNPDF/nnpdf/blob/eeae19b15c960b4f8f974c3ac2ba2af36ed845bc/validphys2/src/validphys/core.py#L162

It will be changed to a python wrapper to the `lhapdf` python interface. It makes sense to do it this way because it allows for certain optimization (such as `grid_values`).

# Attributes

Currently we look inside the info file of the PDF by using `hasattr`, which is confusing and hides what's actually happening.

https://github.com/NNPDF/nnpdf/blob/eeae19b15c960b4f8f974c3ac2ba2af36ed845bc/validphys2/src/validphys/core.py#L101
https://github.com/NNPDF/nnpdf/blob/eeae19b15c960b4f8f974c3ac2ba2af36ed845bc/validphys2/src/validphys/core.py#L149

I will change it to something more explicit.

# Reference to libNNPDF PDF

While there is a few instances where libNNPDF is needed, these are only `t0` situations:

https://github.com/NNPDF/nnpdf/blob/d68954eef2d528c9991e37faa30c2ca477a9208b/validphys2/src/validphys/covmats.py#L476

https://github.com/NNPDF/nnpdf/blob/d68954eef2d528c9991e37faa30c2ca477a9208b/validphys2/src/validphys/covmats.py#L510

https://github.com/NNPDF/nnpdf/blob/d68954eef2d528c9991e37faa30c2ca477a9208b/validphys2/src/validphys/closuretest/multiclosure.py#L36

which is nice because it means I can do all of the above while leaving a `load_nnpdf_t0` interface if needed.

[...]

PR #1477, opened 2021-11-26
Dispatching Python covariance matrices

siranipour:
We now dispatch covmats purely from python. Addresses various points including https://github.com/NNPDF/nnpdf/issues/1476#issuecomment-979823687 and https://github.com/NNPDF/nnpdf/issues/1475#issuecomment-979425362

Let's see what the tests do


scarlehoff:
I've rebased your commit on top of #1430 to see whether the test passes also there https://github.com/NNPDF/nnpdf/runs/4333530141?check_suite_focus=true

siranipour:
Looks like the linux tests are passing on the rebase

validphys2/src/validphys/tests/test_regressions.py
```diff
@@ -69,15 +69,19 @@ def test_expcovmat(data_config):
 
 @make_table_comp(parse_exp_mat)
 def test_t0covmat(data_witht0_config):
-    return API.groups_covmat_no_table(**data_witht0_config)
+    d = dict(data_witht0_config)
+    d['use_cuts'] = 'internal'
```
scarlehoff, line 73: You can use this instead: https://github.com/NNPDF/nnpdf/blob/eeae19b15c960b4f8f974c3ac2ba2af36ed845bc/validphys2/src/validphys/tests/conftest.py#L94

siranipour: Wonderful, good spot.

validphys2/src/validphys/covmats.py
```diff
@@ -472,133 +472,10 @@ def groups_corrmat(groups_covmat):
...
-    if norm_threshold is not None:
-        covmat = regularize_covmat(
-            covmat,
-            norm_threshold=norm_threshold
-        )
-    return covmat
-
-
-@check_dataset_cuts_match_theorycovmat
-def covmat(
```
Zaharid, line 510: Where has all this code and documentation gone? How are  the `norm_threshold` and `use_weights_in_covmat` arguments handled?

siranipour: The weights are used in the lower level functions https://github.com/NNPDF/nnpdf/blob/eeae19b15c960b4f8f974c3ac2ba2af36ed845bc/validphys2/src/validphys/covmats.py#L35

Zaharid:
I am a bit worried that this is going to obscure `--help` for various things a bit too much. Admittedly that is because reportengine sucks in that regard. That could possibly be remedied with a tutorial on chi²s which we clearly need for other reasons.

Zaharid:
It would be good to have tests that cover the various options passed to the covmat. E.g. the thing with the `norm_threshold` clearly should have failed something.

siranipour:
> I am a bit worried that this is going to obscure `--help` for various things a bit too much. Admittedly that is because reportengine sucks in that regard. That could possibly be remedied with a tutorial on chi²s which we clearly need for other reasons.

Yeah I couldn't figure out how to do this without having a dummy function which the `explicit_node` could dispatch on. Let me know if you can think of a better alternative, I was thinking perhaps I've missed something obvious

> It would be good to have tests that cover the various options passed to the covmat. E.g. the thing with the `norm_threshold` clearly should have failed something.

will add now

Zaharid:
One thing I'd like to do someday is have a more restricted version of `explict_node` that looks like

```
produce_covmat_t0_considered = dispatch_node(on="use_t0", {True: covmats.t0_covmat_from_systematics, False: covmat_from_systematics})
```

which does the same thing but additionally knows  how to follow the possibilities and produce help.

siranipour:
Yes that would be nice, since it could then dispatch another `dispatch_node` which is the problem I was having here.

scarlehoff:
Should I / Can I rebase #1430 on top of this one already or are more changes needed?

siranipour:
I’m done with it if @Zaharid is happy

On 29 Nov 2021, at 10:14, Juan M. Cruz-Martinez ***@***.***> wrote:

﻿

Should I / Can I rebase #1430<https://github.com/NNPDF/nnpdf/pull/1430> on top of this one already or are more changes needed?

Zaharid:
It seems ok. I would hope that the documentation challenges are addressed by #1478.

siranipour:
Shall we merge this?

scarlehoff (approved):
I'm in favour of merging since everything seems to work in this one and also down the line (#1430 )

PR #1481, opened 2021-12-01
Correcting f string for raise statement

siranipour:
Quick fix, this was broken


scarlehoff (approved):
A classic.
I wish python f-strings were propagated in strings continuations, I often have this exact problem.

PR #1483, opened 2021-12-02
Handle corrupt figures in report indexing

Zaharid:
There were a couple of reports containing corrupt png images, from
various experiments with plotly. These would break the indexing script,
preventing it from being updated with new reports.

Make the script more robust by allowing it to ignore problems in the
thumbnail processing (while printing the exception) and also display the
report that crashed it, in case of unhandled failure.

### DIFF
--- a/validphys2/serverscripts/index-reports.py
+++ b/validphys2/serverscripts/index-reports.py
@@ -11,6 +11,7 @@ import datetime
 import json
 import re
 import sys
+import traceback
 from collections import ChainMap, defaultdict
 
 import ruamel_yaml as yaml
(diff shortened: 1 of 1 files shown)


siranipour:
Is there an easy way to test this?

Zaharid:
You could try uploading again some of your corrupt reports (which I deleted).

siranipour:
I don't know if I can reproduce the earlier issue, but the following could be uploaded using this branch:
https://vp.nnpdf.science/gicJQktuTf-RzdDN4yndMw==/
https://vp.nnpdf.science/03d1YVMwTxai_OlYYEiifA==/

PR #1485, opened 2021-12-02
Improve postfit error message

### LINKED ISSUE #1479: Improve postfit error message
Currently when trying to run postfit on new fits naively, the error message:

```
[WARNING]: Cannot find info file at /rds/user/jmm232/hpc-work/nnpdf_dev_install/top_fits/no_top_fit/nnfit/no_top_fit.info
[ERROR]: Error in postfit:
Postfit cannot find a valid LHAPDF info file
```

is given. It would be useful if this error message was more descriptive and pointed to the appropriate tutorial in the docs.

Zaharid:
A missing info file is a symptom of not having executed the evolution
code. This has cause some confusion and it is therefore useful to
suggest it as a possibility in the error message.

Closes #1479.

### DIFF
--- a/validphys2/src/validphys/scripts/postfit.py
+++ b/validphys2/src/validphys/scripts/postfit.py
@@ -140,7 +140,11 @@ def _postfit(results: str, nrep: int, chi2_threshold: float, arclength_threshold
         if not fitdata.check_nnfit_results_path(result_path):
             raise PostfitError('Postfit cannot find a valid results path')
         if not fitdata.check_lhapdf_info(result_path, fitname):
-            raise PostfitError('Postfit cannot find a valid LHAPDF info file')
+            raise PostfitError(
+                'Postfit cannot find a valid LHAPDF info file\n'
+                'Note: Perhaps you forgot to run evolven3fit? See\n'
+                'https://docs.nnpdf.science/tutorials/run-fit.html#running-the-fitting-code'
+            )
 
         nrep = int(nrep)
         if at_least_nrep:


validphys2/src/validphys/scripts/postfit.py
```diff
@@ -140,7 +140,12 @@ def _postfit(results: str, nrep: int, chi2_threshold: float, arclength_threshold
         if not fitdata.check_nnfit_results_path(result_path):
             raise PostfitError('Postfit cannot find a valid results path')
         if not fitdata.check_lhapdf_info(result_path, fitname):
-            raise PostfitError('Postfit cannot find a valid LHAPDF info file')
+            raise PostfitError(
+                'Postfit cannot find a valid LHAPDF info file\n'
+                'Note: Perhaps you forgot to run evolven3fit? Run\n'
+                f'    evolven3fit {result_path.name} {nrep}\n'
```
scarlehoff, line 146: I'm not entirely sure we want to give the user this line because it is not necessarily what they want to run (namely, they might be doing `postfit 100` but they want to do `evolven3fit 150` and reading this msg they might do `evolven3fit 100` then `postfit 100` will fail etc. I think removing this line solves the problem:

1) The people who already know and just forgot will immediately realise and do the correct thing.
2) The people who don't, it is better if they go and look at the docs anyway.
 

```suggestion
```

Zaharid: Ok fair enough.

scarlehoff (approved):

Issue #1284, opened 2021-06-10
Prefer mamba and conda forge channels


Zaharid:
The whole issue with #1279 brings to attention that conda forge now has a lot of public documentation, one can see who maintains the recipes and probably so many variations that something like  #1279  would be more apparent. 

On top of that there is the mamba project which solves dependencies a lot faster than conda (which can be borderline unusable). With that I would suggest to update https://github.com/NNPDF/binary-bootstrap/blob/master/bootstrap.sh as well as the CI code and relevant pieces of the documentation  to have conda forge as higher priorities than defaults and use mamba instead of conda.

Zaharid:
@RoyStegeman @scarlehoff what do you think about trying to switch the channel order in ci builds and see how much of a problem that causes. Then there is https://github.com/NNPDF/binary-bootstrap/pull/4 but I suppose I am not sure about the order in which to do various things. including potentially having to ask people to edit their .condarc file.

scarlehoff:
Well, it is becoming increasingly clear that we will need to do this so let's just do it and when people complain things don't work we make a point of asking first to see their `.condarc` file.

Zaharid:
Ok @RoyStegeman could you please make PR switching the order of conda-forge and defaults in various places. Hint:
```
 $ rg --hidden conda-forge
.github/workflows/fitbot.yml
42:        conda config --append channels conda-forge

.github/workflows/tests.yml
41:        conda config --append channels conda-forge

.github/workflows/docker.yml
30:        conda config --append channels conda-forge

n3fit/runcards/nnpdfenv-tag-4.0.4.yml
7:  - conda-forge

n3fit/runcards/reproduce_nnpdf40/nnpdf40_env.yml
6:  - conda-forge

docker/dockerfile
33:RUN conda config --append channels conda-forge                                    && \

doc/validphys2/guide.md
391:`defaults` channel, as well as from the community `conda-forge`
```

If that looks not too terrible I'll merge the binary bootsrap PR.

RoyStegeman:
Yes I was planning to try this over the weekend. Given that switching them locally is fine, hopefully  this should be straightforward (:crossed_fingers: ).

AleCandido:
While exploring for #1564 I found this issue.

I know that this idea might be unpopular, but instead of fiddling too much with Conda channels, or fancy alternatives like Mamba, what about releasing just a "simple" `nnpdf` wheel?

[...]

PR #1487, opened 2021-12-03
prioritize conda-forge over defaults channel

RoyStegeman:
As discussed in #1284. 

Relevant PR in binary-bootstrap: NNPDF/binary-bootstrap#4


Zaharid:
I think the failure on mac has to do with numpy, given all the warnings we see. No idea why though.

scarlehoff:
Hi @RoyStegeman, rebase master because something funny happened (my fault, modified something thinking it was just an example).

What I don’t understand is why does it stop 3.9 from running…

Zaharid:
If I look at the current build failure for mac, I notice that we are pullinkg everything from conda-forge except for

```
    scipy:                         1.7.1-py38hf27351e_2                defaults                             
```

which seems to be the problem (no idea why that is happening).

Also at build time

```
    libarchive:      3.4.2-haa3ed63_0          defaults                             
```


Incidentally I wonder if disabling defaults alrogether would make thing simpler.

RoyStegeman:
> Incidentally I wonder if disabling defaults alrogether would make thing simpler.

Maybe, though I'm not sure if that's what we want to do in binary bootstrap and thus peoples local .condarc. Let me try what happens if we enforce a strict channel priority

scarlehoff:
If I remember correctly in order to upload packages to conda forge you need to set strict channel priority so if it works it's a plus.

PR #1488, opened 2021-12-03
Fix example runcard

scarlehoff:
This runcard in the `closuretest` folder currently doesn’t work. I’ve tried it changing it in master (because it looked trivial and that it should not affect anything) but the tests failed so I’m moving the changes to a PR to investigate…

### DIFF
--- a/validphys2/src/validphys/closuretest/runcards/NNPDF40_closure_test.yml
+++ b/validphys2/src/validphys/closuretest/runcards/NNPDF40_closure_test.yml
@@ -99,6 +99,9 @@ trvlseed: 1727532335
 nnseed: 1737785873
 mcseed: 2123509817
 genrep: true          # true = generate MC replicas, false = use real data
+# Legacy options required by closure tests
+rngalg: 0
+seed: 4
 
 parameters: # This defines the parameter dictionary that is passed to the Model Trainer
   nodes_per_layer: [18, 24, 36, 8]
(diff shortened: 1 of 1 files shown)


scarlehoff:
Is there something going on with Github workflows? Why is it not launching the 3.9 tests?

validphys2/src/validphys/closuretest/runcards/NNPDF40_closure_test.yml
```diff
@@ -99,6 +99,9 @@ trvlseed: 1727532335
 nnseed: 1737785873
 mcseed: 2123509817
 genrep: true          # true = generate MC replicas, false = use real data
+# Legacy options required by closure tests
```
scarlehoff, line 102:
```suggestion
# Deprecated options still required by closure tests
```

scarlehoff:
Ok. The problems I was having were because of github. Good.

scarlehoff:
I'm wondering whether to merge this to master. It is just a change to a runcard and it might be that the errors where due to the workflow problem of the other day. Not sure it needs a review.

(and if it fails in master I can always revert the merge)

Just in case, the error happened when building the docs because the download of the package from the server failed so the fact that the test pass here don't mean anything but I don't think the error was related to these changes.

This was the error:
https://github.com/NNPDF/nnpdf/runs/4409716514?check_suite_focus=true#step:9:852

But as can be seen here https://packages.nnpdf.science/public/linux-64/repodata.json

The sha256 is perfectly correct:

```
    "nnpdf-4.0.4.106+gc6fabc187-py39h9a426be_0.tar.bz2": {
      "build": "py39h9a426be_0",
      "build_number": 0,
      "md5": "537ba86f62b1ddadc31b557966efe474",
      "name": "nnpdf",
      "sha256": "414673db7cba0432915c3b7c5d20e5292d7a1743e492eb3659e526c23e0c19c0",
      "size": 127203204,
      "subdir": "linux-64",
      "timestamp": 1638541045211,
      "version": "4.0.4.106+gc6fabc187"
    },
```

It is likely that the download simply didn't work.

Zaharid (approved):

Issue #1486, opened 2021-12-03
Is `internal_multiclosure_dataset_loader` used?


scarlehoff:
This is one of the functions which are problematic in #1475, I'm trying to find some test or runcard where this function is used in order to remove the call to libNNPDF but cannot find it.

I've tried rerunning some reports tagged as `multiclosure` in the server with no luck (for instance https://vp.nnpdf.science/ebphG4sgSEG2hVt2c5A41g==/)

If anyone has a runcard for it I'd appreciate if it is shared. Otherwise I guess we can just remove (or truly deprecate, with the import of NNPDF and the creation of the libNNPDF PDF inside of the function) that family of functions?

tgiani:
might be related to the closure PRs MIchael opened I while ago. I ll try to look at this next week

tgiani:
so one example where this function is used is in this report produced by Michael some time ago https://vp.nnpdf.science/qSAUPISAQ7qQeh0nL8hIOg==/. The corresponding runcard uses the function ```plot_data_central_diff_histogram```, which requires the function ```data_replica_and_central_diff``` which requires ```internal_multiclosure_data_loader```. These functionalities are implemented in the PR #1321

tgiani:
and I think that ```internal_multiclosure_dataset_loader``` is used basically for all the functions implemented in ```multiclosure.py```, so it is required to compute all the closure test estimators

scarlehoff:
But is the stuff in `multiclosure` used?

tgiani:
it is used for everything concerning closure test estimators. Including results reported in the NNPDF4.0 paper I guess

tgiani:
so I don' t think we want to remove these functions

scarlehoff:
Strange. I tried running a few runcards and that function didn't seem to be used. Maybe I did something wrong, I can try again.

Or better yet, do you see an easy way to avoid the libNNPDF call there? (I.e., the call to pdf load)

tgiani:
uhm maybe I m the one doing smt wrong. if you try for example to run this runcard in the master 
```
it: 210326-mw-001

theoryid: 200
use_cuts: internal
q2min: 3.49                        # Q2 minimum
w2min: 12.5                        # W2 minimum

dataset_inputs:
# ATLAS
- {dataset: DYE906R_dw_ite, frac: 0.75, cfac: [ACC,QCD]}
- { dataset: ATLASWZRAP11CF, frac: 0.75, cfac: [QCD] } # N

[...]

PR #1489, opened 2021-12-09
Use python covmat in multiclosure tests

scarlehoff:
This closes issue #1486 while dealing with another of the not-trivial cases of #1475. And the last preventing me from working on #1475 because I've decided not to touch the filter at this stage.

This "fix" is very tricky because there are no examples or tests where I can check what the intended behavior is exactly. Thanks to @tgiani I have at least a runcard that uses the function:
```
fit: 210326-mw-001

theoryid: 200
use_cuts: internal
q2min: 3.49                        # Q2 minimum
w2min: 12.5                        # W2 minimum

dataset_inputs:
# ATLAS
- {dataset: DYE906R_dw_ite, frac: 0.75, cfac: [ACC,QCD]}
- { dataset: ATLASWZRAP11CF, frac: 0.75, cfac: [QCD] } # N

meta:
 author: TG
 keywords: [test]
 title: Test plot_data_fits_bias_variance

diag_ns:
 - diagonal_basis: True
 - diagonal_basis: False

template_text: |
 # Results with 25 fits

 ## Central diff histogram
 {@with diag_ns@}
 # Diagonal: {@diagonal_basis@}
 {@plot_data_fits_bias_variance@}
 {@endwith@}

use_t0: true

fits:
- 210326-mw-001
- 210326-mw-002

actions_:
 - report(main=True)
```

Honestly, I would've like not to have modified `config.py` or force people to have to use `use_t0` when before it wasn't the case, however due to https://github.com/NNPDF/nnpdf/blob/b44ae914ef494049f9456c8c0fdd267e8386446e/validphys2/src/validphys/closuretest/multiclosure.py#L49 just calling `dataset_inputs_t0_covmat_from_systematics` is non-trivial, doable, but complicates quite a bit the logic (unless there's some shortcut I haven't found).

### Fix
Just using `dataset_inputs_t0_covmat_from_systematics`. I've checked that for the above runcard the results are completely equivalent. This also simplifies the function since it no longer have to take care of doing the loading of the data and setting the t0 set etc (which included a copy not sure why)

### Backwards compatibility
Since this breaks previously working runcards (in that now the user has to add one extra flag) and in that the usage of the t0 covmat is not happening in the dark in some function but done by validphys I've added two checks:
- `use_t0` just telling the user the runcard should be updated
- Ensure that the user doesn't mistakenly set a `t0set` different from the `multiclosure_law`.

### Tests and examples
It would be nice if someone could add tests to this module which is currently lacking. I honestly don't know what to test other than it runs. I'll add the runcard @tgiani gave me but for the tests someone else should do another PR that takes care of this because I'm mostly guessing how things should work so I don't think I'm qualified to write a test specifying _how really_ things should work.

Anyway this is ready for review.

### DIFF
--- a/validphys2/src/validphys/checks.py
+++ b/validphys2/src/validphys/checks.py
@@ -22,6 +22,12 @@ import logging
 log = logging.getLogger(__name__)
 
 @make_check
+def check_use_t0(ns, **kwargs):
+    """Checks use_t0 is set to true"""
+    if not ns.get("use_t0"):
+        raise CheckError("The flag 'use_t0' needs to be set to 'true' for this action.")
+
+@make_check
 def check_pdf_is_montecarlo(ns, **kwargs):
     pdf = ns['pdf']
     etype = pdf.ErrorType
--- a/validphys2/src/validphys/closuretest/closure_checks.py
+++ b/validphys2/src/validphys/closuretest/closure_checks.py
@@ -77,6 +77,14 @@ def check_t0pdfset_matches_law(t0pdfset, fit):
             f"Underlying pdf: {t0_from_fit}, does not match t0pdfset: {t0pdfset}"
         )
 
+@make_argcheck
+def check_t0pdfset_matches_multiclosure_law(multiclosure_underlyinglaw, t0set):
+    """Checks that, if a multiclosure_underlyinglaw is present, it matches the t0set
+    Checks t0set instead of t0pdfset since different mechanisms can fill t0set
+    """
+    if str(t0set) != str(multiclosure_underlyinglaw):
+        raise CheckError(f"The underlying pdf {multiclosure_underlyinglaw} does not match t0pdfset: {t0set}")
+
 
 @make_argcheck
 def check_at_least_10_fits(fits):
--- a/validphys2/src/validphys/closuretest/closure_checks.py
+++ b/validphys2/src/validphys/closuretest/closure_checks.py
@@ -4,9 +4,12 @@ closuretest/checks.py
 Module containing checks specific to the closure tests.
 
 """
+import logging
 from collections import defaultdict
 
 from reportengine.checks import make_argcheck, CheckError
+
+log = logging.getLogger(__name__)
 
 
 @make_argcheck
@@ -83,7 +86,7 @@ def check_t0pdfset_matches_multiclosure_law(multiclosure_underlyinglaw, t0set):
     Checks t0set instead of t0pdfset since different mechanisms can fill t0set
     """
     if str(t0set) != str(multiclosure_underlyinglaw):
-        raise CheckError(f"The underlying pdf {multiclosure_underlyinglaw} does not match t0pdfset: {t0set}")
+        log.warning(f"The underlying pdf {multiclosure_underlyinglaw} does not match t0pdfset: {t0set}")
 
 
 @make_argcheck
--- a/validphys2/src/validphys/closuretest/multiclosure.py
+++ b/validphys2/src/validphys/closuretest/multiclosure.py
@@ -34,7 +34,7 @@ SAMPLING_INTERVAL = 5
 @check_fits_areclosures
 @check_fits_different_filterseed
 def internal_multiclosure_dataset_loader(
-    dataset, fits_pdf, multiclosure_underlyinglaw, fits
+    dataset, fits_pdf, multiclosure_underlyinglaw, fits, dataset_inputs_t0_covmat_from_systematics
 ):
     """Internal function for loading multiple theory predictions for a given
     dataset and a single covariance matrix using underlying law as t0 PDF,
@@ -77,12 +77,6 @@ def internal_multiclosure_dataset_loader(
     typically used in these studies.
 
     """
-    if isinstance(dataset, DataSetSpec):
-        data = dataset.load()  # just use internal loader
-    else:
-        # don't cache result
-        data = dataset.load.__wrapped__(dataset)
-
     fits_dataset_predictions = [
         ThPredictionsResult.from_convolution(pdf, dataset)
         for pdf in fits_pdf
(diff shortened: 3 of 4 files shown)


tgiani:
looks nice, I ll have a look asap

validphys2/src/validphys/closuretest/closure_checks.py
```diff
@@ -77,6 +77,14 @@ def check_t0pdfset_matches_law(t0pdfset, fit):
             f"Underlying pdf: {t0_from_fit}, does not match t0pdfset: {t0pdfset}"
         )
 
+@make_argcheck
+def check_t0pdfset_matches_multiclosure_law(multiclosure_underlyinglaw, t0set):
+    """Checks that, if a multiclosure_underlyinglaw is present, it matches the t0set
+    Checks t0set instead of t0pdfset since different mechanisms can fill t0set
+    """
+    if str(t0set) != str(multiclosure_underlyinglaw):
```
Zaharid, line 85: Do we need the string cast here?

scarlehoff: Will always be a string?
In any case, sice it (could) be a string or a PDF which knows how to `__str__` I think it doesn't do harm (maybe it does but I didn't realised)

tgiani:
it looks good to me

Zaharid:
@tgiani  this is blocking a lot of stuff so a quick review here would be helpful. We were also wondering:

  - Is there a case for multiclosure_underlyinglaw . I seem to recall this was added for some reason but can't remember it or know if it is relevant.
  - how prevalent was the use case of needing `multiclosure_underlyinglaw` at all seeing as it is constrained to be `t0` and also how bad would it be to remove it altogether (i.e. would it break some runcards that were directly useful for e.g. the closure test paper)?

tgiani:
@Zaharid @scarlehoff this PR looks good to me. Just one thing, which is connected to what @zahari was asking above: why do we need the check ```check_t0pdfset_matches_multiclosure_law``` ? I think that for all practical purposes the ```t0``` is indeed always set equal to be the underlying law, and closure tests are not iterated. But in principle one might set as ```t0``` a different pdf set, and check that after few iterations it indeed converges to the underlying law (and to the fit result). Does it make sense?

tgiani:
to answer to the two points @Zaharid asked above: I m not sure if ```multiclosure_underlyinglaw``` is ever used, this was part of the work Michael did. It looks to me that there are no cases where the initial ```t0``` is set to be different from the underying law, but I m not 100% sure. In general I wouldn t remove this, and I would change the error raised by the check ```check_t0pdfset_matches_multiclosure_law``` to a warning

scarlehoff:
It makes sense, the problem is that the underlying law was being used implicitly as the t0. I'm fine with "disconnecting" the two (and then the user has to set the t0 manually and if it sets the wrong one that's on them).

Should I make that change? The runcards that use that action won't work anymore because the t0set will be missing.

tgiani:
uhm when  you say that this would break the existing runcard do ypou mean that the check ```check_t0pdfset_matches_law``` would fail?

scarlehoff:
No, sorry. I meant the actions for multiclosure. They were using implicitly the underlying law to generate the t0 covmat, I just made it explicit.

When I say "won't work" I mean you will need to set the t0 explicitly in the runcard (which as you say will be the same one most of the time). Right now this is not the case (actually, if you were it was being ignored for that specific action).

tgiani:
ah okok sorry I misunderstood. Fine by me

scarlehoff:
I just did as suggested, changed the check to a warning so if someone really wants they can use a different t0. Also now the t0 needs to be explicitly set by adding the `t0pdfset`

scarlehoff:
@tgiani could you check when you have time if you're happy with the changes (and approve the PR if so), thanks!

tgiani (approved):

validphys2/src/validphys/checks.py
```diff
@@ -21,6 +21,12 @@
 import logging
 log = logging.getLogger(__name__)
 
+@make_check
+def check_use_t0(ns, **kwargs):
+    """Checks use_t0 is set to true"""
+    if not ns.get("use_t0"):
+        raise CheckError("The flag 'use_t0' needs to be set to 'true' for this action.")
+
```
RoyStegeman, line 29: Is this the correct behavior? It checks for this `ns` object which I think is just an object of the Config class that is constructed manually as opposed to taken directly in from the runcard (meaning setting `use_t0` flag in the runcard as suggest in the comments does not work).

RoyStegeman: @scarlehoff

scarlehoff: isn't `use_t0` one of the runcard flags? (where are you getting the error to understand the problem?)

RoyStegeman: That's what I understood when reading the comments in this PR. But `ns` does not contain all runcard keys. 

A minimal example to get the error is below:

runcard:
```
use_t0: true

actions_:
  - foo
```

with vp action: 
```
@check_use_t0
def foo():
    pass
```

scarlehoff: I think `check_use_t0` is only supposed to be used in functions that take `use_t0` as input (so with `foo()` it will never find it, but with `foo(use_t0)` will only accept to run `foo` with `use_t0=True`.

Maybe since we added `use_t0_in_fitting` and `_in sampling` (or similar) this is creating a problem somewhere?

scarlehoff: What function in particular is being problematic? `use_t0_in_fitting/sampling` should only be used during the fit and have no effect if you are not fitting.

RoyStegeman: I want to run a closure test but can't. `internal_multiclosure_data_loader` gives the error. 

They shouldn't affect anything, but perhaps something was touched that broke this down the line. On the other hand, CT worked in Michaels branch, but I never used it after it had been merged so maybe it never worked on the main branch...

PR #1492, opened 2021-12-17
Alphas theories: alphas=0.114,0.115,0.121,0.122 and PDF4LHC21 alphas=0.117,0.119

enocera:
This PR modifies the `theory.db` and includes the additional theories detailed in the title. The theories are available from the server.


enocera:
Theories are as follows:
ID: 233; alphas=0.114
ID: 234; alphas=0.115
ID: 235; alphas=0.121
ID: 236; alphas=0.122

siranipour:
Excellent, many thanks for generating these so quickly Emanuele. Could you kindly update the table https://docs.nnpdf.science/theory/theoryindex.html?highlight=theory%20table?

Also, would it be possible to have these new theories be labelled such that they are adjacent to theories 200-209? Otherwise the linked table will become a bit disjointed over time.

enocera:
> Excellent, many thanks for generating these so quickly Emanuele. Could you kindly update the table https://docs.nnpdf.science/theory/theoryindex.html?highlight=theory%20table?

Sure.

> Also, would it be possible to have these new theories be labelled such that they are adjacent to theories 200-209? Otherwise the linked table will become a bit disjointed over time.

That's true. However I'm afraid that this will cause some trouble, as we didn't leave any free IDs adjacent to 200-209. And I would definitely avoid to recast IDs, as this will cause more trouble. My opinion is that we will have to live with a slightly dijointed table.

enocera:
> Excellent, many thanks for generating these so quickly Emanuele. Could you kindly update the table https://docs.nnpdf.science/theory/theoryindex.html?highlight=theory%20table?

@siranipour My understanding is that, as long as the `theory.db` file is updated (which is the case in this PR), the table is automatically generated from it, see https://raw.githubusercontent.com/NNPDF/nnpdf/master/doc/sphinx/source/theory/theoryindex.rst

siranipour:
Ah yes very cool! The PDF4LHC theories are ID 237 and 238 right?

Also I agree that reindexing theories is a bad idea, so let's not do that.

enocera:
> Ah yes very cool! The PDF4LHC theories are ID 237 and 238 right?

Yes, they are.

enocera:
I was too lazy to have two separate PRs.

siranipour (approved):

PR #1494, opened 2021-12-28
do what genrep says in makereplica

scarlehoff:
I've noticed `make_replica` is not honoring `genrep` (which was not an argument to `make_replica` and so the replicas were always generated). This PR fixes that.

Code that reproduce the problem:

```python
from validphys.api import API
import numpy as np
t0 = API.make_replica(
    dataset_inputs=[{"dataset":"NMC"}, {"dataset": "NMCPD"}],
    use_cuts="nocuts",
    theoryid=53,
    replica=1,
    mcseed=123,
    genrep=False,
    )
tnot = API.make_replica(
    dataset_inputs=[{"dataset":"NMC"}, {"dataset": "NMCPD"}],
    use_cuts="nocuts",
    theoryid=53,
    replica=1,
    mcseed=321,
    genrep=False,
    )
np.allclose(t0, tnot)
```
False

What was happening before if `genrep` is used is that `numpy` is seeded with `None` and then a random non-reproducible replica is generated, so basically worsts case scenario. After this PR this code returns `True` (since no replica is generated).

### DIFF
--- a/validphys2/src/validphys/pseudodata.py
+++ b/validphys2/src/validphys/pseudodata.py
@@ -101,7 +101,7 @@ def read_replica_pseudodata(fit, context_index, replica):
     return DataTrValSpec(pseudodata.drop("type", axis=1), tr.index, val.index)
 
 
-def make_replica(groups_dataset_inputs_loaded_cd_with_cuts, replica_mcseed):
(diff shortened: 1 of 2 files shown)


siranipour:
Ah thanks! Can you add a small test?

scarlehoff:
Sure. I'll add exactly what I have in the PR text with some of the test datasets.

siranipour (approved):

PR #1495, opened 2022-01-05
Add an overview of loss functions to the documentation

### LINKED ISSUE #1478: Add tutorial on computing chi²
The way we use and define objective functions and chi² related metrics has many knobs, which makes it difficult to use and interpret correctly without a lot of context. It would be good to have a tutorial that lists the possible options, their applications and the relevant vp flags.

These are at least:

  - t0
  - use_pdferror
  - regularization
  - weights

RoyStegeman:
This addresses #1478.

This adds an overview of the loss functions we use in NNPDF to the documentation. The page itself does not go into much detail but I think it links to the relevant resources so it can be used as a reference in case someone is confused about the definition of the chi squared used in various places. 

@Zaharid is this roughly what you had in mind?

Closes  #1478


doc/sphinx/source/figuresofmerit/index.rst
```diff
@@ -0,0 +1,184 @@
...
+the :math:`t_0` method (discussed below) applies.
+
+Here we we provide an overview of the different figures of merit, and discuss
+when each of them is used.
+
+
+The basis of the loss functions: 𝜒²
+--------------------------------------------------------------------------------
+The :math:`\chi^2` figures of merit used in the NNPDF methodology are all
+based on the chi square distribution:
```
Zaharid, line 18:
```suggestion
based on the chi square statistic:
```

doc/sphinx/source/figuresofmerit/index.rst
```diff
@@ -0,0 +1,184 @@
...
+.. note::
+    This definition of :math:`\chi^2` is not used as a figure of merit
+    anywhere in the NNDPF methodology. Instead, variations of this :math:`\chi^2`
+    are used.
+
+
+Avoiding bias: t₀ method
+~~~~~~~~~~~~~~~~~~~~~~~~
+The :math:`t_0` method introduced in
+`arXiv:0912.2276 <https://arxiv.org/abs/0912.2276>`_ aims to
```
Zaharid, line 42: Please add this to the bibliography.

doc/sphinx/source/figuresofmerit/index.rst
```diff
@@ -0,0 +1,184 @@
...
+
+
+Missing higher order uncertainties
+--------------------------------------------------------------------------------
+Another source of uncertainties that we may want to include in the covariance
+matrix are theoretical uncertainties, particularly missing higher order
+uncertainties estimated through scale variations. These unceratinties can be
+considered in the figure of merit through the implementation of a 'theory
+covariance matrix'. A paper discussing the formalism can be found here:
+`arXiv:1905.04311 <https://arxiv.org/abs/1905.04311>`_. For a tutorial see
```
Zaharid, line 61: Same, please use the citations instead.

Zaharid (commented):
Looks good. Would like to amend the introduction a bit and perhaps think on one or two more substantive comments.

RoyStegeman:
> Looks good. Would like to amend the introduction a bit and perhaps think on one or two more substantive comments.

What comments do you think are missing?

doc/sphinx/source/figuresofmerit/index.rst
```diff
@@ -0,0 +1,184 @@
...
+model, and :math:`C_{ij}` is the covariance matrix between datapoints :math:`i`
+and :math:`j`.
+
+The covariance matrix accounts for correlated systematic uncertainties,
+normalization uncertainties, and statistical uncertainties as provided by the
+experimental collaborations.
+
+.. note::
+    This definition of :math:`\chi^2` is not used as a figure of merit
+    anywhere in the NNDPF methodology. Instead, variations of this :math:`\chi^2`
```
Zaharid, line 35:
```suggestion
    anywhere in NNDPF fits. Instead, variations discussed below
```

doc/sphinx/source/figuresofmerit/index.rst
```diff
@@ -0,0 +1,184 @@
...
+The basis of the loss functions: 𝜒²
+--------------------------------------------------------------------------------
+The :math:`\chi^2` figures of merit used in the NNPDF methodology are all
+based on the chi square statistic:
+
+.. math::
+    \chi^{2}=\sum_{i, j}^{N_{\text {dat }}}(D-P)_{i} C_{i j}^{-1}(D-P)_{j},
+
+where :math:`D_i` is the :math:`i`-th datapoint, :math:`P_i` is the prediction
+of the corresponding datapoint calculated by performing the convolution product
```
Zaharid, line 24: This can be a bit more ambiguous to also be correct for compound observables. Although we might want to delete it altogether.

```suggestion
of the corresponding datapoint calculated from the convolution product
```

doc/sphinx/source/figuresofmerit/index.rst
```diff
@@ -0,0 +1,184 @@
...
+understand the differences between the various figures of merit, and to
+understand which definition we are referring to in a given context. In
+particular, it is worth stressing that whenever a figure of merit is discussed,
+the :math:`t_0` method (discussed below) applies.
+
+Here we we provide an overview of the different figures of merit, and discuss
+when each of them is used.
+
+
+The basis of the loss functions: 𝜒²
```
Zaharid, line 15: I feel that here it is somewhat ambiguous that we are discussing the so called "experimental" chi2, and seems to me the word "experimental" should appear in a few places.

doc/sphinx/source/figuresofmerit/index.rst
```diff
@@ -0,0 +1,184 @@
...
+The :math:`\chi^2` figures of merit used in the NNPDF methodology are all
+based on the chi square statistic:
+
+.. math::
+    \chi^{2}=\sum_{i, j}^{N_{\text {dat }}}(D-P)_{i} C_{i j}^{-1}(D-P)_{j},
+
+where :math:`D_i` is the :math:`i`-th datapoint, :math:`P_i` is the prediction
+of the corresponding datapoint calculated by performing the convolution product
+between the :ref:`FastKernel tables<fktables>` for point :math:`i` and the PDF
+model, and :math:`C_{ij}` is the covariance matrix between datapoints :math:`i`
```
Zaharid, line 26:
```suggestion
model, and :math:`C_{ij}` is the covariance between datapoints :math:`i`
```

doc/sphinx/source/figuresofmerit/index.rst
RoyStegeman: So this is something that confused me a bit about the discussion for the referee reply, perhaps you can clear this up. What we call the "experimental" chi2 in n3fit uses the same t0 covmats as the validation and training losses, however based on the mail exchange this week it seems that the "experimental" chi2 doesn't use the t0 prescription?

siranipour: Perhaps I can chime in here. For all our fits we use the covariance matrix constructed under the t0 prescription. That is: multiplicative uncertainties are multiplied by predictions from the t0 PDF (as opposed to the experimental measurement) to transform them to additive uncertainties before constructive the covariance matrix.

The reason we do this has to do with the d'agostini bias: if we attempt to use the non-t0 covmat (which is the experimental covmat) to compute the loss, then our results will be a biased estimator.

The discussion for the reply to the referee pertains to what covmat we use to compute the chi2 tables in the reports. The experimental chi2 uses the experimental covmat; the t0 chi2 similarly uses the t0 covmat. Really we should look at the t0 chi2 because that is what the replicas are optimising for.

RoyStegeman: Thanks, but this is indeed my point: what you now call the "t0 chi2" is simply called "experimental chi2" (or sometimes "true chi2" iirc) inside n3fit. 

For external communications I suppose that's fine, as long as we're consistent in distinguishing t0 chi2 from experimental chi2, but in the docs this double use of the name can cause problems/confusion.

PR #1499, opened 2022-01-13
Access the PDF info class explicitly

scarlehoff:
One of the changes proposed in #1475.

This is not strictly necessarily for #destroyingc++ but it makes my life easier. Realistically the `PDF` only needs to look into the `.info` file for the error type and eventually the confidence levels of the hessian errors. Now this is done through explicit attributes. The `.info` file is still exposed through an `info` attribute, which among other things allows for the user to look at what's included instead of guessing and hoping for `__get_attr__` to find it.

Since I was effectively rewriting the `PDF` class I've also fixed a `TODO` which asked for the usage of `Path` for `infopath` and I've clean some redundant checks (for instance, `rescale_factor` is only called in one specific situation so no point on checking inside the same reason why it was being called in the first place.

The only quirk here is the fact that some hessian PDFs do not include a confidence level. Is that because it is assumed to be 68%? (@Zaharid could you comment?) I've left it such that it behaves exactly as the previous one in that case but the code could be simplified even more if this is confirmed.

Some notes on the PR itself: all the changes are to the `core.py` (and `test_core.py`) files. All the other files changed have been a sed command for the following substitutions: (`ErrorType` -> `error_type`, `QMin` -> `q_min`) and the removal of redundant `pathlib` calls.

### DIFF
--- a/validphys2/src/validphys/core.py
+++ b/validphys2/src/validphys/core.py
@@ -15,6 +15,7 @@ import functools
 import inspect
 import json
 import logging
+from pathlib import Path
 
 import numpy as np
 
@@ -91,24 +92,18 @@ class _PDFSETS():
 PDFSETS = _PDFSETS()
 
 class PDF(TupleComp):
+    """Base validphys PDF providing high level access to metadata.
+
+    Statistical estimators which depends on the PDF type (MC, Hessian...)
+    are exposed as a :py:class:`Stats` object through the :py:meth:`stats_class` attribute
+    The LHAPDF metadata can directly be accessed through the :py:meth:`info` attribute
+    """
 
     def __init__(self, name):
         self.name = name
         self._plotname = name
+        self._info = None
         super().__init__(name)
-
-
-    def __getattr__(self, attr):
-        #We don't even try to get reserved attributes from the info file
-        if attr.startswith('__'):
-            raise AttributeError(attr)
-        try:
-            return lhaindex.parse_info(self.name)[attr]
-        except KeyError as e:
-            raise AttributeError("'%r' has no attribute '%s'" % (type(self),
-                                                                 attr)) from e
-        except IOError as e:
-            raise PDFDoesNotExist(self.name) from e
 
     @property
     def label(self):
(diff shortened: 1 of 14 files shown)


validphys2/src/validphys/core.py
```diff
@@ -167,48 +191,36 @@ def load_t0(self):
...
-                                              "with confidence"
-                                              " interval %s" % (cl,) )
             else:
+                raise NotImplementedError(f"No hessian errors with confidence interval {cl}")
+        if error == "symmhessian":
+            if cl == 68:
                 return LHAPDFSet.erType_ER_SYMEIG
+            else:
+                raise NotImplementedError(
+                    "No symmetric hessian errors " f"with confidence interval {cl}"
```
Zaharid, line 220:
```suggestion
                    f"No symmetric hessian errors with confidence interval {cl}"
```

Zaharid:
> The only quirk here is the fact that some hessian PDFs do not include a confidence level. Is that because it is assumed to be 68%? (@Zaharid could you comment?) I've left it such that it behaves exactly as the previous one in that case but the code could be simplified even more if this is confirmed.

I think so, yes.

validphys2/src/validphys/core.py
```diff
@@ -121,42 +116,71 @@ def label(self, label):
...
+                    "make sense with 'replicas' error type"
+                )
+            return None
+        return self.info.get("ErrorConfLevel", 68)
 
     @property
     def isinstalled(self):
         try:
-            self.infopath
+            return self.infopath.is_file()
```
Zaharid, line 171: This looks a bit strange now. For one it is maybe raising `PDFDoesNotExist` elsewhere.

scarlehoff: How? It does not change with respect to what it was doing before, right?

scarlehoff:
> I've left it such that it behaves exactly as the previous one in that case but the code could be simplified even more if this is confirmed.

It turns out in order to simplify it I needed to update the regression tests because `scipy.stats.norm.isf((1 - 0.01 * confidence_level) / 2)` = 0.994457883209753 != 1.

validphys2/src/validphys/core.py
Zaharid: You are right.

Zaharid (approved):

Issue #1436, opened 2021-10-15
Create test implementations of new commondata format


Zaharid:
To make things in #1416 more concrete we agreed to try and produce test implementations of the (still fuzzy) current version of the idea for a couple of datasets.

The datasets should be one DIS and one hadronic and one of them should involve multiple (COMPOUND) fktables.

These should have:

  - A metadata file with the informations in the "runcards" folder, PLOTTING file, COMPOUND file and others.
  - A kinematics file with bin edges and central value in some format.
  - A data file with experimental central value and uncertainties (probably in a separate file).

The uncertainties should support a way of specifying correlated systematics across datasets. Probably also some way to deal with additive vs multiplicative.

Regarding formats, I would think  for this a text format of some sort is more adequate (easier to read and exchange around) and also one that trivially round trips with pandas (ie better than csv), so maybe something like json.

enocera:
Let me see whether I can make any of the above a little more concrete. For the time being, let's focus on the dataset currently called `NMCPD`. It corresponds to the F2d/F2p ratio measured by NMC. This is the only COMPOUND DIS dataset.
I understand that we want the following files.
1. >     * A metadata file with the informations in the "runcards" folder, PLOTTING file, COMPOUND file and others.
My understanding is that this is the first file that a developer would write when implementing a new dataset (or migrating an old dataset to the new format). I think that it should be as simple (minimal) as possible. Let's call it `METADATA_NMCPD.json`. A possible incarnation may look like as follows.
```
{
    "setname":  "NMPCPD",
    "arXiv":    "hep-ex/9611022",
    "hepdata":  "https://www.hepdata.net/record/ins426595",
    "ndata":    260,
    "nsys_exp": 5, 
    "proctype": "DIS_F2R",
    "compound": "RATIO",
    "experiment": "NMC",
    "nnpdf31_process": "DIS NC"
}
```
Here I already have some questions:
- how are `COMPOUND` datasets treated by the new tool chain that produces FK tables?
- which kind of information is actually required from the `PLOTTING` file?
- which information should we import from the `metadata.txt` in the "runcard" folder?

2. >     * A kinematics file with bin edges and central value in some format.
3. >     * A data file with experimental central value and uncertainties (probably in a separate file).
[...]

PR #1500, opened 2022-01-19
[WIP] Test implementation of new commondata format

enocera:
This PR addresses #1436.
The proposed layout is as follows. For each data set, there is a folder called `DATASET_NAME`. Each folder contains:
- a `metadata.yaml` file;
- a `filter.py` file.

The `filter.py` script reads the `metadata.yaml` file and generates no less than three `.json` files: 
- `data.json`, for the central values of the data;
- `kinematics.json` for the kinematics of the data;
- `uncertainties.json` for the uncertainties of the data.

There can be as many variants as needed for each of these three files. Variants are specified in the `metdata.yaml`.

TO DO
- [x] include the missing `filter.py`
- [x] have a look at the layout and at its prototypical realisation;
- [x] write a validphys reader that takes as input the `.json` files in lieu of the 'DATA` and `SYSTYPE` files. (#1511)


scarlehoff:
One thing missing from the list (I only realised now that I started with the parser) would be an automatic installation.

I guess to first approximation these should be part of the validphys package, right? Since the fit needs validphys anyway.

(it's not important for the parser, I can hardcode the folder for now)

scarlehoff:
@enocera 
One question, is it the idea that the variants are now selected in a way similar to how we select the kfactors? For instance one can have as input:

```yaml
dataset_input:
  - dataset: "NMCPD"
  - variant: "dw"
```

and if no variant is selected it defauls to the standard one.

enocera:
@scarlehoff Do you mean in a vp runcard?

scarlehoff:
Yes. Instead of having to write NMCPD_dw

enocera:
Then yes, I think that we want to have something similar in spirit to, but more general than, the current `sys` key. cc @Zaharid

enocera:
Or, as you say, the `cfac` key.

scarlehoff:
Great, I like this idea. I'll implement it in this spirit. The idea is then that "sys" will disappear?

enocera:
Yes, definitely. The problem with `sys` is that it only controls the way in which systematic uncertainties are correlated or not. The `variant` is something more general (and cleaner) which is specified by a set of `data.json` and `uncertainty.json` files. This means that you can have different central values and different uncertainties (correlated in whatever way) for each dataset (without the need of renaming the dataset). In my opinion, this is useful to have nuclear variants, but also, e.g., EW variants (and of course also more manageable debugged variants).

scarlehoff:
Do we want to have a `CommonDataSpec` and a `CommonData` as two separate classes or can we unify them as `CommonDataSpec` that will have all the attributes of both? This one is for @Zaharid I guess

scarlehoff:
And for @enocera, this is a perfect and total substitution, right? So I can just scrap all the "old" commondata stuff as I find it in the code? No need to keep legacy wrappers like I have to do for the fktables.

enocera:
> And for @enocera, this is a perfect and total substitution, right? So I can just scrap all the "old" commondata stuff as I find it in the code? No need to keep legacy wrappers like I have to do for the fktables.

I'd answer yes because the idea (my idea) is to rework all the datasets in the new format.

buildmaster/NMCPD/kinematics.json
```diff
@@ -0,0 +1 @@
....36,14.85,19.79,26.49,35.4,47.03,63.53,99.03]},{"y":[0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0]}]"kinematics_min":[{"x":null},{"Q2":null},{"y":null}]"kinematics_max":[{"x":null},{"Q2":null},{"y":null}]}
```
scarlehoff, line 1: Are these `null` expected? What should be returned if an user asks for the "minimums array", just the same thing as the avg?

buildmaster/NMCPD/metadata.yaml
```diff
@@ -0,0 +1,60 @@
...
+setname: "NMPCPD"
+experiment: "NMC"
+#ndata: 260 #not needed because it follows from data_central?
+#nsys_exp: 5 #not needed because it follows from data_uncertainties?
+
+#References
+arXiv: "hep-ex/9611022"
+iNSPIRE: "https://inspirehep.net/literature/424154"
+hepdata: "https://www.hepdata.net/record/ins426595"
+   tables:
```
scarlehoff, line 11:
```suggestion
tables:
```

Zaharid:
> Do we want to have a `CommonDataSpec` and a `CommonData` as two separate classes or can we unify them as `CommonDataSpec` that will have all the attributes of both? This one is for @Zaharid I guess

I like the separation between header and object and also would like the parsing stage of vp to be as fast as possible, so rather prefer the arrangement where we have two classes.

buildmaster/NMCPD/kinematics.json
enocera: I placed a `null` because this is a case in which bin edges are not provided by the experimentalists; I'd say that if `null`, then min=avg.

buildmaster/NMCPD/metadata.yaml
```diff
@@ -0,0 +1,60 @@
...
+y_label: "$F_2^d/F_2^p(x,Q^2)$" 
+kinematics_override: dis_sqrt_scale
+figure_by:
+  - k1
+
+#Variants
+variants:
+  - nuclear_uncertainty_choice:
+       deweighted: uncertainties_dw.json
+       shifted: uncertainties_dw.json
```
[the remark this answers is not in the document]
Zaharid: Can variants change things other than the uncertainties?

Zaharid:
The one class we need to have is the one representing *precisely whatever the user wrote in the runcard*. As @RoyStegeman can tell, not having that e.g. for positivity is a source of of pain.

scarlehoff:
That has nothing to do with this, the kinematics or uncertainties are not written in the runcard but are part of the metadata of the commondata and that needs to be held by the commondata.

buildmaster/NMCPD/kinematics.json
scarlehoff: I see, thanks.
I also I think there might have a bug in the script that outputs the json because the `kinematics_max` and `kinematics_min` are missing a comma in front.

Also, I think `[{"x":null},{"Q2":null},{"y":null}]` could just be a single dictionary: `{"x":null, "Q2":null, "y":null}` (I'm using the `null` as example because the arrays are big, but the same applies.

Zaharid:
One big annoyance with json for scientific applications is that it cannot handle NaN or infinity. I wonder how relevant is that in here. For one we could parse the same files with yaml and write e.g. `.inf` for infinity.

enocera:
So, I don't think that NaN or inf can appear as sensible values of data central values or uncertainties or kinematic bins. Moving to `yaml` is perhaps advisable anyway - incidentally I used `.json` instead of `.yaml` only because this is what was originally suggested.

scarlehoff:
Hi @enocera, I don't understand the format of the uncertainties:

```
{'sys_1': ['ADD', 'CORR', <numbers>}
```
What should I parse from here. Are the numbers following `ADD` and `CORR` the list of ADD and CORR concatenated together? Wouldn't it be better to have maybe `{sys_1: ([list of add], [list of corr])}`?

Or even directly a list of dictionaries 
```
[{`ADD`: [numbers], 'CORR': [NUMBERS]},]
```
and then the number of systematic is just the number of dictionaries in the list (so no need to keep track with the `sys_x` name).

enocera:
So:
- `sys_1` is the 1-st sytematic uncertainty; we can have as many sys as we need, and I thought to enumerate them like `sys_1`, `sys_2`, ... , `sys_i`, ..., `sys_n`, where `n` is the number of systematic uncertainties;
- `sys_1` in this case is additive (`ADD`) and correlated (`CORR`) at the same time; the nature (additive vs multiplicative) and correlation of each uncertainty should be specified for each sys_i (and can be different for different `i`s);
- `<numbers> correspond to `sys_i` over the `j` data points.

enocera:
Hi @scarlehoff can you please have a look at the 'metadata.yaml`, `data.yaml`, `kinematics.yaml` and `uncertainties.yaml` files (the last three files are generated by `filter.py` which is NO WAY a final example of what a filter should look like) and see if:
- the format complies with the various suggestions made in this PR;
- the format can be passed to your readers.
In my opinion, the format of the `.yaml` files (with the exception of the theory specifications in `metadata.yaml`) is very close to what is going to become our default.

scarlehoff:
Hi @enocera I think you forgot to add the `filter.py` to the PR? (or I am not able to find it among the many files)

enocera:
@scarlehoff It should be there now.

enocera:
As I mentioned, this is a quick and dirty thing for the time being.

scarlehoff:
> As I mentioned, this is a quick and dirty thing for the time being.

Sure! But it will help me ask questions about the format :P

buildmaster/NMCPD/filter.py
```diff
@@ -0,0 +1,64 @@
...
+        values = input['dependent_variables'][0]['values']
+        x = float(input['dependent_variables'][0]['qualifiers'][4]['value'])
+        sqrts = float(input['dependent_variables'][0]['qualifiers'][3]['value'])
+        
+        for j in range(len(values)):
+            data_central_value = {'index' : index, 'value' : input['dependent_variables'][0]['values'][j]['value']}      
+            data_central.append(data_central_value)
+            Q2 = input['independent_variables'][0]['values'][j]['value']
+            y = Q2 / ( sqrts * sqrts * x )
+            kin_avg_value = {'index' : index , 'x' : x , 'q2' : Q2 , 'y' : y}
```
scarlehoff, line 34: Are the kinematic values always going to be named x, q2 and y?
(I guess the question is, are x and y standing for "whatever is in the x and y axis and so I should use these two names)

enocera: NO. This si true for DIS, but NOT for, say, W boson production. Would you prefer a more neutral naming, such as Kin_1, Kin_2 etc?

enocera: Where the idea is to have as many Kin_* as we want, depending on the nature of the observable?

scarlehoff: Nono, this is fine. I can just read the name form the file. But if they were going to be always the same I would just hardcode them in the reader.

buildmaster/NMCPD/filter.py
```diff
@@ -0,0 +1,64 @@
...
+            kin_avg.append(kin_avg_value)
+            kin_min_value = {'index' : index , 'x' : None , 'q2' : None , 'y' : None}
+            kin_min.append(kin_min_value)
+            kin_max_value = {'index' : index , 'x' : None , 'q2' : None , 'y' : None}
+            kin_max.append(kin_max_value)
```
scarlehoff, line 39: In the `None` case (if there is expected to be a  `None` case) I would rather have the minimum and the maximum to be equal to the central value (or the entire dictionary set to `None`).
Having the `yaml` file fill with `None`s seems a bit like a waste.

enocera: Done.

scarlehoff:
@enocera would it be possible to add a few more datasets to the example? So that I can make sure that the reader don't have anything that works just by chance.

I would also like to prepare a small script (similar to what @felixhekhorn did for the theory pipeline) that checks automatically that everything works so it would be helpful to have more than one example. I think it is a bit easier in this case since there are less intermediate steps.

enocera:
> @enocera would it be possible to add a few more datasets to the example? So that I can make sure that the reader don't have anything that works just by chance.

@scarlehoff Yes, definitely. It is my goal to have all the DIS data ready by the code meeting.

scarlehoff:
If you have some ready which are more complex that NMCPD it would be great (to check more combinations of additive / multiplicative / systematics / etc

enocera:
Indeed - I think that the HERA data sets are a good example. I'll keep you posted when these are ready.

Zaharid:
I'd like to change a bit the kinematics format: As I understand it we have different lists for average, max and min, and we repeat that for each point. That seems rather hard to put together and verbose.  Also I had to use regular expression searches to find my way in the example kinematics file. 

I would instead do something like:

meta.yaml


```yaml
kinematics:
    q2: {"description":  "Parton momentum transfer", "units": "GeV²"}
    x: {"description": "Bjorken x"}
    y: {"description": "Inelasticity"}
```

with the validobj stub being

```
@dataclass
class KinInfo:
    description: str
    units: Optional[str] = None

KinDesc = Dict[str: KinInfo]
```
(maybe with somethng more explicit for no units...)

And for the kinematics themselves

```yaml
# kinematics.yaml
bins:
    - q2: {min: null, avg: 0.16, max: null}
      x: {min: null, avg: 0.0015, max: null}
      y: {min: null, avg: 0.12085, max: null}

    - q2: {min: null, avg: 0.16, max: null}
      x: {min: null, avg: 0.0015, max: null}
      y: {min: null, avg: 0.12085, max: null}
```

with the stub

```
@dataclass
class BinLimit:
    min: Optional[float]
    avg: Optional[float]
    max: Optional[float]

Bins = List[Dict[str, BinLimit]]

```
with some more or less complicated logic on how to fill min, max and avg (unless we decided that we are going to require them?).  Contrary to what is being discussed, I'd strongly prefer if the files didn't lie and didn't set the min and max values if not available. E.g. if the current format was used to derive integration limits, it would yield zero. 

I would think that the kinematics file doesn't strictly need and `index` as it is going to be the thing that gives identity to the points and defines the index.

Zaharid:
The PLOTTING files had the concept of per process plotting file that can abstract common values. E.g. the kinematic description for DIS.

I am not sure how much we want to replicate that behavior: it is convenient to have some common stuff, but on the other hand we do all sorts of overrides in the plotting code that somewhat defeat the purpose. Guess could be something to discuss.

enocera:
@scarlehoff You should be able to find in this branch the updated filter and `.yaml` files for the prototype data set as per the code meeting discussion.

buildmaster/NMCPD/filter.py
```diff
@@ -0,0 +1,69 @@
+#Filter for NMCPD
+
+import sys
+import yaml
```
Radonirinaunimi, line 4: The function that retrieves the tables in `YAML` format could look like:
```suggestion
import yaml
import requests

from urllib.request import urlopen

def download_hepdata_table(list_dict_tables: list, table_number: int) -> dict:
    # Loop over the various links of the tables
    for table in list_dict_tables:
        if table_number == int(table["name"].split()[-1]):
            table_dict = requests.get(table["@id"]).json()
            # Loop over the various table formats
            for url_tab in table_dict["distribution"]:
                if "YAML" in url_tab["description"]:
                    reads_yaml = urlopen(url_tab["contentUrl"])
    return yaml.safe_load(reads_yaml)
```
And one can choose to still write the raw data into files if needed to.

buildmaster/NMCPD/filter.py
```diff
@@ -0,0 +1,69 @@
...
+
+def filter_NMCPD():
+    with open('metadata.yaml', 'r') as file:
+        metadata = yaml.safe_load(file)
+
+    version = metadata['hepdata']['version']
+    tables  = metadata['hepdata']['tables']
+
```
Radonirinaunimi, line 12:
```suggestion
    tables  = metadata['hepdata']['tables']
    # Retrieves the tables url
    hepdata_url = metadata['hepdata']['url']
    tables_dict = requests.get(hepdata_url).json()["hasPart"]
```

buildmaster/NMCPD/filter.py
```diff
@@ -0,0 +1,69 @@
...
+    data_central = []
+    kin = []
+    error = []
+    error_nuc = []
+    
+    for i in tables:
+        hepdata_tables="rawdata/HEPData-ins426595-"+version+"-Table_"+str(i)+".yaml"
+        with open(hepdata_tables, 'r') as file:
+            input = yaml.safe_load(file)
```
Radonirinaunimi, line 21: I will also avoid using `input` as a variable as that is a built-in function in python.
```suggestion
        input_yaml = download_hepdata_table(tables_dict, i)
```

Radonirinaunimi (commented):
@enocera I was having a slight look at this for the FPF (low energy neutrino) project I am helping a bachelor student with and I have a few suggestions below.

As far as I understand, you still manually download the tables from HepData? This could be automatized in order to avoid this manual work. HepData has a [hepdata-cli/-api](https://github.com/HEPData/hepdata-cli) but this does not seem to easily provide us with what we need. In addition, I don't think we'd rely on a package that is not maintained. Instead, below is a simple sample function that can retrieves the tables for us.

buildmaster/NMCPD/filter.py
```diff
@@ -0,0 +1,69 @@
...
+    with open('uncertainties.yaml', 'w') as file:
+        yaml.dump(uncertainties_yaml, file, sort_keys=False)
+
+    with open('uncertainties_dw.yaml', 'w') as file:
+        yaml.dump(uncertainties_dw_yaml, file, sort_keys=False)
+
+    with open('uncertainties_sh.yaml', 'w') as file:
+        yaml.dump(uncertainties_sh_yaml, file, sort_keys=False)    
+         
+filter_NMCPD()
```
Radonirinaunimi, line 69:
```suggestion
if __name__ == '__main__':
    filter_NMCPD()
```
and specify the interpreter at the beginning.

Radonirinaunimi:
There was also the suggestion to add the information on what is being fitted [here](https://github.com/NNPDF/nnpdf/issues/1559#issuecomment-1102615041). This might be worth adding already?

Radonirinaunimi:
> We have discussed doing this several times and it would indeed be nice if we could have it in a pull request form. A key requirement is the ability to check that the current version of the tables matches the latest version in hepdata.

I agree that this is definitely something that we want. This could be easily implemented and something I can take care of.

Radonirinaunimi:
I now have a module that reads the hepdata URL from `metadata.yaml` and download the corresponding tables. It does so by first checking if the corresponding dataset has already been downloaded, and if yes checks if the version matches the latest version on hepdata.

This can be used to perform what you, @Zaharid, mentioned above. But I just want to make sure that I understand **exactly** what you have in mind. What I have been playing with is a github action that--using the module--checks every now and then hepdata if a new version of a dataset is available. And if so, it opens a PR with the new version and keeps pushing commits whenever new versions are available until the PR is merged. Is this somehow in line with what you think?

scarlehoff:
Hi @Radonirinaunimi, in this branch there are some utilities that might be useful (while the dataset implementation is now outdated). Do you want to keep these tools?

Otherwise we can close this PR.

Radonirinaunimi:
> Hi @Radonirinaunimi, in this branch there are some utilities that might be useful (while the dataset implementation is now outdated). Do you want to keep these tools?
> 
> Otherwise we can close this PR.

I think  it'd be better to close this PR, and if needed I'd implement these utilities in another PR.  Especially since after having re-implemented datasets, I'd approach these utilities in a different way now.

scarlehoff:
ok, thanks!

PR #1501, opened 2022-01-21
Swap libNNPDF LHAPDFSet for a python PDF object

scarlehoff:
Final step of #1472, with the previous changes this is just a drop-in replacement. At least from the point of view of the tests nothing changes (the test will fail because the effective exponent for `smallx` for `t3` changed).

The only thing missing from here in order to close fully #1472 is how to avoid having to run the convolution twice to get the central value for observable predictions (which is a factor of x2 every time one tries to get a prediction). See issue #1502 for more on this.

Note that a `legacy_load` is still left there because the `filter.py` uses a PDF for `MakeClosure`.

### DIFF
--- a/validphys2/src/validphys/lhapdfset.py
+++ b/validphys2/src/validphys/lhapdfset.py
@@ -60,11 +60,6 @@ class LHAPDFSet:
         self._flavors = None
 
     @property
-    def is_monte_carlo(self):
-        """Check whether the error type is MC"""
-        return self._error_type == "replicas"
-
-    @property
     def is_t0(self):
         """Check whether we are in t0 mode"""
         return self._error_type == "t0"
@@ -83,7 +78,7 @@ class LHAPDFSet:
         """
         if self.is_t0:
             return self._lhapdf_set[0:1]
-        if self.is_monte_carlo:
+        if self._error_type == "replicas":
             return self._lhapdf_set[1:]
         return self._lhapdf_set
 
@@ -129,6 +124,6 @@ class LHAPDFSet:
         # Create an array of x and q of equal length for LHAPDF
         xarr, qarr = (g.ravel() for g in np.meshgrid(xgrid, qgrid))
         # Ask LHAPDF for the values and swap the flavours and xgrid-qgrid axes
-        raw = np.array([member.xfxQ(flavors, xarr, qarr) for member in self.members]).swapaxes(1,2)
+        raw = np.array([member.xfxQ(flavors, xarr, qarr) for member in self.members]).swapaxes(1, 2)
         # Unroll the xgrid-qgrid axes
         return raw.reshape(self.n_members, len(flavors), len(xgrid), len(qgrid))
--- a/validphys2/src/validphys/lhapdfset.py
+++ b/validphys2/src/validphys/lhapdfset.py
@@ -118,21 +118,17 @@ class LHAPDFSet:
         Examples
         --------
         >>> import numpy as np
-        >>> from validphys.lhapdfset import LHAPDFSet, ER_MC
-        >>> pdf = LHAPDFSet("NNPDF40_nnlo_as_01180", ER_MC)
+        >>> from validphys.lhapdfset import LHAPDFSet
+        >>> pdf = LHAPDFSet("NNPDF40_nnlo_as_01180", "replicas")
         >>> xgrid = np.random.rand(10)
         >>> qgrid = np.random.rand(3)
         >>> flavs = np.arange(-4,4)
         >>> flavs[4] = 21
         >>> results = pdf.grid_values(flavs, xgrid, qgrid)
         """
-        # Grid values loop
-        ret = np.array(
-            [
-                [[member.xfxQ(flavors, x, q) for x in xgrid] for q in qgrid]
-                for member in self.members
-            ]
-        )
-
-        # Finally return in the documented shape
-        return np.transpose(ret, (0, 3, 2, 1))
+        # Create an array of x and q of equal length for LHAPDF
+        xarr, qarr = (g.ravel() for g in np.meshgrid(xgrid, qgrid))
+        # Ask LHAPDF for the values and swap the flavours and xgrid-qgrid axes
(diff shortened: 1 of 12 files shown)


scarlehoff:
The test that fails (effective exponent) is due to how sensitive NNPDF to precision when the PDF is close to 0 which triggers this:

https://github.com/NNPDF/nnpdf/blob/66abca9954324231d0e38969c711cae23a549ae9/validphys2/src/validphys/eff_exponents.py#L75

In the test itself this has an effect only for T3 (alpha from -0.4 to -0.6) which is the only negative exponent that we have. fwiw I think this version is better than master in that the precision is a bit better (which in turns might mean a PDF that before was 0 now it's just 1e-8 or whatever).

scarlehoff:
For the reviewer, note that this is partial PR and that needs #1504 to fully work in a sensible manner (or in a way that I am comfortable with).
I think comparefits might fail with this PR.

Zaharid:
@scarlehoff Should I start with this one then?

validphys2/src/validphys/lhapdfset.py
```diff
@@ -0,0 +1,138 @@
...
+        flavours, points in x and pointx in q
+        The return shape is
+            (members, flavors, xgrid, qgrid)
+        Return
+        ------
+            ndarray of shape (members, flavors, xgrid, qgrid)
+        Examples
+        --------
+        >>> import numpy as np
+        >>> from validphys.lhapdfset import LHAPDFSet, ER_MC
```
Zaharid, line 121: ImportError: cannot import name 'ER_MC' from 'validphys.lhapdfset' (/home/zah/nngit/nnpdf/validphys2/src/validphys/lhapdfset.py)

scarlehoff: And rightly so. Sorry, I copied the example from the closed PR and missed that. I'll fix it later today.

scarlehoff:
> @scarlehoff Should I start with this one then?

Yes, this one should be ready.

validphys2/src/validphys/lhapdfset.py
```diff
@@ -0,0 +1,138 @@
...
+        # Grid values loop
+        ret = np.array(
+            [
+                [[member.xfxQ(flavors, x, q) for x in xgrid] for q in qgrid]
+                for member in self.members
+            ]
+        )
+
+        # Finally return in the documented shape
+        return np.transpose(ret, (0, 3, 2, 1))
```
Zaharid, line 138: Aparently you can call LHAPDF with arrays. This seems to be a tad faster:

```python
def gv_2(self, flavors: np.ndarray, xgrid: np.ndarray, qgrid: np.ndarray):
    mxgrid, mqgrid = (g.ravel() for g in np.meshgrid(xgrid, qgrid))
    raw = np.array([mem.xfxQ(flgrid, mxgrid, mqgrid) for mem in self.members])
    return raw.reshape(
        len(self.members), len(qgrid), len(xgrid), len(flgrid)
    ).transpose(0, 3, 2, 1)
```

%timeit gv_2(l_pdf ,flgrid, xgrid, qgrid)
82.6 ms ± 1.65 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

vs
%timeit gv_1(l_pdf ,flgrid, xgrid, qgrid)
108 ms ± 1.28 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

Zaharid:
@tgiani some help with the remaining cpp methods, particularly MakeClosure would be appreciated here. Should not be too hard to do in python with all the pseudodata machinery in place.

validphys2/src/validphys/lhapdfset.py
scarlehoff: oh, you're right. I'm sure I tried this (a wrong variation of, clearly,) and didn't work. I'll update it.

Why did you need to use meshgrid? Would a call with xgrid/qgrid not work out of the box?

scarlehoff:
Note that the positivity one for instance won't be needed any longer (in the other PR that convolution is done with python)

validphys2/src/validphys/lhapdfset.py
Zaharid: Because it zips the inputs, which is slightly inconvenient for our use.

scarlehoff: nvm, I had forgotten what our grid values was doing (which is of course the reason you had to use `meshgrid` there)

validphys2/src/validphys/lhapdfset.py
```diff
@@ -0,0 +1,134 @@
...
+        """Check whether we are in t0 mode"""
+        return self._error_type == "t0"
+
+    @property
+    def n_members(self):
+        """Return the number of active members in the PDF set"""
+        return len(self.members)
+
+    @property
+    def members(self):
```
Zaharid, line 78: I take it changing this is the other PR.

scarlehoff: You mean members being 101 instead of 100?

validphys2/src/validphys/lhapdfset.py
```diff
@@ -0,0 +1,134 @@
...
+        self._error_type = error_type
+        if self.is_t0:
+            # If at this point we already know this is a T0 set, load only the CV
+            self._lhapdf_set = [lhapdf.mkPDF(name)]
+        else:
+            self._lhapdf_set = lhapdf.mkPDFs(name)
+        self._flavors = None
+
+    @property
+    def is_monte_carlo(self):
```
Zaharid, line 63: I am not sure if it is worth adding these seeing as there doesn't seem to be a strong case for them anywhere. For example `check_pdf_is_montecarlo` is indeed going to keep looking at `PDF` rather than `LHAPDFSet`

Zaharid:
This looks fine to me, given that it is not doing much dangerous stuff on its own.

validphys2/src/validphys/lhapdfset.py
scarlehoff: Yes, it's useless. Actually in the 101-to-100 PR I removed it. I might as well remove it from here directly.

Zaharid (approved):

PR #1503, opened 2022-01-25
Optimize pseudodata computation

Zaharid:
The function `computed_pseudoreplicas_chi2` in chi2grids.py is unusable in the current master as well as various pseudodata replica utilities.

One showstopper involves

```
 fitted_make_replicas = collect('make_replica', ('pdfreplicas',))
```

Reportengine is unfortunately unable to deal with this very elegantly and will try to resolve all the data for each of the replicas anew. It would be nice to fix that but in the meantime, add the loop manually.

 Once that is fixed it turns out that generating a replica with `make_replica` is unbearably slow now (in fact it didn't finish for me between launching  it early this morning  and looking at it now). I am not sure if this is a old phenomenon or has to do with some regression in pandas.

(Incidentally @enocera  did you notice any problems in that phase while running the fits?)

So currently, on master calling make_replica with an unlucky input can take well in excess of a few minutes per replica. The slow parts are all pandas operations such as some trivial looking concatenation like `special_add_errors = pd.concat(special_add, axis=0, sort=True).fillna(0).to_numpy()` (hence why I think it may e a regression) which takes half a second for some reason. To address that move all these operations outside, to a separate provider so they can be shared across various inputs. With that computing a pseidodata replica takes a few seconds at worst on  my laptop.

### DIFF
--- a/validphys2/src/validphys/chi2grids.py
+++ b/validphys2/src/validphys/chi2grids.py
@@ -60,10 +60,7 @@ def computed_pseudoreplicas_chi2(
     return df
 
 
-# TODO: Probably fitcontext should set all of the variables required to compute
-# this. But better setting
-# them explicitly than setting some, so we require the user to do that.
-fits_computed_pseudoreplicas_chi2 = collect(computed_pseudoreplicas_chi2, ("fits",))
+fits_computed_pseudoreplicas_chi2 = collect(computed_pseudoreplicas_chi2, ("fits", "fitenvironment", "fitpdf"))
 
 dataspecs_computed_pseudorreplicas_chi2 = collect(computed_pseudoreplicas_chi2, ("dataspecs",))
 
--- a/validphys2/src/validphys/chi2grids.py
+++ b/validphys2/src/validphys/chi2grids.py
@@ -25,6 +25,7 @@ log = logging.getLogger(__name__)
 
 def computed_pseudoreplicas_chi2(
     fitted_make_replicas,
+    fitted_replica_indexes,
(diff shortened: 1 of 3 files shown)


scarlehoff:
I've done a few fits in the last few weeks and didn't notice it. What version of pandas are you using?

Zaharid:
```
$ python
Python 3.9.6 | packaged by conda-forge | (default, Jul  6 2021, 08:53:59) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pandas
>>> pandas.__version__
'1.4.0'
```

scarlehoff:
I'm using 1.3.5. Could you write a snippet of code that reproduces the problem? So that I can run it and we can benchmark. I guess in any case the problem "started" when we started doing replicas in python right? That's at this point a while ago so I'm a bit surprised we didn't noticed it earlier.

scarlehoff:
@Zaharid are you planning to work more on this? 

Given that the motivation of this PR was

> Once that is fixed it turns out that generating a replica with make_replica is unbearably slow now (in fact it didn't finish for me between launching it early this morning and looking at it now). I am not sure if this is a old phenomenon or has to do with some regression in pandas.

I'm guessing that whatever the problem was has magically been solved during the last year? (gosh, I was convinced that this PR was only a few months old)

Zaharid:
I think this can be closed but don’t think it has been solved. Probably should be picked again by @RoyStegeman or @comane who are looking at these things. But at this point should be from scratch.

PR #1504, opened 2022-01-27
Utilize stats for predictions or any other PDF product

scarlehoff:
As promised, now all PDFs are essentially the same (you asked for the PDF, you get the full PDF, no tricks) which obviously breaks compatibility with libNNPDF.

As a result everything has to be used through the `stats` class. I've looked around for instances where the mean was taken manually (for instance in effective exponents) or where the `_rawdata` of the predictions was being accessed and tried to use `error_members` or `data` depending on what the question being asked was.

All test pass, which is good (in some of the intermediate steps the regressions test were broken which mean they were picking up the changes) but this needs to be reviewed with a lot of care. Please have a go at #1501 since this builds on top of that one (both are actually not that much code, they are tricky because they touch basically the core of validphys).

The only thing I'm not sure about is lines like this one in the closure tests:

https://github.com/NNPDF/nnpdf/blob/a692709641866ebe6e8a4e799b468b1a14d34bce/validphys2/src/validphys/closuretest/closure_results.py#L213

Because I thought the `_rawdata` would be just the error members (being MC pdfs most of the time) but the tests only works with `.data` (so, including the central replica). I guess either I misunderstood what these functions were doing or there is something wrong somewhere else.

### DIFF
--- a/n3fit/src/n3fit/tests/test_vpinterface.py
+++ b/n3fit/src/n3fit/tests/test_vpinterface.py
@@ -40,7 +40,7 @@ def test_N3PDF(members, layers):
     xsize = np.random.randint(2, 20)
     xx = np.random.rand(xsize)
     n3pdf = generate_n3pdf(layers=layers, members=members)
-    assert len(n3pdf) == members + 1
+    assert len(n3pdf) == members
     assert n3pdf.stats_class == MCStats
     assert n3pdf.load() is n3pdf
     w = n3pdf.get_nn_weights()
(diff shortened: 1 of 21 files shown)


scarlehoff:
The amount of places where python is being abused in many creative ways is making me go crazy.

scarlehoff:
This is exactly the kind of thing I was worried about @scarrazza @Zaharid.
I might start this PR anew with what I know now next week, starting with a number of test that can pick up all the silent problems I've been finding (this is just an obvious one, but I've been picking up quite a few this morning as I was testing)

[matched_positivity_from_dataspecs1_plot_dataspecs_positivity.pdf](https://github.com/NNPDF/nnpdf/files/7958183/matched_positivity_from_dataspecs1_plot_dataspecs_positivity.pdf)

PR #1506, opened 2022-01-28
Add a fallback to using all points for reports that are using old fits with bad filters.

### LINKED ISSUE #1505: Add a catch for old fits where setupfit didn't generate cuts correctly
In the new code we are asking explicitly for cuts where cuts are needed. In the olden times `vp-setupfit` would just skip some datasets if there were no cuts for them so there are some fits in the server that, in order to work, need the filters to be regenerated.

I'll have a look at how add a flag so that if `fromfit` is being used and no cut is found the default is to not cut anything.

This is a problem we saw already at some point in the past, opening the issue so I don't forget (or in case people prefer the exception to be triggered and fixed in a case-by-case basis).

scarlehoff:
One example (I'm using `generate_a_report.yaml` as template) that would fail in current master (because the HERACOMBNCEM cuts do not exist). With this you get a million warnings but it works.
I think many warnings is better than a failure. Fixes #1505 

```
fit: NNPDF31_nnlo_as_0118_DISonly

pdf:
  from_: "fit"

experiments:
  - experiment: TEST
    datasets:
      - {dataset: HERACOMBNCEM , frac: 0.5}

theoryid: 162

use_cuts: "fromfit"

template_text: |
  {@experiments::experiment plot_chi2dist@}

actions_:
  - report(main=True)
```

### DIFF
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -446,6 +446,7 @@ class CoreConfig(configparser.Config):
     def parse_use_fitcommondata(self, do_use: bool):
         """Use the commondata files in the fit instead of those in the data
         directory."""
+        import ipdb; ipdb.set_trace()
         return do_use
 
     def produce_commondata(self, *, dataset_input, use_fitcommondata=False, fit=None):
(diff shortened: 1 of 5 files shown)


Zaharid (changes requested):
Let's discuss in some call.

scarlehoff:
This is now updated with what we talked about yesterday.
I guess we could even cache the cuts instead of loading them every time from the file but I guess it doesn't really matter.

Zaharid (approved):

PR #1507, opened 2022-01-28
Add tests for sum rules

scarlehoff:
The sum rules are important enough to be tested explicitly.


Zaharid:
Could this use the regression test facility so it is easier to regenerate?

validphys2/src/validphys/tests/test_sumrules.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def _regression_sum_rules(pdf_name):
+    known_sumrules = API.sum_rules_table(pdf=pdf_name, Q=Q)
+    central_val = API.central_sum_rules_table(pdf=pdf_name, Q=Q)
+    # First concatenate the central value to the sum rules
+    ret = pd.concat([known_sumrules, central_val], axis=1)
+    # The unknown sum rules need to be treated a bit because
+    # 1) They don't contain min or max
+    # 2) They are of type object?
```
Zaharid, line 37: This is so that the central value and error are displayed with a reasonable number of decimals in all cases. See

https://github.com/NNPDF/reportengine/blob/675036a93ffc853398d169501d7fc40c9c2cb168/src/reportengine/floatformatting.py#L55

I guess one could use `unknown_sum_rules` action instead.

scarlehoff: I see. But the other problem stays so not sure if it changes much (and the others are floats)

Zaharid (commented):
Looks fine to me. Thanks for adding the test.

RoyStegeman (approved):
Seems ok!

PR #1514, opened 2022-02-08
Make saved pseudodata match fitted pseudodata

### LINKED ISSUE #1513: Using the flag `savepseudodata` doesn't save the same data that was fitted
From what I can see for the fit `make_replica` is called separately for each group, instead when it has to save it 1) it generates it again -which takes time- but also `make_replica` is called only once, and that messes up for sure with the seeding and maybe with something else.

Basically, both numbers only agree if just one group is used (so for instance only ATLAS data) since in both cases `make_replica` is called just once.

I think I have a solution for it which I will push when I'm done with making it as minimal as possible.

The training/validation mask seems to be correct instead so not everything is lost!

/cc @J-M-Moore @Zaharid @siranipour @RoyStegeman

scarlehoff:
Closes #1513.

`make_replica` still gets called twice per group though, couldn't find a reasonable solution for that, seems a bit of a waste of CPU cycles but at least now they generate the same array.

I guess it would be good if #1503 fixed that somehow

Also, in the topic of #1503, I've been running a lot `make_replica` in my laptop to fix this issue and I don't find it particularly slow.


scarlehoff:
Regardless of this, I wonder what the actual behavior _should_ be. Right now two groups won't get the exact same replica (meaning, the same random number) as long as their number of datapoints is different. This is a bug waiting to happen.

I would rather use the name of the group as a salt like we do for the training validation split but let me know what people think (because this would make past fits non reproducible).

scarlehoff:
Note that this is what we did for NNPDF4.0: https://github.com/NNPDF/nnpdf/blob/1229126c3383fd8efb3c24fa186c59eadee3da46/validphys2/src/validphys/n3fit_data.py#L189

so the bug is not part of the paper release

validphys2/src/validphys/tests/test_pseudodata.py
```diff
@@ -74,11 +72,17 @@ def test_no_savepseudodata():
 def test_read_matches_recreate():
     reads = API.read_fit_pseudodata(fit=PSEUDODATA_FIT)
     recreates = API.recreate_fit_pseudodata(fit=PSEUDODATA_FIT)
-    for read, recreate in zip(reads, recreates):
+    recreate_pdf = API.recreate_fit_pseudodata(fit=PSEUDODATA_FIT)
```
siranipour, line 75: What's the difference between this and the previous line?

scarlehoff: ups, it should be `recreate_pdf_pseudodata`

scarlehoff:
For the reviewer: what does "accounts for the postfit reshuffling of replicas" actually means here?

https://github.com/NNPDF/nnpdf/blob/d1e0a3ac3caac2ee987fdc94481f7846774548d0/validphys2/src/validphys/pseudodata.py#L277

Because I understood as "it takes the members in the PDF order and makes into the replicas order" but maybe it means the opposite, that the output order is that of the PDFs? (then I should remove said line)

siranipour:
I believe this takes LHAPDF replica 1 and recreates its pseudodata, because of the usual "fit replica 1" =/= "LHAPDF replica 1" thing.

You can see the namespace it makes is from
https://github.com/NNPDF/nnpdf/blob/d1e0a3ac3caac2ee987fdc94481f7846774548d0/validphys2/src/validphys/config.py#L252

which in turn is calling

https://github.com/NNPDF/nnpdf/blob/d1e0a3ac3caac2ee987fdc94481f7846774548d0/validphys2/src/validphys/fitdata.py#L361

so it maps the n3fit replica to whatever it become after postfit. Does that make sense?

scarlehoff:
What is the intended behavior for `recreate_pdf_pseudodata` when it says "reshuffling of replicas"? Is it to be returned in the LHAPDF order or the fit replica order? (basically, in a situation in which member 2 and 1 are replicas 1 and 2 respectively, should the output be [2,1] -LHAPDF- or [1,2] -fit-?)

siranipour:
>  Is it to be returned in the LHAPDF order
> the output be [2,1]

this eone

siranipour (approved):
Good to see the fix is relatively straightforward. Thanks for spotting and fixing this

PR #1515, opened 2022-02-09
Utilize stats everywhere where it can be used

scarlehoff:
I will leave the theory predictions for the end.


validphys2/src/validphys/mc2hessian.py
```diff
@@ -108,9 +108,9 @@ def _create_mc2hessian(pdf, Q, xgrid, Neig, output_path, name=None):
 
 def _get_X(pdf, Q, xgrid, reshape=False):
     pdf_grid = xplotting_grid(pdf, Q, xgrid=xgrid)
-    pdf_grid_values = pdf_grid.grid_values
-    replicas = pdf_grid_values
-    mean = pdf_grid_values.mean(axis=0)
+    pdf_grid_values = pdf_grid.stats_gv
+    replicas = pdf_grid_values.error_members()
+    mean = pdf_grid_values.central_value()
```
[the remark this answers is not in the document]
scarlehoff: Makes sense, I'll add it.

scarlehoff:
In the last commit the ThPredictions are no longer using the NNPDFResult class and instead use the StatsResult. Everyting seems to work fine (I'll send the fit bot again now). Since now the central value is computed by taking the average over the result of the members (instead of taking the result of the central replica as the central value) the regression tests have slightly changed.

I will finish what's missing from this PR (mainly making sure that the `grid_values` are never used by themselves but always as part of the `Stats` class) and then will mark this PR for review.

The change of the MC PDF to include by default all replicas (and then being the job of the StatsClass to decide what to do with them) I will do in a separate PR pointing to this one because although is in principle a smaller change (if I did everything correctly it should only be a small change on the behaviour of the `Stats` and `PDF` classes) but that could have many unforeseeable consequences.

[differences_from_da3069c_to_1fe212432869d.tar.gz](https://github.com/NNPDF/nnpdf/files/8042642/differences_from_da3069c_to_1fe212432869d.tar.gz)

scarlehoff:
This is ready for review.

At this point the raw `rawdata` or `data` is used wherever the `_rawdata` was used, so the results are expected to be equivalent to whatever we got before this PR. Both the tests and the fitbot seems to confirm that.

In order to continue (and deal with #1516) I will open today another PR pointing to this one where instead of the raw data, the `error_members` property/method is used wherever it makes sense. The reason for making it another PR is because, as said in #1516, the code is very inconsistent in many places so it can potentially break havoc and I prefer not to increase the entropy of this (large enough) PR.

validphys2/src/validphys/correlations.py
```diff
@@ -46,28 +43,29 @@ def _basic_obs_pdf_correlation(pdfarr, obsarr):
...
+    obs2: (nreplicas, nbins2)
 
     The result is (nbins1 , nbins2), a mareix containing the correlation
     coefficients between the two sets.
     """
-    #TODO: Do this at Result level taking into account error type
-    x = (obsarr1.T - np.mean(obsarr1, axis=1)).T
-    y = (obsarr2.T - np.mean(obsarr2, axis=1))
+    x = (obs1.data - obs1.central_value()).T
+    y = (obs2.data - obs2.central_value())
```
Zaharid, line 57: This doesn't seem universally correct. In particular obviously not for non symmetric hessian sets. Seems to me this would need to be added per-Stats subclass and throw a check error if not implemented. Or more simply (but more fiddly, use `@check_pdf_is_montecarlo`)

Zaharid (changes requested):

validphys2/src/validphys/correlations.py
scarlehoff: There is such check, I didn't remove it.

scarlehoff: Or do you mean it is also wrong for MC PDFs?

validphys2/src/validphys/pdfgrids.py
```diff
@@ -37,9 +41,26 @@ def xgrid(xmin:numbers.Real=1e-5, xmax:numbers.Real=1,
...
+    xgrid: np.ndarray
+    grid_values: Stats
+    scale: str
 
-XPlottingGrid = namedtuple('XPlottingGrid', ('Q', 'basis', 'flavours', 'xgrid',
-                                             'grid_values', 'scale'))
+    def copy_grid(self, grid_values=None):
+        """Create a copy of the grid with potentially a different set of values"""
+        if not isinstance(grid_values, Stats):
+            log.warning("XPlottingGrid being called with a numpy grid, should be using Stats instead!")
```
Zaharid, line 61: Could this be an error?

validphys2/src/validphys/correlations.py
Zaharid: I should have said that if it is only going to work for monte carlo then it should use arrays rather than stats objects, as it is a bit confusing as it is now.

scarlehoff: You mean making the input be already "error members" and "central_value"? (for each of the observables). I'm not sure I understand what you suggest to do in this particular case?

Zaharid: Originally we had ` def _basic_obs_pdf_correlation(pdfarr, obsarr):` which work with arrays of monte carlo replicas. 

I think it should keep having the same interface (unless it is going to be extended to work for all stats, and even so the mc code would probably call this function...). So I'd suggest reverting the change here and adjusting the caller to pass the array (in the place where we have an explicit check for a mc pdf).

scarlehoff: I see what you mean now, yes, I can pass `pdf.error_members()` instead of the whole of pdf. It should be fine.

Zaharid:
Apart from the things discussed above I think we should merge this to make some progress.

Zaharid (approved):

PR #1517, opened 2022-02-14
use error_members instead of rawdata

scarlehoff:
This is straightforward for MC PDFs (they are the same thing now and this change will make the change from all_members to only_replicas very simple) but for Hessian PDFs is instead very dangerous since the error members are the rawdata-1.

Making this as a separate PR since it makes many changes in the code but are _almost_ a `sed` (a manual one nonetheless) and because I will be running the fit bot quite a bit I guess.

### DIFF
--- a/validphys2/src/validphys/arclength.py
+++ b/validphys2/src/validphys/arclength.py
@@ -49,7 +49,7 @@ def arc_lengths(
         # Finite diff. step-size, x-grid
         eps = (b - a) / npoints
         ixgrid = xgrid(a, b, "linear", npoints)
-        # PDFs evaluated on grid, use the entire thing, the stats class will chose later
+        # PDFs evaluated on grid, use the entire thing, the Stats class will chose later
         xfgrid = xplotting_grid(pdf, Q, ixgrid, basis, flavours).grid_values.data * ixgrid[1]
         fdiff = np.diff(xfgrid) / eps  # Compute forward differences
         res += integrate.simps(1 + np.square(fdiff), ixgrid[1][1:])
--- a/validphys2/src/validphys/arclength.py
+++ b/validphys2/src/validphys/arclength.py
@@ -49,7 +49,7 @@ def arc_lengths(
         # Finite diff. step-size, x-grid
         eps = (b - a) / npoints
         ixgrid = xgrid(a, b, "linear", npoints)
-        # PDFs evaluated on grid
+        # PDFs evaluated on grid, use the entire thing, the stats class will chose later
         xfgrid = xplotting_grid(pdf, Q, ixgrid, basis, flavours).grid_values.data * ixgrid[1]
         fdiff = np.diff(xfgrid) / eps  # Compute forward differences
         res += integrate.simps(1 + np.square(fdiff), ixgrid[1][1:])
--- a/validphys2/src/validphys/deltachi2.py
+++ b/validphys2/src/validphys/deltachi2.py
@@ -213,7 +213,7 @@ class PDFEpsilonPlotter(PDFPlotter):
         # Create a copy of the `Stats` instance of the grid
         # with only the flavours we are interested in
         gv = grid.grid_values.data
-        stats = grid.grid_values.copy_grid(grid_values=gv[:, flindex, :])
+        stats = grid(grid_values=gv[:, flindex, :]).grid_values
 
         # Ignore spurious normalization warnings
         with warnings.catch_warnings():
--- a/validphys2/src/validphys/deltachi2.py
+++ b/validphys2/src/validphys/deltachi2.py
@@ -140,15 +140,12 @@ def pos_neg_xplotting_grids(delta_chi2_hessian, xplotting_grid):
     """
     positive_eigenvalue_mask = delta_chi2_hessian >= 0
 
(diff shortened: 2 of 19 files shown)


scarlehoff:
@Zaharid are you using the paramfits package? Is it tested / are there runcards that I can use as an example that things don't break? (the `mean`, `std` and raw `.data` is used a lot there)

Zaharid:
> @Zaharid are you using the paramfits package? Is it tested / are there runcards that I can use as an example that things don't break? (the `mean`, `std` and raw `.data` is used a lot there)

This should do https://vp.nnpdf.science/rFD9ZeCfSxOwPmrSuyoneg==/input/runcard.yaml

scarlehoff:
oh, that's a pontetially important one, didn't realise it was paramfits stuff.

Maybe using Stats everywhere will create the parabolas :P

Zaharid:
Sorry, this one https://vp.nnpdf.science/dKVAidYgQRW9hPBGQ9z3Rw==/input/runcard.yaml

scarlehoff:
The error seems to be `reportengine` not knowning what to do with a numpy array here: https://github.com/NNPDF/reportengine/blob/fe736351c94b1f0a24b9c695da5bffafa1e5e98b/src/reportengine/floatformatting.py#L80 I think this case should be taken into account (maybe with a nicely-formatted error) since it seems to accept them.

That said, the fact that now they are arrays (of one element) when they used to be floats is a side-effect I was not expecting so I'm actually happy it failed.

scarlehoff:
> Sorry, this one https://vp.nnpdf.science/dKVAidYgQRW9hPBGQ9z3Rw==/input/runcard.yaml

Ok, for now it seems to work and generate the same thing. Thanks!

scarlehoff:
It seems the bot is not able to upload reports to the server? @scarrazza could you have a look at the server?

validphys2/src/validphys/arclength.py
```diff
@@ -49,7 +49,7 @@ def arc_lengths(
         # Finite diff. step-size, x-grid
         eps = (b - a) / npoints
         ixgrid = xgrid(a, b, "linear", npoints)
-        # PDFs evaluated on grid
+        # PDFs evaluated on grid, use the entire thing, the stats class will chose later
```
scarlehoff, line 52:
```suggestion
        # PDFs evaluated on grid, use the entire thing, the Stats class will chose later
```

validphys2/src/validphys/deltachi2.py
```diff
@@ -140,15 +140,12 @@ def pos_neg_xplotting_grids(delta_chi2_hessian, xplotting_grid):
...
+    # The masks do not include replica 0, add it in both grids
     pos_mask = np.append(True, positive_eigenvalue_mask)
     neg_mask = np.append(True, ~positive_eigenvalue_mask)
 
-    pos_grid = xplotting_grid.grid_values.data[pos_mask]
-    neg_grid = xplotting_grid.grid_values.data[neg_mask]
-
-    pos_xplotting_grid = xplotting_grid.copy_grid(grid_values=pos_grid)
-    neg_xplotting_grid = xplotting_grid.copy_grid(grid_values=neg_grid)
+    pos_xplotting_grid = xplotting_grid.mask_replicas(pos_mask)
```
Zaharid, line 147: Is this method expected to be used a lot? If this is going to be the only use I'd rather keep the functionality here than adding the method, mostly to not make xplotting_grid harder than it needs to be. cc @RoyStegeman

scarlehoff: > Is this method expected to be used a lot?

I guess no? But (answering your other question) then the `copy_grid` method needs to work with numpy arrays (or have an access to `__class__` here which is ugly).

RoyStegeman: Probably not, the delta chi2 thing is sometimes used for students projects but I don't plan to do much with it myself. Maybe @scarlehoff  has other uses for mask_replica in mind?

Zaharid: or wrap `pos_mask` in pdf.stats_class, no?

Zaharid:
> The error seems to be `reportengine` not knowning what to do with a numpy array here: https://github.com/NNPDF/reportengine/blob/fe736351c94b1f0a24b9c695da5bffafa1e5e98b/src/reportengine/floatformatting.py#L80 I think this case should be taken into account (maybe with a nicely-formatted error) since it seems to accept them.
> 
> That said, the fact that now they are arrays (of one element) when they used to be floats is a side-effect I was not expecting so I'm actually happy it failed.

How did you trigger that?

scarlehoff:
> How did you trigger that?
```python
from reportengine.floatformatting import ValueErrorTuple
import numpy as np 
aa = ValueErrorTuple(np.array([3]), np.array([1]))
print(aa)
```
I don't think throwing an exception here is a bug as it did help me, but naively I would think it is a common enough situation that warrants its own exception.

Zaharid:
I'd say it is behaving as expected. Not adding that special case.

Zaharid:
This looks good as well. Not testing too much manually but then again hopefully we have automated tests for most stuff.

Zaharid (approved):

PR #1518, opened 2022-02-14
Documentation madgraph+amcfast

enocera:
This PR corrects and complements the discussion of the generation of interpolation tables in the APPLgrid format using madgraph+amcfast.
cc @J-M-Moore @mariaubiali

### DIFF
--- a/doc/sphinx/source/tutorials/APPLgrids.md
+++ b/doc/sphinx/source/tutorials/APPLgrids.md
@@ -66,16 +66,21 @@ below.
 
 ### MadGraph5\_aMC@NLO + amcfast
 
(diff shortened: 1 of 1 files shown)


Zaharid:
On one hand I am hesitant about suggesting that docker is the only way to do this, on the other realistically it is the only reasonable way that can be expected to work (for one because of the python 2 dependency), and apparently it is possible to run docker containers in university clusters nowadays. @J-M-Moore could we also link to the github repo which is less of a black box anyway?

J-M-Moore:
Yes, that's a good idea, we could link my git repository where you can build the Docker image yourself (if you have access to the NNPDF/external repository). Here's the link: https://github.com/J-M-Moore/applgrid_docker.

Issue #1511, opened 2022-02-03
Add vp parser for new commondata format


Zaharid:
We would like to take the input in #1500 and write a parser that results in essentially coredata.Commondata,
https://github.com/NNPDF/nnpdf/blob/ab7f0536cc471bbc53ed6c107ab93a96f7c3dd02/validphys2/src/validphys/coredata.py#L123

with a few adjustments such as supporting n kinematics.

That should come with the accompanying machinery for Loader. We are going to have some more context dependent options, such as the relevant variant that gets selected.

scarlehoff:
Superseded by #1709

PR #1522, opened 2022-02-15
Return all members when using MC PDFs

scarlehoff:
As the title suggest.

After the two previous PR (#1515 and #1517) where `Stats` where used everywhere, the only two things that are problematic are:

1. The sumrules (I guess nobody cared whether the central value was computed because by definition it would fall inside the envelope anyway?)
2. The N3PDF interface (which was a hack to be compatible with what vp was doing, so now the hack has to be undone).

Nothing is evidently broken, but many thing might be broken in not obvious ways. The first commit is just so that the regression for the sumrules passes so I can update it in the next commit (where all members will be computed also there for both Hessian and MC PDFs).

The last commit will be a simplification of the N3PDF interface to validphys. And that should be all.

After all this series of PR nothing should be accessing `libNNPDF` *results* any longer other than the commondata datapoints which are immediately wrapped into a `Stats` class:

https://github.com/NNPDF/nnpdf/blob/0321658bfddfc39338d9635ff0bb4b17e70cff5f/validphys2/src/validphys/results.py#L81

but loading them in python is the next step: #1511 :)

(ofc, this last one is not necessary to destroy C++ but since we were already here we might as well do it, if you guys prefer to keep doing MC pdfs as we were doing up to know please let me know asap so I don't build on top of this PR)


scarlehoff:
almost there

scarlehoff:
This should be ready for review. We can talk this afternoon about who is going to review this because I guess all three PR should be done by the same person and hopefully at the same time.

I'm currently running all the examples in the validphys folder with the current commit. I can gzip them all and send them to the reviewer if they wish.

n3fit/src/n3fit/vpinterface.py
```diff
@@ -46,94 +47,34 @@
...
-    ----------
-        pdf_models: :py:class:`n3fit.backends.MetaModel` (or list thereof)
-            PDF trained with n3fit, x -> f(x)_{i} where i are the flavours in the evol basis
-        fit_basis: list(dict)
-            basis of the training, used for reporting
-        name: str
-            name of the N3PDF object
-    """
+    def error_members(self):
+        return self.data
```
Zaharid, line 55: This is the same as the base class.

n3fit/src/n3fit/vpinterface.py
```diff
@@ -46,94 +47,34 @@
...
 
-    def __init__(self, pdf_models, fit_basis=None, name="n3fit"):
-        self.fit_basis = fit_basis
-        self.basis = check_basis("evolution", EVOL_LIST)["basis"]
+    def central_value(self):
+        return np.mean(self.data, axis=0)
```
Zaharid, line 58: Also the same as the vase class.

n3fit/src/n3fit/vpinterface.py
```diff
@@ -46,94 +47,34 @@
...
-        # If no replica is explicitly requested, get the preprocessing layer for the first model
-        if replica is None:
-            replica = 1
-        preprocessing_layers = self.get_member_model(replica).get_layer_re(r"pdf_prepro_\d")
-        if len(preprocessing_layers) != 1:
-            # We really don't want to fail at this point, but print a warning at least...
-            log.warning("More than one preprocessing layer found within the model!")
-        preprocessing_layer = preprocessing_layers[0]
+    def __init__(self, name, pdf_models):
+        log.info("Creating LHAPDF-like n3fit PDF")
```
Zaharid, line 65: Not a huge fan of printing to screen when objects are created...

But at least:

```suggestion
        log.debug("Creating LHAPDF-like n3fit PDF")
```

Zaharid (changes requested):
Will take me a while to go over the bits.

n3fit/src/n3fit/vpinterface.py
scarlehoff: The base should be `self.data[1:]`

n3fit/src/n3fit/vpinterface.py
scarlehoff: When all members are returned this is no longer true.

scarlehoff:
TODO:

- [ ] Access the arrays directly in `correlations.py` [comment](https://github.com/NNPDF/nnpdf/pull/1515#discussion_r813932478)
- [ ] Check whether `copy_grid` in `XplottingGrid` can be made to only accept `Stats` without breaking anything [comment](https://github.com/NNPDF/nnpdf/pull/1515#discussion_r813938564)
- [ ] Either remove `mask_replicas` completely or make it part of the `Stats` class [comment](https://github.com/NNPDF/nnpdf/pull/1517#discussion_r813958091)

As discussed, moving here the minor comments of the other PRs so that they can be merged

Zaharid:
@scarlehoff do you want to merge the other PR as they are and keep working here, or close the others and only merge this one?

scarlehoff:
> @scarlehoff do you want to merge the other PR as they are and keep working here, or close the others and only merge this one?

Maybe merging them one by one makes doing the "undo" easier in case something is wrong obviously with any of them. But I don't think it makes a difference tbh

scarlehoff:
if you approve the others I'll merge them

scarlehoff:
I've implemented now the changes that were requested on the other two PRs.

n3fit/src/n3fit/vpinterface.py
```diff
@@ -45,9 +46,111 @@
...
+        self._error_type = "replicas"
+        self._name = name
+        self._lhapdf_set = pdf_models
+        self._flavors = None
+        self.basis = check_basis("evolution", EVOL_LIST)["basis"]
+
+    def xfxQ(self, x, Q, n, fl):
+        """Return the value of the PDF member for the given value in x"""
+        return self.grid_values([fl], [x]).squeeze()[n]
```
RoyStegeman, line 74:
```suggestion
    def xfxQ(self, x, Q, replica, fl):
        """Return the value of the PDF member for the given value in x"""
        return self.grid_values([fl], [x]).squeeze()[replica]
```

validphys2/src/validphys/sumrules.py
```diff
@@ -123,24 +123,23 @@ def f(x, lpdf, irep, Q):
...
-               limit=1000,
-               epsabs=1e-4, epsrel=1e-4)[0]
-
-    for irep in range(nmembers):
-        for r, f in enumerate(integrands):
-            res[r,irep] =  (integral(f, 1e-9, 1e-5, irep)  +
-               integral(f, 1e-5, 1e-3, irep) + integral(f, 1e-3, 1, irep))
-    return {k: v for k,v in zip(rules_dict, res)}
+    return {k: [_integral(r, m, Q) for m in lpdf.members] for k,r in rules_dict.items()}
+    # I very much doubt that looping over PDF members is the slow piece here...
```
RoyStegeman, line 140: Sure, the loop is not making it slow, but why does that need to be mentioned?

validphys2/src/validphys/core.py
```diff
@@ -243,19 +243,12 @@ def grid_values_index(self):
         -----
         The range object can be used efficiently as a Pandas index.
         """
-        if self.error_type == "replicas":
-            return range(1, len(self))
-        elif self.error_type in ("hessian", "symmhessian"):
-            return range(0, len(self))
-        else:
-            raise RuntimeError(f"Unknown error type: {self.stats_class}")
+        return range(0, len(self))
```
RoyStegeman, line 246: aii, this is a bit confusing with making rep1 the 0 index, but I guess it's libNNPDF stuff.. Anyway, I guess the docstring needs to be adjusted.

validphys2/src/validphys/correlations.py
```diff
@@ -45,32 +45,37 @@ def _basic_obs_pdf_correlation(pdf_stats, obs_stats):
...
+    obs_val = th.error_members
+    corrs = Stats(_basic_obs_pdf_correlation(pdf_val, obs_val))
     return xplotting_grid.copy_grid(grid_values=corrs)
 
 
 corrpair_results = collect("results", ["corrpair"])
 corrpair_datasets = collect("dataset", ["corrpair"])
 
-
+@check_pdf_is_montecarlo
```
RoyStegeman, line 77: Why? Shouldn't this also work for Hessian pdfs?

RoyStegeman, line 77: Why? Shouldn't this also work for Hessian pdfs?

RoyStegeman (commented):
This seems alright. The diffs are a real mess though..

validphys2/src/validphys/core.py
Zaharid: AFAICT rep0 would be the index 0, not rep1.
Also we should check if we still need this method now that it is trivial.

validphys2/src/validphys/sumrules.py
scarlehoff: I guess I was responding to the previous comment :laughing:

n3fit/src/n3fit/vpinterface.py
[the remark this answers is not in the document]
scarlehoff: The use of `n` for the replica is quite common: https://github.com/NNPDF/nnpdf/blob/2ae49281c5d471be490ed5c7bbd28e9211091c19/libnnpdf/src/lhapdfset.cc#L85 (and then later in vp) and since this should basically substitute a vp PDF it's better to keep the same name. Also because not be a replica but a member of some hessian pdf for instance, it just happens to be always a replica in this subclass.

validphys2/src/validphys/sumrules.py
```diff
@@ -123,24 +121,23 @@ def f(x, lpdf, irep, Q):
...
```
scarlehoff, line 140:
```suggestion
```

scarlehoff:
> This seems alright. The diffs are a real mess though..

You are going to hate me so much when you're made to review #1529...

n3fit/src/n3fit/vpinterface.py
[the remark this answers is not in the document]
RoyStegeman: Okay, not ideal but fair enough

scarlehoff:
I think I addressed everything.

scarlehoff:
@Zaharid are you happy merging this? (it's currently blocked by you)

validphys2/src/validphys/results.py
```diff
@@ -547,8 +547,7 @@ def phi_data(abs_chi2_data):
     1410.8849
     """
     alldata, central, npoints = abs_chi2_data
-    cv = float(alldata.central_value())
-    return (np.sqrt((cv - central) / npoints), npoints)
+    return (np.sqrt((alldata.error_members().mean() - central) / npoints), npoints)
```
Zaharid, line 550: Maybe we should make this and `abs_chi2_data` Stats as well, but that is another PR.

validphys2/src/validphys/sumrules.py
```diff
@@ -11,24 +11,23 @@
...
-from validphys.pdfbases import ALL_FLAVOURS, parse_flarr
-from validphys.lhapdfset import LHAPDFSet
+from validphys.pdfbases import parse_flarr
 
 
-def _momentum_sum_rule_integrand(x, lpdf:LHAPDFSet, irep, Q):
-    return sum([lpdf.xfxQ(x, Q=Q, n=irep, fl=fl) for fl in ALL_FLAVOURS])
+def _momentum_sum_rule_integrand(x, lpdf, Q):
+    res = lpdf.xfxQ(x, Q)
+    return sum([res[f] for f in lpdf.flavors()])
```
Zaharid, line 26:
```suggestion
    xqvals = lpdf.xfxQ(x, Q)
    return sum([xqvals[f] for f in lpdf.flavors()])
```

Zaharid (commented):
If Roy is happy I am happy.

Zaharid (approved):

Issue #1519, opened 2022-02-15
covariance_matrix_plots runcard fails with 5bar pt. prescription 


andreab1997:
Running the covariance_matrix_plots runcard with the 5bar point prescription (as in
[covariance_matrix_plots_5barpt_reduced.txt](https://github.com/NNPDF/nnpdf/files/8067133/covariance_matrix_plots_5barpt_reduced.txt)) results in:
**_Choice of input theories does not correspond to a valid prescription for theory covariance matrix calculation_**
 This could be linked to the absence of the flag `fivetheories` in https://github.com/NNPDF/nnpdf/blob/9840fc69dfcbcf25ebb30236916c4f1fc65a1785/validphys2/src/validphys/config.py#L1515

Issue #1521, opened 2022-02-15
validation_report runcard fails with 5 pts prescriptions


andreab1997:
Running the validation_report runcard with one of the 5 pts prescriptions (as in 
[validation_report_5barpt.txt](https://github.com/NNPDF/nnpdf/files/8067338/validation_report_5barpt.txt))
will result in :
_**For five input theories a prescription bar, nobar or linear for the flag fivetheories must be specified**_

andreab1997:
Note however that when running the covariance_matrix_plot runcard, the fivetheories flag is correctly loaded and therefore https://github.com/NNPDF/nnpdf/blob/9840fc69dfcbcf25ebb30236916c4f1fc65a1785/validphys2/src/validphys/theorycovariance/theorycovarianceutils.py#L15
does not fail

Issue #1520, opened 2022-02-15
covariance_matrix_plot runcard fails when executing procs_chi2_table 


andreab1997:
Running the covariance_matrix_plot runcard fails when entering in the function `groups_chi2_table` https://github.com/NNPDF/nnpdf/blob/9840fc69dfcbcf25ebb30236916c4f1fc65a1785/validphys2/src/validphys/results.py#L666  This could be linked to the wrong shape of `groups_each_dataset_chi2` as constructed by https://github.com/NNPDF/nnpdf/blob/9840fc69dfcbcf25ebb30236916c4f1fc65a1785/validphys2/src/validphys/results.py#L663

scarlehoff:
Note that interestingly enough, `groups_data` is grouped correctly instead:

https://github.com/NNPDF/nnpdf/blob/9840fc69dfcbcf25ebb30236916c4f1fc65a1785/validphys2/src/validphys/results.py#L803

despise using exactly the same `collect` syntax! I guess we need @Zaharid for this one.

andreab1997:
Runcard to reproduce the error:
[covariance_matrix_plots_7pt_reduced.txt](https://github.com/NNPDF/nnpdf/files/8080453/covariance_matrix_plots_7pt_reduced.txt)

Zaharid:
I tried reproducing this and got the error

```python-traceboack
Traceback (most recent call last):
  File "/home/zk261/data2/miniconda3/envs/nnpdfdev/bin/validphys", line 33, in <module>
    sys.exit(load_entry_point('validphys', 'console_scripts', 'validphys')())
  File "/mhome/damtp/s/zk261/nnpdf/validphys2/src/validphys/scripts/main.py", line 10, in main
    vp.main()
  File "/mhome/damtp/s/zk261/nnpdf/validphys2/src/validphys/app.py", line 161, in main
    a.main()
  File "/home/zk261/data2/miniconda3/envs/nnpdfdev/lib/python3.9/site-packages/reportengine/app.py", line 395, in main
    self.run()
  File "/mhome/damtp/s/zk261/nnpdf/validphys2/src/validphys/app.py", line 156, in run
    super().run()
  File "/home/zk261/data2/miniconda3/envs/nnpdfdev/lib/python3.9/site-packages/reportengine/app.py", line 380, in run
    rb.execute_sequential()
[...]

PR #1524, opened 2022-02-17
Fixed bugs of previous theorycovariance runcards and added new working runcards

andreab1997:
Fixes issues #1519  and #1521 . Note that 

1.  The flags _fivetheories_ and _seventheories_ are not used anymore in the runcard. The wanted prescription must be specified in _point_prescription_ instead. Anyway the _point_prescription : '7original point'_ and _point_prescription : '5linear point'_ will not work because they are not implemented in  **validphys2/src/validphys/scalevariations**
2.  Since the correct theoryids are now loaded directly by https://github.com/NNPDF/nnpdf/blob/9840fc69dfcbcf25ebb30236916c4f1fc65a1785/validphys2/src/validphys/config.py#L1515 the _dataspecs_ entry in the runcard seems to have no use. Anyway, since it could be used by other functions, maybe we should add a check to be sure that it corresponds to the correct _point_prescription_.
3. The _covariance_matrix_plots_ family of new runcards uses a new template (_template_matrix_plot_new_) which does not include some of the functions of the original one. That is because the issue #1520 has not been solved yet.
4. All the default values of _point_prescription_, _fivetheories_ and _seventheories_ have been removed.
5. Most of the previous bugs were caused by an incorrect order of arguments in some function calls.


AleCandido:
Re: `dataspecs`

To me, it would make sense to remove it, since it's not used here and so it should be relevant nowhere.
I investigated the places where `"theoryconfig"` is used (since the `dataspecs` we are talking about is the child of that, while `shiftconfig.dataspecs` or any other one is a different thing).

However, it looks like is used also:
- here:
  https://github.com/NNPDF/nnpdf/blob/f8f9452c6de99a3d74bef238ecbb6859ed0def25/validphys2/src/validphys/theorycovariance/tests.py#L189
  and I guess this is used in the function right below, and even in `check_correct_theory_combination_dataspecs` in `theorycovariance/theorycovarianceutils.py`
- and here:
  https://github.com/NNPDF/nnpdf/blob/f8f9452c6de99a3d74bef238ecbb6859ed0def25/validphys2/src/validphys/config.py#L879-L893
  but this looks like it's used nowhere

In any case, if the fix is working, let's keep the PR atomic, and if you wish to keep trying removing `dataspecs` (that is a good thing, since you believe not to be relevant any longer), then open another one.

AleCandido:
About 3., the `template_matrix_plot_new` I believe to be `validphys2/examples/theory_covariance/working_runcards/template_matrix_plots_new.md`, is it correct?

Why some runcards (both in `working_runcards` and outside) are still using the other one?
```
>>> np.finfo(np.float32)
finfo(resolution=1e-06, min=-3.4028235e+38, max=3.4028235e+38, dtype=float32)
>>> np.finfo(np.float16)
finfo(resolution=0.001, min=-6.55040e+04, max=6.55040e+04, dtype=float16)
pkgconfig.exists("lhadpf") and shutil.which("lhapdf")
 tarpath = pathlib.Path(tarname)
if tarpath.suffix != ".tar":
    raise ValueError(f"'{tarname}' is not a valid tar filename, wrong suffix")
with tarfile.open(tarpath, "r") as tar:
    tar.extractall(tmpdir)
innerdir = list(tmpdir.glob("*"))[0]
#include <fastnlotk/fastNLOLHAPDF.h>

using namespace std;

int main()
{
    auto file = fastNLOLHAPDF("h1.tab", "CT10");

    return 0;
}
black = "black ."
isort = "isort ."
format = ["black", "isort"]
l = [{0: 'x'}, {0:'y'}, {0:'z', 1:'c'},]
a = {1: 'a'}
b = {1: 'b'}


d = {'l': l, 'a': a, 'b': b}
validphys2/examples/theory_covariance/covariance_matrix_plots.yaml
76:template: template_matrix_plots.md

validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_5nobarpt_reduced.yaml
44:template: template_matrix_plots_new.md

validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_9pt_reduced.yaml
51:template: template_matrix_plots_new.md

validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_5barpt_reduced.yaml
46:template: template_matrix_plots_new.md

validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_7pt_reduced.yaml
48:template: template_matrix_plots_new.md

validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_3fpt.yaml
77:template: template_matrix_plots.md

validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_3pt.yaml
77:template: template_matrix_plots_new.md

validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_3rpt.yaml
77:template: template_matrix_plots.md
```

And I did not understand the plan. Once #1520 will be solved, will this new template be dropped or not?

andreab1997:
> About 3., the `template_matrix_plot_new` I believe to be `validphys2/examples/theory_covariance/working_runcards/template_matrix_plots_new.md`, is it correct?
> 
> Why some runcards (both in `working_runcards` and outside) are still using the other one?
> 
> ```
> validphys2/examples/theory_covariance/covariance_matrix_plots.yaml
> 76:template: template_matrix_plots.md
> 
> validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_5nobarpt_reduced.yaml
> 44:template: template_matrix_plots_new.md
> 
> validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_9pt_reduced.yaml
> 51:template: template_matrix_plots_new.md
> 
> validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_5barpt_reduced.yaml
> 46:template: template_matrix_plots_new.md
> 
> validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_7pt_reduced.yaml
> 48:template: template_matrix_plots_new.md
> 
> validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_3fpt.yaml
> 77:template: template_matrix_plots.md
> 
> validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_3pt.yaml
> 77:template: template_matrix_plots_new.md
> 
> validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_3rpt.yaml
> 77:template: template_matrix_plots.md
> ```
> 
> And I did not understand the plan. Once #1520 will be solved, will this new template be dropped or not?

The fact that some of the runcards in the working_runcards folder are still using the old template is a bug and I will fix it soon. Anyway, the two templates are actually the same and I've added the new one only to have a directory (i.e. working_runcards) in which the runcards can be executed safely. 
The plan is then to substitute both the old template and the old runcards (at least the ones which are replaced by the newer versions) once we convince ourself that they are working.

andreab1997:
Currently the last not working pt prescription is the _5linear_ but I was not able to fix it because it is not documented (for example it is not clear to me which theoryids it needs). @Zaharid, since as far as I have understood it was proposed by you, do you want to add some docs or we should just delete it?

doc/sphinx/source/vp/theorycov/examples.rst
```diff
@@ -35,121 +34,100 @@ relevant c-factors.
...
-   
+
    meta:
-      author: Rosalyn Pearson
-      keywords: [theory uncertainties, 3-point]
-      title: NLO 3-point variations for 5 process types - DIS CC, DIS NC, DY, Top, Jets
-    
-   metadata_group: nnpdf31_process
-    
+   author: Andrea Barontini
```
AleCandido, line 39: The indentation here is inconsistent with https://github.com/NNPDF/nnpdf/blame/fix_fivetheories_issues/doc/sphinx/source/vp/theorycov/examples.rst#L228-L231

scarlehoff:
> Currently the last not working pt prescription is the _5linear_ but I was not able to fix it because it is not documented (for example it is not clear to me which theoryids it needs). @Zaharid, since as far as I have understood it was proposed by you, do you want to add some docs or we should just delete it?

Given previous discussions (and yesterday's code meeting) l I think @Zaharid is happy with removing things that have not been used or documented so please go ahead.

Zaharid:
Yes, delete it. I don't really care for it any longer.

validphys2/src/validphys/config.py
```diff
@@ -1505,6 +1505,19 @@ def produce_group_dataset_inputs_by_metadata(
...
+        elif point_prescription == "5 point":
+            return 'nobar'
+        return None 
+    
+    def produce_seventheories(self,point_prescription):
+        if point_prescription == "7 point":
+            return None
+        elif point_prescription == "7original point":
+            return 'original'
+        return None
```
scarlehoff, line 1520:
```suggestion
    def produce_fivetheories(self, point_prescription):
        if point_prescription == "5bar point":
            return "bar"
        elif point_prescription == "5 point":
            return "nobar"
        return None 
    
    def produce_seventheories(self, point_prescription):
        if point_prescription == "7 point":
            return None
        elif point_prescription == "7original point":
            return "original"
        return None
```

Why is the first `None` in `7 point` correct?

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -120,7 +120,7 @@ def total_covmat_datasets(
 
 @check_correct_theory_combination
 def total_covmat_diagtheory_datasets(
-    each_dataset_results_bytheory, fivetheories: (str, type(None)) = None
+    each_dataset_results_bytheory, fivetheories
```
scarlehoff, line 123: For this one (and a few more in this file) you can change them to be just one line since now they are more or less small:

`def total_covmat_diagtheory_datasets(each_dataset_results_bytheory, fivetheories)`

validphys2/src/validphys/theorycovariance/output.py
```diff
@@ -270,7 +270,7 @@ def plot_thblockcorrmat_heatmap(theory_blockcorrmat):
 def plot_thcorrmat_heatmap_custom(
     theory_corrmat_custom,
     theoryids,
-    fivetheories: (str, type(None)) = None,
+    fivetheories,
```
scarlehoff, line 273: Same as the previous comment, but for this file you can just run `black` on it. `config.py` is big and modified by many people, this one will be mostly worked by you in the future so it's fine.

scarlehoff (changes requested):
I added a few comments (regarding style). I still have to go through the documentation and test the runcards.

imho for the files inside `theorycovariance/` folder you can just run `black` on all of them.

andreab1997:
> Yes, delete it. I don't really care for it any longer.

Ok perfect

validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_3fpt.yaml
```diff
@@ -0,0 +1,87 @@
...
+  - {dataset: CMSZDIFF12, cfac: [NRM]}
+  - dataset: CMSJETS11
+  - dataset: CMSTTBARTOT
+  - dataset: CMSTOPDIFF8TEVTTRAPNORM
+  - dataset: LHCBZ940PB
+  - dataset: LHCBZEE2FB
+  - {dataset: LHCBWZMU7TEV, cfac: [NRM]}
+  - {dataset: LHCBWZMU8TEV, cfac: [NRM]}
+
+template: template_matrix_plots.md
```
scarlehoff, line 77: For this one should one use the `template_matrix_plots_new` or the old one?

validphys2/src/validphys/config.py
andreab1997: because here 
https://github.com/NNPDF/nnpdf/blob/16e2ac648853ddadf68feb23d288f243c65d5126/validphys2/src/validphys/theorycovariance/construction.py#L395
the Gavin version of the 7pt prescription is the default one. So it is executed when seventheories is None

scarlehoff: Ah, right. I remember now. Could you add a comment? (so that when I get confused again after two weeks I know the answer)

validphys2/src/validphys/theorycovariance/construction.py
andreab1997: I believe that this has been done automatically by black now.

validphys2/examples/theory_covariance/working_runcards/covariance_matrix_plots_3fpt.yaml
andreab1997: The new one ( I fixed now). Anyway all this is just a temporary solution because as you remember we had the problem with the chi2 calculation (now solved I believe). Therefore, the new template is just the same of the old one but without the chi2 functions.

validphys2/src/validphys/config.py
andreab1997: yes, of course

scarlehoff (approved):
thanks!

PR #1526, opened 2022-02-17
[WIP] Reader proposal for the commondata proposal

scarlehoff:
In the root of the repository there is a jupyter notebook that should be able to work with this branch of the code.

I made two small corrections to the proposal yaml/json files (I commented about them in the other PR). The only actual change is the variants of course.

What I did:

Added a `KinematicsSpec` and `UncertaintiesSpec`. These two are loaded inside the commondata file when the user asks explicitly for kinematics or uncertainties since they might be changed by the metadata.

At the moment `CommonDataSpec` takes as input the datafile (as before) but in the light of the discussion today maybe the loader should _only_ read up the metadata (and the information of `dataset_input`) and everything else done afterwards.

`dataset_input` gains a new key: `variant`

The `CommonData` coming from `libNNPDF` has ben removed. The new format I think might please @Zaharid since now not only nothing gets loaded until the user explicitly ask for it but if the user only needs, say, the data (no uncertainties), the kinematics or the uncertainties, that's the only thing that will get loaded. Instead, in the current version, `CommonData` had to load everything.

Note: the "new loader" already works by setting the environment path  `BUILDMASTER_PATH` to wherever the commondata folders are, since at the moment they are not installed.


validphys2/src/validphys/core.py
```diff
@@ -266,83 +264,150 @@ def get_kinlabel_key(process_label):
...
-    with open(commondatafilename) as f:
-        try:
-            l = f.readline()
-            name, nsys_str, ndata_str = l.split()
-            l = f.readline()
-            process_type_str = l.split()[1]
-        except Exception:
-            log.error(f"Error processing {commondatafilename}")
-            raise
+class UncertaintiesSpec(TupleComp):
```
Zaharid, line 268: Maybe use dataclasses here instead?

Zaharid: Probably wouldn't call these things "specs" as they are realized objects.

Zaharid:
> At the moment `CommonDataSpec` takes as input the datafile (as before) but in the light of the discussion today maybe the loader should _only_ read up the metadata (and the information of `dataset_input`) and everything else done afterwards.

I think that would be good.

Zaharid:
One benefit of the python implementation of commondata was that there was an index for everything and you could be sure that point n was the nth point in the order it was implemented regardless of cuts and the like.

I wonder if we could have something like that here. Conceptually the index is the kinematics file and it is "required" in order to load either uncertainties or central values, at least insofar we want to make sense of what the data is.

As it is now, there is not much stopping these things from being mismatched.

scarlehoff:
> One benefit of the python implementation of commondata was that there was an index for everything and you could be sure that point n was the nth point in the order it was implemented regardless of cuts and the like.

What do you mean by this? (do you mean C++ and not python?)

The data is the raw data but we can load the data as a indexed dataframe instead of an array (so if later on it is cut the index is "kept"). Is this what you mean?

>  I wonder if we could have something like that here. 

Yes, we can. To first approximation everything is possible here (and easy to shift around). That first commit is a minimal thing able to read up, but your imagination is the limit.

I wonder whether we should open some collab with the jupyter notebook...

scarlehoff:
I've updated everything taking into account some of the discussion in this thread. It's quite different from the first version.

As @Zaharid suggested, now everything held is a `DataFrame` and the only non-derived quantities are indeed the original `DataFrames` for Kinematics, Uncertainties and Data. This facilitates thing as one doesn't need to keep track of the number of points but rather look explicitly at how many are left after the cuts (for instance).

The uncertainties are not fully working since the proposal is going to change a bit but the base is already there. Please have a look at the jupyter notebook, the output at the end of the day is the same `coredata.CommonData` of current validphys, the only difference is that now `Kinematics` and `Uncertainties` live in their own instances of fancy dataframes.

The loading and parsing I am currently doing in `CommonDataSpec`. It is very self-contained so after we have converged it can be copied into the appropiate functions inside `commondataparser.py`.

scarlehoff:
Reader updated with the latest changes by @enocera
It's much cleaner now with the `yaml` file (while the result are equivalent to the previous ones)

scarlehoff:
There are two parts in this PR:

1. The reading of the data, that's all in `core.py` and it looks quite complicated, probably because I don't know well enough how to create pandas objects. I'm not very happy with this part.
2. The validphys python object for uncertainties, kinematics and commondata. That's all in `coredata.py`. I'm much happier with this part. They are all dataframes and always treated as dataframes. Some of them were already there in the past, some substitute the C++ version.

In all honesty, I think (1) could be moved to the module that generates the commondata, so that we have a module (call it buildmaster) that generates the data and knows how to read the data. We only define the dataframe the data has to be coerced into since this we know won't change (the data will have a central value, it will have kinematic variables, the uncertainties can be systematic or statistical, the treatment can be additive or multiplicative, etc).
@AleCandido this goes in the direction of what we were talking about yesterday, it only depends on `pandas` so one could even probably convince experimentalists that using the format is reasonable)

(2) instead will always stay in vp, they are objects that take dataframe and then manipulate them in a way that makes our life easier (extract additive uncertainties, treat multiplicative errors, extract the central value, etc).

validphys2/src/validphys/core.py
```diff
@@ -266,83 +264,153 @@ def get_kinlabel_key(process_label):
...
-            l = f.readline()
-            process_type_str = l.split()[1]
-        except Exception:
-            log.error(f"Error processing {commondatafilename}")
-            raise
+class CommonDataSpec(TupleComp):
 
-    return CommonDataMetadata(name, int(nsys_str), int(ndata_str),
-                              get_kinlabel_key(process_type_str))
+    def __init__(self, name, variant, metadata_file=None):
```
Zaharid, line 270: Doesn't look like `None` is a valid default for `metadata_file` if you are trying to open it next.

validphys2/src/validphys/core.py
```diff
@@ -266,83 +264,153 @@ def get_kinlabel_key(process_label):
...
+
+        # Prepare to read the defaults
+        dat = metadata["data_central"]
+        unc = metadata["data_uncertainties"]
+        kin = metadata["data_kinematics"]
+
+        # Check whether the variants override any of the files
+        # TODO: various try-except in case things don't work as expected
+        if variant is None:
+            variant_info = metadata[variant]
```
Zaharid, line 282: Is `variant` a required key then?

validphys2/src/validphys/core.py
```diff
@@ -266,83 +264,153 @@ def get_kinlabel_key(process_label):
...
+        metadata = yaml.safe_load(metadata_file.open("r", encoding="utf-8"))
+
+        # Prepare to read the defaults
+        dat = metadata["data_central"]
+        unc = metadata["data_uncertainties"]
+        kin = metadata["data_kinematics"]
+
+        # Check whether the variants override any of the files
+        # TODO: various try-except in case things don't work as expected
+        if variant is None:
```
Zaharid, line 281: not None?

validphys2/src/validphys/core.py
Zaharid, line 270: Do we really need `name` and `metadata_file`? Could we not have just a folder and derive the name from that?

validphys2/src/validphys/core.py
```diff
@@ -266,83 +264,153 @@ def get_kinlabel_key(process_label):
...
+class CommonDataSpec(TupleComp):
 
-    return CommonDataMetadata(name, int(nsys_str), int(ndata_str),
-                              get_kinlabel_key(process_type_str))
+    def __init__(self, name, variant, metadata_file=None):
+        parent = metadata_file.parent
+        metadata = yaml.safe_load(metadata_file.open("r", encoding="utf-8"))
+
+        # Prepare to read the defaults
+        dat = metadata["data_central"]
```
Zaharid, line 275: Are we going to allow commondata without central values, e.g. for positivity?

validphys2/src/validphys/core.py
```diff
@@ -266,83 +264,153 @@ def get_kinlabel_key(process_label):
...
     @property
     def nsys(self):
-        return self.metadata.nsys
+        return self.uncertainties.nsys
 
     @property
     def ndata(self):
-        return self.metadata.ndata
+        """Return the number of datapoints"""
+        return len(self.data)
```
Zaharid, line 368: We should not assume we have `data` for this.

validphys2/src/validphys/core.py
```diff
@@ -266,83 +264,153 @@ def get_kinlabel_key(process_label):
...
     @property
     def process_type(self):
-        return self.metadata.process_type
-
-    @property
-    def metadata(self):
-        if self._metadata is None:
-            self._metadata = peek_commondata_metadata(self.datafile)
-        return self._metadata
+        """Return the number of systematic uncertainties"""
```
Zaharid, line 372: Doesn't look like the right docstring.

validphys2/src/validphys/core.py
```diff
@@ -266,83 +264,153 @@ def get_kinlabel_key(process_label):
...
+        return self.metadata.get("proctype")
```
Zaharid, line 373: I guess we need to discuss what to do with process types, as it is a bit ambiguous. On one hand there is "the string that identifies what kinematics mean" (which is roughly what we have in the current commondata files). We may or may not need that with more extensive metadata. Then there are random groupings that people pull out of their hat (e.g. the process type for thcovmat), which we do need to store somewhere.

cc @enocera

validphys2/src/validphys/core.py
```diff
@@ -266,83 +264,153 @@ def get_kinlabel_key(process_label):
...
         return self.name
 
     def __iter__(self):
-        return iter((self.datafile, self.sysfile, self.plotfiles))
-
-    @functools.lru_cache()
-    def load(self)->CommonData:
-        #TODO: Use better path handling in python 3.6
-        return CommonData.ReadFile(str(self.datafile), str(self.sysfile))
+        return iter((self.data, self.uncertainties))
```
Zaharid, line 379: Don't believe we want this.

validphys2/src/validphys/core.py
```diff
@@ -266,83 +264,153 @@ def get_kinlabel_key(process_label):
...
     def plot_kinlabels(self):
         return get_plot_kinlabels(self)
 
+    def load(self):
+        """Load completely the dataset, creating a full CommonData object"""
+        # Load all information
+        d = self.data
+        k = self.kinematics
+        u = self.uncertainties
+        return CommonData(self.name, self.variant, self.process_type, d, k, u)
```
Zaharid, line 391: Just to note that this is subject to the various comments above.

validphys2/src/validphys/coredata.py
```diff
@@ -9,6 +9,117 @@
...
+    return df
+
+
+@dataclasses.dataclass
+class _FancyDataframe:
+    """Holds a dataframe in a dataclass to which vp cuts can be applied"""
+
+    dataframe: pd.DataFrame
+
+    def __post_init__(self):
```
Zaharid, line 49: Doesn't seem needed

scarlehoff: It looks I forgot that here at some point.

Did you look at the jupyter notebook? I would be interested on your opinion about how the data get saved and "understood" in terms of the dataframes. The methods, attributes and so on we can polish afterwards.

validphys2/src/validphys/core.py
scarlehoff: I guess the question/decision is "it's default a variant?". I like to think it is but I don't have a strong opinion. I should prepare a list with these points.

validphys2/src/validphys/core.py
```diff
@@ -266,83 +264,153 @@ def get_kinlabel_key(process_label):
...
-        return self.metadata.name
+    def data(self):
+        """Reads the data as a 1-d numpy array"""
+        if self._data is None:
+            datayaml = yaml.safe_load(self._data_file.read_text(encoding="utf-8"))
+            self._data = pd.DataFrame.from_records(datayaml["data_central"], index="index")
+            self._data.rename(columns={self._data.columns[0]:"data"}, inplace=True)
+        return self._data
+
+    @property
```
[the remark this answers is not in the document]
AleCandido: `validphys` at least declares support for [3.6 on](https://github.com/NNPDF/nnpdf/blob/7ceadf8259bc361d0a86fcc542ff22b3b454e72d/validphys2/setup.py#L5-L7), while `cached_property` has been introduced in 3.8.
I have no strong opinion, but the two things together are inconsistent, so you can stop supporting 3.6 and 3.7 (3.6 is barely used at this point, and even 3.7 has limited user base) and write it in `validphys`, or you have to give up on `cached_property`.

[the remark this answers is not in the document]
Zaharid: Effectively we support whatever we build packages for, which is 3.8 and 3.9. PRs to specify this in other places would be welcome.

validphys2/src/validphys/core.py
scarlehoff: Yeah, that part is broken. I forgot I had broken it though. Still, related to your other point. I agree, the `CommonData` is (should be) fully defined by its metadata file and variant, no need for the name.

validphys2/src/validphys/core.py
scarlehoff: Yes! I didn't notice because the data is data anyway :_

validphys2/src/validphys/core.py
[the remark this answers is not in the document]
scarlehoff: Yeah, we decided that anything not in conda was not supported. But since I use 3.10 I'm always wary of introducing new stuff :_ 
I should also introduce also the walrus operator everywhere, I really like it and thought it was 3.9 but it's 8!

validphys2/src/validphys/coredata.py
scarlehoff: Ah, no, it's needed because otherwise children classes won't call `__post_init__`. Don't know whether I did something wrong in the children classes, but that's the reason.

Zaharid: The children classes should have the dataclass decorator as well.

AleCandido: I guess the `@dataclass` is a decorator somehow `vp`-like: it reads you're code, and generate some other code, e.g. the `__init__` function, out of it.
You will inherit that `__init__`, but not the `__init__` generation.

PR #1528, opened 2022-02-22
Fit with theory covmat with n3fit

andreab1997:
This will make possible to run a NNPDF4.0 fit including theory_covariance matrix

Edit: it also modifies the replica generation (`make_replica`) to utilize the covmat (with theory errors and multiplicative uncertainties if the proper flags are utilized)


n3fit/src/n3fit/scripts/n3fit_exec.py
```diff
@@ -145,7 +145,13 @@ def from_yaml(cls, o, *args, **kwargs):
             validation_action = namespace + "validation_pseudodata"
 
             N3FIT_FIXED_CONFIG['actions_'].extend((training_action, validation_action))
-
+        N3FIT_FIXED_CONFIG['theory_covmat_flag'] = False
+        N3FIT_FIXED_CONFIG['use_user_uncertainties'] = None
+        N3FIT_FIXED_CONFIG['use_scalevar_uncertainties'] = None
```
scarlehoff, line 150: Is there a better way to set these defaults validphys-wise @Zaharid ?

Zaharid: Maybe use production rules. But this is fine really.

validphys2/src/validphys/config.py
```diff
@@ -1504,6 +1502,11 @@ def produce_group_dataset_inputs_by_metadata(
             {"data_input": NSList(group, nskey="dataset_input"), "group_name": name}
             for name, group in res.items()
         ]
+    
+    def produce_group_dataset_inputs_by_fitting_group(self, data_input, theory_covmat_flag):
+        if theory_covmat_flag is True:
+            return self.produce_group_dataset_inputs_by_metadata(data_input, "custom_group")
```
scarlehoff, line 1508: We need to define a new grouping that just takes everything (if only to avoid the `unset` flag).

A possibility would be to use here `BIGEXP` or `ALL` and then in
https://github.com/NNPDF/nnpdf/blob/7b0ea262b48029e51fcd21a6a7ab83a3a5e6fdba/validphys2/src/validphys/config.py#L1475

change it so that if it receives `ALL` it doesn't look into the metadata and just groups everything together with `name=ALL`.

validphys2/src/validphys/covmats.py
```diff
@@ -227,6 +228,17 @@ def dataset_inputs_covmat_from_systematics(
...
+   #     theory_covmat_path = pathlib.Path.cwd()
+   #     data = pd.read_csv(theory_covmat_path / "prov_moredata" / "tables" / "datacuts_theory_theorycovmatconfig_theory_covmat_custom.csv", sep='\t')
+   #     datael = data.iloc[3:]
+    #    datael = datael.drop(['group'], axis=1)
+    #    datael = datael.drop(['Unnamed: 1'], axis=1)
+    #    datael = datael.drop(['Unnamed: 2'], axis=1)
+    #    theory_covmat = np.copy(datael.values)
+    #except FileNotFoundError:
+    #    theory_covmat = np.zeros(covmat.shape) 
+    #total_covmat = np.add(covmat, theory_covmat)
```
scarlehoff, line 241:
```suggestion
```

validphys2/src/validphys/covmats.py
```diff
@@ -339,6 +351,41 @@ def dataset_inputs_t0_covmat_from_systematics(
...
+            if use_user_uncertainties is True:
+                generic_path = pathlib.Path("datacuts_theory_theorycovmatconfig_total_theory_covmat.csv")
+            else:
+                generic_path = pathlib.Path("datacuts_theory_theorycovmatconfig_theory_covmat_custom.csv")
+        else:
+            if use_user_uncertainties is True:
+                generic_path = pathlib.Path("datacuts_theory_theorycovmatconfig_user_covmat.csv")
+            else:
+                generic_path = pathlib.Path("datacuts_theory_theorycovmatconfig_theory_covmat_custom.csv")
+        theorypath = pathlib.Path(str(output_path/"tables"/generic_path.relative_to(generic_path.anchor)))
```
scarlehoff, line 383: I guess output_path is already a path so this (I think, please test!) should do the trick.

```suggestion
       theorypath = output_path / "tables" / generic_path
```

scarlehoff (changes requested):
Could you also add one example runcard (and maybe a regression test for the invcovmat?)

Also -for @Zaharid - would you be against using `.npy` instead of `.csv`? I'm don't have a strong opinion on this but it would make things a bit faster and leaner (for a 4000x4000 matrix this is 382M vs 123M)

Zaharid:
> Also -for @Zaharid - would you be against using `.npy` instead of `.csv`? I'm don't have a strong opinion on this but it would make things a bit faster and leaner (for a 4000x4000 matrix this is 382M vs 123M)

Yes please, do that.

Zaharid:
Or, compute the covmat dynamically, not sure if that is better.

Zaharid:
We also had https://github.com/NNPDF/nnpdf/pull/1091 which turned out to be a bit more involved and was forgotten about. Also probably @AleCandido convinced me that npy is reasonable enough for this kind of purposes.

scarlehoff:
As a note, I think having a csv in the regression tests is good. Having the changes in human readable form is nice (you can always parse them of course). But anyway, this would only be for the storing of the thcovmat.

Although maybe it makes sense to compute it on the fly indeed. @andreab1997 how long does setupfit take?

andreab1997:
> As a note, I think having a csv in the regression tests is good. Having the changes in human readable form is nice (you can always parse them of course). But anyway, this would only be for the storing of the thcovmat.
> 
> Although maybe it makes sense to compute it on the fly indeed. @andreab1997 how long does setupfit take?

For the runcards I tried, it takes about 40 minutes against the 20 of a fit without theory covmat

scarlehoff:
> For the runcards I tried, it takes about 40 minutes against the 20 of a fit without theory covmat

The I guess it's better to store it and load it.

validphys2/src/validphys/config.py
andreab1997: Do you know where the groups are defined?

scarlehoff: The group themselves in the metadata in the plotting files. But I don't think this should be a group (since it should use all datasets) but rather a special case in `produce_group_dataset_inputs_by_metadata`.

andreab1997: Ok but I believe that the return of produce_group_dataset_inputs_by_metadata is still expecting one group over which iterate. So what I can do is to call this group "ALL" or something like that (at least to avoid the name to be "unset"). Is this what you mean?

scarlehoff: Yes exactly!
Instead of using the `custom_group` (which is a group and doesn't necessarily mean all datasets).

andreab1997:
I believe that the only left change to apply is the one related to .csv vs .npy, right? Anyway, I have not understood the final decision, should we dump the thcovmat in csv or npy? (leaving as given that we do not want to compute it online) 

PS: @scarlehoff there are still your requested changes blocking the merging but I took already care of them. How can I solve this problem?

scarlehoff:
> Anyway, I have not understood the final decision, should we dump the thcovmat in csv or npy? 

Yes, please, using `.npy` will be faster and will use less memory!

> How can I solve this problem?

You can't! I have to look through the changes!

AleCandido:
> PS: @scarlehoff there are still your requested changes blocking the merging but I took already care of them. How can I solve this problem?

I guess the proper solution would be to re-request a review: if you solved them, you should allow your reviewer to check that the solution is satisfactory.

andreab1997:
> > PS: @scarlehoff there are still your requested changes blocking the merging but I took already care of them. How can I solve this problem?
> 
> I guess the proper solution would be to re-request a review: if you solved them, you should allow your reviewer to check that the solution is satisfactory (and not breaking anything else...).

ok done, thank you :)

andreab1997:
> > Anyway, I have not understood the final decision, should we dump the thcovmat in csv or npy?
> 
> Yes, please, using `.npy` will be faster and will use less memory!
> 
> > How can I solve this problem?
> 
> You can't! I have to look through the changes!

I am sorry, I am having problems finding the place where the writing happens. Can you please suggest me the right place where to look?(I believed this should have been in produce_nnfit_theory_covmat but I cannot find the writing) @scarlehoff @Zaharid

validphys2/src/validphys/pseudodata.py
```diff
@@ -159,14 +159,14 @@ def make_replica(groups_dataset_inputs_loaded_cd_with_cuts, replica_mcseed,
     #eliminate negative eignvalues of thcovmat and compute sqrt
     chol_theory_covmat = np.zeros(shape = (1,1))
     if theory_covmat_flag is True:
-        tr = 1.e-6
+        tr = 1.e-3
```
AleCandido, line 162: Why so big?

scarlehoff:
> Although maybe it makes sense to compute it on the fly indeed. 

I'm going to reply to myself here, it's _much_ better to not compute it on the fly even if it's fast because otherwise I have to download all 9 theories to my laptop even for simple debugging. I'm happy that I can use one from a previous fit (or computed in a different computer where all the theories are) so that I only need to download the central theory.

AleCandido:
> I'm going to reply to myself here, it's _much_ better to not compute it on the fly even if it's fast because otherwise I have to download all 9 theories to my laptop even for simple debugging. I'm happy that I can use one from a previous fit (or computed in a different computer where all the theories are) so that I only need to download the central theory.

Can we make a switch?
If you pass a filename(/path) with the thcovmat it attempts to use it, and do not recompute, while if it's `None`, it attempts recomputation.

scarlehoff:
My proposal would be actually to have a repository of covmats for the different prescriptions, then when you need one you just download it (instead of downloading all the theories!) and then mask away the data you don't want.

The 9 points prescriptions will take us up to ~18GB of RAM even with the optimized pineappl grids (so no wonder it currently takes about 1 hour in a laptop).

AleCandido:
> so no wonder it currently takes about 1 hour in a laptop

1h on the cluster

scarlehoff:
Right. But I don't think a laptop would fare much better once you get above what it can hold in RAM.

AleCandido:
> Right. But I don't think a laptop would fare much better once you get above what it can hold in RAM.

However, we don't need all the datasets in memory at the same time:
- we compute the array of shifts process by process (and if needed even point by point, i.e. shift by shift)
- you compute predictions for the central value
- you compute predictions for a given point, and subtract the central value to obtain the shift

So, true that you have to load 18GB, that time is needed, but you don't have to *hold* them in memory for the whole time: once predictions are computed, you can drop it (at least for what thcovmat is concerned).

scarlehoff:
> However, we don't need all the datasets in memory at the same time:

No but they are -minimum- still 18GB that you will have to download and save to disk. Then you have to change how the thcovmat is computed. And then you still 18GB of useless fktables (since they are not used for the fit).

Instead if you just download the covmat you only need to download, load and use a `.npy` file of 200M max.

AleCandido:
Yes, that's fine: I have nothing against the storage of a few thcovmat in `.npy` files, that's perfectly feasible.

Even though computing the covmat itself in `numpy` take a few seconds, it will be dominated by download time (plus sub-leading load time, that is still much more than `numpy` algebra).

Still, whenever/wherever we have the theories at hand (usually not on laptops) we can still allow for on-the-fly recomputation (e.g. if we want to experiment with fancy point prescriptions, instead of storing all of them somewhere).

In any case, we need a function able to compute the thcovmats for initial generation, so I see nothing wrong in making it available on-the-fly. It will be opt-in, you need to specify theories' path in any case (what I would definitely avoid is on-the-fly theories download).

scarlehoff:
Just to clarify, I was not saying that the ability of doing it on demand should be removed, only "enhanced".

validphys2/src/validphys/pseudodata.py
```diff
@@ -148,77 +148,29 @@ def make_replica(groups_dataset_inputs_loaded_cd_with_cuts, replica_mcseed, genr
...
 
     # The inner while True loop is for ensuring a positive definite
     # pseudodata replica
     while True:
         pseudodatas = []
-        special_add = []
-        special_mult = []
-        mult_shifts = []
         check_positive_masks = []
         for cd in groups_dataset_inputs_loaded_cd_with_cuts:
```
scarlehoff, line 160: The changes I did for the additive errors might be savaged to remove this loop: https://github.com/NNPDF/nnpdf/pull/1544/files

which will speed up even more the generation of replicas (which is important given the amount of negative pseudodata we get)

validphys2/src/validphys/covmats.py
```diff
@@ -339,6 +342,102 @@ def dataset_inputs_t0_covmat_from_systematics(
...
+        data_input,
+        use_weights_in_covmat,
+        norm_threshold=norm_threshold,
+        _list_of_central_values=dataset_inputs_t0_predictions
+    )
+    if theory_covmat_flag and use_thcovmat_in_fitting:
+            #Adding regularization to theory covmat to make sqrt possible  
+            diag_enha = 1.e-6
+            return exp_covmat + loaded_theory_covmat*(np.ones_like(loaded_theory_covmat) + diag_enha*np.eye(loaded_theory_covmat.shape[0]))
+    return exp_covmat
```
scarlehoff, line 439: I think for these two functions (and all the branches it created) we need to rethink a bit the names and what we want to have.

Imho we should have `sampling_total_covmat` and `fitting_total_covmat` (or something like that) so that it's clear what they should be used for.

Personally I would just enforce that `sampling` does not include `t0` and `fitting` does but yesterday talking with @Radonirinaunimi he was interested in having the possibility of setting that flag on and off so we might as well included (since you already did).

We can sketch the different names / branches that we want in a blackboard when I'm at the office, it will be faster.

validphys2/src/validphys/pseudodata.py
andreab1997: I don't understand, we don't use anymore the additive and multiplicative errors here. We only loop to collect the pseudodata, right?

validphys2/src/validphys/covmats.py
andreab1997: Ok sure. Till then I leave the names as they are now

andreab1997: BTW I believe that we should leave the flags for t0 as well

validphys2/src/validphys/pseudodata.py
scarlehoff: Yes, but you are still looping over all datasets for every iteration of the outer `while`. My point is that you now can do that just once as I did here https://github.com/NNPDF/nnpdf/pull/1544/files and extract all the central values and the positivity mask.

Zaharid:
I am a bit uncomfortable with the pseudodata generation changing so much in this PR. Would probably prefer that to be in a separate PR. Also would be good to 

1) Have a summary of what is changing in the pseudodata.
2) Know how compatible is it, in particular if one could still reproduce the random data given the seeds.
3) Decide if we want to have several ways to generate pseudodata e.g. for compatibility.

scarlehoff:
> Know how compatible is it, in particular if one could still reproduce the random data given the seeds.

What do you mean by this? You won't be able to get the previous one of course but they should be reproducible.

(this will be checked by the tests once they are updated)

> Decide if we want to have several ways to generate pseudodata e.g. for compatibility.

I don't think we need this because the python `make_replica` has not been used for any release (NNPDF4.0 was using the C++ version). We are just changing what `make_replica` is before it gets used.
What's more, anything that was using the python `make_replica` was broken until a month ago so I don't think even side projects will need this compatibility layer.

--------

wrt separation, @andreab1997, could you do it? It will be a bit tricky (on the bright side, you will only need to regenerate regression tests on the other branch), let me know if you need help.

andreab1997:
> > Know how compatible is it, in particular if one could still reproduce the random data given the seeds.
> 
> What do you mean by this? You won't be able to get the previous one of course but they should be reproducible.
> 
> (this will be checked by the tests once they are updated)
> 
> > Decide if we want to have several ways to generate pseudodata e.g. for compatibility.
> 
> I don't think we need this because the python `make_replica` has not been used for any release (NNPDF4.0 was using the C++ version). We are just changing what `make_replica` is before it gets used. What's more, anything that was using the python `make_replica` was broken until a month ago so I don't think even side projects will need this compatibility layer.
> 
> wrt separation, @andreab1997, could you do it? It will be a bit tricky (on the bright side, you will only need to regenerate regression tests on the other branch), let me know if you need help.

You mean putting the pseudodata modification in another PR? Yes, of course I can do that

scarlehoff:
> You mean putting the pseudodata modification in another PR? Yes, of course I can do that

Yes (only with the exp covmat)

Zaharid:
> I don't think we need this because the python `make_replica` has not been used for any release (NNPDF4.0 was using the C++ version). We are just changing what `make_replica` is before it gets used.

If I read this explicitly https://docs.nnpdf.science/releases.html we would like the 4.0 runcards to be strictly reproducible until 4.1.

scarlehoff:
> > I don't think we need this because the python `make_replica` has not been used for any release (NNPDF4.0 was using the C++ version). We are just changing what `make_replica` is before it gets used.
> 
> If I read this explicitly https://docs.nnpdf.science/releases.html we would like the 4.0 runcards to be strictly reproducible until 4.1.

The 4.0 runcards were using the C++ `MakeReplica` so that was broken as soon as we changed to `make_replica`. And that was even further broken when the bug in `make_replica` was fixed.

Anywasy, I don't think reproducible should mean even random-number-reproducible as that can change due to other libraries. Only the tag can give you that.

Zaharid:
> The 4.0 runcards were using the C++ `MakeReplica` so that was broken as soon as we changed to `make_replica`. And that was even further broken when the bug in `make_replica` was fixed.

Don't see how that is related: We are not changing the random seed but also the random generation method. As I read it, we are saying that a 4.0 runcard should use multiplicative uncertainties.

scarlehoff:
> Don't see how that is related: We are not changing the random seed but also the random generation method. As I read it, we are saying that a 4.0 runcard should use multiplicative uncertainties.

I see what you mean now. I was confused because you were talking about `seeds` and "strictly reproducible" and both points were already out.

@andreab1997 you can just recover the branch you had with the `_only_additive` with the goodies that you are added here as that is exactly what Zahari is looking for.

scarlehoff:
Now the question is, what should be the `default` behaviour? I would say we add to the runcard (with these defaults):

```yaml
sampling:
   use_t0: false
   use_thcovmat: false # this is added in this PR instead.
   separate_multiplicative: false
```

@andreab1997 probably the best thing is to use the branch with `_only_additive` that you have and create a `sampling_covariance_matrix` method that takes this as input and generates the covmat with the right options according to it?

andreab1997:
> Now the question is, what should be the `default` behaviour? I would say we add to the runcard (with these defaults):
> 
> ```yaml
> sampling:
>    use_t0: false
>    use_thcovmat: false # this is added in this PR instead.
>    separate_multiplicative: false
> ```
> 
> @andreab1997 probably the best thing is to use the branch with `_only_additive` that you have and create a `sampling_covariance_matrix` method that takes this as input and generates the covmat with the right options according to it?

Now the old behaviour should have been restored trough the flag `separate_multiplicative`. In this way when it is set to False we have the new behaviour, namely we use the full covmat (or covmat+thcovmat) to generate replicas, while if it is set to True we construct the exp_covmat for the replica generation without multiplicative errors which are then restored directly in make_replica (as we used to do in the past) after the additive errors. I assumed the default behaviour to be the new one so that `separate_multiplicative` is False by default. @Zaharid, @scarlehoff what do you think about that?

scarlehoff:
> Do have a look at https://github.com/NNPDF/nnpdf/pull/991, seems like some work and discussion happened there already.

What are you referring to exactly? That's a big PR and the mentioned flags have nothing to do with the theory covmat (just whether the multiplicative uncertainties should be separated).

Zaharid:
Ah, sorry, I quoted a comment but didn't work: I referred to the part where you discuss storing a theory covmat and downloading it for later use.

Zaharid:
> 

As discussed, I think the default behaviour should be the current one until after 4.1 so that nothing should change in existing runcards.

scarlehoff:
Ah, yes. That we should do. But that's independent of this PR (we should do that plus use something other than `csv` to save space) but I think for the first working version it makes sense to leave the `csv`.

andreab1997:
> > Now the old behaviour should have been restored trough the flag `separate_multiplicative`. In this way when it is set to False we have the new behaviour, namely we use the full covmat (or covmat+thcovmat) to generate replicas, while if it is set to True we construct the exp_covmat for the replica generation without multiplicative errors which are then restored directly in make_replica (as we used to do in the past) after the additive errors. I assumed the default behaviour to be the new one so that `separate_multiplicative` is False by default. @Zaharid, @scarlehoff what do you think about that?
> 
> As discussed, I think the default behaviour should be the current one until after 4.1 so that nothing should change in existing runcards.

Now the default has been changed

andreab1997 (commented):
This should be ok now

andreab1997:
Since now the tests have been fixed, I request again a review

validphys2/src/validphys/tests/test_pseudodata.py
```diff
@@ -38,8 +38,17 @@ def test_read_pdf_pseudodata():
...
-
+    config = {}
+    config["use_t0_sampling"] = False
+    config["separate_multiplicative"]= True
+    config["output_path"] = None
+    config["theory_covmat_flag"] = False
+    config["use_user_uncertainties"] = None 
+    config["use_scalevar_uncertainties"] = None
+    config["use_thcovmat_in_sampling"]=None
+    config["dataset_t0_predictions"] = None
```
scarlehoff, line 49: Why do you need this? `recreate_fit_pseudodata` should be ale to recreate it from the fit only.

validphys2/src/validphys/tests/test_pseudodata.py
```diff
@@ -49,7 +58,17 @@ def test_recreate_fit_pseudodata():
...
+
+    pdf_pseudodata = API.recreate_pdf_pseudodata(fit=PSEUDODATA_FIT, **config)
```
scarlehoff, line 71: Same as above (and same below)

validphys2/src/validphys/tests/test_pythonmakereplica.py
```diff
@@ -21,9 +22,9 @@
 SINGLE_SYS_DATASETS = [
     {"dataset": "DYE886R"},
     {"dataset": "D0ZRAP", "cfac": ["QCD"]},
-    {"dataset": "ATLAS_SINGLETOP_TCH_R_13TEV", "cfac": ["QCD"]},
-    {"dataset": "CMS_SINGLETOP_TCH_R_8TEV", "cfac": ["QCD"]},
-    {"dataset": "CMS_SINGLETOP_TCH_R_13TEV", "cfac": ["QCD"]},
+    {"dataset": "NMC"},
+    {"dataset": "NMCPD"},
```
scarlehoff, line 26: Could you put another `ATLAS` and a `CMS`? I'm not sure what the logic of the previous one was but better if changes to the tests (which is always dangerous) are as minimal as possible in every sense.

validphys2/src/validphys/tests/test_pythonmakereplica.py
```diff
@@ -38,14 +39,24 @@ def test_commondata_unchanged(data_config, dataset_inputs, use_cuts):
...
     config["use_cuts"] = use_cuts
+    config["replica_mcseed"] = SEED
+    config["use_t0_sampling"] = False
+    config["separate_multiplicative"]= True
+    config["output_path"] = None
+    config["theory_covmat_flag"] = False
+    config["use_user_uncertainties"] = None 
+    config["use_scalevar_uncertainties"] = None
+    config["use_thcovmat_in_sampling"]=None
+    config["dataset_t0_predictions"] = None
```
scarlehoff, line 50: Same as before. I don't like the fact that you need all these `None` and `False`. I need to check in the vp part but I think the problem is that you should not arrive to the functions in which you check that it is `False` or `None`.

validphys2/src/validphys/tests/test_regressions.py
```diff
@@ -54,6 +54,7 @@ def f_(*args, **kwargs):
 
 @make_table_comp(parse_data_cv)
 def test_mcreplica(data_config):
+    ####
```
scarlehoff, line 57:
```suggestion
```

validphys2/src/validphys/tests/test_regressions.py
```diff
@@ -138,10 +139,12 @@ def test_datasetchi2(data_singleexp_witht0_config):
 
 @make_table_comp(sane_load)
 def test_art_rep_generation(data_config):
+    ###
```
scarlehoff, line 142:
```suggestion
```

validphys2/src/validphys/tests/test_regressions.py
```diff
@@ -138,10 +139,12 @@ def test_datasetchi2(data_singleexp_witht0_config):
...
     config = dict(data_config)
     config["dataset_inputs"] = CORR_DATA
     config["mcseed"] = 123456
     config["genrep"] = True
     config["nreplica"] = 1
+
```
scarlehoff, line 148:
```suggestion
```

validphys2/src/validphys/tests/test_pythonmakereplica.py
```diff
@@ -38,14 +42,16 @@ def test_commondata_unchanged(data_config, dataset_inputs, use_cuts):
...
     ld_cds = API.dataset_inputs_loaded_cd_with_cuts(**config)
 
     # keep a copy of all dataframes/series pre make replica
     pre_mkrep_cvs = [deepcopy(cd.central_values) for cd in ld_cds]
     pre_mkrep_sys_tabs = [deepcopy(cd.systematics_table) for cd in ld_cds]
     pre_mkrep_cd_tabs = [deepcopy(cd.commondata_table) for cd in ld_cds]
-
-    make_replica(ld_cds, SEED)
+    
+
```
scarlehoff, line 53:
```suggestion
```

validphys2/src/validphys/tests/test_pseudodata.py
```diff
@@ -70,8 +71,10 @@ def test_no_savepseudodata():
 
 
 def test_read_matches_recreate():
+    
+
     reads = API.read_fit_pseudodata(fit=PSEUDODATA_FIT)
-    recreates = API.recreate_fit_pseudodata(fit=PSEUDODATA_FIT)
+    recreates = API.recreate_fit_pseudodata(fit=PSEUDODATA_FIT,)
```
scarlehoff, line 77: I think this file you can revert to its original state (so that it obvious there is no changes)

validphys2/src/validphys/config.py
```diff
@@ -693,6 +696,83 @@ def produce_experiment_from_input(
...
+    def produce_sep_mult(self,separate_multiplicative=None):
+        if separate_multiplicative is False:
+            return False
+        return True
+
+    @configparser.explicit_node   
+    def produce_dataset_inputs_fitting_covmat(self,theory_covmat_flag=False,
+    use_thcovmat_in_fitting=False,
+    use_t0_fitting=True,
+    ):
```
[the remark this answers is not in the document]
andreab1997: OK yes!

validphys2/examples/theory_covariance/Fit_with_theory_covmat.yml
```diff
@@ -0,0 +1,149 @@
...
+############################################################
+theory:
+  theoryid: 163        # database id
+theorycovmatconfig:
+  point_prescription: "9 point"
+  theoryids:
+    from_: scale_variation_theories
+  pdf: NNPDF31_nlo_as_0118
+  use_thcovmat_in_fitting: true
+  use_thcovmat_in_sampling: true
```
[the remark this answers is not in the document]
andreab1997: Yes, it is needed. You can specify if you want to use the theory_covmat in sampling or not.  If you do not specify it,the default is true anyhow but maybe is better to keep it explicit in this example runcard.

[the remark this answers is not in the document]
scarlehoff: ah, so this is not in the sampling dictionary then?

validphys2/src/validphys/config.py
```diff
@@ -911,13 +991,13 @@ def parse_use_t0(self, do_use_t0: bool):
...
-    def produce_t0set(self, use_t0=False, t0pdfset=None):
+    def produce_t0set(self, t0pdfset=None, use_t0_sampling=False, use_t0_fitting=True, ):
         """Return the t0set if use_t0 is True and None otherwise. Raises an
         error if t0 is requested but no t0set is given.
         """
-        if use_t0:
+        if use_t0_sampling or use_t0_fitting:
             if not t0pdfset:
-                raise ConfigError("Setting use_t0 requires specifying a valid t0pdfset")
+                raise ConfigError("Setting use_t0_* requires specifying a valid t0pdfset")
```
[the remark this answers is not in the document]
andreab1997: Yes, exactly. The difference is only whether you put use_t0 below fitting or sampling.

validphys2/examples/theory_covariance/Fit_with_theory_covmat.yml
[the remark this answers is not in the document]
andreab1997: No, it is in the thcovmat dictionary. In fact if you do not have thcovmat at all, use_thcovmat_in_sampling is None by default

validphys2/src/validphys/config.py
```diff
@@ -693,6 +696,83 @@ def produce_experiment_from_input(
...
+                generic_path = "datacuts_theory_theorycovmatconfig_user_covmat.csv"
+        else:
+            generic_path = "datacuts_theory_theorycovmatconfig_theory_covmat_custom.csv"
+
+        theorypath = output_path/"tables"/generic_path
+        theory_covmat = pd.read_csv(theorypath, index_col=[0, 1, 2], header=[0, 1, 2], sep="\t|,", engine="python").fillna(0)
+        #change ordering according to exp_covmat (so according to runcard order)
+        tmp = theory_covmat.droplevel(0, axis=0).droplevel(0, axis=1)
+        bb = [str(i) for i in data_input]
+        return tmp.reindex(index=bb, columns=bb, level=0).values
```
scarlehoff, line 773:
```suggestion
        dataset_order = [str(i) for i in data_input]
        return theory_covmat.reindex(index=dataset_order, columns=dataset_order, level=1).values
```

This might fix it.

validphys2/src/validphys/covmats.py
```diff
@@ -340,6 +344,186 @@ def dataset_inputs_t0_covmat_from_systematics(
...
+    loaded_theory_covmat,
+    ):
+    """
+    Function to compute the total covmat to be used for the chi2 by fitting_data_dict.
+    Depending on the theory_covmat_flag value, the theory covmat will be added to the 
+    experimental covmat or not. With the use_t0_fitting flag it is possible to choose
+    if using the t0 prescription in the covmat computation. In order to compute the covmat
+    only using the additive errors, you can set only_add to True.
+    """
+    covmat = generate_exp_covmat(dataset_inputs_loaded_cd_with_cuts, data_input, use_weights_in_covmat, norm_threshold, dataset_inputs_t0_predictions , True)
```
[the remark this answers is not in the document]
andreab1997: Yes but I don't think it would be simpler in any case. Moreover, maybe it would become more difficult to follow. However, if you prefer I can change it.

validphys2/src/validphys/config.py
```diff
@@ -55,6 +55,9 @@
 
 import validphys.scalevariations
 
+import numpy as np
+import pandas as pd
+import pathlib
```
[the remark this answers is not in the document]
AleCandido: Yeah, `isort` would complain otherwise: first section stdlib, second third parties, third section local imports.

P.S.: why do you import from `validphys` instead of:
```py
from . import scalevariations
```
?

validphys2/src/validphys/config.py
```diff
@@ -693,6 +696,83 @@ def produce_experiment_from_input(
             )
         }
 
+    def produce_sep_mult(self,separate_multiplicative=None):
+        if separate_multiplicative is False:
+            return False
+        return True
+
+    @configparser.explicit_node
```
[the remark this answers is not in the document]
AleCandido: https://github.com/N3PDF/eko/blob/master/.pre-commit-config.yaml

validphys2/src/validphys/covmats.py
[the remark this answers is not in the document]
scarlehoff: If you use `dataset_inputs_t0_exp_covmat_separate` then `vp` will read up the arguments and find out whether it can run that part in particular and let you know that that's failed (instead of saying that it was `total_covmat`).

In addition it could be cached so you can potentially save calls to `generate_exp_covmat`.

validphys2/src/validphys/config.py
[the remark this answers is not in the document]
AleCandido: Another benefit of `pre-commit` is that it will be natural to adopt the standard incrementally: by default, it acts only on the modified files.

validphys2/src/validphys/covmats.py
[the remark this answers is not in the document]
andreab1997: Ok then I'll change

validphys2/src/validphys/tests/test_pseudodata.py
```diff
@@ -70,6 +71,7 @@ def test_no_savepseudodata():
 
 
 def test_read_matches_recreate():
+
```
scarlehoff, line 74: this file can be reverted to the original state (since the only changes are new lines)

validphys2/src/validphys/tests/test_pythonmakereplica.py
```diff
@@ -38,14 +42,15 @@ def test_commondata_unchanged(data_config, dataset_inputs, use_cuts):
...
+    config["replica_mcseed"] = SEED
     ld_cds = API.dataset_inputs_loaded_cd_with_cuts(**config)
 
     # keep a copy of all dataframes/series pre make replica
     pre_mkrep_cvs = [deepcopy(cd.central_values) for cd in ld_cds]
     pre_mkrep_sys_tabs = [deepcopy(cd.systematics_table) for cd in ld_cds]
     pre_mkrep_cd_tabs = [deepcopy(cd.commondata_table) for cd in ld_cds]
-
-    make_replica(ld_cds, SEED)
+
```
scarlehoff, line 52:
```suggestion
```

validphys2/src/validphys/pseudodata.py
```diff
@@ -100,8 +101,7 @@ def read_replica_pseudodata(fit, context_index, replica):
 
     return DataTrValSpec(pseudodata.drop("type", axis=1), tr.index, val.index)
 
-
-def make_replica(groups_dataset_inputs_loaded_cd_with_cuts, replica_mcseed, genrep=True):
+def make_replica(groups_dataset_inputs_loaded_cd_with_cuts, replica_mcseed,  dataset_inputs_sampling_covmat, sep_mult, genrep=True, ):
```
scarlehoff, line 104:
```suggestion
def make_replica(
    groups_dataset_inputs_loaded_cd_with_cuts,
    replica_mcseed,
    dataset_inputs_sampling_covmat,
    sep_mult,
    genrep=True,
):
```

validphys2/src/validphys/covmats.py
```diff
@@ -226,7 +230,7 @@ def dataset_inputs_covmat_from_systematics(
         covmat = regularize_covmat(
             covmat,
             norm_threshold=norm_threshold
-        )
+        )
```
scarlehoff, line 233:
```suggestion
        )
```

validphys2/src/validphys/commondata.py
```diff
@@ -34,5 +34,4 @@ def loaded_commondata_with_cuts(commondata, cuts):
 
 groups_dataset_inputs_loaded_cd_with_cuts = collect(
     "loaded_commondata_with_cuts", ("group_dataset_inputs_by_metadata", "data_input")
-)
-
+)
```
scarlehoff, line 37: I'm guessing this things are due to Mac? Or to code?

I'm curious why some of your files just change the last line?

validphys2/src/validphys/config.py
```diff
@@ -6,6 +6,8 @@
 """
 import logging
 import pathlib
+import numpy as np
+import pandas as pd
```
scarlehoff, line 10: They should be below collections to follow the "standard-lib, third-party, first-pary" rule.

AleCandido: And add a blank line to separate the two groups.

validphys2/src/validphys/commondata.py
andreab1997: I don't really know

scarlehoff (approved):
Thanks for this.

@Zaharid do you want to also have a look before merging?

Zaharid:
@scarlehoff yes please.

andreab1997:
> @scarlehoff yes please.

Can this be merged? @Zaharid

andreab1997:
@Zaharid I have solved the conflict. Can this be merged now?

scarrazza:
@RoyStegeman could you please have a quick review of this PR so we can merge?

RoyStegeman:
Sure

n3fit/src/n3fit/scripts/n3fit_exec.py
```diff
@@ -145,7 +145,20 @@ def from_yaml(cls, o, *args, **kwargs):
...
+            N3FIT_FIXED_CONFIG['use_thcovmat_in_fitting'] = thconfig.get('use_thcovmat_in_fitting', True) 
+            N3FIT_FIXED_CONFIG['use_thcovmat_in_sampling'] = thconfig.get('use_thcovmat_in_sampling', True)  
+            N3FIT_FIXED_CONFIG['use_user_uncertainties'] = thconfig.get('use_user_uncertainties', False) 
+            N3FIT_FIXED_CONFIG['use_scalevar_uncertainties'] = thconfig.get('use_scalevar_uncertainties', True) 
+        #Sampling flags
+        if (sam_t0:=file_content.get('sampling')) is not None:
+            N3FIT_FIXED_CONFIG['use_t0_sampling'] = sam_t0.get('use_t0', False) 
+            N3FIT_FIXED_CONFIG['separate_multiplicative'] = sam_t0.get('separate_multiplicative', True) 
+        #Fitting flag
+        N3FIT_FIXED_CONFIG['use_t0_fitting'] = file_content.get('fitting').get('use_t0', True)
```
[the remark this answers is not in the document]
andreab1997: Yes but I believe this is necessary to avoid validphys complaining that, for example, theory_covmat_flag does not exist. The moment it exists we do have defaults inside the function but if it does not, then it is a problem.

validphys2/src/validphys/tests/test_pythonmakereplica.py
```diff
@@ -21,9 +21,13 @@
...
-    {"dataset": "ATLAS_SINGLETOP_TCH_R_13TEV", "cfac": ["QCD"]},
-    {"dataset": "CMS_SINGLETOP_TCH_R_8TEV", "cfac": ["QCD"]},
-    {"dataset": "CMS_SINGLETOP_TCH_R_13TEV", "cfac": ["QCD"]},
+    {"dataset": "NMC"},
+    {"dataset": "NMCPD"},
+    {"dataset": "ATLASZPT8TEVMDIST", "cfac": ["QCD"]},
+    {"dataset": "ATLASWZRAP36PB"},
+    {"dataset": "ATLASZHIGHMASS49FB"},
+    {"dataset": "CMSWEASY840PB"},
+    {"dataset": "CMSWMASY47FB"}
```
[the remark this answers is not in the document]
andreab1997: Oh, It is just to also test dataset by dataset. I add a comment

validphys2/src/validphys/config.py
```diff
@@ -693,6 +696,108 @@ def produce_experiment_from_input(
...
         }
 
+    def produce_sep_mult(self, separate_multiplicative=None):
+        """
+        Specifies whether to separate the multiplicative errors in the 
+        experimental covmat construction. The default is True. 
+        """
+        if separate_multiplicative is False:
+            return False
+        return True
```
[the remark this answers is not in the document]
andreab1997: I believe this was to make sure that if one does not care about `separate_multiplicative` and uses a runcard without it, then can get the "old" behaviour where the multiplicative part is separated.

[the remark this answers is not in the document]
andreab1997: I believe this was to make sure that if one does not care about `separate_multiplicative` and uses a runcard without it, then can get the "old" behaviour where the multiplicative part is separated.

validphys2/src/validphys/covmats.py
```diff
@@ -190,7 +191,6 @@ def dataset_inputs_covmat_from_systematics(
     special_corrs = []
     block_diags = []
     weights = []
-
```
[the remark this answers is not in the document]
andreab1997: I believe we need them because we want to separate the case in which only the additive part of the uncertainties are retained in the covmat and the case in which the covmat is complete. But maybe I misunderstood the question.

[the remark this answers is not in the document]
AleCandido: Being the question attached to a deleted blank line, my naive interpretation would be: "Why are you deleting the blank line?"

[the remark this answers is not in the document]
andreab1997: ok easy ty 😂

validphys2/src/validphys/tests/test_pythonmakereplica.py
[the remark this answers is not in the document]
RoyStegeman: Right, I understand they are "datasets to be tested". What I meant was, why specifically these datasets? The same tests are run using DATA and CORR_DATA, why is that not enough? This is what I would like to understand from the comment for if at some point someone else touches these tests, so the test is not unintentionally broken when changes are made.

validphys2/src/validphys/config.py
```diff
@@ -693,6 +696,108 @@ def produce_experiment_from_input(
...
+        """
+        if theory_covmat_flag is False:
+            return np.array([])
+        #Load correct file according to how the thcovmat was generated by vp-setupfit
+        generic_path = "datacuts_theory_theorycovmatconfig_theory_covmat_custom.csv"
+        if use_user_uncertainties is True:
+            if use_scalevar_uncertainties is True:
+                generic_path = "datacuts_theory_theorycovmatconfig_total_theory_covmat.csv"
+            else:
+                generic_path = "datacuts_theory_theorycovmatconfig_user_covmat.csv"
```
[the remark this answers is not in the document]
RoyStegeman: Yes exactly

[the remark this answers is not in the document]
andreab1997: Ok fine :)

validphys2/src/validphys/config.py
```diff
@@ -781,6 +782,8 @@ def produce_loaded_theory_covmat(
         """
         if theory_covmat_flag is False:
             return 0.
+        if len(glob.glob(output_path/"*theory_covmat*")) > 1:
+            raise ValueError("Too many theory_covmat files in folder tables")
```
scarlehoff, line 786: Sorry to be a pain in the neck but since you have a flag that allow you to use two different theorycovmat files I think it is perfectly fine to have more than one in the folder.

Not only that, I don't see why having a file called `I_really_hate_theory_covmat_oh_my_god` should be an impediment for me to run fits with a theory covmat.

All that to say that I would remove these two lines.

scarrazza:
@andreab1997 could we merge this?

andreab1997:
> @andreab1997 could we merge this?

For me yes. I don't know if @RoyStegeman agrees.

andreab1997:
> @andreab1997 could we merge this?

Actually give me till this evening, I have to fix a minor thing and then I will merge.

validphys2/src/validphys/config.py
```diff
@@ -774,35 +773,59 @@ def produce_loaded_theory_covmat(
...
+                / "tables/datacuts_theory_theorycovmatconfig_user_covmat.csv"
+            ),
+        ]
+        paths.remove(str(output_path / "tables" / generic_path))
+        for f in files:
+            for path in paths:
+                if f == path:
+                    raise ValueError(
+                        "More than one theory_covmat file in folder tables"
+                    )
```
RoyStegeman, line 815: @andreab1997 this doesn't seem right?

andreab1997: Why not?

RoyStegeman: The error is not as general as I originally thought. It probably does do what you intended, but I tried it with a custom covmat and got this error because I did not set `use_scalevar_uncertainties` to false, in which case this message does not make a lot of sense since there was only one covmat. 

So I guess we just need to update the example here: https://github.com/NNPDF/nnpdf/blob/278310410398b36b95e45246992c8ed07db6d6a6/doc/sphinx/source/tutorials/general_th_covmat.rst? If that's correct, I can update the docs.

P.S. the check can be written in one line for example as
`
if set(paths)&set(files):
    raise ValueError
`

andreab1997: Ok yes I agree. I would say the check is checking what I wanted (even if it can be written better) but the error message is misleading.

RoyStegeman: I guess there is no situation in which both `use_scalevar_uncertainties` and `use_user_uncertainties` are true? If so, that should probably be checked by vp.

andreab1997: For the moment I would say no but in principle yes (for example one can have a different source of theory uncertainties which are not `user_uncertainties`)

RoyStegeman: Sure, but here `use_scalevar_uncertainties` is just synonym for using a covmat that has been constructed from different theoryID's (regardless of the source of the theory uncertainty) with the instructions in the runcard, while `user_uncertainties` is an external covmat (not necessarily non-scalevar).

So for now I think `use_scalevar_uncertainties` can probably be removed/deprecated as it is just the inverse of `user_uncertainties`?

RoyStegeman: If you agree, I can take care of the changes. Just you know this better than me, so it's good if you can confirm :)

andreab1997: Yes I agree. In my mind we could have a third option, like `use_whatever_uncertainties`, that uses a different prescription from the `scalevar`. However it is true that we do not have now so for the time being `use_scalevar_uncertainties` is just not `user_uncertainties`. So yes, I agree :)

RoyStegeman: Let's leave hypothetical future features  for hypothetical future people to work on ;)

Ok, I'll open a PR some point this week and ask for your review.

PR #1529, opened 2022-02-22
Pineappl integration with validphys

### LINKED ISSUE #1541: n3fit memory usage
The training of the network seems to consume an awful lot of memory (given the complexity of the task), or at least cluster schedulers seem to perceive the memory use as very high. A NNPDF4.0-NNLO run on the Dutch supercomputer Snellius reports over 200GB of peak memory usage. Detailed measurements will follow.

scarlehoff:
EDIT:
This branch is outdated with respect to master as it points to a master pre-thcovmat in n3fit, #1578 is a rebase of this branch to 08/07/2022 master: https://github.com/NNPDF/nnpdf/commit/3098d0aa36c8d90878c35047b77a9c7fa802da84

As promised, the pineappl integration with vp is in preparation.

In this first commit the pineappl tables are already compatible with vp and can be used for instance in `predictions` (like with the commondata reader, [I put a jupyter notebook at the root of the repository](https://github.com/NNPDF/nnpdf/blob/validphys_with_pineappl/FKTable%20Playground.ipynb)).

The fit part I will do tomorrow, which is a bit more complicated (if I want to keep both the old and new theories) since I will have to reconstruct the fktable from the pandas dataframe.

I will also need to modify `FKTableData` to hold more interesting information, but nothing to worrysome.

This will also close #1541 

-----------

This PR has ended up being a tad bigger than one would like (and broader in scope). Introducing the `pineappl` tables meant it was easier to simplify some of the other stuff (such as the reader for the old fktables) rather than have both of them live together. As such, the review will be more painful than I hoped, so in order to help the reviewer (and myself in 5 years time when I have to remember what I did) I've written some notes on the changes.

## TODO before merging
- [x] `pineappl`, `eko` in `conda-forge`
- [ ] approved
- [x] Theory [whatever](https://github.com/NNPDF/fktables/issues/29) in the server so it can be used by everyone
- [x] Remove jupyter notebook
- [x] Remove the `oldmode` flag from the loader

## How to install the necessary packages

Since they are still not in conda you will need to install `pineappl` in order to use this branch, luckily it is available in the python repos:

```bash
pip install pineappl eko
```

or

```bash
conda install pineappl eko
```

(just in case, make sure that the pineappl version that get installed is `0.5.2`)

## Where to get pineappl grids

I've put in the root of the nnpdf server a folder called `pineappl_ingredients.tar` which contains a `yamldb` (which goes into `NNPDF/data` and a `pineappls` which goes into `NNPDF/data/theory_200`

```bash
cd ${CONDA_PREFIX}/share/NNPDF/data
scp vp.nnpdf.science:pineappl_ingredients.tar .
tar -xf pineappl_ingredients.tar yamldb/ 
tar -C theory_200/ -xf pineappl_ingredients.tar pineappls/
rm pineappl_ingredients.tar
cd -
```

## Notes on the changes

- Jupyter notebook: it's there just so we can play with it a bit, but I'll remove it once this is approved for merging.

- all `n3fit` files: since now we have in vp a few objects that contain exactly the same information that was passed to some of the dictionaries in `n3fit`, I've removed said dictionaries. This is the only reason these files have changed so you might as well ignore them.

- `config.py`: separated posdataset and integdataset

- `convolution.py`: in order to keep using `truediv` for `RATIO` (which is nice because that means that tensorflow, numpy, pandas or any other library know what to do with the operation) it is necessary to make the pandas denominator into a numpy array. I think this is reasonable since in this case the denominator is usually a total cross section so the indexes are not supposed to match the numerator.

- `core.py`: 
Added `load_commondata` to `DataSetInput` and `DataGroupSpec` so that one can get just the `CommonData` from `libNNPDF` instead of the whole `DataSet`.
Modified `FKTableSpec` so that it can load both the new and old fktables.
Added an `IntegrabilitySetSpec` which is equal to `PositivitySetSpec`.

- `coredata.py`
I've added to `FKTableData` methods that generate the information needed to fit an fktable: `luminosity_mapping` and `get_np_fktable`.
I've moved the application of the `cfactors` to `FKTableData` to be equivalent to `with_cuts` (so that no funny games need to be played with the dataclass outside).
Added also a `_protected` flag since `cfactors` and `cuts` were done with the old fktables in mind and as explained in `convolution.py` they wil find a different number of points. When the repetition flag was found in apfelcomb, it gets saved into the yaml database and cuts or cfactors are applied accordingly.
Note that `FKTableData` works the same no matter where the fktables came from (pineappl or old)

- `fkparser.py`
Moved the application of `cfactors` away (see above).

- `loader.py`
I've added `check_fkyaml` to load the new `yaml` files. Eventually the yaml information will come from the new commondata format so at the moment I've just hardcoded the path of the yamldb inside the data folder.
I've separated the positivity and integrability loading. This was not strictly necessary but it facilitated the `n3fit_data.py` below and it was something that annoyed me since a long time.
For testing I've added a flag to `check_dataset` such that if you use `oldmode` as a cfactor, the old fktables are used regardless. This is useful for debugging and will be removed alongside the jupyter notebook for before merging (or as soon as the new theory is prepared).
At the moment whether to use pineappl or not does not depend on the theory since the pineappl tables are not a part of any theory at the moment. 

- `n3fit_data.py`
This has been greatly simplified. Most notably, since like `mask_fk_tables` have been removed. `fitting_data_dict` is no longer a dictionary with a list of dictionaries inside but contains a list of `FittableDatasets` which are coming from the outside.
The most interesting part here is that this means that issue #1541 is also solved. I had to do something else which was creating a `TupleComp` class for the masks that depend on the name and the seed, but it is a small price to pay.
For the rest most of the changes in this module are just removing stuff.

- `n3fit_data_utils.py`
Here don't even look at the diff, this module has been basically rewritten from scratch. Some lines just happen to be similar.
I've created the `FittableDataSet` which contains all the information necessary to fit a given dataset other than the central value. 
imho, once we have the new common data format and pure python `datasets` this should be just an extension of those but it's a bit tricky because in its current form this can be shared between replicas and if we add the cv that won't be possible. But anyway, that's a problem for the future weeks.

- `pineparser.py`
Contains all the logic for loading the new fktables into `coredata.FKTableData` objects completely equivalent to what vp creates reading the old fktables with pandas. This is going to probably be moved to `pineko` since we believe building and reading the fktables should be one single thing. I'll do it when this https://github.com/N3PDF/pineko/pull/12 PR is merged. 

- `results.py`
This is the reason for the `load_commondata` needed in `core.py`.
Currently getting results needed not only the central value of the data but also to load the `libNNPDF` fktable even if they were not used anywhere. I could live with that (just a waster of memory) when they exist, but this is no longer the case if you only have the pineappl version.


scarlehoff:
This is now "done" (let's see whether the fit bot is working) save a few details that I will do directly here since I already changed a lot `n3fit_data_utils`.

It is not ready for review until I get to do a full fit with all datasets we have instead just a few selected ones in case there are unforeseen problems (I'm sure there will be), @felixhekhorn is running now the ekos to prepare the full batch of dataset for the "baseline".

Then there is the (important) detail of the fact that I'm probably wasting a lot of memory and time with the spaguetti code that I've put in `get_np_fktable`. If someone who knows pandas better than me could have a look and fix it, please do.

Other than that, I'm quite happy with the result. Some of the layers of the rightfully-hated dictionary-complexity of `n3fit` became useless since there is no longer the mix of C++ and python (for starters no need to parse the `FKTableData` object, it contains everything needed now and can be pased down verbatim).

scarlehoff:
I've updated the secrets, chances are that I've broken things more than they were but we'll discover it together.

scarlehoff:
Got hit by the indexing...

scarlehoff:
**If** the fit bot works I'm happy with the current state. Still not ready for review because I need to review the changes myself with a fresh pair of eyes.

n3fit/src/n3fit/tests/test_layers.py
```diff
@@ -197,12 +203,12 @@ def test_rotation_flavour():
     # Apply the rotation using numpy tensordot
     x = np.ones(8)  # Vector in the flavour basis v_i
     x = np.expand_dims(x, axis=[0, 1])  # Give to the input the shape (1,1,8)
-    mat = fitbasis_to_NN31IC(flav_info, 'FLAVOUR')  # Rotation matrix R_ij, i=flavour, j=evolution
+    mat = fitbasis_to_NN31IC(flav_info, "FLAVOUR")  # Rotation matrix R_ij, i=flavour, j=evolution
```
Zaharid, line 206: FWIW I'd be a lot happier if the `-S` option was default in black. Not only do I find these diffs a bit distracting in the context of a complicated PR but also don't think it is an improvement in any way.

validphys2/src/validphys/coredata.py
```diff
@@ -87,14 +109,71 @@ def with_cuts(self, cuts):
...
         return dataclasses.replace(self, ndata=newndata, sigma=newsigma)
 
+    @property
+    def luminosity_mapping(self):
+        """Return the flavour combinations that contribute to the fktable
+        in the form of a single array
+
+        The return shape is:
+            (nbasis,) for DIS
+            (nbasis*2,) for hadronic
```
Zaharid, line 129:
```suggestion
            (nbasis**2,) for hadronic
```

validphys2/src/validphys/convolution.py
```diff
@@ -120,9 +120,16 @@ def _predictions(dataset, pdf, fkfunc):
...
         )
     cuts = dataset.cuts.load()
-    all_predictions = [
-        fkfunc(load_fktable(fk).with_cuts(cuts), pdf) for fk in dataset.fkspecs
-    ]
+    all_predictions = []
+    for fk in dataset.fkspecs:
+        fk_w_cuts = load_fktable(fk).with_cuts(cuts)
+        ret = fkfunc(fk_w_cuts, pdf)
+        # TODO: old fktables just repeated values to keep the size the same
```
Zaharid, line 127: Which values were repeated?

validphys2/src/validphys/coredata.py
```diff
@@ -51,6 +51,28 @@ class FKTableData:
     sigma: pd.DataFrame
     metadata: dict = dataclasses.field(default_factory=dict, repr=False)
 
+    def __post_init__(self):
+        self._fktable = None
+
+    @property
+    def protected(self):
```
Zaharid, line 58: I'm having trouble understanding this and why we need it. What is a repetition flag?

validphys2/src/validphys/fkparser.py
```diff
@@ -172,56 +192,52 @@ def _parse_hadronic_fast_kernel(f):
...
     return gi, line_number, next_line
 
 
-
 def _parse_header(lineno, header):
     if not _is_header_line(header):
-        raise BadFKTableError(f"Bad header at line {lineno}: First "
-                "character should be either '_' or '{'")
+        raise BadFKTableError(
+            f"Bad header at line {lineno}: First " "character should be either '_' or '{'"
```
Zaharid, line 221: Uh, sorry but this is a bit too much reformatting. I find many of the changes (e.g. in the spaces between multiplications) for the worse. And what it is doing here is senseless. So I am going to quote

> This tool should not be used to aggressively reformat existing files.

https://docs.nnpdf.science/contributing/python-tools.html?highlight=black#python-static-checks-and-code-style

scarlehoff: I rather fix manually the problems that black might have created, `fkparser` is a file where almost all that I haven't touched explicitly I either want to remove or will remove eventually so it doesn't make sense not to reformat for the sake of some decisions taken 5 years ago.

validphys2/src/validphys/coredata.py
Zaharid: Do I understand correctly that the only place where this ever comes up is ratios and everything else could operate transparently?

Zaharid: > In addition in the original grids some of these are stored as grids of n points where everything is 0 but the first.

Which cases? Wouldn't that break the existing convolution code which does nothing to handle that?

felixhekhorn: as @scarlehoff said this is done in apfelcomb, so before convolutions appear

n3fit/src/n3fit/tests/test_layers.py
scarlehoff: i thinl going for "whatever is default in black" is a good idea because it can be done mindlessly I can just tell people to install it and use it.

it's is true that for files like this the reviewer pays the prize of my past sloppyness. I'll undo some of them before marking this for review (I'm adamant on the `fkparser.py` though because it's other people's fault :P)

validphys2/src/validphys/fkparser.py
Zaharid: Seeing as I am the one who wrote the code, seeing the formatting bulldozed makes me somewhat irrationally unhappy. But on top of that the facts that it does obviously silly things as in here and that makes the diff much harder to read are the reasons for the "don't aggressively reformat" bit in the docs, and which I'd argue are valid.

Also I am not sure I'd be comfortable removing this code soon. We would need that all the infrastructure and all the theories are working with the new grids and iron out all the problems that might appear with more widespread and prolonged usage .

n3fit/src/n3fit/tests/test_layers.py
Zaharid: I was not going to complain until some threshold was surpassed...

validphys2/src/validphys/coredata.py
Zaharid: I wonder if some explicit concept of normalization grid that doesn't come with a datapoint index could be made to work. With the rough semantics that it is only one cross section and operations broadcast like numpy would. Implementation wise it would be stored like a normal fktable (and could act as one if we wanted to e.g. fit the total cross setion) but would drop the data index when loaded as a normalization.

Zaharid: ...so e.g. `with_cuts` would be a no op.

validphys2/src/validphys/fkparser.py
scarlehoff: > Also I am not sure I'd be comfortable removing this code soon. We would need that all the infrastructure and all the theories are working with the new grids and iron out all the problems that might appear with more widespread and prolonged usage .

Yes, but it should become "legacy" after this PR. Honestly I would like to regenerate all theories as soon as all issues are ironed out but that might just be a wish.

validphys2/src/validphys/coredata.py
scarlehoff: Yes, I guess this is what I'm achieving with the `protected` property. The table itself can be loaded if there is no repetition flag in the yaml file.

I guess this can be made a fktable-only property by moving the cfactor application to the FKTableData object?

validphys2/src/validphys/core.py
```diff
@@ -534,28 +537,49 @@ def __str__(self):
...
-        self.cfactors = cfactors
+        self.metadata = metadata
+        self.norm = norm
         super().__init__(fkpath, cfactors)
 
-    #NOTE: We cannot do this because Fkset owns the fktable, and trying
-    #to reuse the loaded one fails after it gets deleted.
-    #@functools.lru_cache()
-    def load(self):
+    def _load_legacy(self):
```
Zaharid, line 559: Can we not just delete this?

validphys2/src/validphys/fkparser.py
felixhekhorn: I guess that depends on what you mean with "all theories": all that are in the db? that might be too much - all that are actually used (which I estimated at ~25 - with the biggest chunk being SV, which we have to regenerate in any case)? maybe yes

validphys2/src/validphys/core.py
scarlehoff: No, I tried but it is needed in several places https://github.com/NNPDF/nnpdf/blob/2ae49281c5d471be490ed5c7bbd28e9211091c19/validphys2/src/validphys/dataplots.py#L792

I was "hunting them down" but at a certain point there was closure stuff involved so undid that. I might try again next week, it might have been something easy but I got a bit overwhelmed.

validphys2/src/validphys/fkparser.py
scarlehoff: Why is it too much? If it can be done automatically we can just leave it running for two months if needed and at the end we have our new shiny fktables. Or are some for which this will not be doable?

validphys2/src/validphys/coredata.py
Zaharid: I suppose moving the cfactors application might make it neater. One reason why it wasn't done was that I didn't want to have FKTableData objects in intermediate states lying around and instead there is one way to load once and for all.

That said I was rather thinking in having a slightly different object with the requisite behaviour rather than patching up FKTableData to support both.

validphys2/src/validphys/core.py
Zaharid: Right. Better to not add more stuff here indeed.

validphys2/src/validphys/fkparser.py
Zaharid: I think there are things that just require some time to be ascertained. Even if we compare point by point all fktables in all theories (which btw we have no ability to produce with the new toolchain at the moment) and they are the same today I would still want to see that things remain stable for some iterations of the new toolchain. 

For one the existing fktable format is easy enough to parse by hand and has been stable for some 9 years now  while the new one cannot be parsed without the implementation and might be non trivial to get to work between updates of pinnaple.  While there are ways to deal with format upgrades, having some alternative source of truth to compare with might be helpful for a while.

Lastly I don't think maintaining the two formats should require too much work (as for the time being it is just a matter of not deleting the existing fkparser) seeing as the two things end up in the same memory object at the time. This is different from say the cost of dealing with c++.

validphys2/src/validphys/core.py
scarlehoff: By here you mean this PR? Ok, then I'll add this to my todo list for the (near) future. As said in the code meeting, during the fit it doesn't seem to be used at all so it should be possible to remove it completely.

validphys2/src/validphys/fkparser.py
felixhekhorn: > Why is it too much? If it can be done automatically we can just leave it running for two months if needed and at the end we have our new shiny fktables. Or are there some for which this will not be doable?

no no, technically there is no problem - it's just a matter of resources ...

scarlehoff:
test

scarlehoff:
This is now ready for review. 
In the first comment I've left an explanation of everything that has changed. The PR is not that big (of the ~1700 lines, about ~1000 are "fake" because they are either the jupyter notebook or the `pineparser` file which will go into `pineko`).

I guess @Zaharid and @RoyStegeman are the right people to review this. @Radonirinaunimi and @tgiani might be interested as well?

Radonirinaunimi:
> This is now ready for review (not ready for merging because the need for conda/theories) In the first comment I've left an explanation of everything that has changed. The PR is not that big (of the ~1700 lines, about ~1000 are "fake" because they are either the jupyter notebook or the `pineparser` file which will go into `pineko`).
> 
> I guess @Zaharid and @RoyStegeman are the right people to review this. @Radonirinaunimi and @tgiani might be interested as well?

Yes, I am interested in helping reviewing this PR. This is something I will have to touch/modify later anyway.

conda-recipe/meta.yaml
```diff
@@ -52,6 +52,7 @@ requirements:
         - sphinxcontrib-bibtex
         - docutils =0.16 # This dependency is not explicity needed but https://github.com/NNPDF/nnpdf/issues/1220
         - curio >=1.0
+        - pineappl >= 0.5
```
scarlehoff, line 55:
```suggestion
        - pineappl >=0.5
```

Zaharid:
Testing the notebook I get

````
FileNotFoundError: [Errno 2] No such file or directory: '/mount/storage/Academic_Workspace/NNPDF/source/nnpdf/n3fit/NNPDF40_with_pineappl.yml'
```

scarlehoff:
That's your fault for not having the exact same directory structure that I have!

For now you have to do this:

> I've put in the root of the nnpdf server a folder called pineappl_ingredients.tar.gz which contains a yamldb (which goes into NNPDF/data and a pineappls which goes into NNPDF/data/theory_200

And then for that line point to some runcard you want to test.

n3fit/src/n3fit/layers/observable.py
```diff
@@ -28,8 +28,8 @@ class Observable(MetaLayer, ABC):
 
         Parameters
         ----------
-            fktable_dicts: list
-                list of fktable_dicts which define basis and xgrid for the fktables in the list
+            fktable_data: list(validphys.coredata.FKTableData)
```
Zaharid, line 31:
```suggestion
            fktable_data: list[validphys.coredata.FKTableData]
```

scarlehoff:
I've updated the `pineappl_ingredients.tar` in the server. I've updated the first comment to show exactly what I need to do to try this in my computer.

validphys2/src/validphys/n3fit_data_utils.py
```diff
@@ -1,168 +1,107 @@
...
-    """
-    parsed_datasets = []
-    for dataset_c, dataset_spec in zip(experiment_c.DataSets(), experiment_spec.datasets):
-        parsed_datasets += common_data_reader_dataset(dataset_c, dataset_spec)
-    return parsed_datasets
-
 
-def positivity_reader(pos_spec):
+    # NOTE:
+    # This class tries to be compatible the libNNPDF dataset class
```
RoyStegeman, line 37:
```suggestion
    # This class tries to be compatible with the libNNPDF dataset class
```

validphys2/src/validphys/n3fit_data_utils.py
```diff
@@ -1,168 +1,107 @@
...
+    # after the class is moved to python, this can inherit from dataset
+    # and the dataset should know how to generate its "fittable" version
```
RoyStegeman, line 39:
```suggestion
    # after the class is moved to python, this can inherit from nnpdf.DataSet
    # and the dataset should know how to generate its "fittable" version
```

Not really sure what you mean here

validphys2/src/validphys/pineparser.py
```diff
@@ -0,0 +1,316 @@
...
+    # Check whether the dataset has shifts
+    # Note: this only happens for ATLASZPT8TEVMDIST, if that gets fixed we might as well remove it
+    if metadata["apfelcomb"].get("shifts") is not None:
+        shift_info = metadata["apfelcomb"]["shifts"]
+        ret["shifts"] = [shift_info.get(op, 0) for op in operands]
+
+    return ret
+
+
+###########
```
RoyStegeman, line 154:
```suggestion
```

validphys2/src/validphys/pineparser.py
```diff
@@ -0,0 +1,316 @@
...
+        if apfelcomb is not None:
+            if apfelcomb.get("normalization") is not None:
+                raw_fktable = raw_fktable * apfelcomb["normalization"][i]
+            if apfelcomb.get("repetition_flag", False):
+                raw_fktable = raw_fktable[0:1]
+                n = 1
+                protected = True
+            if apfelcomb.get("shifts") is not None:
+                ndata += apfelcomb["shifts"][i]
+        ###
```
RoyStegeman, line 283:
```suggestion
```

validphys2/src/validphys/pineparser.py
```diff
@@ -0,0 +1,316 @@
...
+            Whether the fktable is a single point which gets repeated up to a certain size
+            (for instance to normalize a distribution)
+        shift: list(int)
+            Shift in the data index for each grid that forms the fktable
+    """
+    if metadata.get("apfelcomb") is None:
+        return None
+
+    # Can't pathlib understand double suffixes?
+    operands = [i.name.replace(f".{EXT}", "") for i in gridpaths]
```
RoyStegeman, line 127:
```suggestion
    operands = [i.with_suffix("").stem for i in gridpaths]
```

Alternatively you can do this. `with_suffix` removes one suffix, `stem` the other (if it exists)

validphys2/src/validphys/pineparser.py
```diff
@@ -0,0 +1,316 @@
...
+    if metadata["apfelcomb"].get("repetition_flag") is not None:
+        if len(operands) == 1:
+            ret["repetition_flag"] = operands[0] in metadata["apfelcomb"]["repetition_flag"]
+        else:
+            # Just for the sake of it, let's check whether we did something stupid
+            if any(op in metadata["apfelcomb"]["repetition_flag"] for op in operands):
+                raise ValueError(f"The yaml info for {metadata['target_dataset']} is broken")
+
+    # Check whether the dataset has shifts
+    # Note: this only happens for ATLASZPT8TEVMDIST, if that gets fixed we might as well remove it
```
RoyStegeman, line 146:
```suggestion
    # NOTE: this only happens for ATLASZPT8TEVMDIST, if that gets fixed we might as well remove it
```

validphys2/src/validphys/pineparser.py
```diff
@@ -0,0 +1,316 @@
...
+
+    return ret
+
+
+###########
+
+
+def _pinelumi_to_columns(pine_luminosity, hadronic):
+    """Makes the pineappl luminosity into the column indices of a dataframe
+    These corresponds to the indices of a flat (14x14) matrix for hadronic observables
```
RoyStegeman, line 159:
```suggestion
    These corresponds to the indices of a flattened (14x14) matrix for hadronic observables
```

validphys2/src/validphys/pineparser.py
```diff
@@ -0,0 +1,316 @@
...
+    return columns
+
+
+def get_yaml_information(yaml_file, theorypath):
+    """Reads the yaml information from a yaml compound file
+
+    Transitional function: the call to "pineko" might be to some other commondata reader
+    that will know how to extract the information from the commondata
+    """
+    # TODO Tell the reader where to look for the pineappl grids
```
RoyStegeman, line 196:
```suggestion
    # TODO: Tell the reader where to look for the pineappl grids
```

RoyStegeman: Where do you plan to move them after debugging?

RoyStegeman (commented):
Still todo:
- Delete the playground ipynb
- Remove debug code "oldmode"
- Move `pineparser.py` to `eko`(?)

Please make this into multiple PRs next time. For some modules the diffs were effectively useless making it pretty hard to trace what is going on (for me anyway).

scarlehoff:
> Move pineparser.py to eko(?)

I guess it will stay here after all? Not sure because it depends a bit on what happens with commondata...

> Still todo:
>    Delete the playground ipynb
>   Remove debug code "oldmode"

I'll do this before merging.

validphys2/src/validphys/core.py
```diff
@@ -545,10 +585,14 @@ def load(self):
         return PositivitySet(cd, fk, self.maxlambda)
 
     def to_unweighted(self):
-        log.warning("Trying to unweight %s, PositivitySetSpec are always unweighted", self.name)
+        log.warning("Trying to unweight %s, %s are always unweighted", self.__class__.__name__, self.name)
         return self
 
 
+class IntegrabilitySetSpec(PositivitySetSpec):
```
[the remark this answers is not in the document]
scarlehoff: > Why did you add this? Just to be explicit?

This should've been done long ago.

But we need two of them because they are treated differently at different stages by checking whether the dataset is call "POS" or not...

validphys2/src/validphys/core.py
```diff
@@ -515,28 +528,55 @@ def __str__(self):
...
     """Extends DataSetSpec to work around the particularities of the positivity datasets"""
 
     def __init__(self, name, commondataspec, fkspec, maxlambda, thspec):
         cuts = Cuts(commondataspec, None)
         self.maxlambda = maxlambda
         super().__init__(name=name, commondata=commondataspec, fkspecs=fkspec, thspec=thspec, cuts=cuts)
-        if len(self.fkspecs) > 1:
-            # The signature of the function does not accept operations either
-            # so more than one fktable cannot be utilized
-            raise ValueError("Positivity datasets can only contain one fktable")
```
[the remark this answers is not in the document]
scarlehoff: Because the positivity fktables in `libNNPDF` were read _almost_ in the same way as the other fktables but not quite. The change to vp meant that one could read wrong positivity tables (and it would've work silently) so this is just so vp was crashing as well.

After this PR the positivity fktables are read in the same way that other fktables so this "legacy crash" is not needed anymore.

validphys2/src/validphys/loader.py
```diff
@@ -358,6 +362,32 @@ def check_fktable(self, theoryID, setname, cfac):
...
+        _, theopath = self.check_theoryID(theoryID)
+        metadata, fklist = pineparser.get_yaml_information(fkpath, theopath)
+        op = metadata["operation"]
+
+        # TODO:
+        #      at the moment there are no pineappl specific c-factors
+        #      so they need to be loaded from the NNPDF names / compounds files
+        cfac_name = metadata["target_dataset"]
+        # check whether there is a compound file
+        cpath = theopath / "compound" / f"FK_{cfac_name}-COMPOUND.dat"
```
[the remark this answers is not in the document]
scarlehoff: I don't think I understand the case that worries you, a dataset is compound or it isn't. At the moment the cfactors are from libNNPDF so they have NNPDF names.
The problem is that the names of the datasets have changed so until we don't create new c-factors for a new pineappl theory there's not much more one can do.

(and we should do this together with the theory 200X or whatever that we said that we were creating)

validphys2/src/validphys/loader.py
```diff
@@ -482,7 +520,21 @@ def check_dataset(self,
...
+        """
+        # TODO: this is just so I can load both types at once during development
+        readyaml = True
+        force_pineappl = False
+        if "oldmode" in cfac:
+            cfac = [i for i in cfac if i != "oldmode"]
+            readyaml = False
+        elif "forcepineappl" in cfac:
+            cfac = [i for i in cfac if i != "forcepineappl"]
+            force_pineappl = True
```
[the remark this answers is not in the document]
scarlehoff: Only for debugging.

validphys2/src/validphys/n3fit_data_utils.py
```diff
@@ -1,168 +1,107 @@
...
+    Parameters
+    ----------
+        datasets: list(:py:class:`validphys.core.DataSetSpec`)
+            List of dataset spec in this group
+        tr_masks: list(np.array)
+            List of training mask to be set for each dataset
+
+    Returns
+    -------
+        loaded_obs: list (:py:class:`validphys.n3fit_data_utils.FittableDataSet`)
```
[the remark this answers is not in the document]
scarlehoff: if I didn't screw up one can just click and they will be directed to ``FittableDataSet``

validphys2/src/validphys/n3fit_data_utils.py
scarlehoff: The class `FittableDataset` tries to be equal to `libNNPDF` since we still have `libNNPDF` stuff (due to commondata). Once `libNNPDF` is no longer used, this can just inherit from the vp dataset.

validphys2/src/validphys/pineparser.py
scarlehoff: Honestly I find the string replacement clearer to say "I'm removing the extension `.pineappl.lz4` completely".

validphys2/src/validphys/pineparser.py
scarlehoff: Hopefully we will have a Theory 200X soonish...

RoyStegeman: In that case. Why keep the fktables around?

scarlehoff: which ones?

merijn:
I've been playing with/profiling this branch in the context of #1541 and its current form it's showing a pretty big improvement over master, both in reduced memory usage and significantly less time wasted loading data, to the extent that most of the profiling now only shows actual tensorflow running.

I had a look over the changes and all the code looks reasonable to me, but I am also not super familiar with it, so take that assessment with a grain of salt.

validphys2/src/validphys/core.py
```diff
@@ -515,38 +528,78 @@ def __str__(self):
...
+    In Legacy Mode each fktable has one single path.
+    For pineappl tables instead a FKTable is formed by any number of grids
+    therefore in order to check whether we have a new-format or old-format table
+    we will just check whether fkpath is a list
+    For now holds the metadata as an attribute to this function.
+    This is useless/transitional since this metadata is already in the new CommonData format
+    """
+    def __init__(self, fkpath, cfactors, metadata=None):
+        self.cfactors = cfactors if cfactors is not None else []
+        # Note: the legacy interface is expected to be removed by NNPDF5.0
```
RoyStegeman, line 541:
```suggestion
        # NOTE: the legacy interface is expected to be removed by NNPDF5.0
```

validphys2/src/validphys/loader.py
```diff
@@ -358,6 +362,32 @@ def check_fktable(self, theoryID, setname, cfac):
...
+        cfac_name = metadata["target_dataset"]
+        # check whether there is a compound file
+        cpath = theopath / "compound" / f"FK_{cfac_name}-COMPOUND.dat"
+        if cpath.exists():
+            # Get the target filenames
+            tnames = [i[3:-4] for i in cpath.read_text().split() if i.endswith(".dat")]
+            cfactors = [self.check_cfactor(theoryID, i, cfac) for i in tnames]
+        else:
+            cfactors = [self.check_cfactor(theoryID, cfac_name, cfac)]
+        ###
```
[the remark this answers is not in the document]
RoyStegeman: Why is this here?

RoyStegeman (commented):
Maybe you can run `black` on all the files you touched before merging? In principle that should have to be done only once per file so it would prevent the many cosmetic changes in the diffs  in the future. Perhaps you already did that for all files though.

scarlehoff:
> Maybe you can run black on all the files you touched before merging?

Before merging I'll do that trying to do it only for the parts of the files I touched.

> Why is this here?

What do you mean?

RoyStegeman:
> What do you mean?

###

scarlehoff:
Ah, to mark the end of the hack for the lack of pineappl specific cfactors.

scarlehoff:
Now it should be possible to simply run the code in this branch with theory id 400. No more need for tricks. In the next few days I'll remove the extra files and pass black through the code I modified.

scarlehoff:
I don't know why we have the segmentation fault in the CI, but I've noticed that the nnpdfcpp tests do not compile in my computer with something similar to this: https://github.com/catchorg/Catch2/issues/2178 
I've commented it for now.

scarlehoff:
@felixhekhorn @AleCandido is there a known problem with eko and python 3.9? The tests only fail when eko is included in the recipe (despite eko never being imported for any of the tests).

Asking before comparing the two logs side by side to see what has been installed differently...

felixhekhorn:
> @felixhekhorn @AleCandido is there a known problem with eko and python 3.9? The tests only fail when eko is included in the recipe (despite eko never being imported for any of the tests).
> 
> Asking before comparing the two logs side by side to see what has been installed differently...

Mmm ... we're not aware of problems and indeed the [unit tests](https://github.com/N3PDF/eko/runs/5478782057?check_suite_focus=true) and the [isolated benchmarks](https://github.com/N3PDF/eko/runs/5478781291?check_suite_focus=true) work in 3.9

I'd say the most likely candidate would be `numba` and friends (e.g. `llvm`) ... they might cause problems outside python (meaning without import)

scarlehoff:
The only big change that I can see is the libgfortran from 7 to 11... and when I do everything locally  (with python 3.9) installing eko makes only the following change to the standard installation:

```
  numpy                               1.22.3-py39he7a7128_0 --> 1.21.5-py39he7a7128_2                         
  numpy-base                          1.22.3-py39hf524024_0 --> 1.21.5-py39hf524024_2
```

Which (locally) is not enough to break anything. Sigh. I really don't understand what's happening (although because if I run the test that seems to fail by itself in the github CI it works)

AleCandido:
> The numpy downgrade is expected since the current numba (0.55.1) only works with `numpy<1.22` - that constrain should be lifted in [0.56](https://github.com/numba/numba/pull/8027)

Actually, a backport for v0.55.2 is even scheduled: https://github.com/numba/numba/pull/8067

scarlehoff:
I'm sorry for the people subscribed to this PR...

For future reference, the action to open a terminal inside the CI does work but as soon as I try to do something slightly complicated (and that includes getting the conda package with scp) it crashes :/ I thought it was my connection but changing it didn't fix it.

AleCandido:
> @felixhekhorn @AleCandido is there a known problem with eko and python 3.9? The tests only fail when eko is included in the recipe (despite eko never being imported for any of the tests).
> 
> Asking before comparing the two logs side by side to see what has been installed differently...

Actually, looking back at the logs:
- it fails on 3.9 only because arrived first, and stopped the others before they could. I'm pretty confident it would also crash on those before or after (or both)
- I really believe it is a problem of environment and dependencies, so you actually need to reproduce the exact environment of the CI...

However, I hope that  N3PDF/eko#122 will solve it before.

AleCandido:
> It passed in 3.8 a few times.

Compatible, but I don't see where

scarlehoff:
https://github.com/NNPDF/nnpdf/runs/6545874878?check_suite_focus=true

Here it passed the entire test, but in general for quite a few you can see the 3.8 reached a few steps further than 3.9 (before 3.9 crashes the entire thing)

scarlehoff:
Closing in favour of #1578

PR #1534, opened 2022-03-01
adjust hyper_scan step sizes

RoyStegeman:
Since we were talking about this yesterday.


scarlehoff (approved):
It looks ok to me. I assume this is what you are currently running? (meaning, we know this works)

RoyStegeman:
Well, not now. But indeed, some variation of this is what I always did.

PR #1544, opened 2022-03-22
[WIP] Do additive errors with covmat and then do multiplicative

scarlehoff:
<testing branch>


scarlehoff:
@andreab1997 closing this since I believe the r*sqrt(covmat) is the only right way to do it

PR #1545, opened 2022-03-25
[WIP] Index metadata in fits

### LINKED ISSUE #1512: Search feature for fits
Remembering what exact name you used for a given fit is a pain in the neck (specially since we use non very descriptive names by convention) so it would be nice to have a tool, even if it is only cli, that looks for both the description and the name of the fits.

Something like `vp-search` that would look at a `.json` file in the server with pairs `fit: description`.

For that 1) All fits currently in the server have to be opened and the fit: description pair extracted
2) The same operation must be performed on every fit that we upload

Zaharid:
Access the yaml inside the tar file and pull the metadata.

This is slow and inefficient, and simply lumps all the metadata in the
index file, although it should remain compatible with the old one.

It however really should be a different filename on the server probably.

Questions include how to make it fast and whether to pull all or some of
the metadata, and in which format.

Also probably to have some usable validphys UI.

Closes #1512.

### DIFF
--- a/validphys2/serverscripts/index-fits.py
+++ b/validphys2/serverscripts/index-fits.py
@@ -6,8 +6,14 @@ Generate an index with the existing fits
 import pathlib
 import json
 import itertools
+import tarfile
+import sys
+import pickle
 
-root = '/home/nnpdf/WEB/fits'
+import ruamel_yaml as yaml
+
+root = 'WEB/fits'
+seen_cache_path = "seen_fits.pkl"
 
 #TODO: Find a better way
 glob1 = '*.tar.gz'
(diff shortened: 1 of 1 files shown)


Zaharid:
FWIW this is mostly to get the ball rolling, happy for anyone to push here.

I tried running this on the server but the ssh connection stopped before it could finish, so I don't know how it takes and so on. Which suggest that a progress bar would be useful.

scarlehoff:
> simply lumps all the metadata in the index file

Maybe since it is a lot of information, maybe it makes sense to have it separated?

For instance, the index file stays as it is now and we have another file `fit_metadata.db` which is a database with `{fit: metadata}` (can even be an actuall db like sqlite).

Then to check the database the user (in the background) downloads the database, does the operations offline and asks the server for the exact fit it wants (and then downloads such a fit).

Otherwise we will need to either keep a ssh connection alive (which seems problematic given your second comment) or do many ssh connections (which if i remember correctly cannot be done outside of the right network, don't know whether that was ever fixed).

scarlehoff:
To move this forward a bit, I think we might want to just save from a fit: `name`, `description`, `theory` and `keywords` (we don't have any at the moment but they can be added).

If I loop over all the fits in the server and create a `.json` file with that information it's a bit below 500K for 3200 fits which is ok I think (I've left the `.json` file in `/tmp/jason.json`).

Now the problem is how to search inside, because I'm guessing something `vp-get search nnpdf` will return 3100 of the 3200 fits. One possibility is having `ugrep` in the server (or locally and then download the `.json` file) `ugrep -B4 -Q jason.json` drops you in a relatively nice interface that can be used for this use-case.

Zaharid:
> To move this forward a bit, I think we might want to just save from a fit: `name`, `description`, `theory` and `keywords` (we don't have any at the moment but they can be added).
> 
> If I loop over all the fits in the server and create a `.json` file with that information it's a bit over 500K for 3200 fits which is ok I think (I've left the `.json` file in `/tmp/jason.json`).

This looks sensible.

> 
> Now the problem is how to search inside, because I'm guessing something `vp-get search nnpdf` will return 3100 of the 3200 fits. One possibility is having `ugrep` in the server (or locally and then download the `.json` file) `ugrep -B4 -Q jason.json` drops you in a relatively nice interface that can be used for this use-case.

I suppose  there is a sense in which that is a problem for later. I would like to teach vp how to do `theoryid_of_fit(fitname)` so I'd like that to be easy but otherwise we might be fine with "here is the json file good luck finding anything" right now. Later on, I am probably in the minority, but I find that the interface in vp.nnpdf.science is perfectly usable (the other day I could easily find some report from 2017). Namely a single text field that does a fuzzy match on all fields. I'd be happier to add that if it weren't for the endless complaints and precisely zero contributions.

scarlehoff:
> Later on, I am probably in the minority, but I find that the interface in vp.nnpdf.science is perfectly usable

I also like it very much as well. That's also a good option.

scarlehoff:
I think I agree. For now we can be content with having a file `a_lot_info.json` in the server and later on think about having some magic script or a `vp.nnpdf.science`-like site. I can help with that, but is there a repository for it?

Zaharid:
https://github.com/NNPDF/nnpdf/blob/master/validphys2/serverscripts/WEB/validphys-reports/index.html

PR #1546, opened 2022-03-29
Use `<fitname>.json` by default and deprecate the `.fitinfo` file.

### LINKED ISSUE #1543: Read fit information from `json` file
Doing this will break compatibility with `nnfit` but I believe it's about the right time to do this. Once we move to the new theory or the new commondata it won't make sense to use newer updated versions of vp with old fits.

This will also simplify @Radonirinaunimi efforts for the simultaneous fits (since it is better to parse and write info from some `json` file rather than adding extra lines in a particular order to the `.fitinfo` (and other) files).

Since we are doing `yaml` for the data it might make sense to convert the `json` into `yaml` (which is trivial because it's a reasonably simple dictionary so keeping json/yaml compatibility would be easy).

I've marked this as NNDPF4.1 since I think it is reasonably to expect all compatibility with `nnfit` to be dropped by then.

scarlehoff:
Closes #1543 

From now on the fits won't write all the different files (.fitinfo, .version, .preproc, etc) but everything will be part of a `.json` file.

I've left a fallback to the `.fitinfo` in the reader since we want to be able to compare with old fits. I've removed a few functions that were not used by anything else and I've added a check to ensure that the same version was used to run all replicas of a given fit.  The check was already there in the past but for some reason I put it _at the end_ instead of at the beginning. I guess I was younger then  ¯\_ (ツ)_/¯


Zaharid:
> I've left a fallback to the `.fitinfo` in the reader since we want to be able to compare with old fits. I've removed a few functions that were not used by anything else and I've added a check to ensure that the same version was used to run all replicas of a given fit. The check was already there in the past but for some reason I put it _at the end_ instead of at the beginning. I guess I was younger then ¯_ (ツ)_/¯

Actually I am not sure about the  check: Seems a bit expensive to do at the beginning and it is unclear to me if the behavior should be to fail e.g. a comparefits report.

validphys2/src/validphys/checks.py
```diff
@@ -235,6 +236,18 @@ def check_fits_different(fits):
     """Need this check because oterwise the pandas object gets confused"""
     return _check_list_different(fits, 'fits')
 
+@make_argcheck
+def check_fit_versions_equal(fit):
+    """Check that all replicas in the fit used matching versions of the repository
+    and python modules"""
+    original_version = None
+    for json_path in fit.path.glob("nnfit/replica_*/*.json"):
```
Zaharid, line 244: Probably the second glob should be a bit more specific (there could be other json files). Btw a pet peeve is have these variables file names that mke processing that much harder (the right name is `replica_info.json` or something). Is that too difficult to change?

validphys2/src/validphys/fitdata.py
```diff
@@ -187,22 +206,6 @@ def summarise_fits(collected_fit_summaries):
     return pd.concat(collected_fit_summaries, axis=1)
 
 
-def fit_sum_rules(fit, replica_paths):
```
Zaharid, line 190: If we are removing this (btw should we?), can we remove  `FitSumRulesGrid` as well?

scarlehoff: I'll check. `fit_sum_rules` was using the `.sumrules` file that we do create but it has been empty since forever. And they get computed every time anyway.

validphys2/src/validphys/checks.py
scarlehoff: Actually, I wanted to use the `replica_paths` but was not able to make vp understand that I wanted it there?

> the right name is replica_info.json or something

You mean, instead of <fit_name>.json? If that's your suggestion I agree, I'd be happy to change it.

Radonirinaunimi: The following might not be relevant here at all but in the case of proton+nuclear fits there is not only one single `json` file (since each nucleus has its own `json`) and this way of doing it is the easiest.

scarlehoff: Does it make sense to have different `json` files? What information is different between different nucleus?

Radonirinaunimi: Only the `preprocessing` which in principle could actually **all** be dumped into a single `json` file.

scarlehoff: tbh I don't think the preprocessing is used? So I would not fret about it...

Radonirinaunimi: Yes, this information is not used at all currently, however, in the case of nPDFs where the integrability is not enforced in the way it is done in the free-proton, knowing the effective exponents allows one the verify that a posteriori. 

But I agree that there is actually no need for different `json` files and that specifying the name is the correct way to do it.

scarlehoff: If you are happy with `fitinfo.json` I'll do that (with a fallback for compatibility ofc)

scarlehoff:
> If you are happy with fitinfo.json I'll do that (with a fallback for compatibility ofc)

Actually no, I think it makes sense to use `<fitname>.json` otherwise we will need to keep a compatibility layer forever (because all previous fits had `<fintame>.json`) and it actually makes sense since it follows the same logic of `.exportgrid` and `.dat.

wrt nuclear fits, it's a json file so adding more information is trivial

scarlehoff:
This is ready for review with a caveat, while old fits can still be used for things like `comparefits` (since there's a `_old_load_fitinfo` fallback) one can no longer apply `postfit` to them (by old I mean fits done with `nnfit`).

I believe this is reasonable since we are "destroying C++" and soon it won't even be possible to do fits with `nnfit` (for which one need to use a previous tag of the code which means they also have an older version of `postfit`). Adding a fallback also for that is trivial in case it is needed.

Zaharid:
> This is ready for review with a caveat, while old fits can still be used for things like `comparefits` (since there's a `_old_load_fitinfo` fallback) one can no longer apply `postfit` to them (by old I mean fits done with `nnfit`).
> 
> I believe this is reasonable since we are "destroying C++" and soon it won't even be possible to do fits with `nnfit` (for which one need to use a previous tag of the code which means they also have an older version of `postfit`). Adding a fallback also for that is trivial in case it is needed.

I'm getting cold feed here: It doesn't make sense to use nnfit withput postfit and would be confusing to have it in a semi working state. We could just remove nnfit (that is consistent with current compatibility guarantees) but that would have to be documented and have a tag and so on.

The other option would be to restore compatibility with postfit for the moment.

scarlehoff:
@Zaharid you mean we should remove at this point the possibility to run `nnfit` at all with this version of the code, right?

Zaharid:
yes

scarlehoff:
Would you be happy with just a removal of the executable?

Zaharid:
That, and a tag, and adapting the documentation. 

https://docs.nnpdf.science/search.html?q=nnfit&check_keywords=yes&area=default

And making sure we truly don't have legacy studies.

scarlehoff:
> And making sure we truly don't have legacy studies.

What do you mean by this? VP should still be able to deal with old `nnfit` fits.

RoyStegeman (approved):
Looks good

n3fit/src/n3fit/io/writer.py
```diff
@@ -81,18 +71,12 @@ def write_data(self, replica_path_set, fitname, tr_chi2, vl_chi2, true_chi2):
...
-            integrability_numbers,
-            allchi2_lines,
-            preproc_lines,
-            replica_status.positivity_status,
-            self.timings,
         )
 
+        # write the log file for the chi2
+        chi2_log = "\n".join(self.stopping_object.chi2exps_str())
+        (replica_path_set / "chi2exps.log").write_text(chi2_log)
```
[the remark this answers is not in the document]
scarlehoff: Yes, that and this is never read by vp.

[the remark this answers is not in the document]
RoyStegeman: Ok. Though maybe that could be written to a json/yaml as well, to make life easier if we ever want to use it in vp?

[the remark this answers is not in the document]
scarlehoff: I'd rather just remove it at that point and maybe add an option to store _all_ chi2 (not only every 100 epochs).
I don't know, have you ever really found this log to be useful?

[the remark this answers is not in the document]
RoyStegeman: On occasion I've found it useful to scroll through (a while after doing the fit, so the log was no longer available) so I'd rather not get rid of it entirely. Just wondering if it's worth it to make it a yaml since were touching this now anyway. In that case by the time (if ever) someone implements a reader most of the recent fits will be compatible.

Don't really have strong feelings about this though, just a thought.

doc/sphinx/source/tutorials/run-fit.md
```diff
@@ -176,22 +171,10 @@ Output of the fit
 Every time a replica is finalized, the output is saved to the ```runcard/nnfit/replica_$replica```
 folder, which contains a number of files:
 
-- ``chi2exps.log``: a log file with the χ² of the training every 100 epochs.
+- ``chi2exps.log``: a json-like log file with the χ² of the training every 100 epochs.
```
RoyStegeman, line 174:
```suggestion
- ``chi2exps.log``: a json log file with the χ² of the training every 100 epochs.
```

Zaharid (approved):

PR #1548, opened 2022-04-05
[WIP] Reader for new commondata

scarlehoff:
We used #1526 as a scratchpad during the meeting so I'll leave it be and instead move here whatever is needed.

For now I've added just a jupyter notebook that reads the new commondata into pandas dataframes.

@enocera @Zaharid if you are ok with those df I'll port the rest (`coredata.py` and `core.py`, specs and so) using that as the underlying objects.

(why would the test pass here but not in #1500?)


Zaharid:
I think it is entirely proper to deprecate python 3.8 whenever we get around merging this.

Playground for the new commondata format.ipynb
```diff
@@ -483,20 +505,198 @@
...
+    "@Parser\n",
+    "def ValidPath(path_str: str) -> Path:\n",
+    "    \"\"\"Parse strings into paths\"\"\"\n",
+    "    try:\n",
+    "        return Path(path_str)\n",
+    "    except exception as e:\n",
+    "        raise ValidationError(f\"{path_str} is not a valid path\") from e\n",
+    "        \n",
+    "@Parser\n",
+    "def ValidOperation(op_str: str) -> str:\n",
```
Zaharid: Would it make sense to use enums, or Union[Literal[str]] for better structure here?

Playground for the new commondata format.ipynb
```diff
@@ -483,20 +505,198 @@
...
+    "from validobj.custom import Parser\n",
+    "from validobj import ValidationError, parse_input\n",
+    "\n",
+    "\n",
+    "from validphys import convolution\n",
+    "\n",
+    "\n",
+    "# Scalar parsers\n",
+    "@Parser\n",
+    "def ValidPath(path_str: str) -> Path:\n",
```
Zaharid: Do we want things with more checks, like ExistingFilePath? For fun I tried to reach the error using the visible keys on my keyboard and could not.

Playground for the new commondata format.ipynb
```diff
@@ -483,20 +505,198 @@
...
+    "\n",
+    "\n",
+    "\n",
+    "# Object classes wrappers\n",
+    "@dataclass\n",
+    "class TheoryMeta:\n",
+    "    FK_tables: list\n",
+    "    operation: ValidOperation\n",
+    "    conversion_factor: float = 1.0\n",
+    "    apfelcomb: dict = None\n",
```
Zaharid: This should be Optional[dict]. And probably more structured.

scarlehoff: Plan is to remove all dictionaries (if that's what you mean with the structure)

Playground for the new commondata format.ipynb
scarlehoff: I'll see when I actually use it but I'm not sure at this stage the path will be testable for existence

Playground for the new commondata format.ipynb
scarlehoff: For the OPs list you mean?

Zaharid: Yes indeed. Would make for better error messages.

Playground for the new commondata format.ipynb
Zaharid: Should be fine either way. Don't have to check for everything at this stage but it sure is nice when sensible.

Playground for the new commondata format.ipynb
Zaharid: Yes, but also that as it is the default does not type check.

scarlehoff:
Since it is impossible to read the files that have changed I will move the reader to a new branch off master which the commondata added by Emanuele in #1500 but this branch will not target that one nor the old data will be removed. We can merge both branches once this is finished.

PR #1556, opened 2022-04-13
Changing generation of pseudodata

andreab1997:
Currently, the pseudodata are generated again each time even one of them is negative (and not ASY) before going on.  From a theoretical point of view this is strange, because it is equivalent to cut the gaussians distribution before zero.  Moreover, from a performance point of view, this easily becomes a bottleneck (for example in the case of fit with thcovmat in which the shifts are bigger). 
We want to try doing something slightly different, namely taking the absolute value of each negative pseudodata. From a theoretical point of view this makes no sense as the actual method, but, at least, it requires only a single iteration in every case.

### DIFF
--- a/validphys2/src/validphys/pseudodata.py
+++ b/validphys2/src/validphys/pseudodata.py
@@ -230,12 +230,10 @@ def make_replica(
     return shifted_pseudodata
 
 def delta(shifts, data, mask):
-    new_shifts = []
-    for sh, dat, ma in zip(shifts, data, mask):
-        if ma:
-            if sh < 0.:
-                sh = dat*(np.exp(sh/dat)-1.)
-        new_shifts.append(sh)
+    new_shifts = shifts.copy()
+    neg = shifts < 0.
+    negtofix = np.logical_and(neg, mask)
+    new_shifts[negtofix] = data[negtofix] * (np.exp(shifts[negtofix] / data[negtofix]) - 1.)
     return np.array(new_shifts)
 
 
--- a/validphys2/src/validphys/pseudodata.py
+++ b/validphys2/src/validphys/pseudodata.py
@@ -224,7 +224,6 @@ def make_replica(
         mult_part = np.concatenate(mult_shifts, axis=0)*special_mult
     #Shifting pseudodata
     shifted_pseudodata = (all_pseudodata + shifts)*mult_part
-    import ipdb; ipdb.set_trace()
     #positivity control
     for index,dat in enumerate(shifted_pseudodata[full_mask]):
         if dat < 0.:
--- a/validphys2/src/validphys/pseudodata.py
+++ b/validphys2/src/validphys/pseudodata.py
@@ -225,9 +225,7 @@ def make_replica(
     #Shifting pseudodata
     shifted_pseudodata = (all_pseudodata + shifts)*mult_part
     #positivity control
-    for index,dat in enumerate(shifted_pseudodata[full_mask]):
-        if dat < 0.:
-            shifted_pseudodata[index] = np.abs(dat)        
+    shifted_pseudodata[full_mask] = np.abs(shifted_pseudodata[full_mask])        
 
     return shifted_pseudodata
 
--- a/validphys2/src/validphys/pseudodata.py
+++ b/validphys2/src/validphys/pseudodata.py
@@ -198,35 +198,37 @@ def make_replica(
     full_mask=np.concatenate(check_positive_masks, axis=0)
     # The inner while True loop is for ensuring a positive definite
     # pseudodata replica
-    while True:
-        mult_shifts = []
-        # Prepare the per-dataset multiplicative shifts
-        for mult_uncorr_errors, mult_corr_errors in nonspecial_mult:
-        # convert to from percent to fraction
-            mult_shift = (
-                1 + mult_uncorr_errors * rng.normal(size=mult_uncorr_errors.shape) / 100
-            ).prod(axis=1)
 
-            mult_shift *= (
-                1 + mult_corr_errors * rng.normal(size=(1, mult_corr_errors.shape[1])) / 100
-            ).prod(axis=1)
+    mult_shifts = []
+    # Prepare the per-dataset multiplicative shifts
+    for mult_uncorr_errors, mult_corr_errors in nonspecial_mult:
+    # convert to from percent to fraction
+        mult_shift = (
+            1 + mult_uncorr_errors * rng.normal(size=mult_uncorr_errors.shape) / 100
+        ).prod(axis=1)
 
-            mult_shifts.append(mult_shift)
+        mult_shift *= (
+            1 + mult_corr_errors * rng.normal(size=(1, mult_corr_errors.shape[1])) / 100
+        ).prod(axis=1)
+
+        mult_shifts.append(mult_shift)
             
-        #If sep_mult is true then the multiplicative shifts were not included in the covmat
-        shifts = covmat_sqrt @ rng.normal(size=covmat.shape[1])
-        mult_part = 1.
-        if sep_mult:
-            special_mult = (
-                1 + special_mult_errors * rng.normal(size=(1, 
-                special_mult_errors.shape[1])) / 100
-                ).prod(axis=1)
-            mult_part = np.concatenate(mult_shifts, axis=0)*special_mult
-        #Shifting pseudodata
-        shifted_pseudodata = (all_pseudodata + shifts)*mult_part
-        #positivity control
-        if np.all(shifted_pseudodata[full_mask] >= 0):
-            break
+    #If sep_mult is true then the multiplicative shifts were not included in the covmat
+    shifts = covmat_sqrt @ rng.normal(size=covmat.shape[1])
+    mult_part = 1.
+    if sep_mult:
+        special_mult = (
+            1 + special_mult_errors * rng.normal(size=(1, 
+        special_mult_errors.shape[1])) / 100
+        ).prod(axis=1)
+        mult_part = np.concatenate(mult_shifts, axis=0)*special_mult
+    #Shifting pseudodata
+    shifted_pseudodata = (all_pseudodata + shifts)*mult_part
+    import ipdb; ipdb.set_trace()
+    #positivity control
+    for index,dat in enumerate(shifted_pseudodata[full_mask]):
+        if dat < 0.:
+            shifted_pseudodata[index] = np.abs(dat)        
 
     return shifted_pseudodata
 
--- a/validphys2/src/validphys/pseudodata.py
+++ b/validphys2/src/validphys/pseudodata.py
@@ -222,12 +222,23 @@ def make_replica(
         special_mult_errors.shape[1])) / 100
         ).prod(axis=1)
         mult_part = np.concatenate(mult_shifts, axis=0)*special_mult
+    #regularize the negative shifts
+    reg_shifts = delta(shifts, all_pseudodata, full_mask)
     #Shifting pseudodata
-    shifted_pseudodata = (all_pseudodata + shifts)*mult_part
-    #positivity control
-    shifted_pseudodata[full_mask] = np.abs(shifted_pseudodata[full_mask])        
-
+    shifted_pseudodata = (all_pseudodata + reg_shifts)*mult_part
+          
     return shifted_pseudodata
+
+def delta(shifts, data, mask):
+    new_shifts = []
+    for sh, dat, ma in zip(shifts, data, mask):
+        if ma:
+            if sh < 0.:
+                sh = dat*(np.exp(sh/dat)-1.)
+        new_shifts.append(sh)
+    return np.array(new_shifts)
+
+
 
 
 def indexed_make_replica(groups_index, make_replica):


AleCandido:
I change the base of the PR to minimize the diff (and in particular the number of commits), to keep only changes related to the topic.

AleCandido:
Example:
```py
[ins] In [1]: import numpy as np
[ins] In [2]: a = np.zeros(10)
[ins] In [3]: i = np.array([2, 5, 9])
[ins] In [4]: a[i] = 1
[ins] In [5]: a
Out[5]: array([0., 0., 1., 0., 0., 1., 0., 0., 0., 1.])
```

validphys2/src/validphys/pseudodata.py
```diff
@@ -198,35 +198,36 @@ def make_replica(
...
+            1 + special_mult_errors * rng.normal(size=(1, 
+        special_mult_errors.shape[1])) / 100
+        ).prod(axis=1)
+        mult_part = np.concatenate(mult_shifts, axis=0)*special_mult
+    #Shifting pseudodata
+    shifted_pseudodata = (all_pseudodata + shifts)*mult_part
+    #positivity control
+    for index,dat in enumerate(shifted_pseudodata[full_mask]):
+        if dat < 0.:
+            shifted_pseudodata[index] = np.abs(dat)
```
AleCandido, line 230:
```suggestion
    shifted_pseudodata[full_mask] = np.abs(shifted_pseudodata[full_mask])
```

andreab1997: This seems more clever ;)

Zaharid:
FWIW something that we may as well do is setting the negative pseudodata to zero, which incidentally is what we recommend other people to do

http://nnpdf.mi.infn.it/wp-content/uploads/2019/03/NNPDFfits_Positivity_PhysicalCrossSections_v2.pdf

validphys2/src/validphys/pseudodata.py
```diff
@@ -222,13 +222,24 @@ def make_replica(
...
     return shifted_pseudodata
 
+def delta(shifts, data, mask):
+    new_shifts = []
+    for sh, dat, ma in zip(shifts, data, mask):
+        if ma:
+            if sh < 0.:
+                sh = dat*(np.exp(sh/dat)-1.)
+        new_shifts.append(sh)
+    return np.array(new_shifts)
```
AleCandido, line 239:
```suggestion
def delta(shifts, data, mask):
    new_shifts = shifts.copy()
    
    neg = shifts < 0
    negtofix = np.logical_and(neg, mask)
    
    new_shifts[negtofix] = data[negtofix] * (np.exp(shifts[negtofix] / data[negtofix]) - 1.)
    return new_shifts
```

AleCandido: In this way you can avoid looping in Python

AleCandido:
> FWIW something that we may as well do is setting the negative pseudodata to zero, which incidentally is what we recommend other people to do
> 
> http://nnpdf.mi.infn.it/wp-content/uploads/2019/03/NNPDFfits_Positivity_PhysicalCrossSections_v2.pdf

Hi @Zaharid, sorry not to have replied before, there were holidays in the middle: you're right, that's an honest solution, but I like more the ELU regularization: 0 itself is rather a degenerate value, and the resulting distribution has a discrete weight on a single point.

It might be all right, but I prefer to warp everything a bit, and obtain the final distribution deforming the original one (it is not a Gaussian in any case, not even with truncation).

Zaharid:
Fwiw the proposed solution seems to me worse than the status quo.  It has the unpleasant feature that big negative deviations become big postive ones for no reason while it is unclear what should happen with the corrlations. With the truncated gaussian one at least gets a proportional rescaling of the positive part and sensible enough correlations. 

In any case I would be interested in seeing how much of a difference do all of these choices make at the end of a fit.

AleCandido:
> Fwiw the proposed solution seems to me worse than the status quo. It has the unpleasant feature that big negative deviations become big postive ones for no reason while it is unclear what should happen with the corrlations. With the truncated gaussian one at least gets a proportional rescaling of the positive part and sensible enough correlations.
> 
> In any case I would be interested in seeing how much of a difference do all of these choices make at the end of a fit.

I guess you're not catching up with the updates: the reflection was only a first attempt, with the only rationale of applying a transformation instead of retrying.
If you call a "bad" point one for which the variance is so large that the central value is almost zero (such that the Gaussian is cut approximately in half), you see that the status quo is exponentially slow in the number of bad points: since the points are bulk extracted (generated by extracting a vector of iid normal variables and rotating it by the Cholesky factor of the covmat), in order to get all positive you have a chance that is `1/2^b`, with `b` the number of bad points.

The technical improvement was the only improvement we were seeking at the beginning, but the current solution (the [ELU](https://en.wikipedia.org/wiki/Rectifier_(neural_networks)#ELU)) has also the property you desire (not scattering the points around), the property I desire (not putting all the weight at 0), the technical improvement, and it's improving a bit w.r.t. to the status quo (since it's keeping the median, and biasing less the average on the right).

The plan is that the implementation is so simple that is already done [e7fa6ad](https://github.com/NNPDF/nnpdf/pull/1556/commits/e7fa6ad2ed246270fab83dbf28243a6b0bf2fece), you can see that the whole PR is a small change in code, and @andreab1997 is already running a fit to estimate the impact.

The technical improvement is required for further studies with the theory covmat, that can blow up the variance of some points, driving many of them to be baddish.

validphys2/src/validphys/pseudodata.py
```diff
@@ -198,38 +198,46 @@ def make_replica(
...
+    reg_shifts = delta(shifts, all_pseudodata, full_mask)
+    #Shifting pseudodata
+    shifted_pseudodata = (all_pseudodata + reg_shifts)*mult_part
+          
     return shifted_pseudodata
 
+def delta(shifts, data, mask):
+    new_shifts = shifts.copy()
+    neg = shifts < 0.
+    negtofix = np.logical_and(neg, mask)
```
Zaharid, line 235: You can’t count on the data being positive in the general case that includes closure tests. Also not clear to me it is the right scale to set (but maybe the uncertainty is?).

AleCandido: Wait, how come that data are not positive?

In that case I would give up, and just regularize data with a `max(d, 0)`. I'm pretty sure we'll never get a negative central value from experimentalists, for a quantity supposed to be positive.

AleCandido: > Also not clear to me it is the right scale to set (but maybe the uncertainty is?).

This actually I didn't understand, which scale are you talking of? However, the shape has been optimized for:

- mapping `0` shift in `0`
- mapping `-inf` to obtain `0` for the sampled datum (so the shift is `-datum`)
- have continuous first derivative (i.e. derivative equal to `1` for shift equal `0`)

Zaharid: In principle we may want to not have to deal with experimental data but rather with theory predictions, which can logically (and practically) be negative. The most popular application of that is closure tests, but also the kind of thing we do in Cambridge these days (what happens if I add new physics to this pristine self consistent environment where the fake data has been obtained from the predictions?).

AleCandido: Then I'm sorry, but you are asking for something more than the status quo: at the moment data are imposed positive by *retrying* strategy. This PR is not a generalization of that, but simply it's trying to be a more efficient (and possibly slightly less biasing) implementation of that.

If you want to allow for negative elements, then you should bypass the impose positivity (that was pre-existing to this PR), that is a generalization and it's very simple to add to both implementations: bypass it.
Still, it is not in the current scope, but not even farther because of this.

Zaharid: The point is not to allow for negative pseudodata, but rather for negative "experimental" central values. Which works now albeit in a painstaking way.

AleCandido: Ok, you're right, but then it might be even worse than the badness cited above: you can have a rate lower than 50% for a single point, so:
1. if it is an isolated feature, you can decide whatever you wish among `max(d, 0)` and discarding the point
2. if it is consistent over a set of points, then even the former was not a solution: you'd develop an incredibly high rejection rate

If you're in case 2., I would suggest dropping pseudodata positivity (and thus bypass the update above), because it's absurd to fit positive pseudodata for actual negative central values.
Pseudodata positivity is an assumption on the distribution, but it is not mandatory: if the thing you're fitting makes sense to be negative, then use appropriate distributions.

Zaharid: I guess I am saying that I would consider a proposal to either clip negative values at zero or to do nothing and rely on other positivity constraining features to be potentially advantageous. But I think the proposed solution as is now breaks the aforementioned existing use case.

AleCandido: > But I think the proposed solution as is now breaks the aforementioned existing use case.

I believe there is no existing use case available, because while theoretically possible, it requires an exponential time, so it's only useful when it is irrelevant.
The issue we attempted to face in this PR is: is it possible to have a technically light implementation for the current use case, namely positive central values? And the answer seems to be yes.

The scope was a brief technical PR, for a proper extension to negative central values (not supported by the status quo) let's open a new one. I guess we can make both behaviors available to the runcard (clipping and nothing).

AleCandido: I'm in favor to expose to the runcard even current and nothing at pseudodata level, but not *retrying*, that is only bad.

```yaml
pseudodata_positivity: elu/null
central_values_positivity: clip/null
```
(keys are random, you know better than me where to fit it in the runcard)

AleCandido: Of course choosing `elu` and `null` for central values makes no sense, but for the usual fit central values **are** positive, so you will see the difference only in the other cases.

andreab1997:
I have now the reports of the fits done with the ELU, so we can just look to the results.
https://vp.nnpdf.science/JfvnfqGKTvWEfo7oHiFmQw==
https://vp.nnpdf.science/DiO8TovvQiG4oAgqgJB7-g==
The first is the comparison between baselines (so without thcovmat), one with the ELU and the other with the "old" method.
The second is the same but for a thcovmat fit. 
I believe that the results are really similar in both cases, so probably in this case using the ELU gives really just an improvement on the performances.

Zaharid:
Thanks, having a simpler mechanism guaranteed to terminate in a finite amount of time seems like a good idea. That said we could as well use some option that works in various corner cases as per the discussion above.

AleCandido:
> Thanks, having a simpler mechanism guaranteed to terminate in a finite amount of time seems like a good idea. That said we could as well use some option that works in various corner cases as per the discussion above.

I would keep this PR as small as possible (I'd keep PRs in general small, but I'm the first one to fail...). I like this one to be the default, especially now that is proven to work as good as the previous one (at the level of results).

But I'd support a new PR to expose the option of above (actually extended) to the runcard:
```yaml
pseudodata_positivity: elu/clip/null
central_values_positivity: clip/null
```
`clip` is just `max(d,0)`, `null` is no positivity applied, and `elu` is the one of this PR (only makes sense with a reference, so only for pseudodata, not for central values). This should be very simple to implement, about 5 lines of code (just branching and `max`) + parsing and propagating runcards options.

Zaharid:
I’d prefer this to be yet another configuration option that needs to be kept in sync, especially if it is unlikely to make a difference. But we likely would have to maintain the old one in order to be able to make prove that it is equivalent. However the configuration  should be something to remove in the future rather than to expand on. I would think nobody is in the business of launching a big research project into sampling pseudodata that requires permanent infrastructure in the code. Even if those that do, in the past we have made the mistake of keeping too much testing code that was irrelevant to the published work (the theory covmat module is an example if that). 

Even if there end up being more options as far as I am concerned, a hard requirement for the default is to work with negative central experimental values (which have come up despite the apparent silliness).

 I think it is safe to assume that if elu doesn’t make a big difference none of the alternatives will , while personally I am unpersuaded of its benefits. In particular, all things equal, following our own public advice seems like the most sensible choice to me.

AleCandido:
> But we likely would have to maintain the old one in order to be able to make prove that it is equivalent. 

I'm not that sure: we have the old code for reproducibility, it's the purpose of the tags.
So, I'd say we don't have to maintain any weird bit of code written at some point in the past, especially if we have already proven it's not making much difference wrt the new one.

AleCandido:
> However the configuration should be something to remove in the future rather than to expand on. 

> Even if there end up being more options as far as I am concerned, a hard requirement for the default is to work with negative central experimental values (which have come up despite the apparent silliness).

We have already plenty of silly configurations, this is in no way worse than the multiplicative errors. 

And, as said, not even the old one sad working in practice with negative central values. Negative central values are not a feature of baseline fit, so I can agree we can support, but no reason to support them by default.

AleCandido:
> I would think nobody is in the business of launching a big research project into sampling pseudodata that requires permanent infrastructure in the code. 

"Big research project" is for sure too ambitious, but there is some interest to play with pseudodata generation. 
Stefano F mentioned we might like to implement Poisson distributed pseudodata, for a very large x dataset that is waiting since years. 

I see no reason not to keep some variations, especially if they are so simple.
We can document and unit test much better than what has been done, and for testing within the global fit we only need the default to be sensible. The other you can consider part of the "human hyperopt" for each specific problem. 

In particular, since closure tests have to reproduce the actual methodology, I'd clip level 1 pseudodata (central values), rather than allowing for negative ones. However, let's discuss at the code meeting.

AleCandido:
> I think it is safe to assume that if elu doesn’t make a big difference none of the alternatives will , while personally I am unpersuaded of its benefits.

This is not true, as we have already seen with reflection. That was silly, but in principle pseudodata should have been not so negative, such that it would have not "made a big difference".

If you have a suggestion, let's test it, before discussing on conjectures.

Zaharid:
> > But we likely would have to maintain the old one in order to be able to make prove that it is equivalent.
> 
> I'm not that sure: we have the old code for reproducibility, it's the purpose of the tags. So, I'd say we don't have to maintain any weird bit of code written at some point in the past, especially if we have already proven it's not making much difference wrt the new one.

Please let's not re litigate [well established and publicly documented policies](https://docs.nnpdf.science/releases.html) in every thread.

AleCandido:
> Please let's not re litigate [well established and publicly documented policies](https://docs.nnpdf.science/releases.html) in every thread.

Instead of pointing me to the docs, point to the minutes:
image
image

However, we might even maintain the old bugs, but they should be flagged as deprecated, and the next major release can't be something that will happen in 10 years.

Potentially, 4.x series should last for as long as possible, since we roll major only in case of actual major updates, that don't 
have to happen necessarily: the more stable we are, the better it is.
But this should not block smaller updates forever!

Zaharid:
I had not replied because I have no interest in discussing policies. 

I believe there is room for improvement in pseudodata sampling. A better solution would be faster, guaranteed to converge, reflect  a reasonable interpretation of the experimental data, and work in all the same corner cases as the current algorithm. 

Whatever is done, it won’t be a simple two line change. The details would need to be flashed out and discussed. Extensive comparisons would be needed (including the kind of studies Rosalyn and Richard were doing at some point) and backward compatibility as described in the document I linked would have to be maintained.

scarlehoff:
I'm going to close this PR since it has been stale for more than one year and by now the whole set of new fits will be made with the "old" sampling.
By now any change to the sampling will require some long discussions at the PC and wait until 4.1.

AleCandido:
> By now any change to the sampling will require some long discussions at the PC and wait until 4.1.

But do we have any plan of actually discussing it?

scarlehoff:
That's up to you and @andreab1997 to bring it up at, I guess, Amsterdam (if all the other projects are done and we can start thinking about new stuff)

AleCandido:
Then we should keep as an issue, isn't it?

You're right that the PR is stale and most likely incompatible with the present status of the code (so leading to conflicts). But the original issue is unchanged, isn't it?

PR #1557, opened 2022-04-15
Fix the chi2 computation in the report when there is a thcovmat

andreab1997:
At the moment the thcovmat is not considered for the computation of the chi2 in the report even if the flag `use_thcovmat_if_present` is True in `vp-comparefits`.
With this, the full thcovmat is used to compute the total chi2 and an appropriate cut thcovmat is used to compute the chi2 by process or by experiment (or even by dataset).


Zaharid:
Could you please make a PR against master instead so that the changes are easier to review and discuss? Something that should work in simple situations is switching to master, creating a new branch off from there and cherry picking the relevant commits.

andreab1997:
> At the moment the thcovmat is not considered for the computation of the chi2 in the report even if the flag `use_thcovmat_if_present` is True in `vp-comparefits`.
> With this, the full thcovmat is used to compute the total chi2 and an appropriate cut thcovmat is used to compute the chi2 by process or by experiment (or even by dataset).

Yes, I am doing it. Do you want to close this?

PR #1561, opened 2022-04-22
Improve remote keyword functionality

Zaharid:
`vp-comparefits` and other scripts would suggest remote keywords when in  interactive mode. The implementation had some problems as it would raise an exception whenever the user had no permission to access vp server which would in turn crash the whole app. The solution there was to issue the appropriate `raise_for_status` command. This was made worse by the fact that this particular completer would not respect the `--no-net` flag and try to connect anyway.

 While we are on it, tidy up the implementation, moving the bulk of it to RemoteLoader instead. With that make sure that comarefits only looks at that when the remote loader is in use.

cc @J-M-Moore

### DIFF
--- a/validphys2/src/validphys/promptutils.py
+++ b/validphys2/src/validphys/promptutils.py
@@ -58,8 +58,8 @@ def confirm(message, default=None):
     session = PromptSession(complete_message, key_bindings=bindings)
     return session.prompt()
 
-#We need some sort of cache because prompt_toolkit calls the callable
-#every time it tries to complete.
+# We need some sort of cache because prompt_toolkit calls the callable
+# every time it tries to complete.
 class KeywordsWithCache():
     def __init__(self, loader):
         self.loader = loader
@@ -70,7 +70,7 @@ class KeywordsWithCache():
             if hasattr(self.loader, "remote_keywords"):
                 try:
                     self.words = self.loader.remote_keywords
-                # Catch a borad exception here as we don't want the completion
+                # Catch a broad exception here as we don't want the completion
                 # to break the app
                 except Exception as e:
                     self.words= []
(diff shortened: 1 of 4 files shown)


validphys2/src/validphys/promptutils.py
```diff
@@ -57,3 +57,23 @@ def nothing(event):
...
+    def __init__(self, loader):
+        self.loader = loader
+        self.words = None
+
+    def __call__(self):
+        if self.words is None:
+            if hasattr(self.loader, "remote_keywords"):
+                try:
+                    self.words = self.loader.remote_keywords
+                # Catch a borad exception here as we don't want the completion
```
RoyStegeman, line 73:
```suggestion
                # Catch a broad exception here as we don't want the completion
```

validphys2/src/validphys/promptutils.py
```diff
@@ -57,3 +57,23 @@ def nothing(event):
     complete_message = merge_formatted_text([message, yes_no_str(default)])
     session = PromptSession(complete_message, key_bindings=bindings)
     return session.prompt()
+
+#We need some sort of cache because prompt_toolkit calls the callable
+#every time it tries to complete.
```
RoyStegeman, line 62:
```suggestion
# We need some sort of cache because prompt_toolkit calls the callable
# every time it tries to complete.
```

validphys2/src/validphys/promptutils.py
```diff
@@ -57,3 +57,23 @@ def nothing(event):
...
+class KeywordsWithCache():
+    def __init__(self, loader):
+        self.loader = loader
+        self.words = None
+
+    def __call__(self):
+        if self.words is None:
+            if hasattr(self.loader, "remote_keywords"):
```
RoyStegeman, line 70: Why is this? At the moment if someone uses this with anything other than a `RemoteLoader` they're probably doing something strange/wrong, so maybe it should just error in such a case.

Zaharid: `vp-comparefits` could be using a `Loader` rather than a `FallbackLoader` if you pass the `--no-net` flag. But fine, I'll make that more explicit.

RoyStegeman (approved):
Looks good

PR #1563, opened 2022-04-26
Add note on t0 usage

Zaharid:
The docs would not do a good job at specifying how to use the t0 chi2.
Add a reference to the main docs so at least it is somewhere visible.


Zaharid:
Incidentally it would be great if we had an example of e.g. `dataspecs_chi2_table` displaying the various options. Would also be a good test. cc @andreab1997.

PR #1566, opened 2022-05-07
libnnpdf/wrapper: don't try to install to system PYTHONPATH

veprbl:
The standard convention is for the libraries to be installed to python
prefix relative to the installation prefix, not the system prefix. It
is the a reponsibility of the user or their package manager to
relocate the files or set the correct PYTHONPATH.

This avoids installation errors like:

    CMake Error at libnnpdf/wrapper/cmake_install.cmake:49 (file):
      file cannot create directory:
      /path/to/lib/python3.X/site-packages/NNPDF.
      Maybe need administrative privileges.
    Call Stack (most recent call first):
      libnnpdf/cmake_install.cmake:73 (include)
      cmake_install.cmake:50 (include)

### DIFF
--- a/libnnpdf/wrapper/CMakeLists.txt
+++ b/libnnpdf/wrapper/CMakeLists.txt
@@ -17,7 +17,7 @@ endif()
 execute_process(
   COMMAND "${PYTHON_EXECUTABLE}" -c "if True:
     from distutils.sysconfig import get_python_lib;
-    print(get_python_lib())"
+    print(get_python_lib(prefix=\"${CMAKE_INSTALL_PREFIX}\"))"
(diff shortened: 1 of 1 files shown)


Zaharid (commented):
This looks like a good change. Thanks!

Zaharid (approved):

Zaharid:
Not sure why CI is not running here and see no way to trigger it. Did try this locally. cc @scarlehoff

veprbl:
Github actions were disabled in my fork by default. Should work now, I guess.

veprbl:
So they failed, as I wouldn't have access to your secrets:
https://github.com/veprbl/nnpdf/runs/6344065371?check_suite_focus=true

Zaharid:
Seems good enough

PR #1570, opened 2022-05-18
Fix bugs in closuretest

VLTSML26:
Found bugs after merge of PR #1329, #1324 and #1321 (6th April) in `master` when calling functions from `multiclosure_output.py`. This is probably due to call of outdated functions or new behavior of already existing ones, e.g. LHAPDF loaders.

### DIFF
--- a/validphys2/src/validphys/closuretest/closure_checks.py
+++ b/validphys2/src/validphys/closuretest/closure_checks.py
@@ -110,7 +110,7 @@ def check_multifit_replicas(fits_pdf, _internal_max_reps, _internal_min_reps):
     fit.
 
     """
-    n_reps = {pdf.get_members() for pdf in fits_pdf}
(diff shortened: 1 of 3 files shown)


scarlehoff (approved):
Thanks for this.

Sadly the tests cannot be run but I think we can merge this (obviously these functions are not tested because otherwise the tests would've picked it up after the merge to master).

I ran the test in my own computer and the pass FWIW

RoyStegeman:
> Sadly the tests cannot be run but I think we can merge this (obviously these functions are not tested because otherwise the tests would've picked it up after the merge to master).

Yes this should not cause a problem, we have just been sloppy when merging the update to the PDF interface. Don't think it's really worth it to add test for this though?

RoyStegeman (approved):

RoyStegeman (approved):

PR #1573, opened 2022-05-31
Add overfit metric to vp

RoyStegeman:
As discussed in Amsterdam I've implemented the overfit metric to vp, and included a plot and table by default in the comparefit report.

I'll open the metric per dataset grouping in a later PR.

I don't really like how it's implemented in the comparefits report, but I believe including something like `fits::fitcontext` in the loop already means that the default `use_t0=False`  is included in the loop. 

example report: https://vp.nnpdf.science/ihtLSPpKTBa45jD2UEH4bw==

Also to prevent confusion we may went to use something other than "deltachi2", but ideally also something better than "overfit metric"?

### DIFF
--- a/validphys2/src/validphys/checks.py
+++ b/validphys2/src/validphys/checks.py
@@ -154,7 +154,7 @@ def check_have_two_pdfs(pdfs):
 
 
 @make_argcheck
-def check_at_least_two_pdfs(pdfs):
+def check_at_least_two_replicas(pdfs):
     for pdf in pdfs:
         # The get_members function also includes the central value replica, 
         # therefore we need it to be larger than 3
--- a/validphys2/src/validphys/checks.py
+++ b/validphys2/src/validphys/checks.py
@@ -153,6 +153,14 @@ def check_have_two_pdfs(pdfs):
     check(len(pdfs) == 2,'Expecting exactly two pdfs.')
 
 
+@make_argcheck
+def check_at_least_two_pdfs(pdfs):
+    for pdf in pdfs:
+        # The get_members function also includes the central value replica, 
+        # therefore we need it to be larger than 3
+        check(pdf.get_members() >= 3,'Expecting at least two replicas.')
+
+
 #The indexing to one instead of zero is so that we can be consistent with
 #how plot_fancy works, so normalize_to: 1 would normalize to the first pdf
 #for both.
--- a/validphys2/src/validphys/overfit_metric.py
+++ b/validphys2/src/validphys/overfit_metric.py
@@ -1,7 +1,8 @@
 """
 overfit_metric.py
 
-This calculates the overfit metric
+This module contains the functions used to calculate the overfit metric and 
+produce the corresponding tables and figures.
 """
 
 import logging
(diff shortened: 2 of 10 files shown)


validphys2/src/validphys/overfit_metric.py
```diff
@@ -0,0 +1,128 @@
...
+from reportengine.figure import figure
+from reportengine.table import table
+
+log = logging.getLogger(__name__)
+
+preds = collect("predictions", ("pdfs","dataset_inputs",))
+
+
+def array_expected_delta_chi2(calculate_chi2s_per_replica, replica_data):
+
```
scarlehoff, line 24: Please add a docstring here (and all functions in this module)

validphys2/src/validphys/overfit_metric.py
```diff
@@ -0,0 +1,128 @@
...
+    res_over = calculate_chi2s_per_replica
+
+    fitted_val_erf = np.array([info.validation for info in replica_data])
+
+    number_pdfs = res_over.shape[0]
+    list_expected_delta_chi2 = []
+    for _ in range(number_pdfs * 1000):
+        mask = np.random.randint(0, number_pdfs, size=int(0.95 * number_pdfs))
```
scarlehoff, line 32: If `number_pdfs=1` (I know, it is a patological case). This will return an empty list. Is that expected?

Related to that, maybe the `1000` should instead be an input parameter so that it can be modified more easily.

validphys2/src/validphys/overfit_metric.py
```diff
@@ -0,0 +1,128 @@
...
+
+        list_expected_delta_chi2.append(expected_delta_chi2)
+    return np.array(list_expected_delta_chi2)
+
+
+def _create_new_val_pseudodata(pdf_data_index, fit_data_indices_list):
+    vl_data_fitrep = []
+    for fitreplica_info in fit_data_indices_list:
+        vl_data_fitrep.append(fitreplica_info.pseudodata.loc[pdf_data_index.val_idx])
+    return np.array(vl_data_fitrep).squeeze()
```
scarlehoff, line 48: The `squeeze` without argument can be tricky because (4,2,1) will go to (4,2) but (4,1,1) will go to (4,). I think it is better to specify the axis (or axes) that you want to squeeze.

scarlehoff (commented):
A few style changes suggested:

- A vertical line at 0 (so that it is easy to see how far away the result is!)
- Instead of mean/bootstrap error I would call it "overfitteness" (and make it absolute value). Or "sigmas away from 0". It's just easier to read.

validphys2/src/validphys/checks.py
```diff
@@ -153,6 +153,14 @@ def check_have_two_pdfs(pdfs):
     check(len(pdfs) == 2,'Expecting exactly two pdfs.')
 
 
+@make_argcheck
+def check_at_least_two_pdfs(pdfs):
+    for pdf in pdfs:
+        # The get_members function also includes the central value replica, 
+        # therefore we need it to be larger than 3
+        check(pdf.get_members() >= 3,'Expecting at least two replicas.')
```
scarlehoff, line 161: I think here you want `check_at_least_two_replica` or `_two_members`.

validphys2/src/validphys/overfit_metric.py
```diff
@@ -0,0 +1,217 @@
...
+    "predictions",
+    (
+        "pdfs",
+        "dataset_inputs",
+    ),
+)
+
+
+def _create_new_val_pseudodata(pdf_data_index, fit_data_indices_list):
+    """Loads all validation psuedodata replicas used during the fiting of the
```
scarlehoff, line 33:
```suggestion
    """Loads all validation pseudodata replicas used during the fiting of the
```

validphys2/src/validphys/overfit_metric.py
```diff
@@ -0,0 +1,217 @@
...
+    pdf replicas
+
+    Returns
+    -------
+    np.ndarray
+        (nrep,ndata) sized numpy array containingt the validation data used to
```
scarlehoff, line 39:
```suggestion
        (nrep,ndata) sized numpy array containing the validation data used to
```

validphys2/src/validphys/overfit_metric.py
```diff
@@ -0,0 +1,217 @@
...
+
+def array_expected_overfitting(
+    calculate_chi2s_per_replica,
+    replica_data,
+    number_of_resamples=1000,
+    resampling_fraction=0.95,
+):
+    """Calculates the expected difference in chi2 between:
+    1. The chi2 of a PDF replica calculated using the corresponding pseudodata
+        replica using during the fit
```
scarlehoff, line 121:
```suggestion
        replica used during the fit
```

validphys2/src/validphys/overfit_metric.py
```diff
@@ -0,0 +1,217 @@
...
+    2. The chi2 of a PDF replica calculated using an alternative i.i.d random
+        pseudododata replicas
+
+    The expected difference along with an error estimate is obtained through a
+    bootstrapping consisting of `number_of_resamples` resamples per pdf replica
+    where each resampling contains a fraction `resampling_fraction` of all
```
scarlehoff, line 127:
```suggestion
    bootstrapping consisting of ``number_of_resamples`` resamples per pdf replica
    where each resampling contains a fraction ``resampling_fraction`` of all
```

validphys2/src/validphys/overfit_metric.py
```diff
@@ -0,0 +1,217 @@
...
+    replicas.
+
+    Parameters
+    ----------
+    calculate_chi2s_per_replica : np.ndarray
+        validation chi2 per pdf replica
+    replica_data : list(vp.fitdata.FitInfo)
+
```
scarlehoff, line 135:
```suggestion
```

validphys2/src/validphys/overfit_metric.py
```diff
@@ -0,0 +1,217 @@
...
+
+@check_at_least_two_pdfs
+def calculate_chi2s_per_replica(
+    recreate_pdf_pseudodata_no_table,
+    preds,
+    dataset_inputs,
+    groups_covmat_no_table,
+):
+    """Calculates the chi2 for each PDF relica to many pseudodata replicas.
```
scarlehoff, line 57:
```suggestion
    """Calculates, for each PDF replica, the chi2 of the validation with the pseudodata generated
    for all other replicas in the fit
```

I've changed this a bit but please add more stuff if you wish. I think this is the "meat" of the method so I think it should be very precisely described.

validphys2/src/validphys/overfit_metric.py
```diff
@@ -0,0 +1,217 @@
...
+    Parameters
+    ----------
+    calculate_chi2s_per_replica : np.ndarray
+        validation chi2 per pdf replica
+    replica_data : list(vp.fitdata.FitInfo)
+
+    number_of_resamples : int, optional
+        number of resamples per pdf replica, by default 1000
+    resampling_fraction : float, optional
+        fraction of replcias used in the bootstrap resampling, by default 0.95
```
scarlehoff, line 139:
```suggestion
        fraction of replicas used in the bootstrap resampling, by default 0.95
```

validphys2/src/validphys/pseudodata.py
```diff
@@ -296,3 +296,8 @@ def recreate_pdf_pseudodata(_recreate_pdf_pseudodata, pdfreplicas, pdf_tr_masks)
     :py:func:`validphys.pseudodata.recreate_fit_pseudodata`
     """
     return recreate_fit_pseudodata(_recreate_pdf_pseudodata, pdfreplicas, pdf_tr_masks)
+
+pdf_tr_masks_no_table = collect('replica_training_mask', ('pdfreplicas', 'fitenvironment'))
+def recreate_pdf_pseudodata_no_table(_recreate_pdf_pseudodata, pdfreplicas, pdf_tr_masks_no_table):
+    return recreate_fit_pseudodata(_recreate_pdf_pseudodata, pdfreplicas, pdf_tr_masks_no_table)
```
scarlehoff, line 302:
```suggestion
recreate_pdf_pseudodata_no_table = recreate_fit_pseudodata
```

why would this not be exctly the same?

scarlehoff (commented):
lgtm, I've only added corrections to typos.

As a last thing, could you also add some small description of the method in the docs? 
Maybe here at the end? https://docs.nnpdf.science/figuresofmerit/index.html#hyperoptimized-figure-of-merit

I think linking the report as well makes sense as it is very descriptive.

Thanks!

validphys2/src/validphys/pseudodata.py
RoyStegeman: For python it is, for vp I wouldn't be surprised if it's not (but didn't check)

RoyStegeman: When you asked the question you were probably thinking about the decorator? I did as well, that's why I accepted it at first, but the reason it has to be done this way is that the inputs are different (and also the input is what creates the table in one of the cases)

scarlehoff: But I am confused, the inputs look exactly the same to me?

(but yes, I thought you did it because of the decorator and that's why I thought it would be the same)

scarlehoff: ahh! Okokok. Got it now, sorry.

scarlehoff:
@RoyStegeman could you add a small docs (maybe even a simplified form of what you wrote already in the paper/your thesis) so we can merge this PR?

RoyStegeman:
Please have a look to see if this is ok for you

scarlehoff (approved):
We definitely have to redo the fitbot now

https://vp.nnpdf.science/R7mObYsZTTy8M5g1OZr-PA==/#overfitmetric0_plot_overfitting_histogram

(the fitbot fit was done either with the old nnfit replica generation or with the broken python one, so the overfit metric is not supposed to be the same, just wanted to see how different it looked)

PR #1580, opened 2022-07-13
add plotting action for lumi1d per replica

RoyStegeman:
This adds a function to plot luminosities per replica. For an example see https://vp.nnpdf.science/mzAgzDONSg-S3ra2p2ilBg==

### DIFF
--- a/validphys2/src/validphys/pdfplots.py
+++ b/validphys2/src/validphys/pdfplots.py
@@ -730,6 +730,89 @@ def plot_lumi1d_uncertainties(
     return fig
 
 
+
+@figure
+@check_pdf_normalize_to
+def plot_lumi1d_replicas(
+    pdfs,
+    pdfs_lumis,
+    lumi_channel,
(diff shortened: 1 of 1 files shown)


Zaharid:
The motivation was to try to reign in the combinatorial explosion of ways to do PDF plots (e.g, I'd say this https://github.com/NNPDF/nnpdf/blob/ce80309591377ac3fc387eef5b5200ce568911b4/validphys2/src/validphys/pdfplots.py#L355 is rather neat) and also that the layout of these plots is more complicated than it seems, so there would be plenty of duplication.

Having everything spelled out is best to understand one plot in isolation but comes with the trade off against being less maintainable.

RoyStegeman:
Yes that is indeed quite clean and easily readable/easy to understand. I guess it's possible to do something similar for lumis, but I don't really like the idea of adding a function such as `setup_lumi_ax`, that should be copied inside all the lumi plotting functions.

PR #1583, opened 2022-08-03
[WIP] Proposal for the common data reader

scarlehoff:
This is a continuation of #1548 with only the absolutely necessary files so that the code review and maintenance during development is a bit easier.


validphys2/src/validphys/commondataparser.py
```diff
@@ -1,18 +1,319 @@
...
+
+@dataclass
+class CommonMetaData:
+    setname: str
+    ndata: int
+    observable: dict
+    kinematics: ValidKinematics
+    kinematic_coverage: dict
+    data_central: ValidPath
+    data_uncertainties: typing.List[ValidPath]
```
Zaharid, line 165: I'd import things from `typing` than repeating it. Also `list[ValidPath]` works from Python 3.9.

validphys2/src/validphys/commondataparser.py
```diff
@@ -1,18 +1,319 @@
...
+
+    repetition_flag: Optional[list[str]] = None
+    normalization: Optional[dict] = None
+    shifts: Optional[dict] = None
+
+    @classmethod
+    def parser(cls, meta: dict):
+        return parse_input(meta, cls)
+
+ValidApfelComb = Parser(ApfelComb.parser)
```
Zaharid, line 79: Why do we need this? Can it not be apfelcomb?

validphys2/src/validphys/commondataparser.py
```diff
@@ -1,18 +1,319 @@
...
+        to be loaded"""
+        ret = []
+        for operand in self.FK_tables:
+            ret.append([grids_folder / f"{m}.{EXT}" for m in operand])
+        return ret
+    
+    @classmethod
+    def parser(cls, meta: dict):
+        """The yaml databases in the server use "operands" instead of "FK_tables" """
+        if "operands" in meta:
```
Zaharid, line 105: This looks a bit ugly. Can we have an object corresponding to the thing we parse without hacks?

validphys2/src/validphys/commondataparser.py
```diff
@@ -1,18 +1,319 @@
...
-from validphys.core import peek_commondata_metadata
+from validphys.utils import parse_yaml_inp
 from validphys.coredata import CommonData
 
+EXT = "pineappl.lz4"
+
+
+# Auxiliary parser for common types or sanity checks
+@Parser
+def ValidPath(path_str: str) -> Path:
```
Zaharid, line 37: If we are not checking that is "valid" we should probably call this `PathParser`.

validphys2/src/validphys/commondataparser.py
```diff
@@ -1,18 +1,319 @@
...
+    """Parse strings into paths"""
+    return Path(path_str)
+
+
+@Parser
+def ValidOperation(op_str: str) -> str:
```
Zaharid, line 43: I think we could do `ValidOPeration = Literal[*list_of_operations]`.

validphys2/src/validphys/commondataparser.py
```diff
@@ -1,18 +1,319 @@
...
+
+@dataclass
+class KinematicsMeta:
+    """Contains all metadata for the kinematics of the dataset"""
+
+    file: ValidPath
+    variables: dict
+
+    @classmethod
+    def parser(cls, meta: dict):
```
Zaharid, line 118: not sure why do we need these methods: validobj already behaves like this without them.

validphys2/src/validphys/commondataparser.py
```diff
@@ -1,18 +1,319 @@
...
+        return parse_input(meta, cls)
+
+
+@dataclass
+class ReferenceMeta:
+    """Holds literature information for the dataset"""
+
+    url: str
+    version: int = 0
+    tables: list[int] = field(default_factory=list)
```
Zaharid, line 128: Are we sure that "int" is going to cut it?  Also do we need some information on what the tables are?

validphys2/src/validphys/commondataparser.py
scarlehoff: This is due to the example yaml in this branch and the ones in the server being different. There is a mismatch there that needs to be fixed.

validphys2/src/validphys/commondataparser.py
```diff
@@ -1,18 +1,319 @@
...
+
+
+# Define parsers for the more complicated structures
+ValidTheory = Parser(TheoryMeta.parser)
+ValidReference = Parser(ReferenceMeta.parser)
+ValidKinematics = Parser(KinematicsMeta.parser)
+
+
+@Parser
+def ValidVariants(variant_dict: dict) -> dict:
```
Zaharid, line 149: This looks like `Dict[str, Variant]`

validphys2/src/validphys/commondataparser.py
scarlehoff: I think when I tried it didn't work? I'll try again and copy here the error, maybe I'm doing something wrong.

validphys2/src/validphys/commondataparser.py
```diff
@@ -1,18 +1,319 @@
...
+    iNSPIRE: Optional[ValidReference] = None
+    hepdata: Optional[ValidReference] = None
+    variants: Optional[ValidVariants] = None
+
+    _enabled_variant: list[str] = field(default=None, repr=False)
+    _default: dict = field(default=None, repr=False)
+
+    # TODO: these methods will be moved to the main CommonData class
+    # as any change of variant should trigger a reload of the commondata
+    def enable_variant(self, variant):
```
Zaharid, line 183: I think we should have a class CommonMetaData without any mutable state and Commondata(common_meeta_data, variant) also without mutable state. I don;t know if that is what the comment above is about.

validphys2/src/validphys/commondataparser.py
```diff
@@ -1,18 +1,319 @@
...
+class Variant:
+    """Defines the keys of the CommonMetaData that can be overwritten"""
+
+    data_uncertainties: list[ValidPath]
+
+
+# Define parsers for the more complicated structures
+ValidTheory = Parser(TheoryMeta.parser)
+ValidReference = Parser(ReferenceMeta.parser)
+ValidKinematics = Parser(KinematicsMeta.parser)
```
Zaharid, line 145: These should not be needed as e.g. KinematicsMeta works just fine, and recursively, as a type for validobj.

validphys2/src/validphys/commondataparser.py
```diff
@@ -1,18 +1,319 @@
...
+    def metadata(self):
+        ret = parse_yaml_inp(self.metadata_file, CommonMetaData)
+        if self.variant:
+            ret.enable_variant(self.variant)
+        return ret
+
+    @cached_property
+    def data(self):
+        """Pandas DataFrame containing the central data"""
+        datayaml = yaml.safe_load(self.data_file.read_text(encoding="utf-8"))
```
Zaharid, line 268:
```suggestion
        datayaml = yaml.safe_load(self.data_file)
```

validphys2/src/validphys/commondataparser.py
```diff
@@ -1,18 +1,319 @@
...
+        data_df.index.name = "index"
+        return data_df
+
+    @cached_property
+    def uncertainties(self):
+        """Pandas DataFrame containing all uncertainties"""
+        # TODO: uncertainties are complicated enough that they _might_ need their own class
+        all_df = []
+        for ufile in self.uncertainity_files:
+            uncyaml = yaml.safe_load(ufile.read_text())
```
Zaharid, line 281:
```suggestion
            uncyaml = yaml.safe_load(ufile)
```

validphys2/src/validphys/commondataparser.py
scarlehoff: No it is not, but it makes sense, yes.

validphys2/src/validphys/utils.py
```diff
@@ -12,14 +12,18 @@
...
-def parse_yaml_inp(inp, spec, path):
-    """Helper function to parse yaml using the `validobj` library and print
+def parse_yaml_inp(input_yaml, spec):
+    """
+    Helper function to parse yaml using the `validobj` library and print
     useful error messages in case of a parsing error.
 
     https://validobj.readthedocs.io/en/latest/examples.html#yaml-line-numbers
     """
+    input_yaml = pathlib.Path(input_yaml)
```
Zaharid, line 25:
```suggestion
```

this way it should work with string and such.

validphys2/src/validphys/utils.py
```diff
@@ -12,14 +12,18 @@
...
+    inp = yaml.round_trip_load(input_yaml.open("r", encoding="utf-8"))
```
Zaharid, line 26:
```suggestion
    inp = yaml.round_trip_load(input_yaml)
```

validphys2/src/validphys/commondataparser.py
scarlehoff: I don't think this works?
```
from pathlib import Path
yaml.safe_load(Path("run.yml"))
```
results in
```
YAMLStreamError: stream argument needs to have a read() method
```

Zaharid: Right, that is sily. This does work:

```
In [6]: y = yaml.YAML()

In [7]: y.load(Path("xx.yml"))
```

Zaharid: and also e.g.
```
In [12]: y.load("a: 1")
Out[12]: ordereddict([('a', 1)])
```

scarlehoff: I'll stay with `safe_load` since that seems to be what's used in most of vp.

Zaharid: I guess one do `open` to get a stream, which avoids going over the file. In any case I think I'd be nice if `parse_yaml_inp` worked for whatever you feed it.

validphys2/src/validphys/commondataparser.py
scarlehoff: Without the `.parser` I get the following error:

```
ValidationError: Problem processing key at line 50 in .../metadata.yaml:
Cannot process field 'theory' of value into the corresponding field of 'CommonMetaData'
Problem processing key at line 51 in .../metadata.yaml:
Cannot process field 'apfelcomb' of value into the corresponding field of 'TheoryMeta'
No match for any possible type:
Not a valid match for 'Annotated': No match for any possible type:
Not a valid match for 'list': Expecting value to be a list, not CommentedMap
Not a valid match for 'NoneType': Expecting value of type 'NoneType', not CommentedMap.
Not a valid match for 'NoneType': Expecting value of type 'NoneType', not CommentedMap.
```

so I need the explicit call to `validobj.parse_input`.

validphys2/src/validphys/commondataparser.py
scarlehoff: We might want add some checks in the future (and checking whether paths exist and so seem like a check that we might want to do).

validphys2/src/validphys/commondataparser.py
scarlehoff: I'm afraid I don't know what you mean here. How should I write this in `validobj` language?

validphys2/src/validphys/commondataparser.py
scarlehoff: Not sure. I have been given the tables as a list of `int`. If they are supposed to be something else it should be a trivial change.

enocera:
@scarlehoff You'll likely have DIS+DY after Christmas.

tgiani:
yes DY should be finalised shortly after christmas

scarlehoff:
Ok, in the meantime I’ll give a try to reading the old commondata into the new format (so that libNNPDF can be dropped)

validphys2/src/validphys/commondataparser.py
Zaharid: Literally my comment. With ‘list_of_operations’=OP or whatever.

validphys2/src/validphys/commondataparser.py
```diff
@@ -1,18 +1,310 @@
...
+
+@dataclasses.dataclass
+class Variant:
+    """Defines the keys of the CommonMetaData that can be overwritten"""
+
+    data_uncertainties: list[ValidPath]
+
+
+@Parser
+def ValidVariants(variant_dict: dict) -> Dict[str, Variant]:
```
Zaharid, line 133: This is not needed. Can be just the return type and will be processed recursively as done manually in the function.

validphys2/src/validphys/commondataparser.py
```diff
@@ -1,18 +1,310 @@
...
+    plot_x: str
+    figure_by: list[str]
+    theory: ValidTheory
+    nnpdf_metadata: dict
+    version: int
+    version_comment: str = ""
+    arXiv: Optional[ValidReference] = None
+    iNSPIRE: Optional[ValidReference] = None
+    hepdata: Optional[ValidReference] = None
+    variants: Optional[ValidVariants] = dataclasses.field(default_factory=dict)
```
Zaharid, line 160: This should not be optional if the default is not going to be None

scarlehoff: I can put the default to None of course. It is just a question of adding a `If variants is None` in a few places.

validphys2/src/validphys/commondataparser.py
scarlehoff: You mean that it will be able to understand that it is a `dict` of `Variants` by itself or the `Parser` is not needed?

Zaharid: Yes it should.

validphys2/src/validphys/commondataparser.py
Zaharid: Can as well make it a dict. But at the moment if you do set it to none it is broken in several places.

Issue #1586, opened 2022-08-19
Reuse the same grid for all fktables when using a pineappl theory.


scarlehoff:
All pineappl fktables use the same x-grid, which means there's no need to evaluate the PDF once per fktable neither during the fit nor to generate a report.

This will hopefully increase performance and decrease the memory footprint of the code.

PR #1587, opened 2022-08-19
Evaluate only unique xgrids

scarlehoff:
Deals with #1586 in a very generic way so that one is still allow to mix situations in which not all xgrids are equal but some of them are.

For now only done for the fit. Once tests have run and it works there, I'll try to add some kind of cache to the vp PDF so that we can avoid calling `lhapdf` once per dataset (which I cannot promise because it is both trivial to do and trivial to do it in a disastrous way).

### DIFF
--- a/n3fit/src/n3fit/model_gen.py
+++ b/n3fit/src/n3fit/model_gen.py
@@ -9,11 +9,13 @@
 
 
 """
+from itertools import zip_longest
 from dataclasses import dataclass
 import numpy as np
 from n3fit.msr import msr_impose
 from n3fit.layers import DIS, DY, ObsRotation, losses
 from n3fit.layers import Preprocessing, FkRotation, FlavourToEvolution
+from n3fit.layers.observable import is_unique
 
 from n3fit.backends import MetaModel, Input
 from n3fit.backends import operations as op
@@ -23,7 +25,7 @@ from n3fit.backends import base_layer_selector, regularizer_selector
 
 @dataclass
 class ObservableWrapper:
-    """Wrapper to generate the observable layer once the PDF model is prepared
+    """Wraps many observables into an experimental layer once the PDF model is prepared
     It can take normal datasets or Lagrange-multiplier-like datasets
     (such as positivity or integrability)
     """
@@ -59,8 +61,10 @@ class ObservableWrapper:
         return loss
 
     def _generate_experimental_layer(self, pdf):
-        """Generates the experimental layer from the PDF"""
-        # First split the layer into the different datasets (if needed!)
+        """Generate the experimental layer by feeding to each observable its PDF.
+        In the most general case, each observable might need a PDF evaluated on a different xgrid,
+        the input PDF is evaluated in all points that the experiment needs and needs to be split
+        """
         if len(self.dataset_xsizes) > 1:
             splitting_layer = op.as_layer(
                 op.split,
@@ -68,12 +72,12 @@ class ObservableWrapper:
                 op_kwargs={"axis": 1},
                 name=f"{self.name}_split",
             )
-            split_pdf = splitting_layer(pdf)
+            sp_pdf = splitting_layer(pdf)
+            output_layers = [obs(p) for obs, p in zip(self.observables, sp_pdf)]
         else:
-            split_pdf = [pdf]
-        # Every obs gets its share of the split
-        output_layers = [obs(p_pdf) for p_pdf, obs in zip(split_pdf, self.observables)]
-        # Concatenate all datasets (so that experiments are one single entity)
+            output_layers = [obs(pdf) for obs in self.observables]
+
+        # Finally concatenate all observables (so that experiments are one single entitiy)
         ret = op.concatenate(output_layers, axis=2)
         if self.rotation is not None:
             ret = self.rotation(ret)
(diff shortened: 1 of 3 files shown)


scarlehoff:
Actually, loading 100 times (order of number of datasets) the same 100 points xgrid/qgrid from LHAPDF doesn't seem to justify the added complication of caching the result, so probably it doesn't make sense to do it.
```
1.45 s ± 9.35 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
I might play with it at some point in case it does improve things in real use cases.

scarlehoff:
Positivity seems ok (which is the only place where this would have an effect in the standard fit) https://vp.nnpdf.science/tmsUkZzYTfugL2uy5OoNfA==

n3fit/src/n3fit/model_gen.py
```diff
@@ -9,11 +9,13 @@
 
 
 """
+from itertools import zip_longest
```
Zaharid, line 12: Seems unused

scarlehoff, line 12:
```suggestion
```

n3fit/src/n3fit/model_trainer.py
```diff
@@ -327,15 +326,39 @@ def _model_generation(self, pdf_models, partition, partition_idx):
...
+        # Clean the input list
+        inputs_hash = []
+        inputs_unique = []
+        inputs_idx = []
+        for input_grid in self.input_list:
+            ghash = hash(tuple(input_grid.flatten()))
+            if ghash not in inputs_hash:
+                inputs_hash.append(ghash)
+                inputs_unique.append(input_grid)
+            inputs_idx.append(inputs_hash.index(ghash))
```
Zaharid, line 350: I am not feeling brave enough to fully understand this logic right now, but seems to me this approach is flacky and complicated. First of all hashes are not necessarily unique, but also computing it seems unneeded given that is a big expense to do only once and the right interface is already provided by dicts and sets.  Putting something like `tuple(input_grid.flatten())` in a dict would be correct, faster (dicts can avoid computing hashes for the same pointer) and avoid the O(N) list lookup.

scarlehoff: I'm not entirely sure what you mean by the right interface. A set doesn't really work here (I need the order both outside and inside the arrays!) and a dictionary seems to me more complicated than plain lists.

I guess I can do this and remove the hash if you like it better.
```python
inputs_unique = []
inputs_unique_tuple = []
inputs_idx = []
for input_grid in self.input_list:
    tarr = tuple(input_grid.flatten())
    if tarr not in inputs_unique_tuple:
        inputs_unique.append(input_grid)
        inputs_unique_tuple.append(tarr)
    inputs_idx.append(inputs_unique_tuple.index(tarr))
```

Also, an O(N) list lookup is very much not a problem here. If it were this would not be needed in the first place. 

Actually I'm wondering now whether I shouldn't, instead of doing the `in` compare the arrays using `allclose` with some threshold.  I would not want to compute the grid twice just because in one you have `0.0000133333` and in the other `0.0000133334`...

scarlehoff, line 350:
```suggestion
        inputs_unique = []
        inputs_idx = []
        for igrid in self.input_list:
            for idx, arr in enumerate(inputs_unique):
                if np.array_equal(igrid, arr):
                    inputs_idx.append(idx)
                    break
            else:
                inputs_idx.append(len(inputs_unique))
                inputs_unique.append(igrid)
```
This should work. And allows for some fine-tunning on what the equality should be, which I like. 

Although it becomes even more complicated.

Zaharid: The two straight forward way would be to put the frozen versions of the inputs in a set/dictionary (which you were somehow doing minus the dictionary) or to compare manually the items. I guess your approach below would be sensible if you do want the allclose checks. 

I also wonder if it would make sense to "name" the grids somewhere so there is a name associated with a grid and we match by name (as in `pinneaple.default_fkgrid` or somesuch). This might help the toolchain enforce that the grids are actually the same, and require a mindful decision to do anything else. As a counterpoint, negative experience with reportengine suggests that just hashing the values would have been better than the complicated namings there (though I don't expect this to face the same problems).

RoyStegeman: I think this is pretty clear, and the preferable way if you want to use allclose. 

Hashes would also be fine with me (if you don't care about allclose), of course they're technically not unique as Zahari pionts out, but in practice we'd have to work pretty hard to find two objects returning the same hash. 

 I don't really like the idea of checking for default names inside n3fit. That's pineappl business.

scarlehoff: Sadly forcing all of them to be the same might not be the best because at the very least there is integrability which is a very special grid so we would have to have at least one exception.

Rethinking about it, the `allclose` is the best thing because as @RoyStegeman `n3fit` should just take whatever information receives (from pineappl in this case) and if the grids are close they produce the same pdf otherwise they don't.

Zaharid: There is no reason why you would use hashes but not dicts or sets which use hashes and additionally an equality check (quick identity check in the happy case). I think it is best to use none of that.

RoyStegeman (approved):

PR #1588, opened 2022-08-22
Fixing the grouping and the thcovmat flag

andreab1997:
The grouping used in the thcovmat fit is broken because `groups_dataset_inputs_loaded_cd_with_cuts` does not use `group_dataset_inputs_by_fitting_group` and therefore is not grouped as the thcovmat (even if the groupings are similar so sometimes it happens to be correct by chance). 

Moreover, the thcovmat group is used also when both `use_thcovmat_in_sampling` and `use_thcovmat_in_fitting` are false.

### DIFF
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -1584,11 +1584,12 @@ class CoreConfig(configparser.Config):
         # somebody will want to add to this at some point e.g for th. uncertainties
         allowed = {
             "standard_report": "experiment",
+            "thcovmat_fit": "ALL"
         }
         return allowed[spec]
 
     def produce_processed_data_grouping(
-        self, data_grouping=None, data_grouping_recorded_spec_=None
+        self, use_thcovmat_in_fitting, use_thcovmat_in_sampling, data_grouping=None, data_grouping_recorded_spec_=None
     ):
         """Process the data_grouping key from the runcard, or lockfile. If
         `data_grouping_recorded_spec_` is present then its value is taken, and
@@ -1602,6 +1603,8 @@ class CoreConfig(configparser.Config):
         if data_grouping is None:
             # fallback to old default behaviour, but still record to lockfile
             data_grouping = self.parse_data_grouping("standard_report")
+            if use_thcovmat_in_fitting or use_thcovmat_in_sampling:
+                data_grouping = self.parse_data_grouping("thcovmat_fit")
         if data_grouping_recorded_spec_ is not None:
             return data_grouping_recorded_spec_[data_grouping]
         return self.load_default_data_grouping(data_grouping)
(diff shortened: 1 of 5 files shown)


andreab1997:
@scarlehoff as you can see I have only changed `group_dataset_inputs_by_metadata` for `group_dataset_inputs_by_fitting_group` in the collect of `groups_dataset_inputs_loaded_cd_with_cuts`. This was enough to solve our problem. However there a lot of other places in which `group_dataset_inputs_by_metadata` is used. I wonder if would be better to change them all. What do you think?

scarlehoff:
> I wonder if would be better to change them all.

We only want to order by fitting group when we are fitting. I'm not sure this is a good solution actually, I think the "metadata group" should be equal to the "fitting group" during a fit. Is there a reason why they should be different?

andreab1997:
> > I wonder if would be better to change them all.
> 
> We only want to order by fitting group when we are fitting. I'm not sure this is a good solution actually, I think the "metadata group" should be equal to the "fitting group" during a fit. Is there a reason why they should be different?

The point is that by default `metadata group` groups by experiment. So if we use `fitting_group` for the covmat and `metadata_group` for the groups_dataset_inputs_loaded_cd_with_cuts, we will have in general different ordering

scarlehoff:
> The point is that by default metadata group groups by experiment.

Then I think the right thing is for metadata group to default (in the case of a fit) to the fitting group.

andreab1997:
> > The point is that by default metadata group groups by experiment.
> 
> Then I think the right thing is for metadata group to default (in the case of a fit) to the fitting group.

https://github.com/NNPDF/nnpdf/blob/05602d6fa57477689cc96b4cc58274e0a552e082/validphys2/src/validphys/config.py#L1581
This is the point in which the default data_grouping is stated. Do we want to change this?

scarlehoff:
Yes. Indeed, I can even see that @wilsonmr left a comment many moons ago predicting this day would come :P

Zaharid:
FWIW while the while "data keyword" grouping system was great in both allowing complex new uses of the code and maintaining compatibility, I think the resulting trade off in complexity is not favorable to us anymore, in particular since we don't have to support an `experiments` declaration anymore. Simplifying that would be very useful work attacking the base of this problem.

andreab1997:
> Yes. Indeed, I can even see that @wilsonmr left a comment many moons ago predicting this day would come :P

Do you like the new solution more?

validphys2/src/validphys/config.py
```diff
@@ -1602,6 +1603,8 @@ def produce_processed_data_grouping(
         if data_grouping is None:
             # fallback to old default behaviour, but still record to lockfile
             data_grouping = self.parse_data_grouping("standard_report")
+            if use_thcovmat_in_fitting or use_thcovmat_in_sampling:
+                data_grouping = self.parse_data_grouping("thcovmat_fit")
```
scarlehoff, line 1607: Add this new behavior to the docstring.

And, since you removed `by fitting_group` add a comment somewhere to say that the default fitting group is equal to the the `standard_report` (because a fit is a standard report after all, but better to be explicit since it is not very intuitive).

scarlehoff (approved):

PR #1589, opened 2022-08-27
Add NLO and NNLO sv theories in db

andreab1997:
I am adding the new scale variations theories (so 401-418, 420-428, 430-438) to the `theory.db`.


scarlehoff:
The 410 to 418 have scheme C in the comment but FONLL-B in the FNS field.
420 to 428 instead don't have scheme C in the comment (while 430-438 do) but FONLL-C in the field.

Is that correct, or is that mixed up?

andreab1997:
> The 410 to 418 have scheme C in the comment but FONLL-B in the FNS field. 420 to 428 instead don't have scheme C in the comment (while 430-438 do) but FONLL-C in the field.
> 
> Is that correct, or is that mixed up?

No, it is not correct, thanks to have spotted. Something went wrong apparently...

andreab1997:
> The 410 to 418 have scheme C in the comment but FONLL-B in the FNS field. 420 to 428 instead don't have scheme C in the comment (while 430-438 do) but FONLL-C in the field.
> 
> Is that correct, or is that mixed up?

I mean, FONLL "versions" are correct. The problem is just that from 420 to 428 don't have scheme C in the comment.

scarlehoff:
Have a look at all them in case the other are also wrong though (since I believe you are using a script to modify the theory I guess it is easier to rerun the full script)

andreab1997:
> Have a look at all them in case the other are also wrong though (since I believe you are using a script to modify the theory I guess it is easier to rerun the full script)

Yes, the problem was just that for scheme B sv theories the comment did not include "scheme B" instead it did in the scheme C case.

scarlehoff (approved):

scarlehoff:
feel free to merge as soon as the tests are done.

Issue #1582, opened 2022-07-22
Warn the user when loading a positivity fk with the dataset API


scarlehoff:
Edit: changed the title to reflect what the actual bug is (tl;dr, user error)

------
The positivity grids contain 20 points but instead the positivity predictions contain only 12 after the `internal` cuts happen. After the PC we believe this is a bug (or rather an oversight). It should be checked (whether it is a bug or it is intended) and then it should be fixed.

```python
from validphys.api import API
from validphys.convolution import central_predictions
pdf = API.pdf(pdf="NNPDF40_nnlo_as_01180")
dat = API.dataset(dataset_input={"dataset": "POSF2U"}, theoryid=200, use_cuts="internal")
print(f"Positivity predictions {len(central_predictions(dat, pdf)}")
```

Zaharid:
Note we don't use API.dataset to load positivity observables, but rather posdataset

```
In [13]: from validphys.api import API
    ...: from validphys.convolution import central_predictions
    ...: pdf = API.pdf(pdf="NNPDF40_nnlo_as_01180")
    ...: dat = API.posdataset(posdataset={"dataset": "POSF2U", "maxlambda":1}, theoryid=200, use_cuts="internal")
    ...: print(f"Positivity predictions {len(central_predictions(dat, pdf))}")
Positivity predictions 20
```

PR #1591, opened 2022-08-31
Raise an exception for wrongly loaded positivity or integrability datasets

scarlehoff:
Fixes user errors like #1582

### DIFF
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -410,6 +410,12 @@ class CoreConfig(configparser.Config):
             name = dataset["dataset"]
             if not isinstance(name, str):
                 raise ConfigError(f"'dataset' must be a string, not {type(name)}")
+            # Check whether this is an integrability or positivity dataset (in the only way we know?)
+            if name.startswith(("INTEG", "POS")):
+                if name.startswith("INTEG"):
+                    raise ConfigError("Please, use `integdataset` for integrability")
+                if name.startswith("POS"):
+                    raise ConfigError("Please, use `posdataset` for positivity")
         except KeyError:
             raise ConfigError(
                 "'dataset' must be a mapping with " "'dataset' and 'sysnum'"


Zaharid:
I am thinking if this is limiting some interesting cases but can't think of any so this is probably fine for now.

I think it is good to think how we'd like these checks to look in the new commondata format, and also rethink how we'd like the special cases for cuts to work.

I think they would be based on capabilities rather than on e.g. naming conventions. For example it should be fine to use a "full" dataset as a positivity constraint (e.g. because we'd like to exclude it from the fit but still make sure the predictions are positive). I think that should work without e.g. having to rename the set, which we know is a pain. But of course if we try to compute the chi2 of a positivity only commondata that should fail saying that we don't have experimental data.

cc @enocera

Zaharid (approved):

scarlehoff:
> For example it should be fine to use a "full" dataset as a positivity constraint (e.g. because we'd like to exclude it from the fit but still make sure the predictions are positive)

This would be actually useful, yes. I guess right now the hack exists by copying the `fktable` with a POS in front. But it would be nice to be able to do it hackless.

felixhekhorn:
> This would be actually useful, yes. I guess right now the hack exists by copying the `fktable` with a POS in front. But it would be nice to be able to do it hackless.

just to say, that for pinefarm there is indeed no difference between a "normal" dataset and a "positivity" dataset

the current situation is:
- PDF positivity datasets have their own [generator](https://github.com/NNPDF/runcards/blob/master/pinefarm/external/positivity.py)
- DIS positivity datasets are ordinary yadism runcards with `NCPositivityCharge` in the obs card; actually it is a bit more complicated since I need to [detect this](https://github.com/NNPDF/runcards/blob/04572c4124fb1a379ef77805749d726f0cd6a3e9/pinefarm/external/yad.py#L39) and eventually [shut down TMC](https://github.com/NNPDF/runcards/pull/143#issuecomment-1218281454)
- DY positivity datasets are similar to DIS, in that they have a special configuration and act slightly different in the generation

however, the outcome is in any case just an ordinary pineappl grid, as for any dataset

scarlehoff:
No, the grids are also equal from the vp points of view. But we want to know they are different because they are treated different (cuts are not applied, lagrange multipliers, no data associated).

Sadly at some point whether a dataset was positivity (or integrability) was defined by its name (INTEG or POS) and treated differently. Going forward we can burn a tag into the fktable metadata (would be trivial to add to pinefarm and might be relevant information anyway)

felixhekhorn:
But wasn't @Zaharid thinking about using them in a flexible way? i.e. the fk table can be used both as "normal" and positivity table? (@Zaharid correct me, if I misunderstood) and so the information should not be burnt in, because it is dynamic?

scarlehoff:
The information that it is a positivity fktable does not prevent you from using it in a flexible way. It just means that as soon as you try to do something that is not allowed you can catch it without having to resort to it being called "POS".

PR #1592, opened 2022-09-05
update code citation with EPJC info

RoyStegeman:
This is just taken from inspire.


Zaharid (approved):

goord:
Hi just to let you know, there is also an 'official' code citation file format, you could include a [citation.cff](https://citation-file-format.github.io) as well.

Issue #1220, opened 2021-05-13
Error building the docs


Zaharid:
Looking at the doc [failure](https://travis-ci.com/github/NNPDF/nnpdf/jobs/504946613) on master, I think we are being affected by:

https://github.com/readthedocs/recommonmark/issues/220

Maybe we can pin docutils as suggested in there?

PR #1596, opened 2022-09-08
Change recommonmark for myst

scarlehoff:
While it might not be the problem of #1537, this is something that we should've done some time ago given https://github.com/NNPDF/nnpdf/issues/1220#issuecomment-947732053

I think that the documentation produced by this branch is equal to the documentation in master (actually, a bit better because I've fixed a few broken links I've found) but I've only checked by eye.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -47,7 +47,7 @@ requirements:
         - prompt_toolkit
         - validobj
         - sphinx >=4.0.2 # documentation. Needs pinning becasue https://github.com/sphinx-doc/sphinx/issues/9216
-        - recommonmark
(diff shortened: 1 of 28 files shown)


Zaharid:
Since we are touching the md files anyway, how bad would it be to convert them to rst as per the original comment? Supporting the two things was quickly disliked by everyone, hence the note here

https://docs.nnpdf.science/contributing/sphinx-documentation.html?highlight=rst

scarlehoff:
> how bad would it be to convert them to rst as per the original comment?

If `recommonmark` is the cause of #1537, very bad (because it would take time).

scarlehoff:
Closing in favour of #1597

PR #1606, opened 2022-09-29
Kinetic energy of the pdf

scarlehoff:
This PR adds the kinetic energy of the PDF as two new actions:

```
plot_pdfs_kinetic_energy
plot_pdfreplicas_kinetic_energy
```

Example of a resulting report: https://vp.nnpdf.science/IixzVlnyQX2wi9ACRPeniQ==/

I'm using #1605 to get the derivative and then these create the kinetic energy PDF.

To get the correct label I've added a `process_label` method to the `XPlottingGrid` which the PDFPlotter will call before generating the `ylabel`. This is quite generic so it might be helpful for future things.

The addition of `KineticXPlottingGrid` just to get a label might be too much though but I am not sure how to do it more cleanly.

### DIFF
--- a/validphys2/src/validphys/pdfgrids.py
+++ b/validphys2/src/validphys/pdfgrids.py
@@ -48,13 +48,14 @@ class XPlottingGrid:
     The `grid_values` attribute corresponds to a `Stats` instance
     in order to compute statistical estimators in a sensible manner.
     """
+
     Q: float
     basis: (str, Basis)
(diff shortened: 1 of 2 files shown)


scarlehoff:
I think this one and its parent could potentially be merged.

For #1607 we need something more complicated but I guess it will need to be done if we are actually going to publish papers with its content?

RoyStegeman (approved):

PR #1607, opened 2022-09-29
Mix bands and replicas

scarlehoff:
Uses #1605 and #1606

Since the hopscotch replicas are single PDFs but we want to compare them to the "true" NNPDF we need to somehow be able to see in a single plot the hopscotch PDFs and the NNPDF bands. This is what I've come up with:

You might notice there is no label for the replicas, cannot see how to make them cross-talk (the bands and the replicas) but I don't think we will want to have this in the core code anyway? Or do we?

I've also removed the "central replica" since it makes no sense for the hopscotchy ones https://vp.nnpdf.science/avVrWwjGTsOtYFh1n5j66A==/

### DIFF
--- a/validphys2/src/validphys/checks.py
+++ b/validphys2/src/validphys/checks.py
@@ -231,43 +231,43 @@ def check_pdfs_noband(pdfs, pdfs_noband):
 
 
 @make_argcheck
-def check_pdfs_plot_replicas(pdfs, pdfs_plot_replicas):
-    """Same as check_pdfs_noband, but for the pdfs_plot_replicas key.
-    Allows pdfs_plot_replicas to be specified as a list of PDF IDs or a list of
+def check_mixband_as_replicas(pdfs, mixband_as_replicas):
+    """Same as check_pdfs_noband, but for the mixband_as_replicas key.
+    Allows mixband_as_replicas to be specified as a list of PDF IDs or a list of
     PDF indexes (starting from one)."""
 
-    msg = ("pdfs_plot_replicas should be a list of PDF IDs (strings) or a list of "
+    msg = ("mixband_as_replicas should be a list of PDF IDs (strings) or a list of "
            "PDF indexes (integers, starting from one)")
-    msg_range = ("At least one of the choices in pdfs_plot_replicas indexes is out of range. "
+    msg_range = ("At least one of the choices in mixband_as_replicas indexes is out of range. "
                  "Note that pdf_noband indexing starts at 1, not 0.")
 
-    if pdfs_plot_replicas is None:
-        return {'pdfs_plot_replicas': []}
+    if mixband_as_replicas is None:
+        return {'mixband_as_replicas': []}
 
     names = [pdf.name for pdf in pdfs]
     # A list to which PDF IDs can be added, when the PDF is specified by either
     # its PDF ID (i.e. a string) or an index (i.e. an int)
-    pdfs_plot_replicas_combined = []
+    mixband_as_replicas_combined = []
 
-    for pdf_noband in pdfs_plot_replicas:
+    for pdf_noband in mixband_as_replicas:
         if isinstance(pdf_noband, int):
             if not pdf_noband <= len(names) or pdf_noband < 0:
                 raise CheckError(msg_range)
             # Convert PDF index to list index (i.e. starting from zero)
             pdf_noband -= 1
-            pdfs_plot_replicas_combined.append(pdfs[pdf_noband])
+            mixband_as_replicas_combined.append(pdfs[pdf_noband])
 
         elif isinstance(pdf_noband, str):
             try:
                 pdf_index = names.index(pdf_noband)
-                pdfs_plot_replicas_combined.append(pdfs[pdf_index])
+                mixband_as_replicas_combined.append(pdfs[pdf_index])
             except ValueError:
                 raise CheckError(msg, pdf_noband, alternatives=names)
 
         else:
             raise CheckError(msg)
 
-    return {'pdfs_plot_replicas': pdfs_plot_replicas_combined}
+    return {'mixband_as_replicas': mixband_as_replicas_combined}
 
 def _check_list_different(l, name):
     strs = [str(item) for item in l]
--- a/validphys2/src/validphys/checks.py
+++ b/validphys2/src/validphys/checks.py
@@ -238,11 +238,11 @@ def check_pdfs_plot_replicas(pdfs, pdfs_plot_replicas):
 
     msg = ("pdfs_plot_replicas should be a list of PDF IDs (strings) or a list of "
            "PDF indexes (integers, starting from one)")
-    msg_range = ("At least one of your pdf_noband indexes is out of range. "
+    msg_range = ("At least one of the choices in pdfs_plot_replicas indexes is out of range. "
                  "Note that pdf_noband indexing starts at 1, not 0.")
 
     if pdfs_plot_replicas is None:
-        return
+        return {'pdfs_plot_replicas': []}
 
     names = [pdf.name for pdf in pdfs]
     # A list to which PDF IDs can be added, when the PDF is specified by either
--- a/validphys2/src/validphys/checks.py
+++ b/validphys2/src/validphys/checks.py
@@ -230,6 +230,45 @@ def check_pdfs_noband(pdfs, pdfs_noband):
     return {'pdfs_noband': pdfs_noband_combined}
 
 
+@make_argcheck
+def check_pdfs_plot_replicas(pdfs, pdfs_plot_replicas):
+    """Same as check_pdfs_noband, but for the pdfs_plot_replicas key.
+    Allows pdfs_plot_replicas to be specified as a list of PDF IDs or a list of
+    PDF indexes (starting from one)."""
+
+    msg = ("pdfs_plot_replicas should be a list of PDF IDs (strings) or a list of "
+           "PDF indexes (integers, starting from one)")
+    msg_range = ("At least one of your pdf_noband indexes is out of range. "
+                 "Note that pdf_noband indexing starts at 1, not 0.")
+
+    if pdfs_plot_replicas is None:
+        return
+
+    names = [pdf.name for pdf in pdfs]
+    # A list to which PDF IDs can be added, when the PDF is specified by either
+    # its PDF ID (i.e. a string) or an index (i.e. an int)
+    pdfs_plot_replicas_combined = []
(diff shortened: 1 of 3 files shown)


validphys2/src/validphys/pdfplots.py
```diff
@@ -1044,3 +1044,15 @@ def plot_lumi2d_uncertainty(pdf, lumi_channel, lumigrid2d, sqrts:numbers.Real):
...
     return fig
+
+
+class MixMatchPDFPlotter(BandPDFPlotter, ReplicaPDFPlotter):
+    """Special wrapper class to plot, in the same figure, PDF bands and PDF replicas
+    depending on the type of PDF.
+    Practical use: plot together the PDF central values with the NNPDF bands
+    """
+    def draw(self, pdf, grid, flstate):
+        if pdf.label.startswith("HOP"):
```
RoyStegeman, line 1055:
```suggestion
    def draw(self, pdf, grid, flstate):
        # small values corresponding to charm result in a +1 placed on top of
        # the y-axis if axes.formatter.useoffset is not False
        plt.rcParams['axes.formatter.useoffset'] = False 
        if pdf.label.startswith("HOP"):
```

RoyStegeman (changes requested):
On Friday people commented on the +1 appearing in the charm plot

validphys2/src/validphys/pdfplots.py
Zaharid: Should probably make that a default.

RoyStegeman: You mean in the .mpl files? Maybe a good idea, but I'm not sure this PR will ever be merged so maybe that should be done separately.

Zaharid: Yes indeed.

RoyStegeman:
So we want to add a boolean flag indicating whether or not to plot a PDF as a band. i.e. so we can do something like this:
```
pdfs:
  - {id: "NNPDF40_nnlo_as_01180_1000", label: "4.0 NNLO (1000 replicas)", plot_bands: False}
```

RoyStegeman:
Right, also fine for me

RoyStegeman:
We should also think of what to do about replica0

RoyStegeman:
Please have a look if this is acceptable to you.

validphys2/src/validphys/pdfplots.py
```diff
@@ -53,8 +53,9 @@ def __init__(self, pdfs, xplotting_grids, xscale, normalize_to, ymin, ymax):
         self.ymax = ymax
 
     def setup_flavour(self, flstate):
-        pass
-
+        flstate.handles=[]
+        flstate.labels=[]
+        flstate.hatchit=plotutils.hatch_iter()
```
Zaharid, line 58: Why are we moving these things, which were specific to the bands pdfs here?

validphys2/src/validphys/pdfplots.py
```diff
@@ -188,19 +189,29 @@ def _warn_any_pdf_not_montecarlo(pdfs):
 
 class ReplicaPDFPlotter(PDFPlotter):
     def draw(self, pdf, grid, flstate):
+        ax = flstate.ax
```
Zaharid, line 192: This line is duplicated below.

Zaharid:
I wonder if we should expand plot_pdfs do the same thing as plot_pdfs_mixed and remove the later?

validphys2/src/validphys/pdfplots.py
RoyStegeman: Because `MixMatchPDFPlotter` inherits from both  `ReplicaPDFPlotter` and `BandPDFPlotter`, so the way  `ReplicaPDFPlotter` deals with labels and handles has also been changed to be the same.

RoyStegeman:
> I wonder if we should expand plot_pdfs do the same thing as plot_pdfs_mixed and remove the later?

It does the same if you don't specify `pdfs_plot_replicas`. I was thinking about this as well, though since we would still have the `plot_pdfreplicas`  (which we can't get rid off since the `mixed_` variant doesn't print replica0) I don't mind also keeping `plot_pdfs` for symmetry.

validphys2/src/validphys/pdfplots.py
RoyStegeman: But this still uses the flstate handles and labels from `BandPDFPlotter`. The issue here is that `ReplicaPDFPlotter` has different instances for the handles and labels.

Zaharid: Would most things work by inheriting from BandPDFPlotter first?

RoyStegeman: Sure, all BandPDFPlotter and ReplicaPDFPlotter do is add a function on PDFPlotter to draw the figures and add the labels, where for ReplicaPDFPlotter the default matplotlib legend is used. 

What I did so far is move the label generation from BandPDFPlotter to PDFPlotter and add the labels to flstate also for ReplicaPDFPlotter. Instead of moving the label generation to PDFPlotter it is indeed also possible to make ReplicaPDFPlotter inherit from BandPDFPlotter but in my mind that organisation makes less sense.

RoyStegeman: @Zaharid what would be your preferred solution? Feel free to change this PR yourself if that's easier than explaining.

validphys2/src/validphys/checks.py
```diff
@@ -230,6 +230,45 @@ def check_pdfs_noband(pdfs, pdfs_noband):
...
+    Allows pdfs_plot_replicas to be specified as a list of PDF IDs or a list of
+    PDF indexes (starting from one)."""
+
+    msg = ("pdfs_plot_replicas should be a list of PDF IDs (strings) or a list of "
+           "PDF indexes (integers, starting from one)")
+    msg_range = ("At least one of your pdf_noband indexes is out of range. "
+                 "Note that pdf_noband indexing starts at 1, not 0.")
+
+    if pdfs_plot_replicas is None:
+        return
```
scarlehoff, line 245:
```suggestion
        return {'pdfs_plot_replicas': []}
```

(this is in combination with the other comment below, you choose what to do)

validphys2/src/validphys/pdfplots.py
```diff
@@ -188,19 +189,28 @@ def _warn_any_pdf_not_montecarlo(pdfs):
...
-        stats = flavour_grid.grid_values
+        stats = grid.select_flavour(flstate.flindex).grid_values
         gv = stats.data
         ax.plot(grid.xgrid, gv.T, alpha=0.2, linewidth=0.5,
                 color=color, zorder=1)
-        ax.plot(grid.xgrid, stats.central_value(), color=color,
-                linewidth=2,
-                label=pdf.label)
+        cv_line,  = ax.plot(grid.xgrid[0:1], stats.central_value()[0:1], color=color,
+                linewidth=2)
```
scarlehoff, line 202: Is this from `black`? It doesn't look good.

scarlehoff (commented):
To me it looks ok (just realized I cannot approve the PR because I am technically the author). I think for the scope of what we want this PR this is good. The changes it introduces are minimal and the new features are contained in their own functions/checks. 

Could you check that doing a report with replicas doesn't change anything? (in principle it shouldn't but having an explicit example linked in the PR could be useful for future reference).

validphys2/src/validphys/pdfplots.py
RoyStegeman: No I haven't run black (only removed trailing whitespaces). Is it better now?

RoyStegeman:
@Zaharid do you have any idea when you'll look a this?

scarlehoff:
Should we merge this? i.e, is this "merge-ready" from your point of view @RoyStegeman ?

validphys2/src/validphys/pdfplots.py
```diff
@@ -188,19 +189,28 @@ def _warn_any_pdf_not_montecarlo(pdfs):
...
         ax.plot(grid.xgrid, gv.T, alpha=0.2, linewidth=0.5,
                 color=color, zorder=1)
-        ax.plot(grid.xgrid, stats.central_value(), color=color,
-                linewidth=2,
-                label=pdf.label)
+        cv_line = ax.plot(grid.xgrid[0:1], stats.central_value()[0:1], 
+                color=color, linewidth=2)
+        handle = cv_line[0]
+        labels.append(pdf.label)
+        handles.append(handle)
```
Zaharid, line 205: @scarlehoff @RoyStegeman This PR is adding a plot that is used in very limited circumstances, but that makes implementing new subclasses measurably more difficult as demostrated in this bit. I am not too happy with that as per the existing discussion. I see no pressure to merge it.

scarlehoff: > that makes implementing new subclasses measurably more difficult as demostrated in this bit

How does this bit demonstrate that?

scarlehoff: (also, the very limited circumstances are already in the arxiv, which is more than one can say of many of the other plots that we have)

scarlehoff: The only other option that I see is to copy the whole of the `draw` method of `ReplicaPDFPlotter`  into `MixMatchPDFPlotter` instead of reusing `ReplicaPDFPlotter`.

RoyStegeman:
Ok I think now it should be ok for @Zaharid ?

Zaharid (changes requested):
Looks good indeed. Everything works and the error messages are of the expected high quality

```
 $ validphys mixbandreplicas.yaml                             (nnpdfdev) 
[WARNING]: Output folder exists: /home/zah/nngit/nnpdf/validphys2/examples/output Overwriting contents
[ERROR]: Cannot process a resource:
Could not process the resource 'plot_pdfs_mixed', required by:
 - template_text
 - report
pdfs_plot_replicas should be a list of PDF IDs (strings) or a list of PDF indexes (integers, starting from one)
```

However I think we need to change the name of the parameter. Calling it `pdfs_plot_replicas` when we have `plot_pdfreplicas`  to plot replicas is just too confusing.

I am happy to change the name myself if someone suggests a good one.

RoyStegeman:
Good point. How about something along the lines of `mixmatch_as_replicas`?

RoyStegeman:
Feel free to update it again if you have a better idea @Zaharid.

Zaharid (commented):
I don't love that name either but can't really think of much.

RoyStegeman (approved):

PR #1610, opened 2022-10-05
[New commondata format] DY

tgiani:
Implemnetation of DYE605 in the new data format


Radonirinaunimi:
@tgiani Have you by chance used this [script](https://github.com/NNPDF/nnpdf/blob/eea53bc9b191d20a48bac3d5af969e8ac4a4aae9/validphys2/src/validphys/scripts/get_hepdata.py) to download the tables from the `metadata`? And whether it works fine?

tgiani:
also, I still need to compare this data with those in the old format, but in this specific case it looks to me that the tables in the old  rawdata folder are the same as those in hepdata

tgiani:
> @tgiani Have you by chance used this [script](https://github.com/NNPDF/nnpdf/blob/eea53bc9b191d20a48bac3d5af969e8ac4a4aae9/validphys2/src/validphys/scripts/get_hepdata.py) to download the tables from the `metadata`? And whether it works fine?

ah interesting. No I have not, I have written the `metadata` and downloaded the tables by hand. Thank you I will give it a try

enocera:
> @tgiani Have you by chance used this [script](https://github.com/NNPDF/nnpdf/blob/eea53bc9b191d20a48bac3d5af969e8ac4a4aae9/validphys2/src/validphys/scripts/get_hepdata.py) to download the tables from the `metadata`? And whether it works fine?

Not yet, because I told @tgiani to first look at how messy the rawdata are w.r.t. the hepdata and the paper. But thanks @Radonirinaunimi for reminding us that we should use the function that you've implemented.

enocera:
> also, I still need to compare this data with those in the old format, but in this specific case it looks to me that the tables in the old rawdata folder are the same as those in hepdata

So much the better. I bet that for DYE866 you will notice differences.

tgiani:
> > also, I still need to compare this data with those in the old format, but in this specific case it looks to me that the tables in the old rawdata folder are the same as those in hepdata
> 
> So much the better. I bet that for DYE866 you will notice differences.

looking forward

Radonirinaunimi:
> > also, I still need to compare this data with those in the old format, but in this specific case it looks to me that the tables in the old rawdata folder are the same as those in hepdata
> 
> So much the better. I bet that for DYE866 you will notice differences.

Perfect! Btw, if there are more datasets to implement/check or anything in that regard, just let me know.

tgiani:
@enocera this should be ready for review. In this case the old rawdata file look to be the same as those from Hepdata

tgiani:
the only difference is that in the old filter for some reason there is an additional systematic of 10%, but from the Hepdata results this should be 5% instead. Also, in the old filter it is written then it should be taken as uncorrelated, so I did the same in the new filter. However I m not sure why this should be the case. The only information about this systematic I could find in the hepdata tables is its label, which is ```Point-to-point systematic error```

scarlehoff:
Moved to PR #1679

PR #1611, opened 2022-10-06
Fix overfitmetric in comparefits

scarlehoff:
I think due to this collect here
https://github.com/NNPDF/nnpdf/blob/ec6cbaf972240de035b60e5321a4c406e36f78da/validphys2/src/validphys/overfit_metric.py#L23

The overfit metric in the comparefits report gets all predictions from both pdf, i..e, in this function
 https://github.com/NNPDF/nnpdf/blob/ec6cbaf972240de035b60e5321a4c406e36f78da/validphys2/src/validphys/overfit_metric.py#L51

preds is not just a prediction per dataset per replica but the union of what should be preds for both PDFs.

In other words, `preds` is a list of size datasets*2 where the first half are the predictions for one PDF and the second one for the other. The `current` comes first so if the `current` fit has more replicas than the `reference` it will work, but the result for he `reference` will probably be wrong.

This is very obvious if one tries to do a comparefits in which both PDF have a different number of replicas (and crucially, the `current` one has less than the `reference`) because it will fail. 

Note: a better fix would be to modify the offending functions, but I wanted to have something working, so my fix has been overwrite the `pdfs` of the `comparefits` report.

### DIFF
--- a/validphys2/src/validphys/overfit_metric.py
+++ b/validphys2/src/validphys/overfit_metric.py
@@ -181,19 +181,21 @@ def plot_overfitting_histogram(fit, array_expected_overfitting):
 
     f, ax = plt.subplots(1, 1)
 
-    ax.hist(array_expected_overfitting, bins=50, density=True)
-    ax.axvline(x=mean, color="black")
-    ax.axvline(x=0, color="black", linestyle="--")
-    xrange = [
-        array_expected_overfitting.min(),
-        array_expected_overfitting.max(),
-    ]
-    xgrid = np.linspace(xrange[0], xrange[1], num=100)
-    ax.plot(xgrid, stats.norm.pdf(xgrid, mean, std))
-    ax.set_xlabel(r"$\mathcal{R}_O$")
-    ax.set_ylabel("density")
-    ax.set_title(f"{fit.label}")
-    plt.tight_layout()
+    # if array_expected_overfitting is nan it should not produce a histogram
+    if (array_expected_overfitting == array_expected_overfitting).all():
+        ax.hist(array_expected_overfitting, bins=50, density=True)
+        ax.axvline(x=mean, color="black")
+        ax.axvline(x=0, color="black", linestyle="--")
+        xrange = [
+            array_expected_overfitting.min(),
+            array_expected_overfitting.max(),
+        ]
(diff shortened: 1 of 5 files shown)


RoyStegeman (approved):

scarlehoff:
Nothing for the long-training length one.

scarlehoff:
@RoyStegeman do you want to do something more elegant or should I merge this?

RoyStegeman:
Yes I never really liked this. Let me see if I can clean it up a bit

RoyStegeman:
Can this be merged?

validphys2/src/validphys/overfit_metric.py
```diff
@@ -144,22 +147,28 @@ def array_expected_overfitting(
...
-        expected_val_chi2 = res_tmp.mean(axis=0)
-        delta_chi2 = fitted_val_erf_tmp - expected_val_chi2
-        expected_delta_chi2 = delta_chi2.mean()
-
-        list_expected_overfitting.append(expected_delta_chi2)
+    # calculate_chi2s_per_replica is set to zero if the pseudodata generation 
+    # has changed sinc the fit has been performed. As a result the overfitting
+    # metric can no longer be determined.
+    if (calculate_chi2s_per_replica == 0).all():
+        list_expected_overfitting = calculate_chi2s_per_replica
```
scarlehoff, line 154: Maybe it is better than in the case in which it cannot be computed the table is populated with "N/A" instead of [0, 0, NaN] which might confuse people.

scarlehoff:
The histogram is not plotted but the space for it does appear, was this the intended result?

https://vp.nnpdf.science/0OUk5sL5TVevS5BEHO28AA==/#dataspecs1_t0_info0_t0_info0_plot_overfitting_histogram

RoyStegeman:
yes, I need to output some figure for reportengine, so I just went with an empty figure since that's the simplest thing to do. I could replace it with a figure containing some written explanation of what happened instead?

scarlehoff:
> I could replace it with a figure containing some written explanation of what happened instead?

Well, this would be a possibility as well. I was just wondering whether it was intentional or not. You can merge it or add that, as you wish.

PR #1612, opened 2022-10-06
Fix legend of the replica distribution plot (and experiments_chi2_stats)

scarlehoff:
`experiments_chi2_stats` was using `rep_data.central_value().mean()` when it should instead be using `rep_data.error_members().mean()`

The only place where I think this has an effect is in the <X>_rep legend in the distribution of replica-chi2 plot I believe?

### DIFF
--- a/validphys2/src/validphys/results.py
+++ b/validphys2/src/validphys/results.py
@@ -771,7 +771,7 @@ def chi2_stats(abs_chi2_data):
     """
     rep_data, central_result, npoints = abs_chi2_data
     m = central_result.mean()
-    rep_mean = rep_data.central_value().mean()
+    rep_mean = rep_data.error_members().mean()
     return OrderedDict(
         [
             ("central_mean", m),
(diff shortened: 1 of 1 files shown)


Zaharid:
Does chi2_stats need to be changed?

scarlehoff:
True, that one would have the same problem.

scarlehoff:
Are you ok with this @Zaharid ?

RoyStegeman:
I'm happy with the fix, but the labels are pretty unclear and/or just wrong. How about this:
```
chi2_stat_labels = {
    "central_mean": r"$\chi^2_{rep0}$",
    "npoints": r"$N_{data}$",
    "perreplica_mean": r"$\left< \chi^2 \right>_{rep}$",
    "perreplica_std": r"$std_{rep}(\chi^2)$",
    "chi2_per_data": r"$\chi^2 / N_{data}$",
}
```

Zaharid:
These labels look good.

scarlehoff:
Labels changed, please merge once the tests are done.

RoyStegeman (approved):

PR #1613, opened 2022-10-07
Add t0 chi2 to comparefitsreport

RoyStegeman:
As discussed during the meeting today, we should have the t0 chi2 info by default included in the fitreports.

### DIFF
--- a/validphys2/src/validphys/comparefittemplates/report.md
+++ b/validphys2/src/validphys/comparefittemplates/report.md
@@ -11,6 +11,8 @@ We are comparing:
 
 {@ summarise_fits @}
 
+{@ dataspecs::t0_info t0_chi2_info_table @}
+
 Theory covariance summary
 -------------------------
 {@summarise_theory_covmat_fits@}
--- a/validphys2/src/validphys/comparefittemplates/report.md
+++ b/validphys2/src/validphys/comparefittemplates/report.md
@@ -11,6 +11,9 @@ We are comparing:
 
 {@ summarise_fits @}
 
+
+t0 losses
+---------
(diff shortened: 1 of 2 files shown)


validphys2/src/validphys/comparefittemplates/report.md
```diff
@@ -11,6 +11,8 @@ We are comparing:
 
 {@ summarise_fits @}
 
+{@ dataspecs::t0_info t0_chi2_info_table @}
```
Zaharid, line 14: We might want to have a heading here.

RoyStegeman: Now it's under the heading 'summary' where I would say it should be, I think a subsection under summary would be excessive. Do you disagree?

Zaharid: Fair enough. The entries themselves say that is the t0.

RoyStegeman: Exactly, I just made them separate tables since they can rely on different t0pdfsets

scarlehoff:
I'm going to be picky and say that it should be either part of the summary table or completely separated, with its own heading.

RoyStegeman:
A single table would of course be the obvious way to format it, but since the t0 settings would be different between entries reportengine cannot take care of it. So either we do this, or we just calculate the different t0 settings inside one of the summary functions out of reach from validphys (but this seems pretty ugly/bad practice).

But sine you're not alone with this opinion https://github.com/NNPDF/nnpdf/pull/1613#discussion_r990178548 I guess I'l place another heading.

PR #1615, opened 2022-10-18
[WIP][New commondata format] DYE866R 2001

tgiani:
As the title says. Following the old implementation the kinematic variables are y, M and sqrt(s), where y is computed from xF and M. Need to check y definition.

The implementation for the 2003 data, which should be the dataset DYE866P, will be done in a separate PR


enocera:
@tgiani May I ask you a favour: can you please have a single PR with ALL the DY data, instead of one PR for each data set?

tgiani:
@enocera sure. If you prefer a single PR for me there is no difference

enocera:
I definitely do. Thanks.

tgiani:
ok, I ll move everything in one

PR #1616, opened 2022-10-19
New make replica

VLTSML26:
The main version of `make_replica` in `pseudodata.py` takes a sampling covariance matrix in input and uses its square root to generate shifts for pseudodata. This is done through `sqrt_covmat` in `covmats.py` by calling a Cholesky decomposition method from `scipy.linalg` module.
The new version of `make_replica` will take as input only the eigenvalues and eigenvectors and will calculate the square root of the covariance matrix without using the `scipy` module. This is done under the assumption that another scipy module, `scipy.linalg.eig`, returns a orthogonal (or unitary) base change. Let $C$ be the covmat: it follows that $C=UDU^T$, when $D$ is diagonal and $U$ is the eigenvector matrix. Since $D$ is diagonal and $C$ is positive definite, we can write $C=UD^{1/2}D^{1/2}U^T=UD^{1/2}(UD^{1/2})^T$ and therefore $UD^{1/2}$ is a Cholesky decomposition, i.e. a square root of $C$.

I have tested the two methodologies with a level 2 closure test [here](https://vp.nnpdf.science/Bx1JXmgeR6yGwkpICmOIaQ==/) (the fits are also uploaded on the server). Is there another way of testing the right behavior of the new approach? I also would like to know if this would give problems elsewhere.

### DIFF
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -773,6 +773,17 @@ class CoreConfig(configparser.Config):
                 else:
                     return covmats.dataset_inputs_exp_covmat
 
+    @configparser.explicit_node
+    def produce_dataset_inputs_sampling_covmat_eigs(
+        self
+    ):
+        """
+        Produces the diagonalization of the sampling covmat to be used in new
+        approach to make_replica.
+        """
+        from validphys import covmats_utils
+        return covmats_utils.diagonalize_sampling_covmat
+
     def produce_loaded_theory_covmat(
         self,
         output_path,
--- a/validphys2/src/validphys/covmats_utils.py
+++ b/validphys2/src/validphys/covmats_utils.py
@@ -8,6 +8,7 @@ actions/providers.
 """
 import numpy as np
 import pandas as pd
+import scipy.linalg as la
 
 def systematics_matrix(stat_errors: np.array, sys_errors: pd.DataFrame):
     """Basic function to create a systematics matrix , :math:`A`, such that:
@@ -85,3 +86,8 @@ def construct_covmat(stat_errors: np.array, sys_errors: pd.DataFrame):
 
     corr_sys_mat = sys_errors.loc[:, ~is_uncorr].to_numpy()
     return np.diag(diagonal) + corr_sys_mat @ corr_sys_mat.T
+
+def diagonalize_sampling_covmat(
+    dataset_inputs_sampling_covmat
+):
+    return la.eig(dataset_inputs_sampling_covmat)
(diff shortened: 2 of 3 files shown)


Zaharid:
What is the advantage of this approach?

VLTSML26:
This is good whenever we encounter failures in `scipy.linalg.cholesky`, for instance when we perform closure tests with inconsistent data and the covmat is near-singular. If I am not mistaken, the function `sqrt_covmat` itself was introduced to avoid such problems, but sometimes it still fails.

Zaharid:
I don't see the benefit here. If the Cholesky decomposition fails, that means that the matrix is not positive definite and hence not a covariance matrix that one could use to build a Gaussian distribution to sample from. That is a huge telltale that whatever you are doing needs some rethinking (possibly featuring some kind of regularization of the input).

If the eigenvalues were computed as negative then the proposed approach would fail at `np.sqrt(eigval)` as it should. Additionally computing the eigenvalue decomposition is a lot slower than the Cholesky decomposition and unlikely to be better numerically for ill conditioned matrices (do note that there is a more specialized `eigh` but that does not fundamentally change the picture).

VLTSML26:
Still, it happens that the covariance matrix is positive definite in the sense that every eigenvalue $\lambda_i\ge 0$ and `scipy` treats small eigenvalues (e.g. `1e-10`) as zero. This causes the failure of the Cholesky decomposition, but note that does not cause failure in `np.sqrt` since $\sqrt{0}$ is well defined.

As you pointed out, a Gaussian distribution built from a singular covmat is not something that we should use to sample from. But note that if the eigenvalues are near zero and positive we should treat them as they are and not try to regulate them.
Even if a regularization is somehow required (e.g. if $\lambda_\text{max} / \lambda_\text{min} > 10^20$ there is no need to keep such small eigenvalues), I think that it will be faster with the new method since we can directly regularize the eigenvalues and we don't need to transform back into the "nondiagonal basis" and then perform Cholesky decomposition on the covmat.
I will be happy to implement such regularization.

Zaharid:
I don't believe scipy does anything in particular to clip small eigenvalues to zero and you should expect to get anything for ill defined matrices. In any case if the two routines disagree, that means one should trust neither and instead do something before getting to the point of fining out if the matrix is positive definite or not.

What to do depends on the problem. For example, we thought about the problem of keeping small eigenvalues where sensible in the decorrelation paper https://arxiv.org/abs/2207.00690 and the solution there was regularize the correlation matrix instead. That may or may not make sense in the context. One option may be to allow for additional systematics that are presented as something that is not an ndat x ndat matrix if that would help. I don't have a good sense as I never quite understood how this problem arises precisely.

PR #1618, opened 2022-10-24
Tr-vl-split replicas

goord:
Hi @scarlehoff and @Radonirinaunimi this is a branch containing my rebased changes onto the latest master. Initial tests seem to indicate things still work as intended, but again the NNPDF-4 reproduction has yet to be tested.

### DIFF
--- a/n3fit/src/n3fit/backends/keras_backend/MetaModel.py
+++ b/n3fit/src/n3fit/backends/keras_backend/MetaModel.py
@@ -80,6 +80,38 @@ def _fill_placeholders(original_input, new_input=None):
     return x
 
 
+def aggregate_replicas(dictionary, method="append", rep_token_index=1):
+    result = {}
+    if method == "append":
+        for key, value in dictionary.items():
+            tokens = key.split('_')
+            if len(tokens) > rep_token_index and tokens[rep_token_index].startswith("rep"):
+                tokens.pop(rep_token_index)
+                newname = '_'.join(tokens)
+                if newname in result:
+                    result[newname].append(value)
+                else:
+                    result[newname] = [value]
+            else:
+                result[key] = value
+        for key, value in result.items():
+            result[key] = np.array(value)
+    elif method == "sum":
+        for key, value in dictionary.items():
+            tokens = key.split('_')
+            if len(tokens) > rep_token_index and tokens[rep_token_index].startswith("rep"):
+                tokens.pop(rep_token_index)
+                newname = '_'.join(tokens)
+                if newname in result:
+                    result[newname] += value
+                else:
+                    result[newname] = value
+            else:
+                result[key] = value
+    else:
+        raise ValueError(f"Invalid replica aggregation method: {method}")
+    return result
+
 class MetaModel(Model):
     """
     The `MetaModel` behaves as the tensorflow.keras.model.Model class,
@@ -191,8 +223,7 @@ class MetaModel(Model):
         if y is None:
             y = self.target_tensors
         history = super().fit(x=x, y=y, epochs=epochs, **kwargs)
-        loss_dict = history.history
-        return loss_dict
+        return aggregate_replicas(history.history, method="sum")
 
     def predict(self, x=None, **kwargs):
         """ Call super().predict with the right input arguments """
@@ -200,6 +231,7 @@ class MetaModel(Model):
         result = super().predict(x=x, **kwargs)
         return result
 
+#TODO: Fix this!!!!!!!!!!!!!!
     def compute_losses(self):
         """
         This function is equivalent to the model ``evaluate(x,y)`` method of most TensorFlow models
@@ -216,19 +248,44 @@ class MetaModel(Model):
         """
         if self.compute_losses_function is None:
             # If it is the first time we are passing through, compile the function and save it
-            out_names = [f"{i}_loss" for i in self.output_names]
+            names = {}
+            for output_name in self.output_names:
+                tokens = output_name.split('_')
+                if tokens[1].startswith("rep"):
+                    tokens.pop(1)
+                data_name = '_'.join(tokens)
+                if data_name in names:
+                    names[data_name] += 1
+                else:
+                    names[data_name] = 1
+            n_replicas = next(iter(names.values()))
+            for name in names:
+                if names[name] != n_replicas:
+                    raise ValueError(f"Dataset {name} has unexpected no replica output layers {names[name]}")
+
+            out_names = [f"{i}_loss" for i in names]
             out_names.insert(0, "loss")
 
-            # Compile a evaluation function
+            # Compile evaluation function
             @tf.function
             def losses_fun():
                 predictions = self(self._parse_input(None))
                 # If we only have one dataset the output changes
                 if len(out_names) == 2:
                     predictions = [predictions]
-                total_loss = tf.reduce_sum(predictions, axis=0)
-                ret = [total_loss] + predictions
-                return dict(zip(out_names, ret))
+#                total_loss = tf.reduce_sum(predictions, axis=0)
+                total_losses = []
+                if n_replicas > 1:
+                    stacked_predictions = []
+                    for i in range(len(out_names)-1):
+                        reps = predictions[i * n_replicas: (i + 1) * n_replicas]
+                        stacked_predictions.append(tf.concat(reps, axis=0))
+                    for i in range(n_replicas):
+                        rep_losses = [predictions[i + d] for d in range(0, len(predictions), n_replicas)]
+                        total_losses.append(tf.reduce_sum(rep_losses, axis=0))
+                    predictions = stacked_predictions
+                result = [tf.concat(total_losses, axis=0)] + predictions
+                return dict(zip(out_names, result))
 
             self.compute_losses_function = losses_fun
 
@@ -236,7 +293,9 @@ class MetaModel(Model):
 
         # The output of this function is to be used by python (and numpy)
         # so we need to convert the tensors
-        return _to_numpy_or_python_type(ret)
+        ret2 = _to_numpy_or_python_type(ret)
+#        import pdb; pdb.set_trace()
+        return ret2
 
     def compile(
         self,
--- a/n3fit/src/n3fit/backends/keras_backend/MetaModel.py
+++ b/n3fit/src/n3fit/backends/keras_backend/MetaModel.py
@@ -231,7 +231,6 @@ class MetaModel(Model):
         result = super().predict(x=x, **kwargs)
         return result
 
-#TODO: Fix this!!!!!!!!!!!!!!
     def compute_losses(self):
         """
         This function is equivalent to the model ``evaluate(x,y)`` method of most TensorFlow models
--- a/n3fit/src/n3fit/checks.py
+++ b/n3fit/src/n3fit/checks.py
@@ -359,11 +359,11 @@ def check_consistent_parallel(parameters, parallel_models, same_trvl_per_replica
     """
     if not parallel_models:
         return
-    if not same_trvl_per_replica:
-        raise CheckError(
-            "Replicas cannot be run in parallel with different training/validation "
-            " masks, please set `same_trvl_per_replica` to True in the runcard"
-        )
+#    if not same_trvl_per_replica:
+#        raise CheckError(
+#            "Replicas cannot be run in parallel with different training/validation "
+#            " masks, please set `same_trvl_per_replica` to True in the runcard"
+#        )
     if parameters.get("layer_type") != "dense":
         raise CheckError("Parallelization has only been tested with layer_type=='dense'")
 
--- a/n3fit/src/n3fit/model_gen.py
+++ b/n3fit/src/n3fit/model_gen.py
@@ -243,7 +243,7 @@ def observable_generator(
         rotation=obsrot_tr,
     )
     out_vl = ObservableWrapper(
-        f"{spec_name}_val_{name_suffix}",
+        f"{spec_name}_{name_suffix}_val",
         model_obs_vl,
         dataset_xsizes,
         invcovmat=spec_dict["invcovmat_vl"],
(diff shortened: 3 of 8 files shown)


Radonirinaunimi:
Thanks @goord, I will have a look.

goord:
Found an issue with the integrabiliy constraints in the nnlo fit btw

scarlehoff:
> Found an issue with the integrabiliy constraints in the nnlo fit btw

What do you mean?

(thanks for the PR!)

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -80,6 +80,38 @@ def _fill_placeholders(original_input, new_input=None):
     return x
 
 
+def aggregate_replicas(dictionary, method="append", rep_token_index=1):
```
scarlehoff, line 83: Could you add some docstrings to this function? What is it used for, what is the expected input / output etc?
And what is the difference between append and sum.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -216,27 +248,54 @@ def compute_losses(self):
...
+                    tokens.pop(1)
+                data_name = '_'.join(tokens)
+                if data_name in names:
+                    names[data_name] += 1
+                else:
+                    names[data_name] = 1
+            n_replicas = next(iter(names.values()))
+            for name in names:
+                if names[name] != n_replicas:
+                    raise ValueError(f"Dataset {name} has unexpected no replica output layers {names[name]}")
```
scarlehoff, line 264: Is this error ever expected? If so, and depending on the situation, I would prefer to make sure that it is caught before the fit starts loading the data, otherwise is a very expensive crash.

Also, I'm not entirely sure I understand the logic of this loop. What is this used for? If it is just for the check I would rather remove it. When the fit goes to the point of computing the loss there should be no doubt whatsoever that all replicas are working, if there is a problem it should've been caught and fixed earlier.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -216,27 +248,54 @@ def compute_losses(self):
...
+                    stacked_predictions = []
+                    for i in range(len(out_names)-1):
+                        reps = predictions[i * n_replicas: (i + 1) * n_replicas]
+                        stacked_predictions.append(tf.concat(reps, axis=0))
+                    for i in range(n_replicas):
+                        rep_losses = [predictions[i + d] for d in range(0, len(predictions), n_replicas)]
+                        total_losses.append(tf.reduce_sum(rep_losses, axis=0))
+                    predictions = stacked_predictions
+                result = [tf.concat(total_losses, axis=0)] + predictions
+                return dict(zip(out_names, result))
```
scarlehoff, line 288: I'd rather lose some granularity on the loss that complicate (with quite a few loops) something that is going to be computed once per epoch. Take into account that we don't really care about the individual loss of the datasets (actually, when we will be doing th uncertainties, there will be no individual groups).
The loss per replica might be interesting, but again, it is something that can be recovered at the end.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -216,27 +248,54 @@ def compute_losses(self):
...
 
             self.compute_losses_function = losses_fun
 
         ret = self.compute_losses_function()
 
         # The output of this function is to be used by python (and numpy)
         # so we need to convert the tensors
-        return _to_numpy_or_python_type(ret)
+        ret2 = _to_numpy_or_python_type(ret)
+#        import pdb; pdb.set_trace()
```
scarlehoff, line 297:
```suggestion
```

n3fit/src/n3fit/checks.py
```diff
@@ -359,11 +359,11 @@ def check_consistent_parallel(parameters, parallel_models, same_trvl_per_replica
...
-    if not same_trvl_per_replica:
-        raise CheckError(
-            "Replicas cannot be run in parallel with different training/validation "
-            " masks, please set `same_trvl_per_replica` to True in the runcard"
-        )
+#    if not same_trvl_per_replica:
+#        raise CheckError(
+#            "Replicas cannot be run in parallel with different training/validation "
+#            " masks, please set `same_trvl_per_replica` to True in the runcard"
+#        )
```
scarlehoff, line 366:
```suggestion
```

n3fit/src/n3fit/model_gen.py
```diff
@@ -86,7 +86,7 @@ def __call__(self, pdf_layer, mask=None):
 
 
 def observable_generator(
-    spec_dict, positivity_initial=1.0, integrability=False
+    spec_dict, positivity_initial=1.0, integrability=False, name_suffix=''
```
scarlehoff, line 89: I'm guessing the `name_suffix` is so that different observable are treated differently so one can have different trvl masks. Please add to the docstr.

n3fit/src/n3fit/stopping.py
```diff
@@ -532,6 +534,7 @@ def monitor_chi2(self, training_info, epoch, print_stats=False):
             return False
 
         # Step 2. Compute the validation metrics
+#        import pdb; pdb.set_trace()
```
scarlehoff, line 537:
```suggestion
```

n3fit/src/n3fit/model_trainer.py
```diff
@@ -8,7 +8,9 @@
     This allows to use hyperscanning libraries, that need to change the parameters of the network
     between iterations while at the same time keeping the amount of redundant calls to a minimum
 """
+import copy
 import logging
+import pdb
```
scarlehoff, line 13:
```suggestion
```

n3fit/src/n3fit/model_trainer.py
```diff
@@ -349,12 +363,21 @@ def _model_generation(self, pdf_models, partition, partition_idx):
...
         sp_kw = {"axis": 1}
+        splitting_op = op.as_layer(op.split, op_args=sp_ar, op_kwargs=sp_kw, name="pdf_split")
         splitting_layer = op.as_layer(op.split, op_args=sp_ar, op_kwargs=sp_kw, name="pdf_split")
-        splitted_pdf = splitting_layer(full_pdf_per_replica)
+        # splitted_pdf = splitting_layer(full_pdf_per_replica)
+
+        splitted_pdfs = []
+        for layer in all_replicas_pdf:
+            split_layers = splitting_op(layer)
+            splitted_pdfs.append([op.expand(sl) for sl in split_layers])
```
scarlehoff, line 376: I'm confused about this changes. The way it is done now `full_pdf_per_replica` and `splitting_layer` are not needed.

The original logic was to have just one (very big) pdf that would convolute with the fktable for the many replica case. Have you tested that the changes doesn't lead to tf doing many convolutions? (one per replica) We would want to avoid that (and it should have no effect on having a trvl different per replica).

If you have (and it doesn't) then please remove `splitting_layer` and `full_pdf_per_replica` (since this is actually cleaner than what I had anyway).

scarlehoff (commented):
Thank you very much for this PR. I've been looking through it and left some comments as I went.

If I understand correctly the idea is to separate each replica so that instead of having

`<pdf> X (<fktable> X <mask>)`

and, in the case of many replicas:

`<many pdfs> x (<fktable> X <mask>)`

one has 

`<pdf> X (<fktable> X <mask>)` for each replica

The problem with that approach as you well say is that it increases the memory requirement by quite a bit (since right now `(<fktable> X <mask>)` is a singleton). In principle it is a growth of nreplicas* (and eventually we would like to do 50/100 replicas in parallel).

And as you say in some of the comments the solution would indeed be to factor out the fktable and the mask.

The (potential) problem with that approach is that it might be heavy on memory for the single replica fit (which we would like to avoid as it is the main use-case). But I think I would start there, meaning, trying the factorization in single replica fits, profiling that (in memory and time) and then apply said factorization to this branch.

If the factorization works in single replica fits without having an impact in performance it would be fantastic tbh.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
goord: Yes I found myself hard-coding loops to either make lists of partial losses or sum them in other cases, which is why I created this method. I think this entire loss aggregation business should have a round of code improvement because as it is it makes very specific assumptions about the structure of the output layer names. I would like to adopt a more robust design.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
goord: The loop is simply there to compose the out_names and this entire block including the losses_fun() can be simplified by saving some information at initialization. Actually this thing should be handled by a 'aggregate_losses' type of utility method, but it works at the level of tensors, which is why it's still there...

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
scarlehoff: But note that you don't need partial losses and that, moving forward, there will be no partial losses (as all experiments will be grouped together).

About the names, for better or for worse there are several layers of nnpdf code that rely on the names for different things (positivity, integrability, etc) so I guess you can rely on that as well here.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
scarlehoff: But why have the out_names changed with respect to the previous version?

Also, is the error ever expected? As I said, if it is, then it should be handled elsewhere.

n3fit/src/n3fit/model_trainer.py
goord: Yes those are remnants of earlier code and will be cleaned.

I don't think tf is doing more convolutions per se in this new architecture, but it probably cannot do them as efficiently as in the earlier version, where all replica tensors were essentially stacked no?

scarlehoff: Well, it is a difference between doing (pseudo code)

```python
pdf_{ri} X fktable_{i} = result_{r}
```

or

```python
result = []
for i in pdf_list:
    result.append(pdf_{i} X fktable_{i})
```
Specially in a GPU, the first version will be much faster than the second as there is only one copy of the PDF for all replicas (even if it is a bigger array) instead of many copy of smaller PDFs.

goord: In your second example, the allocation for the convolution will probably be the bottleneck on the GPU. But then again, tensorflow uses a big memory pool so this isn't an issue. As for the extra copies: you're right, this architecture has more layers, more tensors being copied and probably less utilization of the device. But we will measure this to avoid speculation.

As I said, this was exploratory from my side and I am going to focus my attention on getting the tr-vl split into the masked convolutions at the level of the Observable implementations. Currently the masks are only there to denote active flavors (uniform per replica of course); the aim will be to mask out data points per replica, right?

scarlehoff: No, there are two masks. The flavour mask is applied to the pdf/luminosity and it is always the same for all replicas for a given fktable.

Instead, the training/validation mask is different per replica. Currently it is applied on the fktable since it reduces its size considerably. However, by doing that, a different trvl mask generates a new fktable, which is a bottleneck for parallelizarion over the number of replicas.

Instead the trvl mask would need to be applied to the data, so that all replicas use the same fktable and generate the same (in shape and operations) output vector and only at the end the validation is data is masked out

goord: Ok so where the observable class now contains a list of fk-tables (I assume length n-replicas, each entry a masked version of the full fk-table), this will be only the full one. When you say 'at the end the validation data is masked out' you mean the observable layer should perform an extra masking of the fk-table convolution. So for instance DY will first execute a masked convolution, then a tensor product with the fk-tables (uniform across replicas) and finally masks out the validation entries per replica.

scarlehoff: The fitting process is as follows (for a DY-type observable which is the most complicated one). The exact order of the shape might not be correct in the example below though. 

x enters as an input to the PDF model, input shape (n_x,)

The PDF model is a concatenation of PDFs, one for replica, its output is then (nreplicas, n_flavours, n_x)

The result of the PDF models is then given to the observable. The observable can be formed by several fktables (for instance, the observable can be the measure of a ratio between two cross sections).
Inside the observable first a luminosity tensor is constructed to account for the two colliding protons, so that we have a luminosity such that:

L = PDF x PDF, shape (nreplicas, n_flavours, n_flavours, n_x, n_x)

However, not all flavour combinations contribute to all observables, so this luminosity tensor is masked such that only active flavours remain (which means the calculation to be done is also lighter).

The new masked luminosity is then: mL = (nreplicas, n_active_flavours_combinations, n_x, n_x)

Now this masked luminosity is convoluted with every fktable in the observable. The output of this convolution is a vector of data points (n_data).

If the observable is composed (for instance, a ratio between two cross sections or whatever) then an operation is applied to all the output arrays.

The final result of the observable is then (n_replicas, n_data).

-----------------------------------------------------------------------------------------------------------

However, in order to do the fit correctly we apply a training/validation mask. This traning validation mask is applied at the level of the data points, so the natural way would be, after the computation is finished, apply the mask, so that one gets:

training_observable = observable X mask
validation_observable = observable X ~mask

But, for backpropagation, we are only interested in the training_observable, and the fktable is a massive object. Therefore what we do is to apply the mask directly to the fktable so that we have

fk_training = fk X mask
fk_validation = fk X ~mask

And then, using those two fktables, we generate two observables (so we no longer have to apply the mask at the end).

Note btw that this training/validation mask to the fktable is applied before the training even starts.

goord: ok thanks for your explanation @scarlehoff, I was a bit confused by the list of fk-tables within the observable class, but your comment clarifies this.

scarlehoff:
Closing this branch in favour of #1661

PR #1620, opened 2022-10-27
Use theory 400 for the fit bot

scarlehoff:
Since pineappl will soon become the default and main provider for fktable, the fit bot should be testing with theory 400.

We might need to create a theory specific for this since theory 400 has a very large number of points and I don't have a feeling of how much memory it will take.

It might also be a good idea to retain a "fitbot-legacy" in order to check that a fit with theory 200 also works.

### DIFF
--- a/n3fit/runcards/examples/developing.yml
+++ b/n3fit/runcards/examples/developing.yml
@@ -60,7 +60,7 @@ datacuts:
 
 ############################################################
 theory:
-  theoryid: 200        # database id
+  theoryid: 400        # database id
 
 sampling:
   use_t0: false


scarlehoff:
@andreab1997 @felixhekhorn how difficult would it be (and how long would it take) to create a theory 400 with something like 20 points in x instead of 100 ?

Basically the "low precision" version of theory 400 to be used in the tests.

andreab1997:
> @andreab1997 @felixhekhorn how difficult would it be (and how long would it take) to create a theory 400 with something like 20 points in x instead of 100 ?
> 
> Basically the "low precision" version of theory 400 to be used in the tests.

It is actually pretty simple but I believe also a bit long (basically everything needs to be recomputed but maybe the ekos that I believe can just be reshaped). At the moment I would not do that and I would prioritize the sv theories. Once they are completed I can of course do that. Otherwise you can do that and I can provide help whenever needed.

scarlehoff:
> basically everything needs to be recomputed

What does everything means?
Only the ekos would need to be recomputed right?

andreab1997:
> > basically everything needs to be recomputed
> 
> What does everything means? Only the ekos would need to be recomputed right?

And the FK of course, which can take quite a bit for non-DIS stuff. Moreover as I said I am not sure if the eko can be just reshaped. So yes, maybe we also need to recompute them.

scarlehoff:
> And the FK of course, which can take quite a bit for non-DIS stuff. Moreover as I said I am not sure if the ekos can be just reshaped. So yes, maybe we also need to recompute them.

But this is something that can happen unsupervised right? We have no rush and I would not want to have dom 100% occupied for this, but some computer in a corner can be left computing this during a week.

andreab1997:
> > And the FK of course, which can take quite a bit for non-DIS stuff. Moreover as I said I am not sure if the ekos can be just reshaped. So yes, maybe we also need to recompute them.
> 
> But this is something that can happen unsupervised right? We have no rush and I would not want to have dom 100% occupied for this, but some computer in a corner can be left computing this during a week.

Yes this is exactly the point. I will use galileo to compute sv theories. I can use dom to compute this but maybe it is better to use another computer as you say (even because in this way we can try to configure everything from scratch and use this exercise as a tutorial)

scarlehoff:
>  I can use dom to compute this

Yes. If you do please make sure that it takes only 4 cores.

> use this exercise as a tutorial

We can _but_ this is something we want to actually have, not leave it forever (like when we give undergrad students some projects) :___

andreab1997:
> > I can use dom to compute this
> 
> Yes. If you do please make sure that it takes only 4 cores.
> 
> > use this exercise as a tutorial
> 
> We can _but_ this is something we want to actually have, not leave it forever (like when we give undergrad students some projects) :___

 Anyway for the moment we need to wait: some refactor happened in eko and so we are currently building some compatibility layer in `pineko` in order to make it work. Most likely, starting from next week I can start computing everything.

PR #1625, opened 2022-11-08
Simplify MetaModel input handling

### LINKED ISSUE #1624: MetaModel should have a dictionary interface only
The meta model class here 
https://github.com/NNPDF/nnpdf/blob/94b80c35618bd55ae463b5806bf2b5886f6ecad6/n3fit/src/n3fit/backends/keras_backend/MetaModel.py#L83
is supposed to be a thin wrapper over a Keras model, where one of the differences is that it works around the peculiarity of keras where one has to bend over backwards to pass constant tensors to layers (i.e. https://github.com/keras-team/keras/issues/11912).

However the way in which it does it is a bit convoluted. One can pass a dictionary like for a keras model, but then the information of the keys is discarded, 

https://github.com/NNPDF/nnpdf/blob/94b80c35618bd55ae463b5806bf2b5886f6ecad6/n3fit/src/n3fit/backends/keras_backend/MetaModel.py#L112

and new keys are added, with some complicated logic

https://github.com/NNPDF/nnpdf/blob/94b80c35618bd55ae463b5806bf2b5886f6ecad6/n3fit/src/n3fit/backends/keras_backend/MetaModel.py#L142

which happens on depend on the keras logic not doing things like

```
In [1]: import tensorflow as tf

In [2]: name = "I really want this name"

In [3]: print(tf.keras.Input(tensor=tf.constant([1,2,3], name=name), name=name).name)

None
```

and then more cumbersome logic to recall the proper inputs, like e.g.

https://github.com/NNPDF/nnpdf/blob/94b80c35618bd55ae463b5806bf2b5886f6ecad6/n3fit/src/n3fit/backends/keras_backend/MetaModel.py#L56

which can get tricky both for the class and the caller.

I think a better way would be to mandate that the input argument is only allowed to be a dict, and similarly allow only dict predictions and similar. Then the keys are the users responsibility, and we can easily produce better errors if some input is not given, rather than some hard to decipher message down the line.

Zaharid:
Mandate that all input is passed as dictionaries. This has the advantage of doing away with a lot of code, having an obvious way to derive the name for each input (the dictionary key) and allowing for better error messages.

Closes #1624

### DIFF
--- a/n3fit/src/n3fit/backends/keras_backend/MetaModel.py
+++ b/n3fit/src/n3fit/backends/keras_backend/MetaModel.py
@@ -53,121 +53,98 @@ def _default_loss(y_true, y_pred): # pylint: disable=unused-argument
     return op.sum(y_pred)
 
 
-def _fill_placeholders(original_input, new_input=None):
-    """
-    Fills the placeholders of the original input with a new set of input
-
-    Parameters
-    ----------
-        original_input: dictionary
-            dictionary of input layers, can contain None
-        new_input: list or dictionary
-            list or dictionary of layers to substitute the None with
-    """
-    if new_input is None:
-        return original_input
-    x = {}
-    i = 0
-    for key, value in original_input.items():
-        if value is None:
-            try:
-                x[key] = new_input[key]
-            except TypeError:
-                x[key] = new_input[i]
-                i += 1
-        else:
-            x[key] = value
-    return x
-
 
 class MetaModel(Model):
     """
-    The `MetaModel` behaves as the tensorflow.keras.model.Model class,
-    with the addition of `tensor_content`:
-
-    - tensor_content:
-    Sometimes when fitting a network the input is fixed, in this case the input can be given
-    together with the input_tensors by setting a `tensor_content` equal to the input value.
-    This is done automatically when using the `numpy_to_input` function from
-    `n3fit.backends.keras_backend.operations`
+    The model wraps keras.Model and adds some custom behaviour. Most notably it
+    allows supplying constant values for input arguments, which are used when
+    training and making predictions with the model (note that constants need to
+    be explicitly registered as inputs, see
+    https://github.com/keras-team/keras/issues/11912). These inputs can be
+    passed in the ``input_values`` parameter, or gathered from the
+    ``tensor_content`` attribute of the ``input_tensors``, which is set
+    automatically when using the ``numpy_to_input`` function from
+    :py:mod:`n3fit.backends.keras_backend.operations`.
 
     Parameters
     ----------
-        input_tensors: tensorflow.keras.layers.Input
+        input_tensors: dict[Any, tensorflow.keras.layers.Input]
             Input layer
         output_tensors: tensorflow.keras.layers.Layer
             Output layer
+        input_values: dict[Any, array_like]
+            Constant values for the input layer, to be supplied when making
+            predictions with the model.
+
         **kwargs:
             keyword arguments to pass directly to Model
     """
 
     accepted_optimizers = optimizers
 
-    def __init__(self, input_tensors, output_tensors, scaler=None, **kwargs):
+    def __init__(self, input_tensors, output_tensors, scaler=None, input_values=None, **kwargs):
         self.has_dataset = False
+        self.required_slots = set()
 
-        input_list = input_tensors
-        output_list = output_tensors
+        if input_values is None:
+            input_values = {}
 
-        if isinstance(input_list, dict):
-            # if this is a dictionary, convert it to a list for now
-            input_list = input_tensors.values()
-        elif not isinstance(input_list, list):
-            # if it is not a dict but also not a list, make it into a 1-element list and pray
-            input_list = [input_list]
+        if not isinstance(input_tensors, dict):
+            raise TypeError("Expecting input_tensors to be a dict")
 
-        if isinstance(output_list, dict):
-            # if this is a dictionary, convert it to a list for now
-            output_list = output_tensors.values()
-        elif not isinstance(output_list, list):
-            # if it is not a dict but also not a list, make it into a 1-element list and pray
-            output_list = [output_list]
+        if not isinstance(input_values, dict):
+            raise TypeError("Expecting input_values to be a dict or None")
 
-        # Note: there used to be two possible options when creating a model:
-        # - Give placeholder tensors (for which the content will be given at run time)
-        # - Give tensors with content* (for which the content is stored with the model
-        # *this option was dropped at some point by TF, the code below keeps this behaviour
-        # We will store within the model the following quantities:
-        #   -> x_in: arrays containing the x-input to the model
-        #   -> tensors_in: when the x-input is known at compile time, we store a reference to the tensor
-        # We pass TensorFlow a dictionary {k: tensor} containing placeholders which will be automatically filled
-        # whenever x_in/tensor_in is known at compile time
 
         x_in = {}
-        tensors_in = {}
-        input_dict = {}
-        for input_tensor in input_list:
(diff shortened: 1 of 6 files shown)


Zaharid:
@scarlehoff if you agree with this in principle I'll work on tests and so on.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -103,71 +77,54 @@ class MetaModel(Model):
...
-            input_dict[name] = input_tensor
-            try:
-                x_in[name] = op.numpy_to_tensor(input_tensor.tensor_content)
-                tensors_in[name] = input_tensor
-            except AttributeError:
-                x_in[name] = None
-                tensors_in[name] = None
-
-        super().__init__(input_dict, output_list, **kwargs)
+        for k, v in input_tensors.items():
```
scarlehoff, line 91: Please add some comments here

scarlehoff (commented):
I'm happy with the change _in principle_.
But the convoluted way of doing this was because of a change at some point on how Tensorflow was handling things so you need to make sure it works in all possible scenarios that we have (in practice, at least with all the runcards in the example folder).

i.e., parallelization over replicas, multiples replicas, feature scaling, with/without sum rules, etc.

I'm actually not even sure whether the dictionary is currently the only option used so that most of the code that was there is currently unused.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -103,71 +77,54 @@ class MetaModel(Model):
...
         """
-        if isinstance(extra_input, dict) and self._scaler is not None:
+        if extra_input is None:
+            if self.required_slots:
+                raise ValueError(
+                    f"The following inputs must be provided: {self.required_slots}"
+                )
+            return self.x_in
+
+        if diff := (self.required_slots - extra_input.keys()):
```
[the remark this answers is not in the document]
Zaharid: Don't remember the context and I do dislike it overall. Maybe something to do with it not working for old python? Or requiring too many parentheses?). I don't mind either way, but looks okay to me in here.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -187,10 +144,10 @@ def perform_fit(self, x=None, y=None, epochs=1, **kwargs):
             loss_dict: dict
                 a dictionary with all partial losses of the model
         """
-        x = self._parse_input(x)
+        params = self._parse_input(x)
         if y is None:
             y = self.target_tensors
-        history = super().fit(x=x, y=y, epochs=epochs, **kwargs)
+        history = super().fit(x=params, y=y, epochs=epochs, **kwargs)
```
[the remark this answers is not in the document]
scarlehoff: parsed x or whatever would also be fine, they are just not parameters so could get confused.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -53,121 +53,87 @@ def _default_loss(y_true, y_pred): # pylint: disable=unused-argument
...
-                x_in[name] = op.numpy_to_tensor(input_tensor.tensor_content)
-                tensors_in[name] = input_tensor
-            except AttributeError:
-                x_in[name] = None
-                tensors_in[name] = None
-
-        super().__init__(input_dict, output_list, **kwargs)
+        # Go over the inputs. If we can deduce a constant value, either because
+        # it is set in input_values or because it has a tensor_content, we
+        # store it. Otherwise we mar the input as required when making
```
scarlehoff, line 98:
```suggestion
        # store it. Otherwise we mark the input as required when making
```

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -53,121 +53,87 @@ def _default_loss(y_true, y_pred): # pylint: disable=unused-argument
...
-        super().__init__(input_dict, output_list, **kwargs)
+        # Go over the inputs. If we can deduce a constant value, either because
+        # it is set in input_values or because it has a tensor_content, we
+        # store it. Otherwise we mark the input as required when making
+        # predictions.
+        for k, v in input_tensors.items():
+            if k in input_values:
+                x_in[k] = input_values[k]
+            elif hasattr(v, "tensor_content"):
+                x_in[k] = v.tensor_content
```
scarlehoff, line 104: You should keep the `op.numpy_to_tensor` here.

The reason for that is having one single place where to cast the numpy array to other types (if we want to do `double` instead of `float` or whatever).

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -53,121 +53,87 @@ def _default_loss(y_true, y_pred): # pylint: disable=unused-argument
...
-        if isinstance(extra_input, dict) and self._scaler is not None:
+        if extra_input is None:
+            if self.required_slots:
+                raise ValueError(
+                    f"The following inputs must be provided: {self.required_slots}"
+                )
+            return self.x_in
+
+        if diff := (self.required_slots - extra_input.keys()):
+            raise ValueError(f"The following inputs must be provided {diff}")
```
scarlehoff, line 131: A few comments here:

- `extra_input` could enter as a list (or as whatever without `keys`). Since we are checking for mistakes, this should be tested for (and the error message should also include the keys that are required).
- The `diff` should also check whether there are `extra_inputs` being provided that is not required.

scarlehoff (commented):
You will also need to modify penalties https://github.com/NNPDF/nnpdf/blob/94b80c35618bd55ae463b5806bf2b5886f6ecad6/n3fit/src/n3fit/hyper_optimization/penalties.py#L58

For the rest I think it is ok.

However, now one cannot simply call the model like one would do for another model with some arrays and get some results but have to actually know the keys, which I don't like.

But I guess the simplification is worth it and this ^ is a developer use case (which used to be only me). 
But please, could you add more stuff to the MetaModel docstring explaining this? With an example with its ```>>>``` and everything.

Zaharid:
> However, now one cannot simply call the model like one would do for another model with some arrays and get some results but have to actually know the keys, which I don't like.

I don't see a way to do this, have the input tracking behavior, and have a reasonable interface.

scarlehoff:
If you force-push every commit it's very difficult to keep track of what is changing.

Zaharid:
Should be everything that you asked and none of what yo didn't ask :P.

scarlehoff:
> Should be everything that you asked and none of what yo didn't ask :P.

Is this ready for review then?

Zaharid:
Yes please.

scarlehoff (approved):
lgtm

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -53,121 +53,98 @@ def _default_loss(y_true, y_pred): # pylint: disable=unused-argument
...
-                tensors_in[name] = None
-
-        super().__init__(input_dict, output_list, **kwargs)
+        # Go over the inputs. If we can deduce a constant value, either because
+        # it is set in input_values or because it has a tensor_content, we
+        # store it. Otherwise we mark the input as required when making
+        # predictions.
+        for k, v in input_tensors.items():
+            if k in input_values:
+                x_in[k] = input_values[k]
```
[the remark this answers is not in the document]
Zaharid: I would not do this: Would prefer it to be to the caller, e.g. if creating a keras Input manually or getting it from somewhere.

scarlehoff:
Stop with the force pushes  :___
We can squash and merge if you want to have everything well organized.

PR #1626, opened 2022-11-09
Remove additional t0 flags

Zaharid:
The work on covariance matrices added new flags to control to t0 setting, namely `use_t0_sampling` and `use_t0_fitting`. These were undocumented, and caused bugs in validphys since t0pdfset could be set by either of them.

Remove both flags and mandate that the fitting covmat always uses t0 while the sampling covmat never uses it.

This removes some footguns and its backwards compatible.

It has the minor disadvantage that the fitting covmat implicitly requires use_t0=True, which is redundant, but it only really matters in n3fit and we set the flag there anyway.

### DIFF
--- a/n3fit/src/n3fit/scripts/n3fit_exec.py
+++ b/n3fit/src/n3fit/scripts/n3fit_exec.py
@@ -158,10 +158,8 @@ class N3FitConfig(Config):
             N3FIT_FIXED_CONFIG['use_scalevar_uncertainties'] = thconfig.get('use_scalevar_uncertainties', True) 
         #Sampling flags
         if (sam_t0:=file_content.get('sampling')) is not None:
-            N3FIT_FIXED_CONFIG['use_t0_sampling'] = sam_t0.get('use_t0', False) 
             N3FIT_FIXED_CONFIG['separate_multiplicative'] = sam_t0.get('separate_multiplicative', True) 
         #Fitting flag
-        N3FIT_FIXED_CONFIG['use_t0_fitting'] = file_content.get('fitting').get('use_t0', True)
         file_content.update(N3FIT_FIXED_CONFIG)
         return cls(file_content, *args, **kwargs)
 
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -713,7 +713,6 @@ class CoreConfig(configparser.Config):
         self,
         theory_covmat_flag=False,
         use_thcovmat_in_fitting=False,
-        use_t0_fitting=True,
     ):
         """
         Produces the correct covmat to be used in fitting_data_dict according
@@ -723,16 +722,9 @@ class CoreConfig(configparser.Config):
         """
         from validphys import covmats
 
-        if use_t0_fitting:
-            if theory_covmat_flag and use_thcovmat_in_fitting:
-                return covmats.dataset_inputs_t0_total_covmat
-            else:
-                return covmats.dataset_inputs_t0_exp_covmat
-        else:
-            if theory_covmat_flag and use_thcovmat_in_fitting:
-                return covmats.dataset_inputs_total_covmat
-            else:
-                return covmats.dataset_inputs_exp_covmat
+        if theory_covmat_flag and use_thcovmat_in_fitting:
+            return covmats.dataset_inputs_t0_total_covmat
+        return covmats.dataset_inputs_t0_exp_covmat
 
     @configparser.explicit_node
     def produce_dataset_inputs_sampling_covmat(
@@ -740,7 +732,6 @@ class CoreConfig(configparser.Config):
         sep_mult,
         theory_covmat_flag=False,
         use_thcovmat_in_sampling=False,
-        use_t0_sampling=False,
     ):
         """
         Produces the correct covmat to be used in make_replica according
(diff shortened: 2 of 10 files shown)


Zaharid:
Note this is less flexible than what we used to have as one cannot meaningfully override to `use_t0: False` in the fit runcard. But I don't think it matters.

scarlehoff:
But we might want to do fits without the t0 covmat, right? (for testing, show how important it is, etc)

Zaharid:
I think when (if) there is a pressing need to run these tests, whoever does it should re-add the options making sure they are documented and tested properly.

It is easy enough to swap for an one off test modifying the source.

n3fit/src/n3fit/scripts/n3fit_exec.py
```diff
@@ -158,10 +158,8 @@ def from_yaml(cls, o, *args, **kwargs):
             N3FIT_FIXED_CONFIG['use_scalevar_uncertainties'] = thconfig.get('use_scalevar_uncertainties', True) 
         #Sampling flags
         if (sam_t0:=file_content.get('sampling')) is not None:
-            N3FIT_FIXED_CONFIG['use_t0_sampling'] = sam_t0.get('use_t0', False) 
             N3FIT_FIXED_CONFIG['separate_multiplicative'] = sam_t0.get('separate_multiplicative', True) 
         #Fitting flag
```
scarlehoff, line 162: Is the only other flag of `sampling` `separate_multiplicative` @andreab1997 ? 

If so it might make sense to remove this extra level.

Also, maybe in the same vein as removing the option to not use t0 we might want to remove the option of doing the sampling with the multiplicative uncertainties separated.

validphys2/src/validphys/config.py
```diff
@@ -723,24 +722,17 @@ def produce_dataset_inputs_fitting_covmat(
...
-            else:
-                return covmats.dataset_inputs_t0_exp_covmat
+        if theory_covmat_flag and use_thcovmat_in_fitting:
+            return covmats.dataset_inputs_t0_total_covmat
         else:
-            if theory_covmat_flag and use_thcovmat_in_fitting:
-                return covmats.dataset_inputs_total_covmat
-            else:
-                return covmats.dataset_inputs_exp_covmat
+            return covmats.dataset_inputs_t0_exp_covmat
```
scarlehoff, line 728: I would remove the `else` given the other changes.

Btw, why are we doing the import inside the method? Circular imports?

scarlehoff (commented):
Please remove also the flags from the runcards and from the documentation.

I don't see any fits without the t0 other than very specific tests so I tend to agree with your last comment.

validphys2/src/validphys/config.py
Zaharid: Probable for no good reason anymore: At some point I wanted to make this file import quickly, but that ships has probably sailed.

Zaharid: What do you mean exactly with removing the else? I would say we want support for enabling and disabling the theorycovmat. Not clear to me why there are two flags.

cc @andreab1997

scarlehoff: Just a stylistic comment, I would do

```python
if this:
    return that
return thet
```

The two flags are there because the th covmat can be used in the fit or in the sampling.
Since seeing the difference was relevant at some point I think that should stay.

Zaharid: > The two flags are there because the th covmat can be used in the fit or in the sampling.
> Since seeing the difference was relevant at some point I think that should stay.

but the two flags are in the fitting action...

scarlehoff: One to say that the th covmat is active and another two to decide whether it is active in sampling or fitting.

A bit like the `use_t0` but in this case there is no the solution of "fitting without t0 is wrong, so let's remove the wrong option".

I guess that's the two flags you are referring to ?

n3fit/src/n3fit/scripts/n3fit_exec.py
andreab1997: > Is the only other flag of `sampling` `separate_multiplicative` @andreab1997 ?

To answer this question, yes so it makes sense to remove completely the sampling level

validphys2/src/validphys/config.py
andreab1997: Indeed you can decide to use the thcovmat in only one of the two places or in both and I believe that one of Stefano's student will actually try to have a look to the different fits you can get with these three options.

Zaharid:
I'd like to merge this as is, and defer discussion of other aspects of the covmat setting.

andreab1997 (commented):
For me it is even okay to merge. I am only confused about the flag `use_t0`. As far as I understood it should always be `True` (otherwise there will be an error), right? And, even if it is `True` we are only using `t0` for the `fitting covmat` and never for the `sampling covmat`. Is my understanding correct? Because in this case I believe it is a bit confusing and we should probably just remove it and rely on the default behaviour.

n3fit/src/n3fit/scripts/n3fit_exec.py
scarlehoff: I would say let's remove both the sampling level and `separate_multiplicative`

scarlehoff:
> I'd like to merge this as is, and defer discussion of other aspects of the covmat setting.

Ok, but please update the documentation and runcards.

Zaharid:
`use_t0` is for any generic covmat in vp. A a special case there are the sampling and fitting functions that as of this pr either always use it or never use it. As pointed above, the fitting covmat requires that use_t0 is explicitly set to True, but we do that in the fit, where it matters most. To me that is the only problem and too minor for me to bother.

andreab1997 (approved):

scarlehoff (changes requested):
I think once runcards/docs are updated this can be merged.

scarlehoff:
https://github.com/NNPDF/nnpdf/blob/94b80c35618bd55ae463b5806bf2b5886f6ecad6/n3fit/runcards/examples/Basic_runcard.yml#L36

https://docs.nnpdf.science/tutorials/thcov_tutorial.html

scarlehoff (approved):
thanks

scarlehoff:
Should I merge this @Zaharid ?

PR #1597, opened 2022-09-08
Remove markdown handling from the docs

Zaharid:
Having markdown required more tooling and more documentation and it is not worth it as it ends up being full of `eval_rst` clauses. 

Convert the existing files to markdown and remove the tooling for it.

One problem is that rst has for some reason no obvious way of having markup inside the links, which is something markdown doesn't know about and needs to be cleaned by hand. Also we had explicit references to markdown source files in varios places. Overall there is quite a big chance of things being missed or getting ugly, even though I tried to remove mos of the warnings.

PR #1628, opened 2022-11-11
Build for Python 3.10

Zaharid:


scarlehoff:
I will be so happy once we can forget about conda :__

Would be lovely to have 3.11 as soon as possible.

Zaharid:
Bottlenecks include that I can't seem to be able to build apfel for osx anymore. No error message to be found:

https://github.com/Zaharid/conda-recipes/actions/runs/3460416136/jobs/5780057654

scarlehoff:
Try first only 3.9, that one should work.

Zaharid:
AFAICT it does not (as per the link above).

scarlehoff:
But it did 23 hours ago! https://github.com/NNPDF/nnpdf/actions/runs/3462322249/jobs/5781058161 That's why I would do step by step. First do just "3.9" to check that somehow the " didn't break everything.

Zaharid:
Note the link is for the apfel recipe repo trying to build apfel. NNPDF should work fine with existing packages.

Zaharid:
No. That would need to be fixed by someone with a mac.

scarlehoff:
Ok, I'm able to compile apfel with brew python 3.10

I'll try with conda as well, but I would need to prepare it (don't have conda installed yet).

However, this is an M1 and I believe workflows are intel ?

Zaharid:
@scarlehoff since you are very clearly having a great time with similar problems, any chance you can look into this as well. I think the specific problem is that apfel is trying to get some metadata from pypi that does not exist.

https://github.com/Zaharid/conda-recipes/actions/runs/3460416136/jobs/5780057654#step:6:1325

but I have no idea why.

scarlehoff:
I'm just hoping we merge #1650 and #1651 soonish and this is not needed.

In any case, in the M1 mac just doing `conda-build` doesn't work but installing by hand does work. The point at which yours fail is that when it needs to install the `apfel` you already have compiled (so, `make install`) it goes to pypi to look for it (and it obviously fails).

Why? That's a good question.

But I have another good question, do we need the python extension of apfel? We only need apfel because libNNPDF is built against it but it is now only used by `evolven3fit` which is a C++ code.

Zaharid:
> I'm just hoping we merge #1650 and #1651 soonish and this is not needed.

We would need this until the next PDF paper is written, at minimum.

> 
> In any case, in the M1 mac just doing `conda-build` doesn't work (but it crashes before it starts) but installing by hand does work. The point at which yours fail is that when it needs to install the `apfel` you already have compiled (so, `make install`) it goes to pypi to look for it (and it obviously fails).
> 
> Why? That's a good question.
> 
> But I have another good question, do we need the python extension of apfel? We only need apfel because libNNPDF is built against it but it is now only used by `evolven3fit` which is a C++ code.

Good point. Will try disabling that.

Zaharid:
FWIW doing the obvious thing with `--disable-pywrap` didn't help: https://github.com/Zaharid/conda-recipes/actions/runs/3950694019/jobs/6763574859

Maybe @scarrazza could have a look?

Zaharid:
Here. This is what was needed rather than "discussion".

@scarlehoff  can you have a look at this now?

scarlehoff:
Let's wait until #1537 is merged just in case.

scarlehoff:
I'm confused about the two 3.8 that are left in "Required". I guess that's due to the change in the test and will disappear upon merging?

I've modified the fitbot to run also under python 3.10, if that works ok I guess there is not much chance anything else will break (I've been using vp with 3.10 for a very long time now)

scarlehoff:
@Zaharid are you fine with me merging master into this branch / rebasing this branch on top of master for the test to run?

Zaharid:
Yes, please. Anything that makes it work.

scarlehoff:
@Zaharid are you ok with me merging this? I think having a 3.10 package might be useful for me tomorrow actually.

Zaharid:
There might be a problem building the markdown docs. Not sure if we can live with that for now. Slightly regretting not finishing #1597.

scarlehoff:
> There might be a problem building the markdown docs (it may also be fixed, haven't tried).

What kind of problem do you mean?

Zaharid:
AFAICT #1220 would still be an issue.

scarlehoff:
Mm... a failure in building the docs would only mean the online docs will be one PR outdated, right?

If so I'd merge. If the docs fail I'll do my best to fix it tomorrow so that people don't see an ugly X but I prefer to have the 3.10 package available for people coming with their own conda installation.

Zaharid:
Makes sense to me. Maybe we can resurrect #1596 although ideally we would remove md. Unfortunately pandoc leaves  many things to be fixed manually.

scarlehoff:
@Zaharid why did you remove the pin to docutils here? (thinking about adding it again just to get the green check... or skipping the docs for python 3.10, we only need them for one of the two after all...)

PR #1632, opened 2022-11-14
Propagate target info from the runcard

Radonirinaunimi:
The following propagates the information on the target(s) from the run card into the data specs. The information are passed as a stringified dictionary into the `target_info` key:
```yaml
- { dataset: NMCPD, frac: 0.5, target_info: '{"A": [1, 2], "Z": [1, 2]}'}
```
These changes will be reverted in the future when the target info are added as metadata in the pineappl grids/FK tables in which they will be part of `validphys.coredata.FKTableData`.

@scarlehoff After some thoughts, for the time being, I have left it the way I did it before, ie adding the info as an attribute of `validphys.n3fit_data_utils.FittableDataSet` instead of `validphys.coredata.FKTableData`. Adding it into the later will require more changes in various places inc. `fkparser` and `pineparser`. Does this make sense?

### DIFF
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -405,7 +405,7 @@ class CoreConfig(configparser.Config):
     def parse_dataset_input(self, dataset: Mapping):
         """The mapping that corresponds to the dataset specifications in the
         fit files"""
-        known_keys = {"dataset", "sys", "cfac", "frac", "weight", "custom_group"}
+        known_keys = {"dataset", "sys", "target_info", "cfac", "frac", "weight", "custom_group"}
         try:
             name = dataset["dataset"]
             if not isinstance(name, str):
@@ -424,6 +424,7 @@ class CoreConfig(configparser.Config):
         sysnum = dataset.get("sys")
         cfac = dataset.get("cfac", tuple())
         frac = dataset.get("frac", 1)
+        target_info = dataset.get("target_info", {})
         if not isinstance(frac, numbers.Real):
             raise ConfigError(f"'frac' must be a number, not '{frac}'")
         if frac < 0 or frac > 1:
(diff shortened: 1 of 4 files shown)


Zaharid:
I think even the temprary solution should go into the PLOTTING files rather than complicating dataset object. Also, why is the dict a string?

scarlehoff:
> I think even the temprary solution should go into the PLOTTING files 

No please, I need to come back to the "new commondata" and would strongly prefer if nothing changes there.
I'd rather have copies of the datasets with a different name like we do with POS INTEG as a temporary solution.

Radonirinaunimi:
> I think even the temprary solution should go into the PLOTTING files rather than complicating dataset object. Also, why is the dict a string?

This has to do with the checks failing if the value is not `hashable` but this can be fixed.

> No please, I need to come back to the "new commondata" and would strongly prefer if nothing changes there.
I'd rather have copies of the datasets with a different name like we do with POS INTEG as a temporary solution.

Another hacky option is to manually add these information to the non-free proton FK tables, after all that's where these information will be stored once they are propagated through pineappl. In this case, the only modifications required will be in `fkparser.`

scarlehoff:
> Another hacky option is to manually add these information to the non-free proton FK tables, after all that's where these information will be stored once they are propagated through pineappl. In this case, the only modifications required will be in fkparser.

This would be better. Although that means that we have to create a new (temporary) theory in which the information is in the fktables metadata. 
Eventually that information (maybe in some other format) will be in the fktables right? So I guess the new theory is hacky but the code will be close to what the final code will look like.

Radonirinaunimi:
> This would be better. Although that means that we have to create a new (temporary) theory in which the information is in the fktables metadata. Eventually that information (maybe in some other format) will be in the fktables right? So I guess the new theory is hacky but the code will be close to what the final code will look like.

Yes, exactly, the advantage of this is that the code will look like the final version. I will explore this now.

Radonirinaunimi:
Closing this as it will be superseded by https://github.com/NNPDF/fktables/issues/34#issuecomment-1324980967

PR #1636, opened 2022-11-25
Train on 1-point datasets

### LINKED ISSUE #1634: One point datasets always end up in validation split
When the dataset has one data point and a `frac` < 1 the code here

https://github.com/NNPDF/nnpdf/blob/43e7e752f04852bd9f2b458e6d03fb088586e762/validphys2/src/validphys/n3fit_data.py#L95

rounds the number of training points to zero, and so the dataset end in validation for all replicas. This may cause some total cross sections like  ATLASTTBARTOT7TEV  to end up with a somewhat worse chi2.

scarlehoff:
closes #1634 

(creating the pr just to send the bot)


RoyStegeman:
In that case, why not just always do this? e.g. 
`(np.random.rand() < ndat*frac-int(ndat*frac)) + int(ndat*frac)`

validphys2/src/validphys/n3fit_data.py
```diff
@@ -92,7 +92,10 @@ def tr_masks(data, replica_trvlseed):
         cuts = dataset.cuts
         ndata = len(cuts.load()) if cuts else dataset.commondata.ndata
         frac = dataset.frac
-        trmax = int(frac * ndata)
+        if ndata == 1:
+            trmax = int(np.random.rand() < 0.75)
+        else:
+            trmax = int(frac * ndata)
```
scarlehoff, line 98:
```suggestion
        trmax = int(frac*ndata)
        if ndata == 1:
            trmax = int(np.random.rand() <= frac)
```

I'll merge it after the bot is done. Let's see what happens to the two 1-point datasets.

Why would you want to do it for all datasets @RoyStegeman ?

RoyStegeman: Why not? I know it probably won't make a difference in practice, but at least we'd be consistent in the treatment of all dataset, and technically it's what we should do anyway to really get the correct fraction.

scarlehoff:
https://vp.nnpdf.science/sq18cmllT9-9udqY5Nb2sg==

validphys2/src/validphys/n3fit_data.py
```diff
@@ -92,7 +92,9 @@ def tr_masks(data, replica_trvlseed):
         cuts = dataset.cuts
         ndata = len(cuts.load()) if cuts else dataset.commondata.ndata
         frac = dataset.frac
-        trmax = int(frac * ndata)
+        trmax = int(frac*ndata)
+        if ndata == 1:
+            trmax = int(np.random.rand() <= frac)
```
Zaharid, line 97: Can we use math.ceil everywhere instead?

scarlehoff: If we do then the 1-point datasets will always go to training though.

RoyStegeman:
Sorry, but why not do this for all datasets?

scarlehoff:
> Sorry, but why not do this for all datasets?

You mean doing `mask = np.random.rand(ndata) < frac`? If so the reason for using shuffle instead is to ensure a 75% for all datasets, otherwise you can potentially have times where by change all points are masked away.

Not that it would be wrong (at some point I even wanted to apply the mask per dataset but that was not received well), but that's the reason.

RoyStegeman:
No I mean the same proposal I made before. For `ndat==1` you now make an exception such that there there is a probability of either including it or not and as more replicas are generated the fraction will approach whatever you put (so 0.75 in this case). Why not do this for all datasets instead of always rounding down? It's weird to have an `if` statement while the best thing to do would be to treat all datasets the same.

validphys2/src/validphys/n3fit_data.py
scarlehoff, line 97:
```suggestion
    trmax = (np.random.rand() < ndat * frac - int(ndat * frac)) + int(ndat * frac)
```

scarlehoff:
Is your worry having a fraction such that also ndata > 1 will suffer from the same problem?

(because other than that I don't really see the difference of the new line with the old one, you are implicitly doing the if inside the parenthesis but with `int(ndata*frac)` instead -which is more correct than just doing it for ndata-).

In any case I've commited your version since the previous one was only correct for `frac > 0.5`.

validphys2/src/validphys/n3fit_data.py
```diff
@@ -92,7 +92,7 @@ def tr_masks(data, replica_trvlseed):
         cuts = dataset.cuts
         ndata = len(cuts.load()) if cuts else dataset.commondata.ndata
         frac = dataset.frac
-        trmax = int(frac * ndata)
+    trmax = (np.random.rand() < ndat * frac - int(ndat * frac)) + int(ndat * frac)
```
scarlehoff, line 95:
```suggestion
        trmax = (np.random.rand() < ndat * frac - int(ndat * frac)) + int(ndat * frac)
```

RoyStegeman:
> Is your worry having a fraction such that also ndata > 1 will suffer from the same problem?

That is one example indeed, don't see when we would go to very low training fractions but you never know. Alsoe, even in general for datasets of e.g. 10 points it's better to have 7 training points 75% of the time and 8 training points the other 25% rather than always having 7 (or at least that's closer to what we say/intend to do). 

It's not a big complaint or anything - I never bothered to change it before - but I think this way of calculating trmax is more correct for all datasets, so I just think the exception for ndat==1 is a bit strange. 

It's not really for any practical reason, if that's what you're wondering

scarlehoff:
> It's not really for any practical reason, if that's what you're wondering

It was indeed, yes.

validphys2/src/validphys/n3fit_data.py
```diff
@@ -92,7 +92,7 @@ def tr_masks(data, replica_trvlseed):
         cuts = dataset.cuts
         ndata = len(cuts.load()) if cuts else dataset.commondata.ndata
         frac = dataset.frac
-        trmax = int(frac * ndata)
+        trmax = (np.random.rand() < ndat * frac - int(ndat * frac)) + int(ndat * frac)
```
scarlehoff, line 95:
```suggestion
        trmax = (np.random.rand() < ndata * frac - int(ndata * frac)) + int(ndata * frac)
```

ups. Also, now there is the problem that the regression tests (should) fail but oh well. I'll update them next week

RoyStegeman:
Is this ready to be merged?

scarlehoff:
Yes, but regressions tests and probably data rebuilding etc will need to be regenerated since we are effecticely changing the random seed.

validphys2/src/validphys/n3fit_data.py
Zaharid: Is that a bad thing?

scarlehoff: Yes, in the sense that it is not what we say we do.

Zaharid (changes requested):
I think we should discuss this a bit. As it is now, it is a change in behaviour. We used to have a fixed number of datapoints for every dataset in either split and that makes it variable. Even if the change was fine, we should be sure the splits are reproducible.

Zaharid:
No it isn’t 

> They should be reproducible, as the seed will affect the same.

No it isn’t: the call to np.random.rand uses the global, unitialized, uncontrolled random state.

scarlehoff:
Are you sure? I think `random.seed(x)` affects `random.rand()` as well.

Zaharid:
Uh, sorry, I assumed we already had done the todo saying to use random generators rather than dealing with the global state elsewhere.

Zaharid:
As an aside, I find how this https://jax.readthedocs.io/en/latest/jax.random.html (http://www.thesalmons.org/john/random123/papers/random123sc11.pdf) works pretty interesting.

scarlehoff:
Ah, no.
Maybe we should given that we are already dealing with this.

Then, the two possibilities are

- Everything gets a 0.75 chance.
- We isolate the cases in which that equals 0, in those special cases `trmax` is manually set to 1 if rand() > 0.75 (closer to current implementation)

In both cases changing to a generator is a possibility.

I don't have any strong opinion but would default to the first one since I already have some of the regressions fits ready (tomorrow I'll check why the tests didn't work in the ci... maybe I forgot vp-upload...)

validphys2/src/validphys/n3fit_data.py
```diff
@@ -84,19 +84,23 @@ def tr_masks(data, replica_trvlseed):
...
         # all data if cuts are None
         cuts = dataset.cuts
         ndata = len(cuts.load()) if cuts else dataset.commondata.ndata
         frac = dataset.frac
-        trmax = int(frac * ndata)
+        # We do this so that a given dataset will always have the same number of points masked
+        trmax = int(ndata*frac)
+        if trmax == 0:
+            # If that number is 0, then get 1 point with probability frac
+            trmax = rng.random() < frac
```
scarlehoff, line 99:
```suggestion
            trmax = int(rng.random() < frac)
```

RoyStegeman:
I mean, I still don't understand what makes the situation of trmax==0 so special, but since it's won't affect the outcome I'm not going to push this further. If it has been decided within the collaboration (where I would be surprised if anyone even cared), and you and Zahari are happy, then I'm also fine.

Zaharid:
Personally I still slightly prefer my thing, mostly owing to the aesthetics of not having special cases, but similarly am not too sure it matters enough anyway. I guess the remaining concern now is that this will break the rebuilding of pseudodata which various things rely on.

scarlehoff:
What is your thing? This is your thing (but not singling out ndata == 1 which would be wrong).

scarlehoff:
The last commit is an attempt to have a bit more stability. From the numpy docs, compatibility between versions is not guaranteed but I hope that algorithm in particular is stable enough (since it right now their default).

Zaharid:
Hmm. I am going to want to version tag this…

scarlehoff:
Feel free. From 4.0.4 we have had a roller coaster of breaking backward compatibility changes... but this should've stopped now (famous last words)

Zaharid:
Which are these changes?

scarlehoff:
I'm not sure but I remember having to update regression tests more than once.

Issue #1622, opened 2022-10-28
luxqed recipe


scarrazza:
In order to obtain a NNPDF fit using luxqed one needs the following:

**Requirements**
- [ ] provide QED evolution with [eko](https://github.com/N3PDF/eko/).
- [x] provide the photon PDF using [fiatlux](https://github.com/scarrazza/fiatlux).
- [ ] provide F2, FL and F2LO contributions for a given input PDF set from [yadism](https://github.com/N3PDF/yadism).

**Algorithm (simplified version)**
1. take an input QCD fit and generate the photon PDF using fiatlux, at large Q value.
2. down-evolve the photon obtained in 1 to Q=1GeV
3. use this photon as fixed parton in a new PDF fit, it should enters all FK tables
4. repeat 1 until the photon shape has converged.
5. care about the extra uncertainties identified by luxqed.

**Missing software requirements**
- [x] fiatlux in python
- [ ] n3fit:
  - load the photon obtained in 2, e.g. using the T0 format, see [here for loading](https://github.com/NNPDF/nnpdf/blob/master/nnpdfcpp/src/nnfit/src/fitbases.cc#L963) and [here for the convolution](https://github.com/NNPDF/nnpdf/blob/master/nnpdfcpp/src/nnfit/src/fitbases.cc#L1166).
- [ ] validphys:
  - implement the T0/photon generation by integrating fiatlux/yadism, i.e. by following all the steps [summarized here](https://github.com/NNPDF/nnpdf/blob/master/nnpdfcpp/src/fiatlux/src/fiatlux.cc).

AleCandido:
I would say that things now have to move in parallel. We should use some pseudo-structure functions, and make it working also in the fit.

In the meanwhile, exact (EXA) QED evolution is working, but slow, @niclaurenti is taking care of it. Concerning YADISM, we should just choose how many points we want, and prepare a runcard. Then we compute, and finally we plug interpolation.
2D spline interpolation is available also in [`scipy.interpolate`](https://scipy.github.io/devdocs/reference/interpolate.html#id1), so we just need to obtain the numbers, and plug them in.

scarlehoff:
For the `n3fit` part, it should be close to trivial, as the fit itself is completely agnostic to the number of flavours and it is the fktable who decides which flavours is convoluted with what.

Currently the photon in the PDF is set to 0, this needs to be instead changed to a value-per-x

[...]

PR #1644, opened 2022-12-03
Separate xgrid and model generation within model trainer

scarlehoff:
Right now the xgrid that is used as input for the PDF (which is a concatenation of all the points in x the fktables ask for) was being prepared just before the observable generation.

There was no reason for that, it just _happened_ to be there. For #1622 we need to generate the photon in all those `x`-points. While technically it doesn't matter where luxqed is called, I think it is better to call it from the top(est?) possible level. In this case, just after the grid of x-points is prepared.

### DIFF
--- a/n3fit/src/n3fit/model_trainer.py
+++ b/n3fit/src/n3fit/model_trainer.py
@@ -33,7 +33,8 @@ PUSH_POSITIVITY_EACH = 100
 # Each how many epochs do we increase the integrability Lagrange Multiplier
 PUSH_INTEGRABILITY_EACH = 100
 
-InputInfo = namedtuple("InputInfo", ["input_l", "split_l", "input_idx"])
+# See ModelTrainer::_xgrid_generation for the definition of each field and how they are generated
+InputInfo = namedtuple("InputInfo", ["input", "split", "idx"])
 
 def _pdf_injection(pdf_layers, observables, masks):
     """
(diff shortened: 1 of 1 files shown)


n3fit/src/n3fit/model_trainer.py
```diff
@@ -32,6 +33,8 @@
 # Each how many epochs do we increase the integrability Lagrange Multiplier
 PUSH_INTEGRABILITY_EACH = 100
 
+# Create a tuple for the input information
+InputInfo = namedtuple("InputInfo", ["input_l", "split_l", "input_idx"])
```
Zaharid, line 37: I find the description of these fields rather unclear. I end up having to read the code to understand what it does.

n3fit/src/n3fit/model_trainer.py
```diff
@@ -289,7 +292,60 @@ def _fill_the_dictionaries(self):
...
+
+        Detailed implementation of the union of xgrids:
+            let's assume an input [x1, x1, x1, x2, x2, x3]
+            where each xi is a different grid, this will be broken into two lists:
+            [x1, x2, x3] (unique grids) and [0,0,0,1,1,2] (index of the grid per dataset)
+            The pdf will then be evaluated to concatenate([x1,x2,x3]) and then split (x1, x2, x3)
+            Then each of the experiment, looking at the indexes, will receive one of the 3 PDFs
+            The decision whether two grids (x1 and x1) are really the same is decided below
+
+        The necessary information to redistribute the x-grid is held by a ``InputInfo`` tuple
```
Zaharid, line 315: Perhaps we could have a Returns field here explaining each element of the tuple?

n3fit/src/n3fit/model_trainer.py
```diff
@@ -32,6 +33,8 @@
 # Each how many epochs do we increase the integrability Lagrange Multiplier
 PUSH_INTEGRABILITY_EACH = 100
 
+# Create a tuple for the input information
```
Zaharid, line 36:
```suggestion
```

Zaharid:
Should this be cheched by the fitbot?

scarlehoff:
It is actually just a reordering of the code with absolute no changes, but it is "free" so... why not.

Zaharid:
LGTM.

RoyStegeman (approved):

PR #1645, opened 2022-12-05
Alphas determination

enocera:
This PR serves the purposes of making fits with a covariance matrix computed from alphas variations.


enocera:
@RoyStegeman This is in preparation of the chat later today.

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -363,7 +370,9 @@ def covs_pt_prescrip(
             deltas1 = list((other - central1 for other in others1))
             central2, *others2 = process_info.theory[name2]
             deltas2 = list((other - central2 for other in others2))
-            if l == 3:
+            if point_prescription in ["alpha_s", "alpha_s_reduced", "alpha_s_extended"]:
+                s = covmat_alphas(name1, name2, deltas1, deltas2)
```
RoyStegeman, line 374: I guess this should go inside l==3 as is aldo done in theorycovarianceutils.py. 

I understand you probably quickly coded this up and we may change this PR anyway, but it's just something I noticed and wanted to make a note of

RoyStegeman (approved):
I'm happy to merge this if you are

scarlehoff:
Out of curiosity, what is the effect of this `alpha_s` covmat with respect to a covmat of `muR` variations? (do we have examples?)
I would expect it to be negligible since the alpha_s variations are in general much smaller than the change on alpha_s due to muR variations.

enocera:
The effect is indeed negligible. The aim of this PR is not to include the alphas cov matrix to see how PDF uncertainties increase and/or how the fit quality improves. The aim is to determine alphas through correlations between a fit that includes an alphas cov matrix and alphas as a prediction, following the reasoning in  https://inspirehep.net/literature/1862837.

Zaharid:
Could anyone please summarize what is this about?

enocera:
@Zaharid Richard has asked @RoyStegeman and myself to study correlations between theoretical uncertainties due to alphas incorporated in PDF fits, and the theoretical uncertainties in the prediction for alphas made using these PDFs. The reasoning follows the paper by Richard and Rosalyn https://arxiv.org/pdf/2105.05114.pdf. Richard's idea is to see if one can use correlations to make a prediction for alphas. I understand that the idea is a semi-analytical Gaussian version of the correlated replica method. In comparison to the correlated replica method, the disadvantage is the Gaussian hypothesis, which may invalidate the reliability of Richard's idea; the advantage is a lighter computational burden. Whether Richard's method is competitive or not should be investigated. But we cannot investigate it without this PR. Now, @Zaharid , you may consider that this PR should only be merged if the method works, otherwise it can stay in a branch.

andreab1997:
As we discussed during the code meeting, in this case we have certain point prescriptions that are not compatible with all the theoryids than you can use (so either 200 or 205). This is because if you use 205 you should use either `alpha_s_0119` or `alpha_s_extended_0119`, as far as I understood. So in this case the test `check_correct_theory_combination_internal` could be useful to check if you are loading the correct theories for the point prescription you chose. Another possibility could be to write another test that just checks if the chosen point prescription is compatible with the theoryid. There is also another possibility that is to write the point prescription in an "agnostic" way (with respect to the central theoryid). In fact, you are always vary the central by the same factor, regardless the actual central value of alpha_s. This last solution is probably the best but I am not sure how easy is to implement in practice. Let me know what you think @enocera @RoyStegeman

RoyStegeman:
I'd like to merge this, if covmat options need to be expanded that can be done in some other PR (or directly in master)

Issue #1646, opened 2022-12-08
Closure tests use t0pdfset as underlying law


Zaharid:
The closure test code is supposed to read from a key called `fakepdf` as documented here

https://docs.nnpdf.science/tutorials/closuretest.html?highlight=closure+tests#brief-background

however when the way the code actually works is by setting the t0 pdf as underlying law. That is done here:

https://github.com/NNPDF/nnpdf/blob/74f9413d7ef59bf75511a750ea543138aedd1548/validphys2/src/validphys/filters.py#L123

this was probably not noticed as the two things are usually set to be the same, but does cause surprising results when not, and may have affected some studies.

Zaharid:
cc @ldd69 @tgiani

Zaharid:
@RoyStegeman @scarlehoff  would you look into this?

RoyStegeman:
Thanks for noticing. Indeed there is not really any reason for the fakepdf to be different from the t0pdf, I imagine Michael implemented it because he needed it at some point. 

Is this an option we want to keep?

Zaharid:
The separation was  explicitly added at some point (with some leftover apparently). The docs have to say

> Setting fakedata to True will cause closure test pseudodata to be generated and subsequently fitted. The PDf which the pseudodata will be generated from is specified by the fakepdf key. It is strongly advised to set the fakepdf and t0pdfset, found under datacuts to be the same PDF, unless specifically testing the impact of the t0 procedure.

RoyStegeman:
Yes I read that, I was wondering wether we still want to keep it since I can't immediatly think of a use-case (though Michael undoutebly had one)

scarlehoff:
> Yes I read that, I was wondering wether we still want to keep it since I can't immediatly think of what exactly that would teach us about the t0 procedure (though Michael undoutebly had some idea).

By having t0 = underlying law then the t0-chi2 is perfect and unbiased without the need for iterations. So I understand the motivation for keeping it the same. But if we keep both keys, each should do the thing it says it does.

Zaharid:
I think this issue is about bringing the behaviour in line with docs and runcards, and we should do that. I think it is reasonable to want separate toggles for these things (even thought perhaps it should emit a warning when they are different). For one, strictly speaking, using the same pdfs is "cheating" as some information would be passed that we don't have in a normal fit (which in practice is insignificant).

PR #1647, opened 2022-12-09
bufix: use fakepdf to generate pseudodata instead of t0pdfset

RoyStegeman:
addresses #1646


Zaharid:
I guess the image errors are unrelated? Would be good to look at what they are.

RoyStegeman:
Yes locally they pass

RoyStegeman:
It also fails on a commit that previously passed: https://github.com/NNPDF/nnpdf/actions/runs/3631739847/jobs/6183718848

I guess that in the new version of matplotlib something changed

PR #1649, opened 2022-12-13
fix default elements for ccbar_asymm basis

tgiani:
```V15``` was missing from the ```defualts_elemets``` of the ```CCBAR_ASYMM``` and this was breaking ```vp-nextfitruncard```

### DIFF
--- a/validphys2/src/validphys/pdfbases.py
+++ b/validphys2/src/validphys/pdfbases.py
@@ -4,6 +4,7 @@ pdfbases.py
 This holds the concrete labels data relative to the PDF bases,
 as declaratively as possible.
 """
+import copy
 import inspect
 import functools
 import abc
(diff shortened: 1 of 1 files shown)


tgiani:
I think it s probably better to redefine the basis altogether? I mean without doing ```CCBAR_ASYMM = evolution``` ?

scarlehoff:
I think this would be good to be merged (and since we are discussing ccbar I'd feel better if we are all in the same page -or branch-)

@Zaharid what kind of tests would you add?

RoyStegeman:
Yes this should be merged, don't want to keep the current bugged (buggish?) implementation in main.

Also wondering about what tests you had in mind. There's not really a clean way to test it, unless we want to do so for all bases explicitly just to prevent us from doing something like this again. Perhaps instead we should just remove `EVOLUTION = evolution` (and instead do `EVOL = LinearBasis....` ) to make it less likely to accidentaly intruduce bugs like this, or we can also add `EVOL = copy.deepcopy(evolution)`, mostly code is just copy-paste with a minor edit anyway.

RoyStegeman (approved):

PR #1650, opened 2022-12-15
Use python `CommonData`

scarlehoff:
There has been a half-finished, half-used `python` version of the `CommonData` object for a while https://github.com/NNPDF/nnpdf/blob/9d57bb0df15b7269d00d7d56ad3cba63b966c9b0/validphys2/src/validphys/coredata.py#L202

This PR completes so that it can be used everywhere.

This still doesn't allow to remove `libNNPDF` because the generation of closure test data is still done from `libNNPDF`. Is there someone in Cambridge dealing with that at the moment @Zaharid ?

If not I could also have a go at that after this is merged.

Note that I have not attempted to fix any possible design problems there could be with the python version of the `CommonData` or with how the `libNNPDF` was used. This is just a conversion so that when we move to the new commondata I only have to port one type of `CommonData` instead of two.

#### TODO

- [x] Check all validphys runcards in the example folder to check that I have not missed some corner action not covered by tests.

### DIFF
--- a/validphys2/src/validphys/commondataparser.py
+++ b/validphys2/src/validphys/commondataparser.py
@@ -6,7 +6,7 @@ interfaces with common Python libraries.
 
 The validphys commondata structure is an instance of :py:class:`validphys.coredata.CommonData`
 """
-from collections import namedtuple
+import dataclasses
 from operator import attrgetter
 import logging
 
@@ -16,7 +16,7 @@ from validphys.coredata import CommonData
 
 log = logging.getLogger(__name__)
 
-kinlabels_latex = {
+KINLABEL_LATEX = {
     "DIJET": ("\\eta", "$\\m_{1,2} (GeV)", "$\\sqrt{s} (GeV)"),
     "DIS": ("$x$", "$Q^2 (GeV^2)$", "$y$"),
     "DYP": ("$y$", "$M^2 (GeV^2)$", "$\\sqrt{s} (GeV)$"),
@@ -41,9 +41,6 @@ kinlabels_latex = {
     "PHT": ("$\\eta_\\gamma$", "$E_{T,\\gamma}^2 (GeV^2)$", "$\\sqrt{s} (GeV)$"),
     "SIA": ("$z$", "$Q^2 (GeV^2)$", "$y$"),
 }
-
-
-_kinlabels_keys = sorted(kinlabels_latex, key=len, reverse=True)
 
 
 def load_commondata(spec):
@@ -124,12 +121,18 @@ def parse_systypes(systypefile):
     return systypetable
 
 
-CommonDataMetadata = namedtuple("CommonDataMetadata", ("name", "nsys", "ndata", "process_type"))
+@dataclasses.dataclass
+class CommonDataMetadata:
+    """Contains metadata information about the data being read"""
+    name: str
+    nsys: int
+    ndata: int
+    process_type: str
 
 
 def peek_commondata_metadata(commondatafilename):
-    """Check some basic properties commondata object without going though the
-    trouble of processing it on the C++ side"""
+    """Read some of the properties of the commondata object as a CommonData Metadata
+    """
     with open(commondatafilename) as f:
         try:
             l = f.readline()
@@ -149,15 +152,17 @@ def get_plot_kinlabels(commondata):
     """Return the LaTex kinematic labels for a given Commondata"""
     key = commondata.process_type
 
-    return kinlabels_latex[key]
+    return KINLABEL_LATEX[key]
 
 
 def get_kinlabel_key(process_label):
-    # Since there is no 1:1 correspondence between latex keys and GetProc,
-    # we match the longest key such that the proc label starts with it.
+    """
+    Since there is no 1:1 correspondence between latex keys and GetProc,
+    we match the longest key such that the proc label starts with it.
+    """
     l = process_label
     try:
-        return next(k for k in _kinlabels_keys if l.startswith(k))
+        return next(k for k in sorted(KINLABEL_LATEX, key=len, reverse=True) if l.startswith(k))
     except StopIteration as e:
         raise ValueError(
             "Could not find a set of kinematic "
--- a/validphys2/src/validphys/commondataparser.py
+++ b/validphys2/src/validphys/commondataparser.py
@@ -2,7 +2,7 @@
 This module implements parsers for commondata  and systype files into useful
 datastructures, contained in the :py:mod:`validphys.coredata` module, which are
 not backed by C++ managed memory, and so they can be easily pickled and
-interfaces with common Python libraries. 
+interfaced with common Python libraries. 
 
 The validphys commondata structure is an instance of :py:class:`validphys.coredata.CommonData`
 """
--- a/validphys2/src/validphys/commondataparser.py
+++ b/validphys2/src/validphys/commondataparser.py
@@ -2,16 +2,49 @@
 This module implements parsers for commondata  and systype files into useful
 datastructures, contained in the :py:mod:`validphys.coredata` module, which are
 not backed by C++ managed memory, and so they can be easily pickled and
-interfaces with common Python libraries.  The integration of these objects into
-the codebase is currently work in progress, and at the moment this module
-serves as a proof of concept.
+interfaces with common Python libraries. 
+
+The validphys commondata structure is an instance of :py:class:`validphys.coredata.CommonData`
 """
+from collections import namedtuple
 from operator import attrgetter
+import logging
 
 import pandas as pd
 
-from validphys.core import peek_commondata_metadata
 from validphys.coredata import CommonData
+
+log = logging.getLogger(__name__)
+
+kinlabels_latex = {
+    "DIJET": ("\\eta", "$\\m_{1,2} (GeV)", "$\\sqrt{s} (GeV)"),
+    "DIS": ("$x$", "$Q^2 (GeV^2)$", "$y$"),
+    "DYP": ("$y$", "$M^2 (GeV^2)$", "$\\sqrt{s} (GeV)$"),
+    "EWJ_JPT": ("$p_T (GeV)$", "$M^2 (GeV^2)$", "$\\sqrt{s} (GeV)$"),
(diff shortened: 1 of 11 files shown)


Zaharid:
We are going to have a pr for the closure test data soon.

scarlehoff:
Good. I guess is this branch? https://github.com/NNPDF/nnpdf/compare/python_closure_sampling

If so there seem to be no overlap with the changes of this PR so when both are merged it should be possible to remove libNNPDF at last.

Zaharid:
Cc @comane

Zaharid:
Would be good if the export functions were more layered, providing ways to dump the individual files to any streams, mirroring what we have in commondataparser. Among other things, this help us to test easily that writing and reading  roundtrip. These lower level things might go in commondata parser.

scarlehoff:
I didn't check whether they were the same but are they not included in the #1651? Or those are yet something different which is closure test only?

Zaharid:
I guess that part does overlap.

scarlehoff:
Yes, but it is not destructive overlap. When that one is merged the `export` can be changed to use the two functions and done!

scarlehoff:
This is now ready for review. As far as I can test there's nothing broken.

Added @andreab1997 as a reviewer in case this touches the covmat in some way. I think it shouldn't because the thcovmat was freed from `libNNPDF`, but just in case.

scarlehoff:
@Zaharid did you have a look at this?

@andreab1997 please, have a look as well

scarlehoff:
With #1660 it should be possible to merge this to master?

scarlehoff:
I'd like to merge this by the end of the week if there are no blocking issues with it.

I'd create a tag before the merge as the last version that still uses `libNNPDF` in the python code since I expect we will be able to remove all `libNNPDF` imports from `validphys` after this one and #1651 (or #1660) are merged.

validphys2/src/validphys/coredata.py
```diff
@@ -353,3 +363,22 @@ def systematic_errors(self, central_values=None):
...
+        sys_path = path / "systypes" / f"SYSTYPE_{self.setname}_DEFAULT.dat"
+        sys_path.parent.mkdir(exist_ok=True)
+
+        dat_string_raw = self.commondata_table.to_string(index=False, header=False, float_format="{:.8e}".format)
+        header = f"{self.setname}    {self.nsys} {self.ndata}"
+        dat_string = "\n".join([f" {i+1}    {r}" for i, r in enumerate(dat_string_raw.split("\n"))])
+        dat_path.write_text(f"{header}\n{dat_string}\n")
+
+        sys_raw = self.systype_table.to_string(index=True, header=False, index_names=False)
+        sys_path.write_text(f"{self.nsys}\n{sys_raw}\n")
```
AleCandido, line 384: Same note about the encoding.

validphys2/src/validphys/filters.py
```diff
@@ -9,13 +9,39 @@
...
 import numpy as np
 
-from NNPDF import CommonData
 from reportengine.checks import make_argcheck, check, check_positive, make_check
 from reportengine.compat import yaml
 import validphys.cuts
 
 log = logging.getLogger(__name__)
 
+KIN_LABEL = {
```
AleCandido, line 18: I guess the plural would be more consistent with the content of `commondataparser.py` (btw it is also not separating by underscore - both fine, but I would pick just one).

validphys2/src/validphys/filters.py
```diff
@@ -488,7 +516,7 @@ def get_cuts_for_dataset(commondata, rules) -> list:
 
     Parameters
     ----------
-    commondata: NNPDF CommonData spec
+    commondata: :py:class:`validphys.coredata.CommonData`
```
AleCandido, line 519: Sphinx is inferring the `:py:class:` part here, if you wish you can simplify with
```suggestion
    commondata: validphys.coredata.CommonData
```

validphys2/src/validphys/core.py
```diff
@@ -472,7 +436,8 @@ def __init__(self, *, name, commondata, fkspecs, thspec, cuts,
 
     @functools.lru_cache()
     def load(self):
-        cd = self.commondata.load()
+        """Load the libNNPDF version of the dataset"""
+        cd = LegacyCommonData.ReadFile(str(self.commondata.datafile), str(self.commondata.sysfile))
```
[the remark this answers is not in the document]
scarlehoff: Yes. I think the last dependency will be dropped by #1651, but now after applying the changes I'll check whether a `load` is left anywhere else. Maybe it can be removed also from here.

validphys2/src/validphys/core.py
```diff
@@ -508,7 +473,9 @@ def load_commondata(self):
             loaded_cuts = self.cuts.load()
             if not (hasattr(loaded_cuts, '_full') and loaded_cuts._full):
                 intmask = [int(ele) for ele in loaded_cuts]
-                cd = CommonData(cd, intmask)
+#                 cd_old = LegacyCommonData(cd.old, intmask)
+                cd = cd.with_cuts(intmask)
+#                 cd.old = cd_old # DEBUG
```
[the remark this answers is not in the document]
scarlehoff: Oh, damn, this PR was in the freezer for so long that I forgot to remove them. Thanks for noticing :__
```suggestion
                cd = cd.with_cuts(intmask)
```

validphys2/src/validphys/coredata.py
```diff
@@ -284,10 +284,20 @@ def with_cuts(self, cuts):
         new_commondata_table = self.commondata_table.loc[cuts]
         return dataclasses.replace(self, ndata=newndata, commondata_table=new_commondata_table)
 
+    @property
+    def kinematics(self):
+        return self.commondata_table[KINNAMES]
+
+    def get_kintable(self):
+        return self.kinematics.values
```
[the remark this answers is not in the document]
scarlehoff: Both this (and the one below) are left as a method because they were used for the old commondata as methods (with those names).

I prefer to leave them like that while the `LegacyCommonData` still exist.

validphys2/src/validphys/coredata.py
```diff
@@ -353,3 +363,22 @@ def systematic_errors(self, central_values=None):
             central_values = self.central_values.to_numpy()
         converted_mult_errors = self.multiplicative_errors * central_values[:, np.newaxis] / 100
         return pd.concat((self.additive_errors, converted_mult_errors), axis=1)
+
+    def export(self, path):
```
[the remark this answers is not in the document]
scarlehoff: This is the one I was discussing with @Zaharid during the code meeting. This function will be removed by #1660 or #1651

(but yes, the name is for compatibility with the old one, it does exactly the same under the same method)

scarlehoff:
RE `LegacyCommonData`, while the fit is not using it, there are too many `load` scattered around the code. Once the pseudodata one is merged and I'm sure that no libNNPDF _should_ be use I'll remove all of them.

If then there are functions that are just there and not being used (it wouldn't be the first time) I'll just remove them.

validphys2/src/validphys/commondataparser.py
```diff
@@ -70,22 +100,72 @@ def parse_commondata(commondatafile, systypefile, setname):
...
     # Some datasets e.g. CMSWCHARMRAT have no systematics
     except pd.errors.EmptyDataError:
         systypetable = pd.DataFrame(columns=systypeheader)
 
     systypetable.set_index("sys_index", inplace=True)
 
     return systypetable
+
+
+@dataclasses.dataclass
```
Zaharid, line 124: These should be frozen by default. We may want to be caching based on this a.t some point

scarlehoff:
```suggestion
@dataclasses.dataclass(frozen=True)
```

Zaharid:
I am inclined to merge this as is, provided various tests pass.

validphys2/src/validphys/commondataparser.py
```diff
@@ -2,17 +2,47 @@
 This module implements parsers for commondata  and systype files into useful
 datastructures, contained in the :py:mod:`validphys.coredata` module, which are
 not backed by C++ managed memory, and so they can be easily pickled and
-interfaces with common Python libraries.  The integration of these objects into
-the codebase is currently work in progress, and at the moment this module
-serves as a proof of concept.
+interfaces with common Python libraries.
```
RoyStegeman, line 5:
```suggestion
interfaced with common Python libraries. 
```

validphys2/src/validphys/dataplots.py
```diff
@@ -960,12 +960,13 @@ def plot_positivity(pdfs, positivity_predictions_for_pdfs, posdataset, pos_use_k
...
 
     posset = posdataset.load_commondata()
-    ndata  = posset.GetNData()
+    ndata  = posset.ndata
     xvals = []
 
     if pos_use_kin:
-        ax.set_xlabel('kin1')
-        xvals = [posset.GetKinematics(i, 0) for i in range(0, ndata)]
+        kin_name = "kin1"
```
RoyStegeman, line 967: why not do `KINNAMES[0]"? Hardcoding it in one place but sharing/using KINNAMES in another place is a bit confusing.

AleCandido: I agree, and hard-coded strings make it also harder to refactor. I plainly missed this.

validphys2/src/validphys/coredata.py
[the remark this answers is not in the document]
scarlehoff: it is already removed in a different PR, don't want to have to update #1660  because there is a conflict :P

validphys2/src/validphys/dataplots.py
scarlehoff: > why not do `KINNAMES[0]"?

Because I wasn't really looking at the strings but rather at removing the C++ methods. I've changed this together with the `_` suggestion.

RoyStegeman: I'll stop phrasing my comments as questions ;)

RoyStegeman (approved):

AleCandido (approved):

Issue #1623, opened 2022-10-28
Extra parameters in closure test runcards


Zaharid:
The closure tests depend on the prepare_nnpdf_rng function here 
https://github.com/NNPDF/nnpdf/blob/94b80c35618bd55ae463b5806bf2b5886f6ecad6/validphys2/src/validphys/filters.py#L80

which uses the c++ rng. As far as I remember this was kept so that we could use older closure tests while analyzing other parts of the code. This doen't make a lot of sense any longer. Not only are we keeping a dependency on the old code, risking having the same problem when people produce new batches of closure tests, but we end up with required keys that are not documented, e.g. here

https://docs.nnpdf.science/tutorials/closuretest.html

We should get rid of that function and use the same rng as for nfit.

comane:
@Zaharid, I think this issue can be closed since in the new python closure test branch the function
`prepare_nnpdf_rng(filterseed:int, rngalgo:int, seed:int): ` has been removed

PR #1651, opened 2022-12-15
Python closure sampling

comane:
Main modification: Level 1 data used for closure tests generated by validphys.pseudodata.make_replica instead of libnnpdf.experiments.MakeClosure.
Closure tests no longer depend on the prepare_nnpdf_rng function which uses the c++ rng (see https://github.com/NNPDF/nnpdf/issues/1623 ).  As a consequence "seed" and "rngalgo" are no longer needed in a closure test runcard, the random initializer only depends on "filterseed"


validphys2/src/validphys/commondataparser.py
```diff
@@ -89,3 +92,87 @@ def parse_systypes(systypefile):
...
+        # path
+        path = str(filter_path) + f'/{commondata_instance.setname}'
+        path_data = str(path) + f"/DATA_{commondata_instance.setname}.dat"
+        commondata_tab = commondata_instance.commondata_table.reset_index(drop = True) # do not use maskcut index
+        commondata_tab.index += 1 # index starting from 1
+        
+        header = f"{commondata_instance.setname} {commondata_instance.nsys} {commondata_instance.ndata}\n"
+        
+        #==== write DATA =====#
+        with open(path_data, "w+") as f:
```
Zaharid, line 124: Would be good to have these this as a separate function, e.g. `def write_commondata_data(commondata, buffer)` that could be used as

```
from io import StringIO

sio = StrinIO()
write_commondata_data(commondata, sio)
print(sio.get_value()) # Contents of a commondata file
```

also, you don't want to have the `"w+"` mode, but the `"w"` one
    
```

File: validphys2/src/validphys/commondataparser.py [hunk 2]
```diff
@@ -89,3 +92,87 @@ def parse_systypes(systypefile):
...
+def write_commondata(commondata_list, filter_path):
+    """
+    GENERAL DESCRIPTION:
+
+    writes dataset data in filter folder using the commondata file format
+
+    Parameters
+    ----------
+
+    commondata_list : list
```
Zaharid, line 105: This doesn't need to be a list, but should take a single instance.

validphys2/src/validphys/filters.py
```diff
@@ -168,24 +171,88 @@ def _filter_real_data(filter_path, data):
...
         nfull, ncut = _write_ds_cut_data(path, dataset)
         total_data_points += nfull
         total_cut_data_points += ncut
+        # Rescale errors 
         loaded_ds = loaded_data.GetSet(j)
         if errorsize != 1.0:
             loaded_ds.RescaleErrors(errorsize)
-        loaded_ds.Export(str(path))
+    
+    from validphys.commondataparser import write_commondata, write_systype
```
scarlehoff, line 240: Does this create a circular import?

validphys2/src/validphys/commondataparser.py
```diff
@@ -89,3 +92,87 @@ def parse_systypes(systypefile):
...
+
+    filter_path : str
+                 path to filter folder
+
+    
+    """
+
+    for commondata_instance in commondata_list: 
+        path = filter_path / commondata_instance.setname / 'systypes'
+        make_systype_dir(path)
```
scarlehoff, line 170: If it is truly necessary I think it is better to do here the check for existence. Having a function just to create a folder is adding too many moving parts.

validphys2/src/validphys/commondataparser.py
```diff
@@ -89,3 +92,87 @@ def parse_systypes(systypefile):
...
+        path_data = str(path) + f"/SYSTYPES_{commondata_instance.setname}_DEFAULT.dat"
+        systype_tab = commondata_instance.systype_table
+        header = f"{len(systype_tab.index)}\n"
```
scarlehoff, line 173: If this is an instance of `coredata:CommonData` it includes already a property with the `nsys` value
```suggestion
        header = f"{commondata_instance.nsys}\n"
```

validphys2/src/validphys/commondataparser.py
```diff
@@ -89,3 +92,87 @@ def parse_systypes(systypefile):
...
+        
+        #==== write DATA =====#
+        with open(path_data, "w+") as f:
+            f.write(header)
+            systype_tab.to_csv(f, sep="\t", header=None)
```
scarlehoff, line 178:
```suggestion
            systype_tab.to_csv(f, sep="\t", header=None)
```

Is this the exact format that was used before? (I used `to_string` instead of `to_csv`, but I had to add `index_names=False` to make it truly equivalent).

validphys2/src/validphys/tests/test_pseudodata.py
```diff
@@ -80,3 +81,22 @@ def test_read_matches_recreate():
...
         pd.testing.assert_index_equal(read.val_idx, recreate.val_idx, check_order=False)
+
+
+def test_make_level0_data():
+    from validphys.loader import Loader
+    from validphys.covmats import dataset_t0_predictions
+
+    dataset='NMC'
+    pdfname='NNPDF40_nnlo_as_01180'
+    theoryid=162
```
scarlehoff, line 92: You can take that information from the `conftest.py` (if there's already a set that contains that, otherwise you can create it)

validphys2/src/validphys/filters.py
Zaharid: It rather prevents it by being inside a function.

comane: It does indeed create a circular import error if imported in the header of the module

validphys2/src/validphys/commondataparser.py
comane: Ok, I am not sure I completely understand what you mean. Would you rather prefer to have something like:

```
def write_commondata_pd(commondata,buffer):
       commondata_table = commondata.commondata_table
       commondata_table.to_csv(buffer, sep = "\t",header=None)
```

```
def write_commondata_to_file(commondata,path):
    from io import StringIO
    
    sio = StringIO()
    write_commondata_pd(commondata,sio)
    
    header = f"{commondata.setname} {commondata.nsys} {commondata.ndata}\n"
    
    with open(path,"w") as file:
        file.write(header)
        file.write(sio.getvalue())
```

Zaharid: I’d like something that  writes the whole commondata file to an arbitrary buffer that could be an open file, a stringio object or anything else.

Zaharid:
I think it should be like

```
def write_commondata_data(commondata, buffer):
    header = f"{commondata.setname} {commondata.nsys} {commondata.ndata}\n"
    buffer.write(header)
    commondata.table.to_csv(buffer, sep="\t", header=None)
```

Then we may (or not) have a simple function like

```
def write_commondata_to_file(commondata, path):
    with(open(path, "w")) as f:
        write_commondata_data(commondata, f)
```

or else have the equivalent functionality in the export function or whatever.

Zaharid:
I think it should be like

```
def write_commondata_data(commondata, buffer):
    header = f"{commondata.setname} {commondata.nsys} {commondata.ndata}\n"
    buffer.write(header)
    commondata.table.to_csv(buffer, sep="\t", header=None)
```

Then we may (or not) have a simple function like

```
def write_commondata_to_file(commondata, path):
    with(open(path, "w")) as f:
        write_commondata_data(commondata, f)
```

or else have the equivalent functionality in the export function or whatever.

validphys2/src/validphys/filters.py
```diff
@@ -168,24 +171,91 @@ def _filter_real_data(filter_path, data):
     return total_data_points, total_cut_data_points
 
 
-def _filter_closure_data(filter_path, data, fakepdf, fakenoise, errorsize):
-    """Filter closure test data."""
+def _filter_closure_data(filter_path, data, fakepdf, fakenoise, filterseed, errorsize):
+    """
+    GENERAL DESCRIPTION:
+
```
Zaharid, line 177:
```suggestion
```

I'd remove these: They don't follow the exsisting style, and it is already clear from the context.

validphys2/src/validphys/filters.py
```diff
@@ -168,24 +171,91 @@ def _filter_real_data(filter_path, data):
...
+    Returns
+    -------
+    tuple
+         total data points and points passing the cuts
+
+    """
+    
     total_data_points = 0
     total_cut_data_points = 0
     fakeset = fakepdf.legacy_load()
```
Zaharid, line 221: Why are we still loading the pdf in cpp? Should't we get rid of this?

validphys2/src/validphys/tests/test_pseudodata.py
```diff
@@ -80,3 +81,23 @@ def test_read_matches_recreate():
...
+    l = Loader()
+    datasetspec = l.check_dataset(list(dataset.values())[0],theoryid=THEORYID)
+    t0set = l.check_pdf(pdfname)
+    
+    l0_cd = API.make_level0_data(dataset_inputs = [dataset], 
+                                use_cuts="internal", theoryid=THEORYID, fakepdf = pdfname)
+                                
+    l0_vals = l0_cd[0].central_values
+  
+    assert(np.abs(np.sum(dataset_t0_predictions(dataset = datasetspec, t0set = t0set) / l0_vals) / len(l0_vals) - 1) <= 1e-12)
```
Zaharid, line 103: See

```
In [1]: from numpy.testing import assert_allclose
```

and similar functions.

validphys2/src/validphys/pseudodata.py
```diff
@@ -236,7 +236,132 @@ def indexed_make_replica(groups_index, make_replica):
...
+        level0_commondata_instances_wc.append(commondata_wc.with_central_value(t0_prediction))
+
+    return level0_commondata_instances_wc
 
+
+
+def make_level1_data(data,commondata_wc,level0_commondata_wc,filterseed):
+    """
+    GENERAL DESCRIPTION:
+
```
Zaharid, line 292:
```suggestion
```

validphys2/src/validphys/pseudodata.py
```diff
@@ -236,7 +236,132 @@ def indexed_make_replica(groups_index, make_replica):
...
+                                use_weights_in_covmat=False, norm_threshold=None,
+                                _list_of_central_values=None, _only_additive=False,)
+
+    #================== generation of pseudo data ======================#
+    from validphys.results import groups_index
+    #= generate pseudo data starting from theory predictions
+    level1_data = make_replica(level0_commondata_wc, filterseed, covmat, 
+                            sep_mult=False, genrep=True)
+
+    group_index = groups_index([data]) # already set cuts
```
Zaharid, line 351: Why not having group_index as a provider here instead?

validphys2/src/validphys/pseudodata.py
```diff
@@ -236,7 +236,132 @@ def indexed_make_replica(groups_index, make_replica):
...
+        #== Generate a new CommonData instance with central value given by Level 0 data generated with fakepdf ==#
+        from validphys.covmats import dataset_t0_predictions
+        t0_prediction = dataset_t0_predictions(dataset=dataset, t0set=fakepdf)  # N.B. cuts already applied to th. pred.
+        level0_commondata_instances_wc.append(commondata_wc.with_central_value(t0_prediction))
+
+    return level0_commondata_instances_wc
 
+
+
+def make_level1_data(data,commondata_wc,level0_commondata_wc,filterseed):
```
Zaharid, line 289: I think we should have a vp provider that does `commondata_wc`  provider, so that this function can be used as an action.

validphys2/src/validphys/pseudodata.py
comane: I think there is already one called experiments_index in results.py.
In the new commits I pass it to the filter_closure_data_by_experiment function which loops over the experiments so that the right indexes are passed to the _filter_closure_data function

validphys2/src/validphys/pseudodata.py
comane: I added a `commondata_wc` function in `results.py`. Also, in the new commits the `make_level0_data` has been renamed `level0_commondata_wc`

Zaharid:
This now lives in #1660.

PR #1653, opened 2022-12-23
Change numpy to python for `int`, `float` and `bool`.

scarlehoff:
Following deprecation notices for `np.int`, `np.float` and `np.bool` which will be removed in 1.24. They were already aliases for the underlying python types.


Zaharid:
I'm surprised they decided this was worth the incredible amounts of noise in every project.

Zaharid (approved):

PR #1654, opened 2022-12-23
fixed bug following from introduction of parse_fakepdf in config.py

comane:
Because of the introduction of the parse_fakepdf method in config.py, the val corresponding to the "fakepdf" key of the datacuts dict in produce_fitunderlyinglaw function is not type str anymore but core.PDF. This leads to an error when running vp-comparefits:

[ERROR]: Bad configuration encountered:
Bad input type for parameter 'pdf': Value 'NNPDF40_nnlo_as_01180' is not of type str, but of type 'PDF'.

### DIFF
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -321,8 +321,8 @@ class CoreConfig(configparser.Config):
         pdf
         """
         with self.set_context(ns=self._curr_ns.new_child({"fit": fit})):
-            _, datacuts = self.parse_from_("fit", "closuretest", write=False)
-        underlyinglaw = self.parse_pdf(datacuts["fakepdf"].name)
+            _, datacuts = self.parse_from_("fit", "closuretest", write=False)       
+        underlyinglaw = datacuts["fakepdf"]
         return {"pdf": underlyinglaw}
 
     @element_of("hyperscans")
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -321,7 +321,7 @@ class CoreConfig(configparser.Config):
         pdf
         """
         with self.set_context(ns=self._curr_ns.new_child({"fit": fit})):
-            _, datacuts = self.parse_from_("fit", "closuretest", write=False)       
+            _, datacuts = self.parse_from_("fit", "closuretest", write=False)
         underlyinglaw = datacuts["fakepdf"]
         return {"pdf": underlyinglaw}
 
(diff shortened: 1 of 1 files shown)


Zaharid:
Do we need to parse the pdf again or can we set it to the value we already get?

RoyStegeman (changes requested):
Thanks for pointing this out. I agree with Zahari, I don't see why we need to parse it seperately in each function, in this case we should just use the pdf directly.

comane:
Thank you for the comments. The new commit should do so. The fakepdf stored in datacuts is taken instead of parsing it again.
There might be other issues with the config.produce_multiclosure_underlyinglaw() function but I don't know in which context this function is to be used.

validphys2/src/validphys/config.py
```diff
@@ -321,8 +321,8 @@ def produce_fitunderlyinglaw(self, fit):
         pdf
         """
         with self.set_context(ns=self._curr_ns.new_child({"fit": fit})):
-            _, datacuts = self.parse_from_("fit", "closuretest", write=False)
-        underlyinglaw = self.parse_pdf(datacuts["fakepdf"])
+            _, datacuts = self.parse_from_("fit", "closuretest", write=False)
```
Zaharid, line 324:
```suggestion
            _, datacuts = self.parse_from_("fit", "closuretest", write=False)
```

Zaharid (approved):

comane:
> What problems are you anticipating? I guess that function would be used to calculate closure test estimators such as these: https://vp.nnpdf.science/5m-aHiGmRVGsOjeZ58AtyA==/ which are based on multiple fits (25 in this case) and to correctly evaluate these estimators the underlying law for each of the fits needs to have been the same.

I guess the problem could be that
`laws.add(closuretest_spec["fakepdf"])`
is adding a PDF object to the `laws` set instead of a string, therefore
`return self.parse_pdf(laws.pop())`
would give the same problem as above.

I think one could just do the following change:
`return laws.pop()`

RoyStegeman:
Did you try? I don't think it's a problem here because the `fakepdf` is obtained by reading  the `filter.yml` file directly (i.e. it's not parsed by reportengine and thus not placed in a namespace).

comane:
Yes, I agree with you. Thanks for the comment. I think with the last commit the branch can be merged

RoyStegeman (approved):
Looks good. Thanks for spotting this!

Zaharid:
Thanks both!

PR #1656, opened 2023-01-04
Add theory_440 to theory.db

andreab1997:
Just added theory_440 (that is a copy of theory 400 only for `HERA` with the b threshold shifted by a factor 1.2) to the theory db. @enocera


enocera (changes requested):
@andreab1997 The value of mb has not been changed, as far as I can see. I know that this is immaterial for all practical purposes, but shouldn't we multiply it by 1.2?

P.S. at some point we should get rid of this `theory.db` file and use the `<theoryid>.yaml` files instead.

andreab1997:
> @andreab1997 The value of mb has not been changed, as far as I can see. I know that this is immaterial for all practical purposes, but shouldn't we multiply it by 1.2?
> 
> P.S. at some point we should get rid of this `theory.db` file and use the `<theoryid>.yaml` files instead.

Actually `mb` is ok, what should change is `kbThr` (in particular from `1.0` to `1.2`). Do you agree?

enocera:
> Actually `mb` is ok, what should change is `kbThr` (in particular from `1.0` to `1.2`). Do you agree?

Yes, I do.

enocera (approved):

scarlehoff:
Please rebase on top of master (which in practice mean, re-add theory_440 on top of master) to ensure that this is not removing a theory from the database (I'm guessing the conflict is theory 600 which was the last one added? Unless someone else added another one afterwards?)

andreab1997:
Actually @giacomomagni already booked untill theory 520 so I will change the name of this theory to 521.

scarrazza:
Tests performed, no needed anymore.

PR #1660, opened 2023-01-12
Python closure sampling (with the only python commondata branch)

scarlehoff:
Hi @Zaharid @comane 

As discussed yesterday during the Code Meeting, I've merged PR #1651 and #1650 so that #1650 can be merged since they are mostly orthogonal as previously mentioned. Indeed, the only difference is the `export` function that can be simply removed so that it uses the new ones, the relevant commit is https://github.com/NNPDF/nnpdf/pull/1660/commits/cc868fb00786ee70a1c563e015f3bb94fce2012a

I would suggest the write functions are put elsewhere so that circular imports are not generated, but I guess for that you can discuss among yourselves where to put what. With this the two branches should be without conflict.

(unless people requests changes in #1650 of course, but I'm not taking responsibility for that xd)


Zaharid:
@scarlehoff is there anything else to do here?

Zaharid:
@scarlehoff @RoyStegeman  What is this

```
Cannot request that multiple replicas are fitted and that pseudodata is saved. Either set `fitting::savepseudodata` to `false` or fit replicas one at a time.
```

https://github.com/NNPDF/nnpdf/actions/runs/4125518187/jobs/7126182659

scarlehoff:
> @scarlehoff is there anything else to do here?

Please use `black` to format the changes. 

Also, the fact that all the commondata writing are imported within the functions to avoid circular imports makes me think they should be moved to an external module that does not depend on core.

Zaharid:
> Also, the fact that all the commondata writing are imported within the functions to avoid circular imports makes me think they should be moved to an external module that does not depend on core.


We looked into that and don't think there is anything to do. 

```
def parse_file_format() -> DataStructure
   ...
```

and

```
class DataStructure:
   ...
   @clasmethod
    def from_fileformat(cls): ...
```

are inherently circular unless everything is in the same file, which is worse. One could remove things like the convenience `export` function but that would not make anything better.

scarlehoff:
I refer to the write functions. As far as I can see they do not contain anything imported from the outside?

Zaharid:
Sorry which write functions where?

validphys2/src/validphys/filters.py
```diff
@@ -194,24 +203,112 @@ def _filter_real_data(filter_path, data):
...
         total_cut_data_points += ncut
-        loaded_ds = loaded_data.GetSet(j)
-        if errorsize != 1.0:
-            loaded_ds.RescaleErrors(errorsize)
-        loaded_ds.Export(str(path))
+
+    from validphys.commondataparser import (
+        write_commondata_to_file,
+        write_systype_to_file,
+    )
```
scarlehoff, line 272: These two. 

Apologies for not writing the comment just below, I was writing from my phone. They are also in `export`.

Ideally also the ones in `validphys.pseudodata`, I haven't looked into those ones so it might be impossible.

validphys2/src/validphys/pseudodata.py
```diff
@@ -236,10 +236,146 @@ def indexed_make_replica(groups_index, make_replica):
...
+        if dataset.cuts is None:
+            commondata_wc = dataset.commondata.load_commondata_instance()
+        else:
+            cuts = dataset.cuts.load()
+            commondata_wc = dataset.commondata.load_commondata_instance().with_cuts(
+                cuts
+            )
+
+        # == Generate a new CommonData instance with central value given by Level 0 data generated with fakepdf ==#
+        from validphys.covmats import dataset_t0_predictions
```
scarlehoff, line 280: Please move this to the beginning of the function, not inside a loop.

validphys2/src/validphys/tests/test_pseudodata.py
```diff
@@ -80,3 +82,27 @@ def test_read_matches_recreate():
         )
         pd.testing.assert_index_equal(read.tr_idx, recreate.tr_idx, check_order=False)
         pd.testing.assert_index_equal(read.val_idx, recreate.val_idx, check_order=False)
+
+
+def test_level0_commondata_wc():
+    from validphys.loader import Loader
+    from validphys.covmats import dataset_t0_predictions
```
scarlehoff, line 89: Please put this outside and add a comment to explain what this is testing.

Can't we use only the loader or only the api?

validphys2/src/validphys/core.py
```diff
@@ -618,6 +626,24 @@ def load(self):
...
+        """
+        commodata_list = []
+        for dataset in self.datasets:
+            if dataset.cuts is None:
+                commodata_list.append(dataset.commondata.load_commondata_instance())
+            else:
+                commodata_list.append(
+                    dataset.commondata.load_commondata_instance().with_cuts(
+                        dataset.cuts.load()
+                    )
```
scarlehoff, line 643: I would take `dataset.commondata.load_commondata_instance()` outside of the if condition since it will ease future debugging and it is used in both branches.

validphys2/src/validphys/tests/test_pseudodata.py
comane: Thank you for the comment, I agree, although I think in order to use the API for both we would probably need to have  a parse_dataset function

validphys2/src/validphys/filters.py
```diff
@@ -119,22 +122,20 @@ def prepare_nnpdf_rng(filterseed:int, rngalgo:int, seed:int):
...
     being shifted away from the underlying law.
 
     """
     log.info('Filtering closure-test data.')
-    return _filter_closure_data(
-        filter_path, data, fakepdf, fakenoise, errorsize)
+    return _filter_closure_data(filter_path, data, fakepdf, fakenoise, filterseed)
 
 
-@check_positive("errorsize")
```
scarlehoff, line 135: Please remove `check_positive` also from the imports.

validphys2/src/validphys/pseudodata.py
```diff
@@ -236,10 +236,146 @@ def indexed_make_replica(groups_index, make_replica):
...
+    >>> API.level0_commondata_wc(dataset_inputs = [{"dataset":"NMC"}], use_cuts="internal", theoryid=200,fakepdf = "NNPDF40_nnlo_as_01180")
+
+    [CommonData(setname='NMC', ndata=204, commondataproc='DIS_NCE', nkin=3, nsys=16)]
+    """
+    from validphys.covmats import dataset_t0_predictions
+    level0_commondata_instances_wc = []
+
+    # ==== Load validphys.coredata.CommonData instance with cuts ====#
+
+    for j, dataset in enumerate(data.datasets):
```
scarlehoff, line 271:
```suggestion
    for dataset in data.datasets:
```

validphys2/src/validphys/pseudodata.py
```diff
@@ -236,10 +236,146 @@ def indexed_make_replica(groups_index, make_replica):
...
+    >>> from validphys.api import API
+    >>> dataset='NMC'
+    >>> l1_cd = API.make_level1_data(dataset_inputs = [{"dataset":dataset}],use_cuts="internal", theoryid=200,
+                             fakepdf = "NNPDF40_nnlo_as_01180",filterseed=1)
+    >>> l1_cd
+    [CommonData(setname='NMC', ndata=204, commondataproc='DIS_NCE', nkin=3, nsys=16)]
+    """
+
+    # =============== generate experimental covariance matrix ===============#
+    from validphys.covmats import dataset_inputs_covmat_from_systematics
```
scarlehoff, line 337: Please put this import at the top (there are already other `covmats` imports so it should be ok)

validphys2/src/validphys/results.py
```diff
@@ -224,6 +224,25 @@ def procs_data_values(proc_result_table):
...
+    ----------
+
+    data : validphys.core.DataGroupSpec
+    
+    Returns
+    -------
+    list containing commondata instances with cuts
+    
+    """
+    return data.load_commondata_instance()
```
scarlehoff, line 244: I'm very confused about this function, there is nothing enforcing the cuts here, right? How is this different from simply doing `data.load_commondata_instance()` inside `make_level1_data`?

Maybe there's something I've missed along the way?

scarlehoff (commented):
Thank you very much @comane! I'm looking forward to merging this.

I'm already using a pip-only version of NNPDF for some fits so we already almost there for a pip-only installation of the code in master (and even a `validphys` only installation separated from `n3fit` which would be _very_ useful!)

One thing I would like to link in this PR is some reports of closure tests done with this branch (I guess you are already using it for that). I'm guessing some of the last CTs in the server are done with this branch but wouldn't want to select the wrong ones.

As a final thing, it would be _great_ to have a comparison between a closure test done with this branch and another one done with master, we already had some trouble reproducing the CT of the 4.0 paper so I wouldn't like to add another layer of changes.

On the other hand, I don't want to inflict CT on anybody so I won't insist if you don't feel like doing it.

validphys2/src/validphys/coredata.py
```diff
@@ -364,21 +369,21 @@ def systematic_errors(self, central_values=None):
...
-        Use the same format as libNNPDF:
 
         - A DATA_<dataset>.dat file with the dataframe of accepted points
         - A systypes/STYPES_<dataset>.dat file with the error types
         """
+        from validphys.commondataparser import (
+            write_systype_to_file,
+            write_commondata_to_file,
+        )
+
```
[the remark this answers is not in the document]
Zaharid: I disagree that they don't belong in the same place. The place I would look for write functions is the same where the read functios are. Besides, unless we have the write functions in coredata.py its not going to be trivial to avoid circular imports.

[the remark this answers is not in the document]
scarlehoff: None of the lines added to `commondataparser.py` need to import anything. You can just create a `commondatawriter.py` and then the problem is fixed. It is trivial to avoid circular imports.

> The place I would look for write functions ...

If there is a file called `commondatawriter.py` next to it I'm sure you will open the right file.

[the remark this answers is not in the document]
Zaharid: The reader functions need to import the data definitions they are reading into. Also I don't think there is anything wrong with circular imports in this case that justifies adding files with ten lines worth of code.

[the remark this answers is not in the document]
scarlehoff: > The reader functions need to import the data definitions they are reading into.

Which reader functions? I think you are thinking of a different thing?

validphys2/src/validphys/filters.py
```diff
@@ -194,24 +206,106 @@ def _filter_real_data(filter_path, data):
...
+            path_cd = filter_path / l1_cd.setname / f"DATA_{l1_cd.setname}.dat"
+            path_sys = (
+                filter_path
+                / l1_cd.setname
+                / "systypes"
+                / f"SYSTYPE_{l1_cd.setname}_DEFAULT.dat"
+            )
+            write_commondata_to_file(commondata=l1_cd, path=path_cd)
+            write_systype_to_file(commondata=l1_cd, path=path_sys)
+
```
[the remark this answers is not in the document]
comane: Hi @scarlehoff thank you very much for this comment, I will simplify the logic of the function asap.

validphys2/src/validphys/results.py
comane: Hi @scarlehoff, I think you are right. This is completely unnecessary. I changed it in the new commits

comane:
> As a final thing, it would be _great_ to have a comparison between a closure test done with this branch and another one done with master, we already had some trouble reproducing the CT of the 4.0 paper so I wouldn't like to add another layer of changes.


Hi @Zaharid, @scarlehoff following the link below you can find a comparison between a CT done with the master and one done with this branch. Note that the initialisation seeds for level 1 noise are not the same. This is because the current branch does not depend on `filters.prepare_nnpdf_rng` anymore.

https://vp.nnpdf.science/jj9pitd5Q_KJ6zB7lBGKkg==

scarlehoff:
Thanks! Let me ping @RoyStegeman who is much more experienced than I am looking at CTs.

RoyStegeman:
I'd have to properly this to understand it better, but all this does is replace the cpp CT sampling with the new python implementation, right? If so, I'm a bit surprised by the difference between the two fits - are the runcards exactly the same (up to seeds)?

@comane would you mind uploading the fits to the nnpdf server? In general it's a good habit to upload the fit if you're sharing a corresponding result, even just for possible later reference.

Zaharid:
@RoyStegeman the random numbers are not the same since one is using the cpp rng while the other is using numpy.

RoyStegeman:
Thanks, that's what I thought. If that's the only difference in the code, the difference between the fits seems large to me, wouldn't you agree?

Zaharid:
@RoyStegeman I guess a lot of the studies were based on "multiclosure" to avoid the dependency on the L1 seed; Looking at 

https://vp.nnpdf.science/P0SSEuO2RdWW753vlYG_sA==

and particularly at

https://vp.nnpdf.science/P0SSEuO2RdWW753vlYG_sA==/plot_delta_chi2_report_report.html

I can be easily convinced that the difference is just the seed. I am not sure if we have the detailed stats somewhere.

comane:
> @comane would you mind uploading the fits to the nnpdf server? In general it's a good habit to upload the fit if you're sharing a corresponding result, even just for possible later reference.

Hi @RoyStegeman, I just uploaded the fits and they can be found at

https://data.nnpdf.science/fits/NNPDF40_nnlo_as_01180_CT_python.tar.gz

https://data.nnpdf.science/fits/NNPDF40_nnlo_as_01180_CT_master.tar.gz


I am currently running another closure test using the branch master but with different seed. This will probably allow us to understand better how much the CT depends on the L1 seed.

RoyStegeman:
Yes I realised my thinking was wrong after sending that last message, hence the edit ;). You're right that looking at MW's fits the difference could well be explained due to a different L1 seed, though his fits were also done using the NNPDF3.1 dataset. On top of that fluctuations are relatively large so I guess a single L1 CT fit is just not very useful. 

> I am not sure if we have the detailed stats somewhere.

I suppose these would just be the estimators quoted in the nnpdf40 paper, or redetermined by Samuele. But they are all based on ~25 fits of 40 replicas...

RoyStegeman:
@comane thanks. 

Indeed, it's probably a good idea to do at least one more fit. I don't know how much computational resources you have, but since we really don't want any surprises when it comes to CT, it may be worth simply doing a full CT so at least we have a reference point in case it gets messed up in the future.

Zaharid:
cc @MaeveMadiganMM

comane:
Hi @Zaharid, @MaeveMadiganMM,  @RoyStegeman and @scarlehoff  this is a comparison between two CT's both performed with the Master branch but with different L1 seeds.

https://vp.nnpdf.science/t2Z6P2ctS5eQaSzbvUS2sg==

On 

https://data.nnpdf.science/fits/13_02_23_MNC_NNPDF40_nnlo_as_01180_CT_master_new_seed.tar.gz

You can find the fit corresponding to the new seed.

comane (approved):

scarlehoff:
Are "we" happy with the closure test result?

From my side I would like these two points to be addressed:

https://github.com/NNPDF/nnpdf/pull/1660/files#r1102438850

https://github.com/NNPDF/nnpdf/pull/1660/files#r1102438226

validphys2/src/validphys/coredata.py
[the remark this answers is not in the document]
Zaharid: I was answering to

> None of the lines added to commondataparser.py need to import anything.

saying that the functions that take files and return data structures do need to import the data structure, and there would be a circular import if you want a convenience method to read.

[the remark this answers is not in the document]
scarlehoff: Sorry, but I really cannot see which of the lines added to `commondataparser.py` is using anything imported form the outside. They are all self-contained functions.

There are no reads methods in the changes introduced by this Pull Request. I just want to have a `commondatawriter.py` with those functions.

comane:
In my opinion it looks like the deviation that one sees between CT(python) and CT(master) is due to stat fluctuations, meaning that if one would repeat the CT with the new branch n times then 0.68 * n of the times the CT (python) would lie within the uncertainty bands of CT(master).
The report below compares a CT(master) against 4 CT(python) each runned with a different seed.

https://vp.nnpdf.science/jrWmrlYNTPikrEfm8BNOCg==

https://data.nnpdf.science/fits/17_02_23_MNC_NNPDF40_nnlo_as_01180_CT_python_ns1.tar.gz
https://data.nnpdf.science/fits/17_02_23_MNC_NNPDF40_nnlo_as_01180_CT_python_ns2.tar.gz
https://data.nnpdf.science/fits/18_02_23_MNC_NNPDF40_nnlo_as_01180_CT_python_ns3.tar.gz

PR #1661, opened 2023-01-13
Trvl mask layers

goord:
Implementation of training-validation split per parallel replica via a set of masking layers. This means there is only one (set of) observables, but still training-validation observable wrappers which mask out the correct data.

Currently, I still need to do:

- [x] Run the realistic nnpdf4 use case
- [x] Compare results of a basic run
- [x] Compare results of a realistic run
- [x] Monitor memory use, investigate scaling with no. parallel replicas
- [x] Test on GPU's


goord:
I am facing problems with datasets with single datapoint, some replicas wll mask these out, other won't. That is a poor fit for the design I followed, which assumes the training/validation masks can be represented as block-wise boolean arrays with the same number of 'True' values per row, such that the output tensor is not strided

RoyStegeman:
This was recently changed in #1636. 

In principle I would be happy with a gpu implementation that treats datasets with a single point in the old way (so including it in the tr set). The difference in the result is negligible and as such we could still use the gpu for most purposes, only for a very final fit before a release might it be better to have the new treatment of single point datasets. @scarlehoff has many gpus available, so perhaps he has an opinion on this.

scarlehoff:
I'd be happy with the solution of just ignoring those datasets for the time being.

RoyStegeman:
Also fine for me. Point is (to address @goord's concerns), this is not a showstopper for the parallel fits.

goord:
Yes that would be the best solution, as things become _very_ complicated if we can't assume equal number of masked data points across replicas. I will try to include (or exclude?) all single-point datasets in the fit if parallel replicas is set and same_trvl_split is unset...

n3fit/src/n3fit/model_gen.py
```diff
@@ -76,8 +80,16 @@ def _generate_experimental_layer(self, pdf):
...
+        masked_output_layers = []
+
+        for output_layer, mask_layer in zip(output_layers, self.trvl_mask_layers):
+            if mask_layer is None:
+                masked_output_layers.append(output_layer)
+            else:
+                masked_output_layers.append(mask_layer(output_layer))
+
+        # Finally concatenate all observables (so that experiments are one single entity)
+        ret = op.concatenate(masked_output_layers)
```
scarlehoff, line 92: Why is the axis removed? (I guess the default is exactly the right axis, but I'd like to have it explicit, it makes debugging easier)

n3fit/src/n3fit/model_gen.py
```diff
@@ -138,9 +150,11 @@ def observable_generator(
...
     dataset_xsizes = []
     model_inputs = []
-    model_obs_tr = []
-    model_obs_vl = []
-    model_obs_ex = []
+    model_observables = []
+    tr_mask_layers = []
+    vl_mask_layers = []
+    offset = 0
+    apply_masks = spec_dict.get("data_transformation_tr") is None and mask_array is not None
```
scarlehoff, line 157: I'm a bit worried. If `mask_array` is not `None` but there is a `data_transformation_tr` then the masks will not be applied. If this is necessary then it should fail at the beginning.

We usually do that by adding a check before the fit starts. In this case it should check whether the run options are a parallel fit and data_transformation and if so validphys will raise an exception telling the user which options are inconsistent.

(for the time being you can put just a raise Exception here to stop it and create the check at the end)

n3fit/src/n3fit/model_gen.py
```diff
@@ -155,56 +169,27 @@ def observable_generator(
         # Set the operation (if any) to be applied to the fktables of this dataset
         operation_name = dataset.operation
 
+        # Extract the masks that will end up in the observable wrappers...
+        trmask = mask_array[:, offset:offset + dataset.ndata] if apply_masks else None
+        tr_mask_layers.append(Mask(trmask, axis=1, c=1) if apply_masks else None)
+        vl_mask_layers.append(Mask(~trmask, axis=1, c=1) if apply_masks else None)
```
scarlehoff, line 175:
```suggestion
        if apply_masks:
            trmask = mask_array[:, offset:offset + dataset.ndata]
            tr_mask_layers.append(Mask(bool_mask=trmask, axis=1))
            vl_mask_layers.append(Mask(bool_mask=~trmask, axis=1))
        else:
            trmask = None
            tr_mask_layers.append(None)
            vl_mask_layers.append(None)
```

I don't like the idea of having a list of `None`, I think with the check above you will always be in a consistent state and you might be able to check whether to apply the mask somewhere else (so you don't need the `None`).

n3fit/src/n3fit/model_gen.py
```diff
@@ -213,9 +198,10 @@ def observable_generator(
             model_inputs += xgrids
             dataset_xsizes.append(sum([len(i) for i in xgrids]))
 
-        model_obs_tr.append(obs_layer_tr)
-        model_obs_vl.append(obs_layer_vl)
-        model_obs_ex.append(obs_layer_ex)
+        model_observables.append(obs_layer)
+
+        # shift offset for new mask array
+        offset = offset + dataset.ndata
```
scarlehoff, line 204: Would there be a way to have a list of arrays from the onset (instead of having an offset that we move?) such that to each dataset in the list correspond an array.

n3fit/src/n3fit/model_gen.py
```diff
@@ -37,6 +40,7 @@ class ObservableWrapper:
 
     name: str
     observables: list
+    trvl_mask_layers: list
```
scarlehoff, line 43: Maybe we can have this be optional = None, and if it is None then it doesn't get applied.

Because (I think, maybe I'm wrong!) that we should never be in a situation in which some of the masks exist and some are `None`, that way you can avoid the list of `None`.

n3fit/src/n3fit/model_trainer.py
```diff
@@ -11,6 +11,8 @@
 import logging
 from collections import namedtuple
 from itertools import zip_longest
+
+import numpy
```
scarlehoff, line 15:
```suggestion
```

scarlehoff (commented):
I think I like this solution. It is more or less what I had in mind I think.

RE the one-point datasets, I think we can eventually find a way around it but let's ignore them for the moment.

Did you run any parallel fits with this code? Did they work? Is the memory footprint greatly impacted?

goord:
Hi @scarlehoff I haven't replied to your suggestions because I propose to first get the validation in order. After testing with the basic_parallel test case with only a single dataset, I believe I still have a bug somewhere in the implementation because my chi2 values differ significantly from the master sequential (or parallel) runs.

|variable | master-sequential | master-parallel | trvl mask-layers parallel |
|------------|----------------------------|------------------------|-------------------------------------|
| chi2       | 1.02                          | 1.01                     | 20.0                                    |
| erf_tr     | 1.61                          | 2.0                       | 3.77                                    |
| erf_vl     | 1.81                           | 55.4                    | 1.46                                   |

These are means over 500 replicas.

goord:
It looks like the concatenation of the masked observables is giving some issues... to be continued

goord:
I have solved a few problems with the implementation: the first - obvious - one was that the masked truth values weren't correctly propagating to the loss function. Another issue was that a replica-specific inverse covariance was not taken into account.

Latest commits result in a basic runcard fit that is bitwise identical to the sequential run for small no. epochs. More validation results will follow.

goord:
Memory use is now ok, slightly lower than the master branch:file:///home/gvdoord/Downloads/memuse-plot.png

n3fit/src/n3fit/layers/mask.py
```diff
@@ -25,23 +25,27 @@ class Mask(MetaLayer):
...
         super().__init__(**kwargs)
 
     def build(self, input_shape):
         if self.c is not None:
             initializer = MetaLayer.init_constant(value=self.c)
+            output_shape = list(input_shape)
+            output_shape[-1] = self.last_dim
             self.kernel = self.builder_helper(
-                "mask", (1,), initializer, trainable=False
+                "mask", output_shape, initializer, trainable=False
```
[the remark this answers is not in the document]
goord: The implicit shape argument leads to a tensorflow graph execution error
`Node: 'meta_model_1/tf.concat_12/concat'
ConcatOp : Dimension 1 in both shapes must be equal: shape[0] = [1,3,3] vs. shape[7] = [1,1,0]
	 [[{{node meta_model_1/tf.concat_12/concat}}]] [Op:__inference_losses_fun_110454]`
(this is for the NNPDF4.0 runcard with 3 parallel replicas).

Can we expect tensorflow to be able to infer the output shape from the mask? I would be surprised if it can, I haven't been able to find useful documentation about this, so I tried to be as explicit as I could be.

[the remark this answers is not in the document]
goord: Ok, apparently tensorflow inferred the output shape from the multiplication constant (which was no longer a scalar but a const tensor, and the multiplication was being done point-wise). I removed that and made the output shape explicit upon reshaping in commit 303024e7a1cade4150caa2154a36f78d19d04257

n3fit/src/n3fit/layers/mask.py
```diff
@@ -25,23 +25,27 @@ class Mask(MetaLayer):
...
             )
         super(Mask, self).build(input_shape)
 
     def call(self, ret):
         if self.mask is not None:
-            ret = op.boolean_mask(ret, self.mask, axis=self.axis)
+            ret = op.reshape(op.boolean_mask(ret, self.mask, axis=self.axis), shape=(1, -1, self.last_dim))
```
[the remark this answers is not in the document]
goord: Implemented in 303024e7a1cade4150caa2154a36f78d19d04257

n3fit/src/n3fit/model_gen.py
```diff
@@ -10,17 +10,20 @@
...
-from n3fit.layers import Preprocessing, FkRotation, FlavourToEvolution
+from n3fit.layers import Preprocessing, FkRotation, FlavourToEvolution, Mask
 from n3fit.layers.observable import is_unique
 
 from n3fit.backends import MetaModel, Input
 from n3fit.backends import operations as op
 from n3fit.backends import MetaLayer, Lambda
 from n3fit.backends import base_layer_selector, regularizer_selector
 
+from keras.layers import Masking
```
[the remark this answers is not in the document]
goord: Fixed in 303024e7a1cade4150caa2154a36f78d19d04257

n3fit/src/n3fit/model_gen.py
```diff
@@ -10,17 +10,20 @@
 
 """
 from dataclasses import dataclass
+
+import numpy
```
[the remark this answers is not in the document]
goord: Fixed in 303024e7a1cade4150caa2154a36f78d19d04257

n3fit/src/n3fit/model_gen.py
goord: I thought it was a bit explicit, but since all tensor shapes have to be as explicit as possible for tensorflow to do the correct thing, I will re-insert it.

scarlehoff: More than for tensorflow is for the person reading the code in this case, since it is hard to keep track of which axis is what :P

n3fit/src/n3fit/model_gen.py
goord: Fixed in rev. bc568761540fa4a96e91963ad314041fa81dc782

n3fit/src/n3fit/model_gen.py
goord: Fixed in rev. bc568761540fa4a96e91963ad314041fa81dc782. I did not insert the default value yet though

goord:
Regarding memory usage: due to the different masks per replica, the `lru_cache` for `fittable_datasets_masked` in n3fit_data.py is not triggered anymore when `same_trvl_per_replica` is set to `false`. This causes the following loop in n3fit_data_utils.py to be executed for each replica:

```
for dspec, mask in zip_longest(datasets, tr_masks):
        # Load all fktables with the appropiate cuts
        fktables = [fk.load_with_cuts(dspec.cuts) for fk in dspec.fkspecs]
        # And now put them in a FittableDataSet object which
        loaded_obs.append(FittableDataSet(dspec.name, fktables, dspec.op, dspec.frac, mask))
```
Although the FK table loading is in its own `lru_cache` and not re-executed, the `with_cuts` is not and will trigger a copy of the FK-tables for each replica, although the cuts are identical and independent of the mask, and inflates the memory footprint.

The easiest fix is to wrap the `fk.load_with_cuts(dspec.cuts)` in a dedicated function with a `lru_cache` attribute. I wouldn't call this an elegant solution though, any thoughts @scarlehoff ?

scarlehoff:
Yes. I think that should work since all replicas are using the same fktables now (i.e., it should work if the `datasets` are the same for all replicas and the only thing that changes is `tr_masks` but I think this is indeed the case).

validphys2/src/validphys/n3fit_data_utils.py
```diff
@@ -94,7 +95,12 @@ def validphys_group_extractor(datasets, tr_masks):
...
         # Load all fktables with the appropiate cuts
-        fktables = [fk.load_with_cuts(dspec.cuts) for fk in dspec.fkspecs]
+        fktables = [load_cached_fk_tables(fk, dspec.cuts) for fk in dspec.fkspecs]
         # And now put them in a FittableDataSet object which
         loaded_obs.append(FittableDataSet(dspec.name, fktables, dspec.op, dspec.frac, mask))
     return loaded_obs
+
+@functools.lru_cache
+def load_cached_fk_tables(fk, cuts):
+    return fk.load_with_cuts(cuts)
```
scarlehoff, line 105: Wouldn't it be better to wrap directly `fk.load_with_cuts`?

https://github.com/NNPDF/nnpdf/blob/f02b49a8e6eb785af6a56dc3195133abff571046/validphys2/src/validphys/core.py#LL442C4-L442C4

(in practical terms it should be the same)

Given an fktable and a set of cuts, I see no reason why we would we want two different objects so that probably is a positive change in other parts of the code as well*


*hopefully

goord: Yes that would be the more elegant solution, but needs thorough testing because this indeed impacts the code in more places, there will be shared objects where there were copies before. I can have a look.

goord:
New validation and performance tests are underway

goord:
A 100-replica fit comparison with the recent fits by @APJansen is done here: https://vp.nnpdf.science/Hz7Gwu95TzCUCH4oYoQOQA==

RoyStegeman:
Can this be closed in favor of #1788 ?

scarlehoff:
Let me echo @RoyStegeman's question

> Can this be closed in favor of https://github.com/NNPDF/nnpdf/pull/1788 ?

goord:
done

PR #1666, opened 2023-01-18
Fix extra labels slicing when cuts are given

Zaharid:
plot_smpdf can break with some mismatch error like

```python-traceback
  File "/home/zah/nngit/nnpdf/validphys2/src/validphys/dataplots.py", line 869, in plot_smpdf
    table = kitable(dataset, info)
  File "/home/zah/nngit/nnpdf/validphys2/src/validphys/plotoptions/core.py", line 328, in kitable
    table[label] = value
  File "/home/zah/anaconda3/envs/nnpdfdev/lib/python3.9/site-packages/pandas/core/frame.py", line 3978, in __setitem__
    self._set_item(key, value)
  File "/home/zah/anaconda3/envs/nnpdfdev/lib/python3.9/site-packages/pandas/core/frame.py", line 4172, in _set_item
    value = self._sanitize_column(value)
  File "/home/zah/anaconda3/envs/nnpdfdev/lib/python3.9/site-packages/pandas/core/frame.py", line 4912, in _sanitize_column
    com.require_length_match(value, self.index)
  File "/home/zah/anaconda3/envs/nnpdfdev/lib/python3.9/site-packages/pandas/core/common.py", line 561, in require_length_match
    raise ValueError(
ValueError: Length of values (34) does not match length of index (30)

----
```

This is caused by cuts not being used anywhere when calling PlotInfo.from_commondata, specifically not to being used to filter the extra_labels.

I am a bit unsure why this was noticed before, but part of the answer is that plot_fancy use the plotting information without cuts and applies them later itself.

Fix that by filtering the extra labels.

### DIFF
--- a/validphys2/src/validphys/tests/test_loader.py
+++ b/validphys2/src/validphys/tests/test_loader.py
@@ -17,18 +17,22 @@ l = FallbackLoader()
 #The sorted is to appease hypothesis
 dss = sorted(l.available_datasets - {"PDFEVOLTEST"})
 
+class MockCuts():
+    def __init__(self, arr):
+        self.arr = arr
+    def load(self):
+        return self.arr
+
 
 @composite
-def commodata_and_cuts(draw):
+def commondata_and_cuts(draw):
     cd = l.check_commondata(draw(sampled_from(dss)))
     ndata = cd.metadata.ndata
-    #TODO: Maybe upgrade to this
-    #https://github.com/HypothesisWorks/hypothesis/issues/1115
-    mask = sorted(draw(sets(sampled_from(range(ndata)))))
+    mask = sorted(draw(sets(sampled_from(range(ndata)), min_size=1)))
     return cd, mask
 
 
-@given(arg=commodata_and_cuts())
+@given(arg=commondata_and_cuts())
 @settings(deadline=None)
 def test_rebuild_commondata_without_cuts(tmp_path_factory, arg):
     # We need to create a new directory for each call of the test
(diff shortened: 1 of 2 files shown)


Zaharid:
Here an example that breaks for me without this fix:

https://vp.nnpdf.science/01965VhhQqOHdHR0K2RLZw==/

RoyStegeman (approved):
Thanks

validphys2/src/validphys/tests/test_loader.py
```diff
@@ -17,18 +17,22 @@
...
 
 @composite
-def commodata_and_cuts(draw):
+def commondata_and_cuts(draw):
     cd = l.check_commondata(draw(sampled_from(dss)))
     ndata = cd.metadata.ndata
-    #TODO: Maybe upgrade to this
-    #https://github.com/HypothesisWorks/hypothesis/issues/1115
-    mask = sorted(draw(sets(sampled_from(range(ndata)))))
+    mask = sorted(draw(sets(sampled_from(range(ndata)), min_size=1)))
```
scarlehoff, line 31: Seeing a variable 6 functions removed from its input without even a comment makes me nervous.

scarlehoff (approved):
If you add a comment in the test I'd be happier.

Other than that I'm ok with this. Merge at will.

validphys2/src/validphys/tests/test_loader.py
Zaharid: Could you elaborate? Didn’t get what you mean here.

scarlehoff: I'd like have a comment for this function to understand what it does without needing to mentally parse 5 functions. It's even an effort tu understand which function is taking `min_size=1` as an argument.

So I'd by happy either with a comment / docstring `# Creates a mask by selecting a subset of datapoints (eg. [0, 2, 4, 5]` or breaking it down with obvious names.

Otherwise it is not clear at first sight whether `mask` should be `[0, 1, 0, 0]` or `[1]`

PR #1668, opened 2023-01-26
Fix penalties and add tests

scarlehoff:
These have been broken since a while, but since nobody used them it didn't really matter.

Don't think they will be used but who knows, I added a few tests since (even if they are not used in practice) they ensure that several pieces of the code work well together as they require the correct flavours being set, `n3fit` model to be compatible with validphys, etc.

### DIFF
--- a/n3fit/src/n3fit/hyper_optimization/penalties.py
+++ b/n3fit/src/n3fit/hyper_optimization/penalties.py
@@ -20,7 +20,7 @@ The name in the runcard must match the name used in this module.
 """
 import numpy as np
 from validphys import fitveto
-from n3fit.vpinterface import N3PDF
+from n3fit.vpinterface import N3PDF, integrability_numbers
 
 
 def saturation(pdf_models=None, n=100, min_x=1e-6, max_x=1e-4, flavors=None, **_kwargs):
@@ -43,8 +43,8 @@ def saturation(pdf_models=None, n=100, min_x=1e-6, max_x=1e-4, flavors=None, **_
     -------
     >>> from n3fit.hyper_optimization.penalties import saturation
     >>> from n3fit.model_gen import pdfNN_layer_generator
-    >>> fake_fl = [{'fl' : i, 'largex' : [0,1], 'smallx': [1,2]} for i in ['u', 'ubar', 'd', 'dbar', 'c', 'cbar', 's', 'sbar']]
-    >>> pdf_model = pdfNN_layer_generator(nodes=[8], activations=['linear'], seed=0, flav_info=fake_fl)
+    >>> fake_fl = [{'fl' : i, 'largex' : [0,1], 'smallx': [1,2]} for i in ['u', 'ubar', 'd', 'dbar', 'c', 'g', 's', 'sbar']]
+    >>> pdf_model = pdfNN_layer_generator(nodes=[8], activations=['linear'], seed=0, flav_info=fake_fl, fitbasis="FLAVOUR")
     >>> isinstance(saturation(pdf_model, 5), float)
     True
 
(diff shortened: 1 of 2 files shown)


n3fit/src/n3fit/tests/test_penalties.py
```diff
@@ -0,0 +1,40 @@
+"""
+    Test the penalties for n3fit hyperopt
+"""
+from types import SimpleNamespace
+from n3fit.hyper_optimization.penalties import integrability, patience, saturation
+from n3fit.model_gen import pdfNN_layer_generator
+
+
+def test_saturation():
+    """Check the saturation penalty runs and returns a float"""
```
RoyStegeman, line 10:
```suggestion
    """Check that the saturation penalty runs and returns a float"""
```

n3fit/src/n3fit/tests/test_penalties.py
```diff
@@ -0,0 +1,40 @@
...
+    fake_stopping = SimpleNamespace(
+        e_best_chi2=1000, stopping_patience=500, total_epochs=5000, vl_loss=2.42
+    )
+    res = patience(stopping_object=fake_stopping, alpha=1e-4)
+    assert isinstance(res, float)
+
+
+def test_integrability_numbers():
+    """Check that the integrability penality can be run
+    and that it returns indeed a float"""
```
RoyStegeman, line 32:
```suggestion
    """Check that the integrability penalty runs and returns a float"""
```

RoyStegeman (commented):
I don't think anyone is using this, though @Zaharid did contact me about it recently so he may want to be aware of this.

n3fit/src/n3fit/hyper_optimization/penalties.py
```diff
@@ -82,15 +82,15 @@ def patience(stopping_object=None, alpha=1e-4, **_kwargs):
...
+    >>> patience(fake_stopping, alpha=1e-4)
     3.434143467595683
 
     """
     epoch_best = np.take(stopping_object.e_best_chi2, 0)
     patience = stopping_object.stopping_patience
     max_epochs = stopping_object.total_epochs
     diff = abs(max_epochs - patience - epoch_best)
-    vl_loss = np.take(stopping_object.vl_chi2, 0)
+    vl_loss = np.take(stopping_object.vl_loss, 0)
```
RoyStegeman, line 93: I don't believe this is correct

scarlehoff: What do you mean? In principle given the description it should be `loss` (so before it was wrong).

RoyStegeman: Because it's a `Stopping` object, not a `Fitstate` object (though I guess it might actually make life easier to just make it one object). The former does not have the `vl_loss` property.

scarlehoff, line 93:
```suggestion
    vl_loss = np.take(stopping_object.vl_chi2, 0)
```

I see. Then let me revert that.

n3fit/src/n3fit/tests/test_penalties.py
```diff
@@ -0,0 +1,39 @@
...
+    pdf_model = pdfNN_layer_generator(
+        nodes=[8], activations=["linear"], seed=0, flav_info=fake_fl, fitbasis="FLAVOUR"
+    )
+    assert isinstance(saturation(pdf_model, 5), float)
+
+
+def test_patience():
+    """Check that the patience penalty runs and returns a float"""
+    fake_stopping = SimpleNamespace(
+        e_best_chi2=1000, stopping_patience=500, total_epochs=5000, vl_loss=2.42
```
scarlehoff, line 24:
```suggestion
        e_best_chi2=1000, stopping_patience=500, total_epochs=5000, vl_chi2=2.42
```

RoyStegeman (approved):

RoyStegeman:
I think this is ready to be merged?

Issue #1641, opened 2022-12-01
Clarify pseudodata rebuilding workflow 


Zaharid:
We have a number of tools relying on having explicit pseudodata for the fitted replicas (both the fluctuation and the training validation split, and then also implicitly the cuts). We have a way of storing all that which is not used often by default. 

And we have a way of “rebuilding” the data relying on things like the algorithms for splitting being unmodified and random streams being stable.

We have a tendency to break the later one which also adds constraints on how to modify things. I believe we should discuss what to do with it. As it is, I consider any modifications that change the pseudodata a breaking change requiring a version bumps (but covered by the compatibility policy which explicitly allows for randomly state changes). But perhaps we should simply deprecate that and mandate that pseudodata needs to be stored if we intend to use it later.

Zaharid:
Cc @J-M-Moore

scarlehoff:
> and then also implicitly the cuts

Yet another thing that has changed recently for several datasets.

---

Storing the pseudodata by default -instead of not doing it- would be the best thing we can do. One never knows when it will be neded.
One question, is it possible to make `reportengine` / `validphys` dump the pseudodata only after the fit is finished? 

(also, maybe we should dump it as one file to reduce cluttering filesystems)

Zaharid:
You can make a provider depending on the fit for example. Or use the feature where you specify post actions like the figures do.

scarlehoff:
I guess that's an answer for the last point?

For the rest, I'm happy with saving by default for now.

Zaharid:
To me the core issue at hand is what to do with the existing rebuilding functionality. Do we delete it, document is as very unreliable, commit to making it reliable, something else?

scarlehoff:
Documenting it as reliable seems the right thing to me.
A warning of "Rebuilding is only fully reliable when used with the exact commit the fit was made with" or something like that.

PR #1669, opened 2023-02-01
Save pseudodata by default

RoyStegeman:


doc/sphinx/source/tutorials/pseudodata.rst
```diff
@@ -2,9 +2,58 @@
...
+With the postfit reshuffling handled instead by :py:func:`validphys.pseudodata.read_pdf_pseudodata`.
+
+
+Reconstructing pseudodata
+--------------------------
+
+.. warning::
+
+  The functionality described here is not guaranteed to work between different versions of the code
+  or its dependencies. Specifically, if anything breaks the pseudodata geneartion between commits,
```
scarlehoff, line 49:
```suggestion
  or its dependencies. Specifically, if anything breaks the pseudodata generation between commits, 
```

n3fit/src/n3fit/scripts/n3fit_exec.py
```diff
@@ -133,7 +133,7 @@ def from_yaml(cls, o, *args, **kwargs):
 
         N3FIT_FIXED_CONFIG['actions_'].append(namespace + "performfit")
 
-        if file_content["fitting"].get("savepseudodata"):
+        if file_content["fitting"].get("savepseudodata", True) == True:
```
scarlehoff, line 136: Why the `==True`? Since it is a default I think it would be good to have it be true unless explicitly set to `False`.
```suggestion
        if file_content["fitting"].get("savepseudodata", True):
```

scarlehoff (approved):
Thanks! As per #1641, I'm happy with this solution.

Zaharid:
LGTM. Thanks.

n3fit/src/n3fit/scripts/n3fit_exec.py
RoyStegeman: Because of the difference between the key existing and being true.

RoyStegeman:
Note that I also wrote "since version 4.0.7". Are you ok with doing a new release for this?

scarlehoff:
I don't think we need a new release for this. The last one was just one or two PRs ago, simply the next release will have quite a few important changes this, python commondata and closure test and eko evolution, maybe also photon... sadly we cannot call it 4.1 yet I guess but it would be justified)

n3fit/src/n3fit/scripts/n3fit_exec.py
scarlehoff: Yes. My point is that if it exist and it is not explicitly `False` then it should default to True.

RoyStegeman:
I know 4.0.6 was very recently, but I also thought it would be useful to have a clear reference point since if we just say in the docs that it's stored by default, that might be a bit misleading. Anyway, I don't really care, if you want I can just remove the mention of 4.0.7 and merge?

n3fit/src/n3fit/scripts/n3fit_exec.py
RoyStegeman: But it doesn't work that way further down the pipeline (I didn't look at the code, just checked by running it)

scarlehoff:
Ah, maybe I'm optimistic but I was actually thinking that 4.0.7 will be out soon!
If you want you can write "up to 4.0.6" and that will be true in any case.

n3fit/src/n3fit/scripts/n3fit_exec.py
scarlehoff: Ah, then it should fail if it is not a boolean.

RoyStegeman:
I see, I guess 4.0.7 can be mentioned before it's out. As long as it's not skipped, it's fine for me.

n3fit/src/n3fit/scripts/n3fit_exec.py
RoyStegeman: Well, I don't disagree but we do barely any typechecks, so I don't think that's for this PR to address. There must be lots of these situations in this repo if you'd actively go looking.

scarlehoff: Yes, there will be. But given that we just found one of those situations let's fix it.

This is happening at the level of `n3fit_exec` so nothing has been computed yet. It is a great place to fix the problem.

RoyStegeman:
Is this what you had in mind?

n3fit/src/n3fit/scripts/n3fit_exec.py
```diff
@@ -133,7 +133,9 @@ def from_yaml(cls, o, *args, **kwargs):
 
         N3FIT_FIXED_CONFIG['actions_'].append(namespace + "performfit")
 
-        if file_content["fitting"].get("savepseudodata"):
+        if file_content["fitting"].get("savepseudodata", True):
+            if file_content["fitting"]["savepseudodata"] != True:
+                raise TypeError("fitting::savepseudodata is neither True nor False")
```
scarlehoff, line 138: Yes, I'm happy enough with that.

Also a nice opportunity to add a walrus...
```suggestion
        if fps := file_content["fitting"].get("savepseudodata", True):
            if fps != True:
                raise TypeError(f"fitting::savepseudodata is neither True nor False ({fps})")
```
(feel free to ignore this whilst sighing :P)

RoyStegeman: Fair enough, we don't support 3.7 anyway

PR #1671, opened 2023-02-17
only build docs for 3.9

scarlehoff:
As a temporary solution to the issues with the docs in 3.10 ?


Zaharid:
I think you are going to have to have the constraint selectively for python 3.9.

scarlehoff:
> I think you are going to have to have the constraint selectively for python 3.9.

What do you mean? Isn't this what the == does?

PR #1673, opened 2023-02-18
Tensorflow in macos

scarlehoff:
I don't remember why we had this constraint here, but it bit me yesterday... so let us remove it. The fit works well in mac when you install things manually so this limitation is artificial.

I'll see whether I can get my hands onto a few different types of mac so I can test it in more than one computer before merging.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -24,8 +24,7 @@ requirements:
         - python
         - numpy
     run:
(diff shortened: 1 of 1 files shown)


scarlehoff:
@Radonirinaunimi checked that n3fit works in mac os so I'll merge this so that the 3.9 and 3.10 conda packages for mac include tensorflow. Then I'll check whether something has to be done for M1 (since we only compile for x86)

PR #1678, opened 2023-02-23
New CommonData Reader

scarlehoff:
This branch / PR is the _final_ reader for the new commondata files.

As you can see, this is a bit less ambitious than the previous PR. I've added a reader and _only_ a reader. The changes to other parts of the code are minimal and well contained*. The reader is able to read the new commondata but the output is an object of the old type. This will allow me to keep this branch on track with `master`.

@enocera in principle people can branch out of this branch to implement the new dataset so that they can test. As long as they don't change the code itself there will be no conflicts. If people find it difficult I can test the new commondatas myself as they get implemented.

@Zaharid since this branch is only going to add a reader and thus there will be no conflicts with master, feel free to comment or even modify the parsers to your heart's content. This is even _almost_ orthogonal to people's implementation  of the new data so it should be fine.

### For people implementing new commondata:

A quick way of testing whether the new data that you implemented is compatible with this branch is just to checkout this branch (or even clone it in a different folder) and then run the following snippet:

```python
from pathlib import Path
from validphys.commondataparser import parse_commondata_folder
# Put here the path to your new implemented dataset
path_to_folder = Path("/path/to/your/data/DATASET")

new_cd = parse_commondata_folder(path_to_folder)

print(new_cd.commondata_table)
```

The result should be equal to that of the old version:
```python
from validphys.api import API
old_cd = API.dataset(dataset_input={"dataset": "DATASET"}, theoryid=200, use_cuts="internal").load_commondata()
print(old_cd.commondata_table)
```

*note that this is possible only now that there is no longer the need to keep C++ compatible or literal objects in the pipeline.


scarlehoff:
@enocera how can I check that I'm reading all data and uncertainties correctly? The number of systematics in this commondata is not the same as in the old one, isn't it? Or the old one contains all the variations in the same table?

Screenshot 2023-02-23 at 14 50 52

vs

Screenshot 2023-02-23 at 14 51 05

so, data and statistics seem correct but the systematic part not really?

Also, the process name is different. Is this fine? I'm using the `nnpdf31_process` key.

Zaharid:
I think you may have merged something in a funny way: There are a bunch of duplicated classes such as `ValidKinematics`, `Variant` and so on.

enocera:
> @enocera how can I check that I'm reading all data and uncertainties correctly? The number of systematics in this commondata is not the same as in the old one, isn't it? 

It isn't indeed. I realise that all the data set variants are actually the same (they shouldn't) and that they correspond to the variant of the data set with no nuclear uncertainties. So the extra systematics in the old data set are for the extra nuclear uncertainties that have not been implemented (yet?) in any of the variants of the new data set.

> 
> Also, the process name is different. Is this fine? I'm using the `nnpdf31_process` key.

That is just a mistake.

scarlehoff:
> one goal I definitively want to keep is that data files work fine in an incomplete state

Having the theory be optional (so that one can deal with the data in _some ways_ before having a prediction for it) I think it is a good idea.
However, this  

> with only kinematics and metadata being strictly required

is too lenient.

Of the current set of required options (I've taken out theory)

```yaml
    setname: str
    ndata: int
    observable: dict
    kinematics: ValidKinematics
    kinematic_coverage: dict
    data_central: ValidPath
    data_uncertainties: list[ValidPath]
    dataset_label: str
    plot_x: str
    figure_by: list[str]
    nnpdf_metadata: dict
    version: int
```

I think only `nnpdf_metadata` could be made optional. All the others I believe are important from the first minute to have a valid implementation.

Zaharid:
I meant that out of the possible external per-point files (fk-{stuff}, uncertainties, experimental values, kinematics), only the kinematics are intrinsic to what the dataset it, with the rest being optional and subject to being overwritten by variants.

I don't have strong opinions on what should be required wrt the top level metadata but, if we start more strict, we can always add defaults later on.

Zaharid:
I guess it follows that the current `coredata.Commondata` structure is but one of the possible views of the data, which is enabled when there are central values and uncertainties. I guess that suggests that the parser for commondata (or at any rate something in config) should be more complicated (by having checks for having the relevant files).

We could also use a bunch of boolean methods for checks like `has_experimental_values`, `has_theory` and so on.

Zaharid:
Would you like to move the notebook to some example directory or similar?

scarlehoff:
The notebook is just there for people to use it while they are implementing new cd. It won't be merged to master.

> should be more complicated (by having checks for having the relevant files)

I was planning to check on construction. But if we allow for them to be empty I guess yes, the parser will need to do that.

> We could also use a bunch of boolean methods for checks like has_experimental_values, has_theory and so on.

sure

>  the current coredata.Commondata structure is but one of the possible views of the data

And will be the only one for the time being I guess. It is just the CommonMetaData that will be allowed to have missing pieces (so that if in the future there's an use case for a half-constructed CommonData it can be done) but I consider the current* `coredata.Commondata` as the final realisation of the dataset.
 
*understanding current as "whatever is in master". I'm happy changing that and as I said at the beginning I'll keep rebasing this branch so that it is able to track whatever the CommonData needs to have in master.

Zaharid:
> And will be the only one for the time being I guess. It is just the CommonMetaData that will be allowed to have missing pieces (so that if in the future there's an use case for a half-constructed CommonData it can be done) but I consider the current* `coredata.Commondata` as the final realisation of the dataset.


I guess there are a few possible ways to go about it, which have different trade offs in terms of how much code we break and how elegant they are. For example, we have said that we want  positivity predictions to be basically commondata files without experimental central values (which is why the central values are in a separate, optional, file, different from the uncertainties). We could make that be a "normal" commondata file that just returns zero when asked about central values,. That is close what we have now, but less nice.  We could make the commondata object itself have only optional central values, but that would break a lot of stuff unless we walk over the whole code and add checks as needed. 

Or we could have a bunch of different "final realisations" that depend on what capabilities an action needs (e.g. the inputs would be things like `theory_predictions`, `experimental_values`, `systematics_matrix` and so on; maybe wrapped with the relevant metadata), in which case, we would have to have several data structures. This is closes to how I envision it. We would start only with commondata but move positivity to that framework fairly quickly (likely before any of this gets merged into master). This is closer to how I envision it.

Zaharid:
Actually would be kind of nice to have some example notebook. People do seem to like those.

scarlehoff:
> People do seem to like those.

But then the example notebooks need to be maintained. And so they would need to be tested, etc, etc.

> Or we could have a bunch of different "final realisations" that depend on what capabilities an action needs (e.g. the inputs would be things like theory_predictions, experimental_values, systematics_matrix and so on; maybe wrapped with the relevant metadata)

This I like. It was also my original idea (two branches ago)

> We would start only with commondata but move positivity to that framework fairly quickly (likely before any of this gets merged into master)

If you mean, "this is done _in master_ and then this is modified to follow that master as required" I completely agree. I'm not planning to merge this until all the new commondata is ready.

I'm currently pruning the dangling methods, attributes and functions that have been left from the python commondata and closure test. That will allow us to take out the C++. Then I'll try to unify the interfaces (so we don't have things like the `len` / `ndata` problem of the other day or the fact that currently `CommonData` gives the central value with two different methods). I'm hoping that will also help me with the to-do in the OP about the commonmetadata.

scarlehoff:
The reader is now updated for the new format.

The only thing I'm not happy with how I've solved (@Zaharid) is the way I'm keeping track of the parent since with validphys I'm loading the SetMetaData which can include many ObservableMetaData. Then I select a particular observable.

But I would like this Observable to still know everything about its parent Set. My easy solution is to say, at loading time (since there is only one way to load it and it is the parent doing it), `observable._parent = self` but that can be broken. Since it is an attribute I could be evil and say somewhere else `._parent = None`... but maybe I'm overthinking this.

For the rest, nothing changes with respect to the previous loader. @t7phy if you update the top data to use the new naming format (maybe you already did, I don't remember right now) and @enocera adds a few DIS datasets (with variants, which is something not well tested yet!) I'll update the notebook to do a test to ensure that the data/uncertainties/kinematics have not changed (I would like to do such a test before merging this into master).

I believe it might be interesting to merge this into master in a way in which both commondata formats are allowed at the same time since it will surely help people with debugging and porting (but this is a discussion for another moment ofc)

validphys2/src/validphys/commondataparser.py
```diff
@@ -41,6 +67,363 @@
...
+
+@dataclasses.dataclass
+class Variant:
+    """Defines the keys of the CommonMetaData that can be overwritten"""
+
+    data_uncertainties: list[ValidPath]
+
+
+@Parser
+def ValidVariants(variant_dict: dict) -> Dict[str, Variant]:
```
Zaharid, line 168: This is the same as

```
ValidVariants = Dict[str, Variant]
```

validphys2/src/validphys/commondataparser.py
```diff
@@ -41,6 +67,363 @@
 }
 
 
+@Parser
+def ValidPath(path_str: str) -> Path:
+    """Parse strings into paths"""
+    return Path(path_str)
+
+
+@Parser
```
Zaharid, line 76: If we don't care about supporting random-case stuff, we can as well have `Literal[*OP]` instead of this.

scarlehoff: You mean you want to support them or you don't? I don't mind either way tbh.

t7phy:
> For the rest, nothing changes with respect to the previous loader. @t7phy if you update the top data to use the new naming format (maybe you already did, I don't remember right now) and @enocera adds a few DIS datasets (with variants, which is something not well tested yet!) I'll update the notebook to do a test to ensure that the data/uncertainties/kinematics have not changed (I would like to do such a test before merging this into master).

#1684 is now updated for testing with the reader based all new agreed conventions. Have a go at the reader test, in case I have missed something, please let me know. @scarlehoff

scarlehoff:
> At this point it would really help to have https://github.com/NNPDF/nnpdf/issues/1691 , because it is starting to confuse me.

I think we are close to be able to document it.

> Do I remember correctly that "all the things about the parent" are the metadata?

No. The metadata we are interested in within vp is the `ObservableMetaData` (plotting information, ndata, etc). The parent only contains information about the folder where the observable is, the experiment and the hepdata/arxiv/etc information.

t7phy:
@scarlehoff could you please test the reader on this specific dataset/folder: https://github.com/NNPDF/nnpdf/tree/gluon_pdf_ncd/buildmaster/ATLAS_TTBAR_8TEV_LJ_DIF   As you know, for the other PR i still need to delete `theory`, add 3rd kinematic variable and so on. This particular dataset should have every single point that we have agreed on in the last few days. Have a go and please let me know if this is fine.

t7phy:
> You have to remove the `file` key.
fixed it, thanks!
> But I'll try it tomorrow

Ok, you can test this and also #1684 is now fixed. Besides these, the following two can also be tested:
https://github.com/NNPDF/nnpdf/tree/gluon_pdf_ncd/buildmaster/H1_1JET_319GEV_351PB-1
https://github.com/NNPDF/nnpdf/tree/gluon_pdf_ncd/buildmaster/H1_2JET_319GEV_351PB-1

so I expect the 4 of these so far to be fully compatible with the reader (I hope).

enocera:
> > You have to remove the `file` key.
> > fixed it, thanks!
> > But I'll try it tomorrow
> 
> Ok, you can test this and also #1684 is now fixed. Besides these, the following two can also be tested: https://github.com/NNPDF/nnpdf/tree/gluon_pdf_ncd/buildmaster/H1_1JET_319GEV_351PB-1 https://github.com/NNPDF/nnpdf/tree/gluon_pdf_ncd/buildmaster/H1_2JET_319GEV_351PB-1
> 
> so I expect the 4 of these so far to be fully compatible with the reader (I hope).

@t7phy Although these are for new data sets (not in NNPDF4.0), aren't they? So perhaps @scarlehoff wants to try first the old data sets.

t7phy:
> > > You have to remove the `file` key.
> > > fixed it, thanks!
> > > But I'll try it tomorrow
> > 
> > 
> > Ok, you can test this and also #1684 is now fixed. Besides these, the following two can also be tested: https://github.com/NNPDF/nnpdf/tree/gluon_pdf_ncd/buildmaster/H1_1JET_319GEV_351PB-1 https://github.com/NNPDF/nnpdf/tree/gluon_pdf_ncd/buildmaster/H1_2JET_319GEV_351PB-1
> > so I expect the 4 of these so far to be fully compatible with the reader (I hope).
> 
> @t7phy Although these are for new data sets (not in NNPDF4.0), aren't they? So perhaps @scarlehoff wants to try first the old data sets.
https://github.com/NNPDF/nnpdf/tree/gluon_pdf_ncd/buildmaster/ATLAS_TTBAR_8TEV_LJ_DIF and #1684 are old, the other 2 are new.

scarlehoff:
RE: testing other datasets, let's keep a document in which you can list the datasets that are already implemented together with their old name. That way I can test them (and we keep them better organized). I've opened such an issue here: #1709

@enocera one question about kinematics, is their order important? Also, do we want to keep the same variables as the old ones? (for instance, for E605 @tgiani has used s instead of sqrt(s)... not really relevant as a difference, but again, to know how strict should I be when checking).

enocera:
@scarlehoff I'd say that the order and the values of the kin variables are important insofar as we want to have validphys deal correctly with data/theory comparisons and plot labelling. So I'd make an effort to keep the old convention as much as possible.

t7phy:
> @scarlehoff I'd say that the order and the values of the kin variables are important insofar as we want to have validphys deal correctly with data/theory comparisons and plot labelling. So I'd make an effort to keep the old convention as much as possible.

@scarlehoff  @enocera just to understand, the consistency in the order of kinematic variables just needs to be maintained in the metadata, correct? because I assume the kinematics vars from files will be taken by their key name, which is obtained from the metadata

scarlehoff:
Yes. The `variables` field has to be in the right order.

validphys2/src/validphys/commondataparser.py
Zaharid: I think we can make `ValidOperation = typing.Literal[*OP]`. The last version  is incorrect in that it would break if you pass a non sting (when calling `upper`)

scarlehoff: At the moment OP is in another module and would create circular dependencies. I've casted the str (and I've also covered the case `None` since the `null` as entering `None`). We can decide at the end whether to support
 random-case or not. Same for `ValidPath` above, might get converted to Path with no checks.

Zaharid: Then can we not copy the dict keys (or even move that dict and related to a separate module?)

Why do we want to support None at all? Can we not make it "NULL" only?

Also the latest version will sill crash for e.g. `op_str = ["patata"]`; we really want to have an annotation for the input argument (which could be `str`, and you could have that the type is `Optional[ValidOperation]`).

scarlehoff: My idea is to move the operations when I'm done with the parser. They should not be part of `convolutions` anyway, but of the coredata or parser. But I want to limit changes to other parts of the code until I'm done so that I can rebase on top of master without suffering too much.

> Why do we want to support None at all?

Because `yaml.safe_load("null")` is `None` so I'm making it "NULL" when it comes as `None` since it's likely that people will forget the " when writing the operations. 

Maybe we want to be very strict but I just tried with your suggestion and it doesn't work but even if it worked, I think an error msg which will say something like `you tried to use None, accepted values are "NULL"` when they have written `operation: null` is only going to annoy people.

Zaharid: > My idea is to move the operations when I'm done with the parser. They should not be part of `convolutions` anyway, but of the coredata or parser. But I want to limit changes to other parts of the code until I'm done so that I can rebase on top of master without suffering too much.

I'd suggest just copying the keys then. For one, it leads to better error messages.

scarlehoff: You mean now or as the final version?

> Personally, I could live with that on my conscience.

I don't know. Given that it is a mistake so easy to anticipate it seems a bit mean :__ Also, it's on us for choosing "NULL" as the name of the operation I guess.

Zaharid: > You mean now or as the final version?

Not sure what the question is referring to, but if it is the aliasing thing, then definitively yes, before there are more of these files and conflicting versions of the format. For using Literal, it doesn't matter that much, but it would be a mild improvement for some duplication.

scarlehoff: Just to make sure I understand the proposal:

Change all new commondata files to use `ID` instead of `NULL`.
For old theories, since they will have `NULL` in their compound files, change it directly to `ID` upon reading.

Zaharid: The proposal would be

```diff
diff --git a/validphys2/src/validphys/convolution.py b/validphys2/src/validphys/convolution.py
index 94e4c7fcd..3b54b26db 100644
--- a/validphys2/src/validphys/convolution.py
+++ b/validphys2/src/validphys/convolution.py
@@ -93,6 +93,7 @@ OP = {
     "COM": _com,
     "SMT": _smt,
     "NULL": _id,
+    "ID": _id,
 }
 
 class PredictionsRequireCutsError(Exception): pass
```

and the banning "NULL" from new files.

validphys2/src/validphys/core.py
```diff
@@ -215,25 +216,37 @@ def get_members(self):
...
         return self.metadata.name
 
     @property
     def nsys(self):
-        return self.metadata.nsys
+        if self.legacy:
+            return self.metadata.nsys
+        else:
+            cd = self.load()
+            return cd.nsys
```
comane, line 241: Hi @scarlehoff, I think that loading a Commondata instance within CommondataSpec goes against the idea of CommondataSpec.

The reason is that CommondataSpec has a production rule, `produce_commondata`, that its being loaded during `reportengine`'s "compilation". As such it should only contain metadata and not actual data.

There might have been already a discussion about it, so sorry if this has been already discussed, but isn't it possible to add `nsys` to metadata.yaml as well?
I mean one can do it after having generated the artificial systematics with `filter.py`

validphys2/src/validphys/core.py
```diff
@@ -215,25 +216,37 @@ def get_members(self):
...
 
     @property
     def ndata(self):
-        return self.metadata.ndata
+        if self.legacy:
+            return self.metadata.ndata
+        else:
+            cd = self.load()
+            return cd.nsys
```
comane, line 249:
```suggestion
            cd = self.load()
            return cd.ndata
```

Same reasoning applies here. Here it is even easier since `ndata` is always known a priori, and hence is always included in metadata.yaml

validphys2/src/validphys/core.py
```diff
@@ -251,9 +264,13 @@ def __str__(self):
     def __iter__(self):
         return iter((self.datafile, self.sysfile, self.plotfiles))
 
+    # TODO: one of the two functions below needs to go
```
comane, line 267: I agree with this, and I would tend to say that we can just keep `load`.

However, do we want to change this in this PR, or in a later one?

Note that there are some dependencies on `load_commondata_instance` in several modules, e.g. `pseudodata.py`

validphys2/src/validphys/coredata.py
```diff
@@ -317,10 +318,16 @@ def multiplicative_errors(self):
         mult_systype = self.systype_table[self.systype_table["type"] == "MULT"]
         # NOTE: Index with list here so that return is always a DataFrame, even
         # if N_sys = 1 (else a Series could be returned)
-        mult_table = self.systematics_table.loc[:, ["MULT"]]
-        # Minus 1 because iloc starts from 0, while the systype counting starts
-        # from 1.
-        mult_table = mult_table.iloc[:, mult_systype.index - 1]
+        mult_table = self.systematics_table.filter(like = "MULT")
+        # TODO: check whether the `filter` creates a problem with n_sys = 1, if not remove both
+        # note and todo
```
comane, line 323: seems to work well when `n_sys=1`
```
In [19]: cd
Out[19]: CommonData(setname='ATLAS_SINGLETOP_TCH_R_8TEV', ndata=1, commondataproc='INC', nkin=3, nsys=1, legacy=True)

In [20]: cd.systematics_table.filter(like = "MULT")
Out[20]: 
           MULT
entry          
1      4.069767

```

validphys2/src/validphys/coredata.py
```diff
@@ -317,10 +318,16 @@ def multiplicative_errors(self):
...
+        #mult_table = self.systematics_table.loc[:, ["MULT"]]
+
+        if self.legacy:
+            # Minus 1 because iloc starts from 0, while the systype counting starts
+            # from 1.
+            mult_table = mult_table.iloc[:, mult_systype.index - 1]
```
comane, line 329: Why is this done for legacy datasets only?

validphys2/src/validphys/core.py
scarlehoff: At the end of the PR, just before merging, I can do a search or replace of `load_commondata_instance` by `load`.

validphys2/src/validphys/coredata.py
scarlehoff: The uncertainties in the new format are either ADD or MULT. In the previous format you had all the uncertainties in one file and then another file which was the one deciding what they were (and that's why the two steps were needed)

validphys2/src/validphys/core.py
comane, line 249:
```suggestion
            return cd.ndata
```
shouldn't this be `ndata`?

scarlehoff: yes! Thanks!

validphys2/src/validphys/commondataparser.py
```diff
@@ -41,20 +83,585 @@
...
+    return pd.concat(kin_dict, axis=1, names=[_INDEX_NAME]).swaplevel(0, 1).T
+
+
+def parse_new_metadata(metadata_file, observable_name, variants=[]):
+    """Given a metadata file in the new format and the specific observable to be read
+    load and parse the metadata and select the observable.
+    If any variants are selected, apply them.
+    """
+    # Note: we are re-loading many times the same yaml file, possibly a good target for lru_cache
+    set_metadata = parse_yaml_inp(metadata_file, SetMetaData)
```
comane, line 537:
```suggestion
    metadata_file = Path(metadata_file)
    set_metadata = parse_yaml_inp(metadata_file, SetMetaData)
```

Otherwise if `metadata_file` is a standard str pointing at the path of metadata.yaml, the parser would crush since strings have no `.parent` attribute

validphys2/src/validphys/commondataparser.py
```diff
@@ -41,20 +83,585 @@
...
+
+    # And apply variants
+    for variant in variants:
+        metadata = metadata.apply_variant(variant)
+
+    return metadata
+
+
+def parse_commondata_new(metadata):
+    """
```
comane, line 551: A metadata `ObservableMetaData` object has a `variants`  attribute.
How is this specified/enforced when loading a new commondata instance?

If I remember correctly there was a parse_commondata_new function in which the variant could be specified
(I think it was something like parse_commondata_new(datasetname, variants=['decorrelated']))

scarlehoff: When you ask for a dataset with a variant, the metadata is already the "variant" metadata. It is applied in `parse_new_metadata(metadata_file, observable_name, variants=[]):`

I'm doing it like this because (when variants are selected) it should be applied immediately so I moved it to just after the `metadata_file` is loaded.

scarlehoff:
Everything that's left now are regression tests that I want to wait until the end to regenerate. As far as I can see all of the failures are due to the names being different.

They might hide actual problems because the name can actually have an effect (for instance, when sorting dataframes) so I'll go through them carefully anyway. However, for the rest this is finished.

If you want to test it, you can install this branch in complete isolation by doing the following:

```
python -m venv nnpdf_venv
source nnpdf_venv/bin/activate
python -m pip install git+https://github.com/NNPDF/nnpdf.git@final_reader_for_new_commondata_mk2
```

Then you can try your favourite `n3fit` or `validpyhs` runcards. If you get errors, please report them.

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,21 +108,805 @@
...
+        """
+        # Check that the data_central is empty if and only if the dataset is a positivity set
+        if self.data_central is None and not self.is_lagrange_multiplier:
+            raise ValidationError(f"Missing `data_central` field for {self.name}")
+
+        # Check that plotting.plot_x is being filled
+        if self.plotting.plot_x is None:
+            ermsg = f"No variable selected as x-axis in the plot for {self.name}. Please add plotting::plot_x."
+            if self.plotting.x is not None:
+                ermsg += "If you are using `plotting::x` please change it to `plotting::plot_x`"
```
Radonirinaunimi, line 428:
```suggestion
            ermsg = f"No variable selected as x-axis in the plot for {self.name}. Please add `plotting::plot_x`."
            if self.plotting.x is not None:
                ermsg += "Please replace `plotting::x` with `plotting::plot_x`."
```

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,21 +108,805 @@
...
+            if var not in self.kinematics.variables:
+                raise ValidationError(
+                    f"Variable {var} is in `kinematic_coverage` but not included in `kinematics` for {self.name}"
+                )
+
+        if len(self.kinematic_coverage) > 3:
+            raise ValidationError(
+                "Only a maximum of 3 variables can be used for `kinematic_coverage`"
+            )
+
```
Radonirinaunimi, line 442: One thing that I have found myself often when implementing new commondata was the fact that the number of of elements inside the `bins` key of `uncertainties.yaml` do not correspond to the number of datapoints. Would it also be possible to check this here and outputs instructive errors if not.

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,21 +108,805 @@
...
+                self.kinematic_coverage += [f"extra_{i}" for i in range(diff_to_3)]
+
+        object.__setattr__(self, 'process_type', self.process_type.upper())
+
+    def check(self):
+        """Various checks to apply manually to the observable before it is used anywhere
+        These are not part of the __post_init__ call since they can only happen after the metadata
+        has been read and the observable selected.
+        """
+        # Check that the data_central is empty if and only if the dataset is a positivity set
```
Radonirinaunimi, line 420: This also applies to integrability dataset, not only to positivity.
```suggestion
        # Check that the data_central is empty if and only if the dataset is a positivity/integrability set
```

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,21 +108,805 @@
...
+        else:
+            datayaml = _quick_yaml_load(self.path_data_central)
+            data = datayaml["data_central"]
+        data_df = pd.DataFrame(data, index=range(1, self.ndata + 1), columns=["data"])
+        data_df.index.name = _INDEX_NAME
+        return data_df
+
+    @property
+    def paths_uncertainties(self):
+        return [self._parent.folder / i for i in self.data_uncertainties]
```
Radonirinaunimi, line 498: I maybe missing the rational here but why should all the uncertainty variants be loaded once?

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,21 +108,805 @@
...
+        -------
+        pd.DataFrame
+            a dataframe containing the kinematics
+        """
+        kinematics_file = self.path_kinematics
+        kinyaml = _quick_yaml_load(kinematics_file)
+
+        kin_dict = {}
+        for i, dbin in enumerate(kinyaml["bins"]):
+            bin_index = i + 1
```
Radonirinaunimi, line 554:
```suggestion
        for bin_index, dbin in enumerate(kinyaml["bins"], start=1):
```

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,21 +108,805 @@
...
+
+    uncertainties_df.columns = new_columns
+
+    commondata_table = pd.concat([kin_df, data_df, uncertainties_df], axis=1)
+    systype_table = pd.DataFrame(systypes, index=range(1, nsys + 1))
+    systype_table.index.name = "sys_index"
+
+    # TODO: Legacy compatibility
+    # 1. Add a stat column if it doesn't exist
+    # 2. Transform multiplicatie uncertainties into % as it was done in the older version
```
Radonirinaunimi, line 858:
```suggestion
    # 2. Transform multiplicative uncertainties into % as it was done in the older version
```

validphys2/src/validphys/config.py
```diff
@@ -586,7 +633,7 @@ def produce_dataset(
             raise ConfigError(e)
 
         if check_plotting:
-            from validphys.plotoptions import get_info
+            from validphys.plotoptions.core import get_info
```
Radonirinaunimi, line 636: Already imported above.
```suggestion
```

validphys2/src/validphys/datafiles/__init__.py
```diff
@@ -1,5 +1,65 @@
...
+                f"I can not translate the combination of {dataset_name} and sys: {sys}. Please report this."
+            )
+        variant += f"_{sys}"
+
+    return new_name, variant
+
+
+@lru_cache
+def new_to_legacy_map(dataset_name, variant_used):
+    """Loop over the dictionary and fing the right dataset"""
```
Radonirinaunimi, line 44:
```suggestion
    """Loop over the dictionary and find the right dataset"""
```

validphys2/src/validphys/filters.py
```diff
@@ -43,6 +44,41 @@
...
+
+    def substitute(pr_v, cd_x):
+        if pr_v in process_vars and cd_x in variables:
+            variables[variables.index(cd_x)] = pr_v
+
+    substitute("eta", "y")
+    substitute("eta", "eta")
+    substitute("etay", "eta")
+    substitute("etay", "y")
+    substitute("p_T2", "pT_sqr")
```
Radonirinaunimi, line 72: In hindsight, this variable can be easily confused with `pT_sqrt`. I am not sure if is still easily possible to change this.
```suggestion
    substitute("p_T2", "pT_sqrd")
```

validphys2/src/validphys/loader.py
```diff
@@ -326,26 +331,66 @@ def available_theories(self):
...
         }
 
+    @property
+    @functools.lru_cache()
+    def available_datasets(self):
+        """Provide all available datasets.
+        At the moment this only returns old datasets for which we have a translation available
+        """
+        skip = ("POS", "INTEG")
+        old_datasets = [i for i in legacy_to_new_mapping.keys() if not i.startswith(skip)]
```
Radonirinaunimi, line 354: I guess this is no longer necessary?
```suggestion
        old_datasets = [i for i in legacy_to_new_mapping.keys()]
```

validphys2/src/validphys/plotoptions/kintransforms.py
```diff
@@ -69,6 +69,31 @@ def xq2map(self, k1:np.array,k2:np.array,k3:np.array,**extra_labels) -> (np.arra
...
+    """
+    def __call__(self, k1, k2, k3):
+        return k1, k2, k3
+
+    def new_labels(self, k1, k2, k3):
+        return k1, k2, k3
+
+    def xq2map(self, k1, k2, k3, **extra_labels):
+        # This is going to be a problem
+        return k1, k2
```
Radonirinaunimi, line 94: What exactly is this supposed to do?

validphys2/src/validphys/plotoptions/plottingoptions.py
```diff
@@ -0,0 +1,105 @@
...
+            for el in self.line_by:
+                if el in labeler_functions:
+                    self.func_labels[el] = labeler_functions[el]
+
+    def parse_x(self):
+        if self.x is not None and self.x not in self.all_labels:
+            raise ValidationError(
+                f"The label {self.x} is not in the set of known labels {self.all_labels}"
+            )
+
```
Radonirinaunimi, line 87: Shouldn't [results_transform](https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/plotoptions/resulttransforms.py) also be parsed here? Some datasets require this for the plotting.

Radonirinaunimi (commented):
These are some first-round comments. Given that these are huge amount of changes, I'd need to go over them one more time.

validphys2/src/validphys/commondataparser.py
scarlehoff: You mean, you'd like an explicit error when that's the case? (I guess right now it fails in some magic way?)

validphys2/src/validphys/commondataparser.py
scarlehoff: This is a list to allow people to separate uncertainties. This was requested at some point in the past and have stayed like that. I don't think anybody used it.

The idea was that you would have a number of uncertainties that are shared between variants, and then some that are unique, then you can have variants like:

```
my_variant_one:
    uncertainties: 
         - file_shared
         - file_unique_one
my_variant_two:
    uncertainties: 
         - file_shared
         - file_unique_two
```

validphys2/src/validphys/filters.py
scarlehoff: The pT squared is now `pT2`

That list of substitutions is there only for debugging, that function will be removed at the end (once I've checked everything and can then substitute the variables at the top of the file).

validphys2/src/validphys/plotoptions/kintransforms.py
scarlehoff: Making sure that the code doesn't crash. But the results are wrong (the xq2map is the one used for the kinematic coverage plot, so when a new data with `identity` is being used, it just outputs k1 and k2 so that the program doesn't crash but the results are wrong).

validphys2/src/validphys/plotoptions/plottingoptions.py
scarlehoff: Results transforms hasn't changed (at least I think it hasn't) for any of the new datasets so it can be taken raw from the metadata (like 49).

This is akin to what was done before, only then it was taken from the plotting.

As I said, _I think_ it hasn't failed, I went over the plots and didn't see anything broken, but there were many plots so the changes that I missed something are high.

validphys2/src/validphys/commondataparser.py
Radonirinaunimi: In what I observed, it fails only when computing the $\chi^2$ due to a mismatch in shapes but with non-instructive errors making it hard understanding what went wrong. On the other hand, I don't want to open the pandora's box here as one could argue that a test that checks that the number of kinematic points and datapoints also match is required here, and other tests as well.

validphys2/src/validphys/commondataparser.py
Radonirinaunimi: I see, it is good to have such flexibility.

validphys2/src/validphys/plotoptions/plottingoptions.py
Radonirinaunimi: Could you share the link to the plots? I can try to have a look as well.

scarlehoff: I thought I had but I cannot find the links so I clearly didn't.

Atlas: https://vp.nnpdf.science/0YL4eaPTT7eR3wM9IQwj5w==
BCDMS: https://vp.nnpdf.science/E0IDIgWFRF6tvdFk3pD0mA==
Chorus: https://vp.nnpdf.science/wCFRKBNsSA2U2O_6Sa75Zw==
CMS: https://vp.nnpdf.science/oSZLrPg3Tyyf-i2mE33G-Q==
HERA: https://vp.nnpdf.science/UHU-TYLJQCuE-8lRHe8Aog==
FTDY: https://vp.nnpdf.science/Toy_r6uFRm-h1oUFEZF6hw==
NMC:  https://vp.nnpdf.science/-2VyNN3CTHWnb26fUQkUJQ==
Nutev: https://vp.nnpdf.science/YPAQHvMtTeyBuNq12nl6RQ==
Slac: https://vp.nnpdf.science/chwFM_lJR025vREJ1I9VPQ==

I haven't been able to upload tevatron because the INFN has decided that I was connecting many times (I just looped over all the reports) and I'm banned for a few minutes or hours xD
LHCb I didn't do because those datasets were not ported automatically.

validphys2/src/validphys/commondataparser.py
scarlehoff: Ok, I'll add the check. This can only be done upon load though (so you can use it wrongly for a while before realizing)

validphys2/src/validphys/core.py
scarlehoff: There's no way around this so I'm going to mark it as resolved.

Hopefully there are no many utilities using nsys without having loaded the data first so it should have no effect.

validphys2/src/validphys/plotoptions/plottingoptions.py
Radonirinaunimi: Looks good! I went through the reports and didn't notice anything wrong.

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,22 +108,825 @@
...
+    def is_integrability(self):
+        return self.setname.startswith("NNPDF_INTEG")
+
+    @property
+    def is_lagrange_multiplier(self):
+        return self.is_positivity or self.is_integrability
+
+    @property
+    def path_data_central(self):
+        return self._parent.folder / self.data_central
```
Radonirinaunimi, line 486:
```suggestion
```
Do we really need this to be a separate attribute (same for `paths_uncertainties`)? This is only used in `load_data_central` function and therefore can be defined there.

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,22 +108,825 @@
...
+                f"The number of bins in {self.path_data_central} does not match ndata={self.ndata}"
+            )
+
+        data_df = pd.DataFrame(data, index=range(1, self.ndata + 1), columns=["data"])
+        data_df.index.name = _INDEX_NAME
+        return data_df
+
+    @property
+    def paths_uncertainties(self):
+        return [self._parent.folder / i for i in self.data_uncertainties]
```
Radonirinaunimi, line 513:
```suggestion
```
See comment in `path_data_central`.

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,22 +108,825 @@
...
+
+            # I'm guessing there will be a better way of doing this than calling  dataframe twice for the same thing?
+            final_df = pd.DataFrame(bin_list, columns=mindex, index=range(1, self.ndata + 1))
+            final_df.index.name = _INDEX_NAME
+            all_df.append(final_df)
+        return pd.concat(all_df, axis=1)
+
+    @property
+    def path_kinematics(self):
+        return self._parent.folder / self.kinematics.file
```
Radonirinaunimi, line 546: Here as well!

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,22 +108,825 @@
...
+        """Return True if this is an automatically ported dataset that has not been updated"""
+        if self.ported_from is None:
+            return False
+
+        # If it is using a legacy variant and has a ported_from field, then it is a ported one
+        if self.applied_variant is not None and self.applied_variant.startswith("legacy"):
+            return True
+
+        # If not using a legacy variant, we consider it ported if the kin variables are still k1,k2,k3
+        return "k1" in self.kinematic_coverage and "k2" in self.kinematic_coverage
```
Radonirinaunimi, line 634: Maybe the following is more correct?
```suggestion
        return set(["k1", "k2", "k3"]) == set(self.kinematic_coverage)
```

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,22 +108,825 @@
...
+        ndata=metadata.ndata,
+        commondataproc=procname,
+        nkin=3,
+        nsys=nsys,
+        commondata_table=commondata_table,
+        systype_table=systype_table,
+        legacy=False,
+        legacy_name=legacy_name,
+        kin_variables=metadata.kinematic_coverage,
+    )
```
Radonirinaunimi, line 904: What about the `systematics_table` argument? In the dataclass definition, that is not Optional.

validphys2/src/validphys/config.py
```diff
@@ -364,14 +365,32 @@ def produce_fitpdfandbasis(self, fitpdf, basisfromfit):
...
+        Old-format commondata will be translated to the new version in this function.
+        """
+        accepted_keys = {"dataset", "sys", "cfac", "frac", "weight", "custom_group", "variant"}
         try:
             name = dataset["dataset"]
             if not isinstance(name, str):
                 raise ConfigError(f"'dataset' must be a string, not {type(name)}")
-            # Check whether this is an integrability or positivity dataset (in the only way we know?)
+            #! Check whether this is an integrability or positivity dataset (in the only way we know?)
             if name.startswith(("INTEG", "POS")):
```
Radonirinaunimi, line 394: I guess these should also change.
```suggestion
            #! Check whether this is an integrability or positivity dataset (in the only way we know?)
            if name.startswith(("NNPDF_INTEG", "NNPDF_POS")):
```

validphys2/src/validphys/datafiles/__init__.py
```diff
@@ -1,5 +1,65 @@
...
+    """Find the new dataset name and variant corresponding to an old dataset
+    and systematics choice"""
+    if dataset_name not in legacy_to_new_mapping:
+        return dataset_name, None
+
+    new_name = legacy_to_new_mapping[dataset_name]
+    if isinstance(new_name, str):
+        if sys is not None:
+            raise KeyError(
+                f"I can not translate the combination of {dataset_name} and sys: {sys}. Please report this."
```
Radonirinaunimi, line 26:
```suggestion
                f"I cannot translate the combination of {dataset_name} and sys: {sys}. Please report this."
```

validphys2/src/validphys/datafiles/__init__.py
```diff
@@ -1,5 +1,65 @@
...
+            )
+        return new_name, None
+
+    variant = new_name.get("variant")
+    new_name = new_name["dataset"]
+    if sys is not None:
+        if variant is None:
+            raise KeyError(
+                f"I can not translate the combination of {dataset_name} and sys: {sys}. Please report this."
```
Radonirinaunimi, line 35:
```suggestion
                f"I cannot translate the combination of {dataset_name} and sys: {sys}. Please report this."
```

Radonirinaunimi (commented):
Here are some more comments after a fresh new look. This is not the full thing though, I still need to continue from `filters.py`. I'll continue again tomorrow.

validphys2/src/validphys/commondataparser.py
```diff
@@ -1,16 +1,83 @@
...
 import dataclasses
+from functools import cached_property, lru_cache
 import logging
 from operator import attrgetter
+from pathlib import Path
+from typing import Any, Dict, Optional
 
+import numpy as np
 import pandas as pd
+from validobj import ValidationError, parse_input
```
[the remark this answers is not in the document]
scarlehoff: Why separate it?

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,22 +108,825 @@
...
+    Similarly to kinlabel, some special cases are taken into account.
+    """
+    if process_type in PROCESS_DESCRIPTION_LABEL:
+        return PROCESS_DESCRIPTION_LABEL[process_type]
+    # If not, is this a DYP- or DIS-like dataset?
+    if process_type[:3] in ("DIS", "DYP"):
+        return _get_process_description(process_type[:3])
+    # Remove pieces of "_" until it is found
+    if len(process_type.split("_")) > 1:
+        return _get_process_description(process_type.rsplit("_", 1)[0])
```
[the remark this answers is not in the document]
scarlehoff: > do you expect a process type to be DISXX or DYXX in which case it shouldn't be because it should always be DIS_XX or DY_XX.

To be completely honest I don't know. This is what was being done in the old reader so I just copied it, I didn't check whether one of them was DISXX.

Hopefully through #1949 this won't be necessary anymore. For now I'll leave this (and put it in _get_ported_kinlabel, since as you say it might occur as well).

validphys2/src/validphys/commondataparser.py
scarlehoff: While in vp it is not used, it is the only way to have access to the actual path which might be useful (for instance, if from pineko you want to read the kinematics or data without loading them) since the metadata itself only contains the name of the file but not the full path (it is also the reason we cannot check for existence while parsing).

validphys2/src/validphys/commondataparser.py
scarlehoff: It is autogenerated during the `__post_init__` of that class.

Radonirinaunimi: Right, but then it should set as Optional (?)

validphys2/src/validphys/commondataparser.py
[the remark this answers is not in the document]
Radonirinaunimi: Maybe for clarity, as these come from different modules, otherwise the linter will complain a lot :/

validphys2/src/validphys/commondataparser.py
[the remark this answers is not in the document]
Radonirinaunimi: At least in the documentation (https://docs.nnpdf.science/data/exp-data-files.html#process-types-and-kinematics) the convention should be that the naming should be constructed in segments - I interpreted this as separated by `_`.

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,22 +108,825 @@
...
+    @classmethod
+    def parser(cls, yaml_file):
+        """The yaml databases in the server use "operands" instead of "FK_tables" """
+        if not yaml_file.exists():
+            raise FileNotFoundError(yaml_file)
+        meta = yaml.safe_load(yaml_file.read_text())
+        # Make sure the operations are upper-cased for compound-compatibility
+        meta["operation"] = "NULL" if meta["operation"] is None else meta["operation"].upper()
+        if "operands" in meta:
+            meta["FK_tables"] = meta.pop("operands")
```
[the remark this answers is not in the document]
scarlehoff: Not going to do it yet (just in case) but it should be already possible.

validphys2/src/validphys/commondataparser.py
Radonirinaunimi: Ah, then it's ok to leave these as they are.

validphys2/src/validphys/commondataparser.py
scarlehoff: tbf it does have a default value set, but I can mark it explicitly

validphys2/src/validphys/commondataparser.py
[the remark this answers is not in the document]
scarlehoff: It should not complain, I think it might be a linter problem: https://github.com/Zaharid/validobj/blob/master/validobj/__init__.py

validphys2/src/validphys/commondataparser.py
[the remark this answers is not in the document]
scarlehoff: Yes, but we also have DIJET for instance so maybe there's some DISHERA or something somewhere?

(I honestly don't know, I first found we  have a few `process_type[:3] == DIS` scattered around the code not that long ago...)

validphys2/src/validphys/coredata.py
```diff
@@ -370,17 +385,31 @@ def systematic_errors(self, central_values=None):
...
-        - A systypes/STYPES_<dataset>.dat file with the error types
-        """
-
-        dat_path = path / f"DATA_{self.setname}.dat"
-        sys_path = path / "systypes" / f"SYSTYPE_{self.setname}_DEFAULT.dat"
-        sys_path.parent.mkdir(exist_ok=True)
-
-        write_systype_to_file(self, sys_path)
-        write_commondata_to_file(self, dat_path)
+    def export_data(self, buffer):
```
Radonirinaunimi, line 388: Since `commondata.export` has been removed, its previous occurrences need to be adjusted as well. Here for example (?):
https://github.com/NNPDF/nnpdf/blob/6d05ace3aa4aa421fecb4702f92cbe6ced0b412a/validphys2/src/validphys/filters.py#L214

Radonirinaunimi (commented):
I think that the following might be a bug.

validphys2/src/validphys/commondataparser.py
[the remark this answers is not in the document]
Radonirinaunimi: Okey, I can live with that.

validphys2/src/validphys/commondataparser.py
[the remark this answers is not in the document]
Radonirinaunimi: Good, thanks!

validphys2/src/validphys/datafiles/__init__.py
```diff
@@ -1,5 +1,65 @@
...
 
+from reportengine.compat import yaml
+
 path_vpdata = pathlib.Path(__file__).parent
-path_commondata = pathlib.Path(__file__).with_name('commondata')
+path_commondata = path_vpdata / "new_commondata"
 path_theorydb = pathlib.Path(__file__).with_name('theory.db')
+
+# VP should not have access to this file, only to the products
+_path_legacy_mapping = path_commondata / "dataset_names.yml"
```
RoyStegeman, line 11: Maybe this should just be made into one line

validphys2/src/validphys/datafiles/__init__.py
```diff
@@ -1,5 +1,65 @@
+from functools import lru_cache
 import pathlib
 
+from reportengine.compat import yaml
+
 path_vpdata = pathlib.Path(__file__).parent
-path_commondata = pathlib.Path(__file__).with_name('commondata')
+path_commondata = path_vpdata / "new_commondata"
 path_theorydb = pathlib.Path(__file__).with_name('theory.db')
```
RoyStegeman, line 8:
```suggestion
```

validphys2/src/validphys/scripts/vp_rebuild_data.py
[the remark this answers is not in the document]
scarlehoff: Same as before (hopefully) but without the need for rebuild (actually, this command was really unnecessary already since many actions would do it automatically anyway). I'll remove the reference to rebuild from the docs.

validphys2/src/validphys/tests/test_pyfkdata.py
```diff
@@ -90,13 +96,13 @@ def test_positivity(pdf_name):
...
         # Use the loader to load the positivity dataset
-        ps = l.check_posset(setname=posset, theoryID=THEORYID, postlambda=1e6)
+        ps = l.check_posset(setname=posset, theoryID=THEORYID_NEW, postlambda=1e6)
         preds = predictions(ps, pdf)
         core_predictions = PositivityResult.from_convolution(pdf, ps)
         assert_allclose(preds.values, core_predictions.rawdata)
         # Now do the same with the API
         api_predictions = API.positivity_predictions_data_result(
-            theoryid=THEORYID, pdf=pdf_name, posdataset={"dataset": posset, "maxlambda": 1e6}
+            theoryid=THEORYID_NEW, pdf=pdf_name, posdataset={"dataset": posset, "maxlambda": 1e6}
```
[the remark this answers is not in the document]
scarlehoff: In a previous iteration of this PR, when both old/new commondata were active, I needed to test with the new theories.
I'll put it back to minimize the unnecessary changes. At some point all the tests will need to be updated anyway.

validphys2/src/validphys/plotoptions/kintransforms.py
RoyStegeman, line 1: Are the kinlables in the module-level docstrings still up-to-date?

validphys2/src/validphys/plotoptions/kintransforms.py
```diff
@@ -69,6 +69,31 @@ def xq2map(self, k1:np.array,k2:np.array,k3:np.array,**extra_labels) -> (np.arra
 import numpy as np
 
 
+# TODO
+class identity:
```
RoyStegeman, line 73:
```suggestion
class Identity:
```
And the corresponding change in all the yaml files

scarlehoff: The convention in this module is that the classes that are to be used directly by the commondata are to be written in lowercase.

btw, once the process options is completed (and all commondata get actually process labels instead of k1, k2, k3...) this whole module will be removed. So probably we are keeping it forever.

RoyStegeman: >  So probably we are keeping it forever.

Seems about right

validphys2/src/validphys/pseudodata.py
```diff
@@ -101,8 +102,21 @@ def read_replica_pseudodata(fit, context_index, replica):
...
+    context_datasets = context_index.get_level_values("dataset").unique()
+    for dsname in pseudodata.index.get_level_values("dataset").unique():
+        if dsname not in context_datasets:
+            new_name, _ = legacy_to_new_map(dsname)
+            mapping[dsname] = new_name
+
+    pseudodata.rename(mapping, level=1, inplace=True)
+
+    pseudodata.sort_index(level=range(1, 3), inplace=True)
     pseudodata.index = sorted_index
```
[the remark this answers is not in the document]
scarlehoff: I don't want to remove it because I'm not sure why that's there and don't want to add (more) noise to this PR.

validphys2/src/validphys/plotoptions/kintransforms.py
scarlehoff: They should.

They don't correspond to what you would see with the new commondata though because the labels are decided at the data level...

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,22 +108,828 @@
...
+
+
+@dataclasses.dataclass(frozen=True)
+class ValidKinematics:
+    """Contains the metadata necessary to load the kinematics of the dataset.
+    The variables should be a dictionary with the key naming the variable
+    and the content complying with the ``ValidVariable`` spec.
+
+    Only the kinematics defined by the key ``kinematic_coverage`` will be loaded,
+    which must be three.
```
RoyStegeman, line 336: Why must they be three? With this new commondata it should be possible to store any number

validphys2/src/validphys/pineparser.py
```diff
@@ -322,6 +299,13 @@ def pineappl_reader(fkspec):
     # Finallly concatenate all fktables, sort by flavours and fill any holes
     sigma = pd.concat(partial_fktables, sort=True, copy=False).fillna(0.0)
 
+    # Check whether this is a 1-point normalization fktable and, if that's the case, protect!
+    if fkspec.metadata.operation == "RATIO" and ndata == 1 and len(pines) == 1:
+        # it _might_ be, check whether it is the divisor fktable
+        divisor = fkspec.metadata.FK_tables[-1][0]
+        name = fkspec.fkpath[0].name.replace(".pineappl.lz4", "")
```
Radonirinaunimi, line 306:
```suggestion
        name = fkspec.fkpath[0].name.replace(f".{EXT}", "")
```

validphys2/src/validphys/commondataparser.py
scarlehoff: https://www.youtube.com/watch?v=-IOMNUayJjI

(more seriously, expecting 3 values is hardcoded in so many places that I decided to leave it like that for now and then remove this limitation in the future)

RoyStegeman:
What is the status of https://github.com/NNPDF/nnpdf/pull/1708?

validphys2/src/validphys/coredata.py
```diff
@@ -370,17 +386,45 @@ def systematic_errors(self, central_values=None):
...
+        definitions["stat"] = {
+            "description": "Uncorrelated statistical uncertainties",
+            "treatment": "ADD",
+            "type": "UNCORR",
+        }
+
+        ret = {"definitions": definitions, "bins": bins}
+        yaml.safe_dump(ret, buffer)
+
+    def export(self, folder_path):
```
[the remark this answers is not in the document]
scarlehoff: The fact that it is yaml is not important (and the previous interface was export, so I think it is better to leave the same name)

scarlehoff:
> What is the status of https://github.com/NNPDF/nnpdf/pull/1708?

Not bad, but not fantastic.

Radonirinaunimi (commented):
A very tiny one (again) that I've just stumbled upon when skimming through.

RoyStegeman:
> Not bad, but not fantastic.

Okay I think that would deserve relatively high priority, especially once this is merged.

validphys2/src/validphys/commondataparser.py
RoyStegeman: > expecting 3 values is hardcoded in so many places

Fair enough, can't say that surprises me

validphys2/src/validphys/pseudodata.py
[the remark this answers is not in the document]
RoyStegeman: Well currently it would overwrite whatever the index of the psuedodata is, so unless you trust the `sorted_index` more than the one stored in the file, I'd say this is not a good idea to keep around

Radonirinaunimi (approved):

validphys2/src/validphys/commondataparser.py
```diff
@@ -40,22 +109,856 @@
...
+    return metadata
+
+
+def load_commondata_new(metadata):
+    """
+
+    TODO: update this docstring since now the load_commondata_new takes the information from
+    the metadata, and the name -> split is done outside
+
+    In the current iteration of the commondata, each of the commondata
```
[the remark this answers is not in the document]
scarlehoff: we cannot because of the closure tests... what I will do though is concentrate all these functions in a single module so that they are not in the way

validphys2/src/validphys/filters.py
```diff
@@ -45,6 +45,41 @@
...
+    substitute("sqrts", "sqrt_s")
+    substitute("sqrt(s)", "sqrts")
+    substitute("sqrt(s)", "sqrt_s")
+    substitute("yQQ", "y_ttBar")
+    substitute("yQ", "y_t")
```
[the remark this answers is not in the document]
scarlehoff: I will do a search & replace of all the variables in `cuts/filters.yaml` and this module in a later PR, since that will break runcards using extra filter rules.

For now I will remove the ones that I know are not necessary anymore.

validphys2/src/validphys/commondataparser.py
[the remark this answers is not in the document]
RoyStegeman: Ah you're right, I forgot about the closure tests :(

scarlehoff:
The previous fitbot failed because of a problem in the server, but the fit was uploaded.
I just submitted the fitbot so that we have a final report, and then will merge this PR.

validphys2/src/validphys/plotoptions/kintransforms.py
```diff
@@ -13,48 +12,56 @@
...
+            "HQP_MQQ": ("$M^{QQ} (GeV)$", "$\\mu^2 (GeV^2)$", "$\\sqrt{s} (GeV)$"),
+            "HQP_PTQ": ("$p_T^Q (GeV)$", "$\\mu^2 (GeV^2)$", "$\\sqrt{s} (GeV)$"),
+            "HQP_PTQQ": ("$p_T^{QQ} (GeV)$", "$\\mu^2 (GeV^2)$", "$\\sqrt{s} (GeV)$"),
+            "HQP_YQ": ("$y^Q$", "$\\mu^2 (GeV^2)$", "$\\sqrt{s} (GeV)$"),
+            "HQP_YQQ": ("$y^{QQ}$", "$\\mu^2 (GeV^2)$", "$\\sqrt{s} (GeV)$"),
+            "INC": ("$0$", "$\\mu^2 (GeV^2)$", "$\\sqrt{s} (GeV)$"),
+            "JET": ("$\\eta$", "$p_T^2 (GeV^2)$", "$\\sqrt{s} (GeV)$"),
+            "DIJET": ("$\\eta$", "$m_{12} (GeV)$", "$\\sqrt{s} (GeV)$"),
+            "PHT": ("$\\eta_\\gamma$", "$E_{T,\\gamma}^2 (GeV^2)$", "$\\sqrt{s} (GeV)$"),
+            "SIA": ("$z$", "$Q^2 (GeV^2)$", "$y$"),
```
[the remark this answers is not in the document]
scarlehoff: They _should_ be removed or no longer used anyway in the new commondata.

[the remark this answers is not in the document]
RoyStegeman: Even better

RoyStegeman (approved):
Thanks, you put a lot of work in this

PR #1680, opened 2023-02-24
Remove libNNPDF from python

scarlehoff:
This PR removes all `libNNPDF` imports from python. The compilation still happens (not planning to remove that in this PR) so one can still do `import NNPDF` if needed.
 
I've tested the examples in the `validphys/examples` folder and they seem to work ok.

The main thing this PR actually does is to remove the `load` functions which in the past were loading the full dataset (fktables, cuts, commondata) just to do things like getting the number of points because that's what C++ asked for. I've changed these for `load_commondata` functions which seem to be enough everywhere (the only information requested by the calling function was whatever was contained in the commondata). Since the commondata is now pure python, no C++ is needed anymore.

### DIFF
--- a/validphys2/examples/mc_gen_report.md
+++ b/validphys2/examples/mc_gen_report.md
@@ -1,4 +1,7 @@
 %CHORUSNB 100 replicas
+Mean table
+----------
+{@art_data_mean_table@}
 Data replica histograms 
 -----------------------
 {@art_data_comparison@}
@@ -9,6 +12,3 @@ Residuals
 ---------
 {@art_data_residuals@}
 {@one_art_data_residuals@}
-Mean table
-----------
-{@art_data_mean_table@}
--- a/validphys2/src/validphys/app.py
+++ b/validphys2/src/validphys/app.py
@@ -131,9 +131,6 @@ including the contents of the following file:
             if self.args["loglevel"] <= logging.DEBUG:
                 cout = True
         if not cout:
-            import NNPDF
-
-            NNPDF.SetVerbosity(0)
             lhapdf.setVerbosity(0)
 
     @staticmethod
@@ -147,11 +144,11 @@ including the contents of the following file:
         return contextlib.ExitStack()
 
     def run(self):
-        if sys.version_info < (3, 6):
+        if sys.version_info < (3, 9):
             log.warning(
-                "validphys 2 is discontinued on Python<3.6 and will "
+                "validphys 2 is discontinued on Python<3.9 and will "
                 "not be longer updated. Please run\n"
-                "conda install python=3.6\n\n"
+                "conda install python=3.9\n\n"
                 "If you have any problems, please open an issue "
                 "on https://github.com/NNPDF/nnpdf/issues."
             )
--- a/doc/sphinx/source/data/plotting_format.md
+++ b/doc/sphinx/source/data/plotting_format.md
@@ -85,9 +85,8 @@ a substring). Currently they are:
 'SIA': ('$z$', '$Q^2 (GeV^2)$', '$y$')
 ```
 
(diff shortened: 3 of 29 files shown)


scarlehoff:
The only thing left in this PR is to add some tests to substitute the ones I've removed (because the test was basically a cross-check with the C++ version). I guess make them into regression tests is the right thing to do (because at the end of the day it was what they were). But suggestions welcome.

scarlehoff:
Special thanks to @comane and @andreab1997 who made this possible with their work on closure tests and th covmat!

Zaharid:
Thanks for this. The diff looks nice and obvious, which is good.

@comane maybe you would like to poke around this a little bit (e.g. try to see if there is something that obviously crashes).

Someone should go over the docs as well; I would guess some changes would be needed, but haven't checked.

Zaharid:
> Someone should go over the docs as well; I would guess some changes would be needed, but haven't checked.

```
$ git grep "C++" -- doc/sphinx/
```

Looks like a good place to start.

RoyStegeman (commented):
Thanks, the changes look fine. There are still a few mentions of cpp in the docstrings that can be removed.

validphys2/examples/mc_gen_report.md
```diff
@@ -1,4 +1,7 @@
 %CHORUSNB 100 replicas
+Mean table
+----------
+{@art_data_mean_table@}
```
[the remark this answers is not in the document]
scarlehoff: Because there is a huge image below that was making the computer in which I was running crash. But once it gets to creating the img everything that I've modified has already been checked.

validphys2/src/validphys/app.py
```diff
@@ -147,11 +144,11 @@ def upload_context(do_upload, output):
         return contextlib.ExitStack()
 
     def run(self):
-        if sys.version_info < (3, 6):
+        if sys.version_info < (3, 9):
```
AleCandido, line 147: Did we officially drop 3.8 as well?

scarlehoff: Yes. We are no longer building conda packages for 3.8.

AleCandido (approved):
I had a look at all the diffs, and they look like a careful use of `grep` for a few specific keywords :D

However, to me, it is perfect. I'm also amazed that you have been able to change code in several places while updating very little tests, so the code changed/dropped was either so well-tested to be implementation independent, or just untested.
In both cases, there is a good reason to be confident in the modifications :)

validphys2/src/validphys/app.py
AleCandido: Ok, I wonder if we should drop also from the other projects...

Point is that I'm fully happy in dropping 3.7, that will go EOL in June (now that I'd like to support 3.11, I don't feel guilty dropping an *almost* EOL version, even if it's the one running on `galileo`), but I'm not as sure for 3.8.

In particular, I believe @cschwan for example is still using it. Most likely he doesn't care too much about installing vp, but it makes me think that some still very relevant versions of major distributions are still on it.

scarlehoff:
> I had a look at all the diffs, and they look like a careful use of grep for a few specific keywords :D

Indeed.  This is also an answer your question on why I left some variables with "old convention names". That would've required extra greps/searches (or looking two lines below). Too much effort for my old tired hands.

validphys2/src/validphys/app.py
scarlehoff: Yes, some are. I still have 3.8 in a few places. And if you really want to use it you can just install n3fit/validphys manually and it will work. But since we are not supporting it anymore in "the official conda package" we should warn people.

comane (approved):
Seems good to me.
None of the modifications done in validphys.core.py affects the closure tests (I have explicitly checked this).

PR #1681, opened 2023-02-27
vp-comparefit issue, due to arclength errors

comane:
@Zaharid ,  @enocera Issue with vp-comparefit due to arclength function.
my guess is that the problem arises from these lines of the code within `validphys.arclength.plot_arc_lengths`:
`yvalues = arclengths.stats.central_value()`
`ylower, yupper = arclengths.stats.errorbar68()`
`ylower = yvalues - ylower`
`yupper = yupper - yvalues`
Some of the replicas might be shorter than the central replica. The problem is thus probably fixed by taking the absolute value of the difference.

Another possible issue might be in `validphys.arclength.arc_lengths`, namely, in
`res += integrate.simps(1 + np.square(fdiff), ixgrid[1][1:])`
It looks like a square root is missing here:
`res += integrate.simps(np.sqrt(1 + np.square(fdiff)), ixgrid[1][1:])`


Zaharid:
@comane thanks for this!

> @Zaharid , @enocera Issue with vp-comparefit due to arclength function. my guess is that the problem arises from these lines of the code within `validphys.arclength.plot_arc_lengths`: `yvalues = arclengths.stats.central_value()` `ylower, yupper = arclengths.stats.errorbar68()` `ylower = yvalues - ylower` `yupper = yupper - yvalues` Some of the replicas might be shorter than the central replica. The problem is thus probably fixed by taking the absolute value of the difference.

This is rather the result of mixing means with quantiles. Incidentally I believe it was fine before and we changed it because people complained, but then we got this bug.  I don't think taking the cleanest solution for one we would like the length of the error bar to be yupper-ylower in any case. 

> 
> Another possible issue might be in `validphys.arclength.arc_lengths`, namely, in `res += integrate.simps(1 + np.square(fdiff), ixgrid[1][1:])` It looks like a square root is missing here: `res += integrate.simps(np.sqrt(1 + np.square(fdiff)), ixgrid[1][1:])`

Uh. @scarlehoff @RoyStegeman, @enocera  could you have a look at this please?

scarlehoff:
This is not branching from master, is it?

> It looks like a square root is missing here:
> res += integrate.simps(np.sqrt(1 + np.square(fdiff)), ixgrid[1][1:])

Looking at the blame, the square was removed six years ago. Was it intentional? 
In any case, I would not change this. The fact that we are not computing _an actual arclength_ is less relevant than keeping the quantity unchanged.

RoyStegeman:
> Looking at the blame, the square was removed six years ago. Was it intentional?

Yes this is a bit odd... I'd still be in favor of doing the correct thing now though, even if we're mainly just comparing relative lengths anyway. We hardly compare this between different reports, especially older ones, so I don't see huge issues arise if we fix it

scarlehoff:
(Let me tag @scarrazza as well maybe he has some memory of this)

validphys2/src/validphys/arclength.py
```diff
@@ -91,8 +121,9 @@ def plot_arc_lengths(
         ]
         yvalues = arclengths.stats.central_value()
         ylower, yupper = arclengths.stats.errorbar68()
-        ylower = yvalues - ylower
-        yupper = yupper - yvalues
+        ylower = np.abs(yvalues - ylower)
+        yupper = np.abs(yupper - yvalues)
```
scarlehoff, line 125:
```suggestion
        ylower = np.maximum(yvalues - ylower, 0)
        yupper = np.maximum(yupper - yvalues, 0)
```

I think this is more consistent with what we do in other cases of negative numbers that we don't like (it should not matter to much though)

scarlehoff (commented):
Please rebase on the current master so that you get the tests running for python 3.9/3.10.

Also, you will need to update also the regression tests since you changed the form of the arclength.

wrt the `np.sqrt`, as @RoyStegeman says it doesn't really matter since we never use the actual value but I would prefer if some of the people who was around when that change was made could comment just in case it was actually intentional for some obscure reason.

validphys2/src/validphys/arclength.py
Zaharid: As said above, I think the line should be always of length `yupper - ylower`. We could even remove the point in the middle and be done with it.

scarlehoff: That would also work for me.

comane: > As said above, I think the line should be always of length `yupper - ylower`

This is what the modification already does I think.
@Zaharid, @scarlehoff  I pushed the rebased version, but it's not running the tests.

scarlehoff: I think it was just that github needs a few minutes to reset (and the test failed because of the regression test since the AL has changed).

> This is what the modification already does I think.

Maybe I misunderstood @Zaharid but I think what he would like is just to plot `ylower`-----`yupper` instead of  `yvalues +/- (yupper, ywloer)`

validphys2/src/validphys/arclength.py
```diff
@@ -134,7 +134,8 @@ def plot_arc_lengths(
             xvalues + shift,
             yvalues,
             yerr=(ylower, yupper),
-            fmt=".",
+            fmt='',
+            linestyle = '',
```
comane, line 138: Ok I am still not completely sure on whether I understood. However, the following modification should do it.

scarlehoff:
> FAILED tests/test_plots.py::test_plot_xq2

@Zaharid we should just remove the plot tests or change the thresholds because they are a nightmare.

Zaharid:
The plot tests have been fairly stable for years, and seem to have gotten unstable only lately (due to I guess a bigger update in mpl). Would suggest just dealing with it for now by updating the plots. Let's revisit the issue if this keep happening.

validphys2/src/validphys/arclength.py
Zaharid: Except that yvalues should be something like (yupper - ylower)/2. Otherwise we are back to the original problem.

validphys2/src/validphys/arclength.py
Zaharid: See https://github.com/NNPDF/nnpdf/pull/1681/files#r1124281268.

Zaharid:
Looks okay to me the way it is now.

scarlehoff:
I'm also happy with the new look https://vp.nnpdf.science/L4HDQ463SFK7-nQpiYfrTQ==/#Basespecs0_plot_arc_lengths

Zaharid (approved):

Zaharid:
@scarlehoff could you see what the current test failure on master is?

PR #1685, opened 2023-03-06
Update to the plot

scarlehoff:
I needed two commits because having a slightly different version of matplotlib was _slightly_ different. I'm really hoping this one last.


scarlehoff:
(I'm going to take the liberty of merging this without waiting because the red X in master makes me nervous)

Zaharid:
This is actually changing the markers in the plots and not just moving things a few pixels, which I am not sure should be left up to the mpl version. Perhaps its best to specify it in the plotutils function.

PR #1686, opened 2023-03-06
Move xgrid back to n3fit

scarlehoff:
I'm not entirely sure why was it moved away. I guess with so much back and fort and rebase and merge in the other PR it remained in the wrong side, but it doesn't make much sense to have it in evolven3fit since it adds a few spurious dependencies to n3fit.


andreab1997 (approved):
I guess this is a leftover of one of the comments that I addressed and then reverted for some reason. Thanks for fixing it.

PR #1677, opened 2023-02-22
k-factors for new theories

giacomomagni:
Addressing #1672

PR #1689, opened 2023-03-10
Added EXA theory (250) and new baseline theory with displace b-quark threshold (500)

enocera:
This PR is about the addition of two new theories in `theory.db` (and in the server):
- theory 250, as theory 200, but with the `EXA` solution instead of the `TRN` solution for the RGE;
- theory 500, as theory 400, but with DIS FK tables produced by displacing the b quark threshold by a factor of 4.

A NNPDF4.0-like fit with theory 250 is compared to the published NNPDF4.0 fit [here](https://vp.nnpdf.science/eIAqCjguRNCmmnL9U0P1PQ==/).
A NNPDF4.0-like fit (no F2b) with theory 500 is compared to the published NNPDF4.0 fit [here](https://vp.nnpdf.science/M8ECrh6MSCeGDtsTdazOXQ==).


giacomomagni:
Hi @enocera, 
would it be possible to rename theory 500 with a number higher than 520 ??

I know I haven't pushed any of the N3LO theories (since I don't have them fully working...) 
but I'm currently using IDs from 439 to 509.

enocera:
@giacomomagni @andreab1997 In principle we could even get rid of theory 500 and just update theory 400. What do yu think?
cc @scarlehoff @felixhekhorn @AleCandido

andreab1997:
> @giacomomagni @andreab1997 In principle we could even get rid of theory 500 and just update theory 400. What do you think? cc @scarlehoff @felixhekhorn @AleCandido

I agree. As long as we do not have numerical FONLL, we are forced to use this as baseline so I would make it the official theory 400 for the moment.

scarlehoff:
We can indeed (I think it is a good idea to do so actually) but please branch from #1677 (and use the cfactors there) because I've dropped support there for the combination of new theories and old c-factors.

(and Th 400 in the server now only works with #1677)

enocera:
> We can indeed (I think it is a good idea to do so actually) but please branch from #1677 (and use the cfactors there) because I've dropped support there for the combination of new theories and old c-factors.
> 
> (and Th 400 in the server now only works with #1677)
> 
> Edit: but download the latest version of th 400 to do so.

Thanks @scarlehoff. I'll then proceed as you suggest.

scarlehoff:
Ok! (I still need to run a full fit with the new theory 400 in the server and use it for a comparefit to check that I haven't broken anything so I'd say wait until Monday so that I'm sure you are not using something broken).

enocera:
OK @scarlehoff , thanks. I'll wait for you to give me green light.

scarlehoff:
> OK @scarlehoff , thanks. I'll wait for you to give me green light.

Everything seems to work well with the new th 400 so please go ahead https://github.com/NNPDF/nnpdf/pull/1677#issuecomment-1465292229

enocera:
> > OK @scarlehoff , thanks. I'll wait for you to give me green light.
> 
> Everything seems to work well with the new th 400 so please go ahead [#1677 (comment)](https://github.com/NNPDF/nnpdf/pull/1677#issuecomment-1465292229)

Thanks @scarlehoff - will do so tomorrow morning.

enocera:
@scarlehoff @andreab1997 Given that we decided to update theory 400 with the b-quark-threshold-displaced-FK tables, shouldn't we also update [this](https://github.com/NNPDF/theories/blob/main/data/theory_cards/400.yaml) and the grids in [here](https://github.com/NNPDF/theories/tree/main/data/grids)?

andreab1997:
> > > @scarlehoff @andreab1997 Given that we decided to update theory 400 with the b-quark-threshold-displaced-FK tables, shouldn't we also update [this](https://github.com/NNPDF/theories/blob/main/data/theory_cards/400.yaml) and the grids in [here](https://github.com/NNPDF/theories/tree/main/data/grids)?
> > 
> > 
> > Yes. For the theories in the `theories` repo I can add them.
> 
> Thanks. Could you please also recompute the grid for EMC F2C? We do not need it right now, but better re-compute it before we forget. Thanks.

Done https://github.com/NNPDF/theories/commit/aebc2eccf582099734f6a1839f116502225c6653 :)

enocera:
Thanks @andreab1997. I'll then proceed to update theory 400.

scarlehoff:
Theory 399 should work (and, given that the tests which use it are passing, I'd say it does).

For theory 400 I don't know what the status is, the data inside seems correct. Are you finding problems with it?

scarlehoff (approved):

PR #1692, opened 2023-03-10
Fix label of absolute uncertainty plots

### LINKED ISSUE #1687: Fix label of absolute uncertainty plot
We basically only use uncertainty PDF plots relative to some central value. Yet, if one plots absolute vales for some reason, the y_label reads sigma/sigma_ref which doesn't make much sense since what is plotted is an absolute standard deviation.

The relevant code is

https://github.com/NNPDF/nnpdf/blob/2a0e1382555f0246e43cd110d3b63e4a632d0fcc/validphys2/src/validphys/pdfplots.py#L252

This was flagged since at least

https://github.com/NNPDF/nnpdf/issues/1353#issuecomment-897231907

Is there any reason why we don't call it sigma(x) that I forget, @enocera ?

Zaharid:
Closes #1687.


RoyStegeman (approved):
Thanks @Zaharid. This fix has been agreed during the last code meeting.

PR #1694, opened 2023-03-15
Evolve n3fit with eko 0.12

giacomomagni:
This PR is to update `evolven3fit_new` with `eko=0.12.0`.
This is needed for developing since `eko=0.10.3` is not compatible anymore with N3LO and QED #1643. 
cc @scarlehoff


n3fit/src/evolven3fit_new/eko_utils.py
```diff
@@ -71,7 +94,7 @@ def construct_eko_for_fit(theory_card, op_card, save_path=None):
             raise FileNotFoundError(
                 f"Path where eko should be dumped does not exist: {save_path}"
             )
-    eko_op = run_dglap(theory_card, op_card)
+    eko_op = runner.solve(theory_card, op_card, save_path)
```
scarlehoff, line 97: One question, if the `runner` is already receiving `save_path` we no longer need the lines below, right? @andreab1997

n3fit/src/evolven3fit_new/eko_utils.py
```diff
@@ -34,16 +36,37 @@ def construct_eko_cards(
     theory = Loader().check_theoryID(theoryID).get_description()
     theory.pop("FNS")
     theory.update(theory_card_dict)
-    theory_card = gen_theory.gen_theory_card(theory["PTO"], theory["Q0"], update=theory)
+    if "nfref" not in theory:
+        theory["nfref"] = 5
+    if "nf0" not in theory:
+        theory["nf0"] = 4
+    legacy_class = runcards.Legacy(theory, {})
+    theory_card = legacy_class.new_theory
```
scarlehoff, line 44: What does this legacy class do?

n3fit/src/evolven3fit_new/eko_utils.py
giacomomagni: Yes I'd say so.

andreab1997:
Just to clarify, this version is working (@giacomomagni is using it) but there is still some work to do (example: fixing tests). So let me convert this to Draft and I am going to take care of this.

n3fit/src/evolven3fit_new/eko_utils.py
```diff
@@ -34,16 +36,37 @@ def construct_eko_cards(
     theory = Loader().check_theoryID(theoryID).get_description()
     theory.pop("FNS")
     theory.update(theory_card_dict)
-    theory_card = gen_theory.gen_theory_card(theory["PTO"], theory["Q0"], update=theory)
+    if "nfref" not in theory:
+        theory["nfref"] = 5
+    if "nf0" not in theory:
+        theory["nf0"] = 4
```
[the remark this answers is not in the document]
RoyStegeman: No this does not help readability, not everything needs to be on one line.

n3fit/src/evolven3fit_new/eko_utils.py
giacomomagni: This is unfortunately the simplest way I found to read the nnpdf-like theory card (and mimic what is done in pineko) ... 
@AleCandido might have better suggestions

n3fit/src/evolven3fit_new/eko_utils.py
andreab1997: Yes we can remove those lines

n3fit/src/evolven3fit_new/eko_utils.py
andreab1997: This is the same class that we are using in `pineko`. It is able basically to read a NNPDF theory card and convert it to an eko card.

andreab1997: But it is something to improve

n3fit/src/evolven3fit_new/eko_utils.py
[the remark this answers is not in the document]
scarlehoff: > No this does not help readability, not everything needs to be on one line.

Ah, the irony xd
It's just a suggestion though, I'm ok with both versions.

n3fit/src/evolven3fit_new/eko_utils.py
scarlehoff: Ah, ok. Please add a comment on top "Parse the NNPDF theory information into the eko theory" or something.

n3fit/src/evolven3fit_new/eko_utils.py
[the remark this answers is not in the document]
RoyStegeman: So am I, I just prefer the simpler one.

andreab1997:
For the moment we have a known dependencies problem with conda https://github.com/conda-forge/eko-feedstock/pull/11

scarlehoff:
The dependencies problem is now fixed!

RoyStegeman:
Are these passing locally for you?

andreab1997:
> Are these passing locally for you? Just because if they're not, it's probably easier to run only the regression tests locally as opposed to pushing to test.

Actually the test that is failing should no be linked to evolven3fit so I don't understand why it is failing here

scarlehoff:
If I have to guess it would be because the change in eko came with a change in the numpy version being used so that messed up with the seeding at some stage, and the fit is _very_ sensible to changes. So you would need to rerun the regression test locally and update the results.

(assuming this branched out of master)

andreab1997:
> If I have to guess it would be because the change in eko came with a change in the numpy version being used so that messed up with the seeding at some stage, and the fit is _very_ sensible to changes. So you would need to rerun the regression test locally and update the results.
> 
> (assuming this branched out of master)

So that test is only for linux. If I run locally on my mac it is succeding. Can someone with linux try?

scarlehoff:
> If I run locally on my mac it is succeding

I don't think it is being run at all. Also, you would need to run it with a clean conda installation to be in the CI configuration.

andreab1997:
> > If I run locally on my mac it is succeding
> 
> I don't think it is being run at all. Also, you would need to run it with a clean conda installation to be in the CI configuration.

I know, I commented the line that is making the test running only on mac

scarlehoff:
> I know, I commented the line that is making the test running only on linux

And is it running in an environment which is exactly the one in the CI? (which is in practice a perfectly clean conda installation)

Zaharid:
I would be surprised if the numpy version influences the seeding. The "old" numpy random interface that we use is guaranteed to never change afaict.

scarlehoff:
I don't think it is guaranteed but I agree that is surprising. Maybe something else changed version (I haven't compared side by side all packages, I only noticed the bump in the numpy version).

AleCandido:
> I don't think it is guaranteed but I agree that is surprising. Maybe something else changed version (I haven't compared side by side all packages, I only noticed the bump in the numpy version).

That's what lockfiles are useful for :)

Zaharid:
Compatibility GuaranteeA fixed bit generator using a fixed seed and a fixed series of calls to ‘RandomState’ methods using the same parameters will always produce the same results up to roundoff error except when the values were incorrect. RandomState is effectively frozen and will only receive updates that are required by changes in the internals of Numpy. More substantial changes, including algorithmic improvements, are reserved for GeneratoOn 17 Mar 2023, at 11:27, Juan M. Cruz-Martinez ***@***.***> wrote:﻿
I don't think it is guaranteed but I agree that is surprising. Maybe something else changed version (I haven't compared side by side all packages, I only noticed the bump in the numpy version).

n3fit/src/evolven3fit_new/eko_utils.py
```diff
@@ -34,20 +36,51 @@ def construct_eko_cards(
...
+
+    def update_key_if_exists(dict, update_dict, key):
+        if key in update_dict:
+            dict[key].update(update_dict[key])
+
+    for key in op_card:
+        if isinstance(op_card[key], dict):
+            update_key_if_exists(op_card, op_card_dict, key)
+
+    op_card = runcards.OperatorCard.from_dict(op_card)
```
scarlehoff, line 79: I find this part very confusing.

You start with an `op_card` which is  `default_op_card` which you are importing from `eko`.
But this is not a card, is a dictionary, which you need to then update and modify in 3 different ways.

Then this becomes an actual only at the end.

I'm particularly confused about lines 71-77 because if I were to enter:
`op_card_dict = {"mu0": 24}` this would get silently ignored because `op_card["mu0"]` is not a dictionary. 

Same if I enter new parameters in `op_card_dict` that are not included in the original `op_card`. Is this expected or is it a bug?

scarlehoff:
I just updated the tests https://github.com/NNPDF/nnpdf/pull/1694/commits/9eafa6d7253ef74b5234bfeaacd5f584c23e8b6b 

and I agree this doesn't look like a seed. More like a floating point thing that has changed in one of the dependencies (numpy/pandas/tensorflow, which are the three that are relevant here) but nothing we should worry about.

RoyStegeman:
cpu time is a bit strange though

scarlehoff:
> cpu time is a bit strange though, do we understand why that has changed so much?

I just ran them in a better computer than last time.

n3fit/src/evolven3fit_new/eko_utils.py
andreab1997: I would not say it is a bug but it is something to be improved. I will :)

scarlehoff: What I meant is whether is something you actually wanted (then I'd a check to tell the user so that is not silent!) or something you overlooked (then it should be added!)

Zaharid:
Are we using the new random.Generator anywhere as opposed to the old interface?

scarlehoff:
> Are we using the new random.Generator anywhere as opposed to the old interface?

If I had to bet (looking at the diff). The changes came from TF initialization.

But yes, we are using random.Generator for instance for the trvl split since: 9b845d3a97f0543147f8ed01a14c32944f2a7dce (I remember this one because I think we discussed about whether to do it or not, probably in person in Paris given that I cannot find it anything written in the actual PR).

n3fit/src/evolven3fit_new/eko_utils.py
andreab1997: I hope I made it at least a bit clearer

n3fit/src/evolven3fit_new/evolve.py
```diff
@@ -1,27 +1,27 @@
-import pathlib
 import logging
+import pathlib
 import sys
 
+import eko
 import numpy as np
```
scarlehoff, line 6:
```suggestion
import numpy as np
import eko
```

just to put all the eko imports together.

scarlehoff (approved):
lgtm

@niclaurenti I think you can safely rebase on top of this branch.

n3fit/src/evolven3fit_new/evolve.py
```diff
@@ -1,27 +1,27 @@
...
+import pathlib
 import sys
 
 import numpy as np
-from reportengine.compat import yaml
-
-from ekobox import genpdf, gen_info
-from ekomark import apply
+import eko
 from eko import basis_rotation as br
```
RoyStegeman, line 7:
```suggestion
from eko import basis_rotation
```

Not this PR, but basis_rotation is a bit more descriptive

n3fit/src/evolven3fit_new/eko_utils.py
```diff
@@ -34,20 +43,45 @@ def construct_eko_cards(
...
+    op_card["configs"].update(EVOLVEN3FIT_CONFIGS_DEFAULTS)
+    # User can still change the configs via op_card_dict
+    
+    # Note that every entry that is not a dictionary should not be 
+    # touched by the user and indeed an user cannot touch them            
+    for key in op_card:
+        if key in op_card_dict and isinstance(op_card[key], dict):
+            op_card[key].update(op_card_dict[key])
+        elif key in op_card_dict:
+            _logger.warning("Entry %s is not a dictionary and will be ignored", key)
```
RoyStegeman, line 78: I'm wondering if at this point it should just raise an error. If a code runs and returns something that looks halfway alright I don't check the logs. 

Whenever there is a key in a dictionary there may have been a reason for it to be there and if it was put in the wrong place/mistyped or something like that it's good if it raises an error. Probably outweights the few times it would raise an error when unnecessary.

n3fit/src/evolven3fit_new/eko_utils.py
```diff
@@ -58,26 +92,12 @@ def construct_eko_for_fit(theory_card, op_card, save_path=None):
...
-        except:
-            _logger.error(f"Error saving the eko to : {save_path}")
-            pass
-    return eko_op
+    if not save_path.parent.exists():
+        raise FileNotFoundError(
+            f"Path where eko should be dumped does not exist: {save_path}"
+        )
+    
+    runner.solve(theory_card, op_card, save_path)
```
RoyStegeman, line 103: I don't think this is worth keeping a separate function for

RoyStegeman:
Thanks guys, please give me a sec to have a look

n3fit/src/n3fit/tests/test_evolven3fit.py
```diff
@@ -44,27 +45,26 @@ def test_utils():
...
-    n_cores = 6
-    t_card, op_card = eko_utils.construct_eko_cards(theoryID, q_fin, q_points, x_grid, {'n_integration_cores' : n_cores, 'interpolation_polynomial_degree' : 2}, {'Comments' : comments})
-    assert t_card['Qref'] == qref
-    assert t_card['PTO'] == pto
-    assert t_card['Comments'] == comments
-    assert op_card['n_integration_cores'] == n_cores
-    assert_allclose(op_card["interpolation_xgrid"], x_grid)
-    assert_allclose(op_card["Q2grid"][0], 1.65**2)
-    assert_allclose(op_card["Q2grid"][-1], q_fin**2)
+    t_card, op_card = eko_utils.construct_eko_cards(theoryID, q_fin, q_points, x_grid, op_card_dict={"configs":{"interpolation_polynomial_degree":2}}, theory_card_dict={'Comments' : comments})
```
RoyStegeman, line 56:
```suggestion
    t_card, op_card = eko_utils.construct_eko_cards(
        theoryID,
        q_fin,
        q_points,
        x_grid,
        op_card_dict={"configs": {"interpolation_polynomial_degree": 2}},
        theory_card_dict={"Comments": comments},
    )
```

RoyStegeman (approved):
Some suggestions but looks good otherwise

n3fit/src/evolven3fit_new/evolve.py
AleCandido: @scarlehoff this in general is denied by `isort`, since it is standardizing imports imposing alphabetical sorting.

However, I don't care so much in general, we only use `isort` to avoid imports changing over and over.

n3fit/src/evolven3fit_new/evolve.py
AleCandido: It is, we are just using the abbreviation in EKO since it is used over and over in some contexts (and its members are already telling enough).
But here it is also fine to use the full name :)

RoyStegeman: I indeed assumed the abbreviation it was an eko thing, but in NNPDF/nnpdf it doesn't make much sense to me.

n3fit/src/evolven3fit_new/eko_utils.py
AleCandido: Also, Python philosophy is to pretend everything works fine, as long as something won't break, to save repeated checks.
In this case, EKO will try to bootstrap a new folder at the given path, and as soon as it will try to write (i.e. almost immediately) it will crash.

If you want to manage that error you can catch it with a `try... except:` (and it will be exactly a `FileNotFoundError`, returned by a `pathlib.Path.write_text()` call). But if you just raise, keep the original one, and reduce the amount of code :)

n3fit/src/evolven3fit_new/evolve.py
```diff
@@ -90,20 +87,21 @@ def evolve_fit(
     if eko_path is not None:
         eko_path = pathlib.Path(eko_path)
         _logger.info(f"Loading eko from : {eko_path}")
-        eko_op = output.Output.load_tar(eko_path)
+        eko_op = eko.EKO.edit(eko_path)
```
AleCandido, line 90: The suggested way to use this method is within a context manager:

```py
with eko.EKO.edit(eko_path) as eko_op:
    ...
```

But if you are not using the context manager, you are responsible to call `eko_op.close()` by yourself (otherwise terrible things may happen :P)

AleCandido: Consider that you can also do it in two steps:
```py
eko_op = eko.EKO.edit(eko_path)

with eko_op:
    ...
```

n3fit/src/evolven3fit_new/eko_utils.py
andreab1997: In general I agree. The point is that in this case the only thing that an user can pass and that would be ignored are the Q2 and x grids which, in the particular case of the postfit evolution, I don't thing will be source of confusion.

n3fit/src/evolven3fit_new/eko_utils.py
andreab1997: Yes I agree, this check was done for the previous version of `eko` that computed the operator and dumped it just at the end.

andreab1997: So I would remove the check but I would keep the function, even if it contains just one line, because it is used in 2 places and I believe that in this way is a bit clearer

n3fit/src/evolven3fit_new/eko_utils.py
scarlehoff: What is the use case of `op_card_dict`? Is this something that the user is expected to use or is it more an "advance use case" / for debugging / etc?

n3fit/src/n3fit/scripts/evolven3fit_new.py
```diff
@@ -152,7 +154,7 @@ def main():
         tcard, opcard = eko_utils.construct_eko_cards(
             args.theoryID, args.q_fin, args.q_points, x_grid, op_card_info, theory_card_info
         )
-        _ = eko_utils.construct_eko_for_fit(tcard, opcard, args.dump)
+        _ = runner.solve(tcard, opcard, args.dump)
```
AleCandido, line 157: In the current version of EKO, `runner.solve` has no return value (because it was useless with the new output, as proven here)
```suggestion
        runner.solve(tcard, opcard, args.dump)
```

n3fit/src/evolven3fit_new/evolve.py
```diff
@@ -90,28 +88,30 @@ def evolve_fit(
...
-    info["AlphaS_Qs"] = list(eko_op["Q2grid"])
-    info["XMin"] = float(x_grid[0])
-    info["XMax"] = float(x_grid[-1])
-    dump_info_file(usr_path, info)
-    for replica in initial_PDFs_dict.keys():
-        evolved_block = evolve_exportgrid(initial_PDFs_dict[replica], eko_op, x_grid)
-        dump_evolved_replica(
-            evolved_block, usr_path, int(replica.removeprefix("replica_"))
-        )
+            eko_op = runner.solve(theory, op, dump_eko)
```
AleCandido, line 99: `eko_op` will be `None`, since `runner.solve()` is not returning anything.

n3fit/src/evolven3fit_new/evolve.py
AleCandido: @andreab1997 you resolved the conversation before applying any solution: `eko_op` is not used in a context manager, nor explicitly closed.

Why do you believe it is not required?

n3fit/src/evolven3fit_new/evolve.py
AleCandido: You can simply change to an if condition:
```py
        _logger.info(f"Loading eko from theory {theoryID}")
        theory_eko_path = (Loader().check_theoryID(theoryID).path) / "eko.tar"
        if not theory_eko_path.is_file():
            _logger.info(f"eko not found in theory {theoryID}, we will construct it")
            runner.solve(theory, op, dump_eko)
        eko_op = eko.EKO.edit(theory_eko_path)
```

n3fit/src/evolven3fit_new/eko_utils.py
```diff
@@ -34,50 +43,39 @@ def construct_eko_cards(
     theory = Loader().check_theoryID(theoryID).get_description()
     theory.pop("FNS")
     theory.update(theory_card_dict)
-    theory_card = gen_theory.gen_theory_card(theory["PTO"], theory["Q0"], update=theory)
+    if "nfref" not in theory:
+        theory["nfref"] = 5
+    if "nf0" not in theory:
+        theory["nf0"] = 4
```
AleCandido, line 49: Consider lifting these values top-level `NFREF_DEFAULT`, `NF0_DEFAULT`.

conda-recipe/meta.yaml
```diff
@@ -51,8 +51,8 @@ requirements:
         - sphinxcontrib-bibtex
         - curio >=1.0
         - pineappl >=0.5.8
-        - eko ==0.10.2
-        - banana-hep ==0.6.6
+        - eko >=0.12.0
+        - banana-hep >=0.6.8
         - lz4 >=3.1.10 # see https://github.com/conda-forge/eko-feedstock/issues/10
```
AleCandido, line 56: This issue is actually solved, I believe here you now want greater equal than `4.0` (`4.0.2` to be aligned with that version of EKO)

n3fit/src/evolven3fit_new/evolve.py
```diff
@@ -90,28 +88,30 @@ def evolve_fit(
...
+        eko.io.manipulate.xgrid_reshape(eko_op, targetgrid=x_grid_obj, inputgrid=x_grid_obj)
+        info = info_file.build(theory, op, 1, info_update={})
+        info["NumMembers"] = "REPLACE_NREP"
+        info["ErrorType"] = "replicas"
+        info["AlphaS_Qs"] = eko_op.mu2grid.tolist()
+        info["XMin"] = float(x_grid[0])
+        info["XMax"] = float(x_grid[-1])
+        dump_info_file(usr_path, info)
+        for replica in initial_PDFs_dict.keys():
+            evolved_block = evolve_exportgrid(initial_PDFs_dict[replica], eko_op, x_grid)
```
AleCandido, line 111: Performance tip: if you put the iteration over $\mu^2$ in a loop, you will load the EKO files in memory over and over.

The best way would be to keep the EKO iteration outer, at the price of building the blocks in memory and dump all of them at the end.

However, this only regards performances, from any other point of view it should be perfectly equivalent.

n3fit/src/evolven3fit_new/evolve.py
```diff
@@ -90,28 +88,30 @@ def evolve_fit(
...
-        evolved_block = evolve_exportgrid(initial_PDFs_dict[replica], eko_op, x_grid)
-        dump_evolved_replica(
-            evolved_block, usr_path, int(replica.removeprefix("replica_"))
-        )
+            runner.solve(theory, op, dump_eko)
+            eko_path = dump_eko
+    eko_op = eko.EKO.edit(eko_path)
+    with eko.EKO.edit(eko_path) as eko_op:
+        x_grid_obj = eko.interpolation.XGrid(x_grid)
+        eko.io.manipulate.xgrid_reshape(eko_op, targetgrid=x_grid_obj, inputgrid=x_grid_obj)
```
AleCandido, line 102: Since you're manipulating only here, and then you're using it inside the loop, you consider closing here the `edit` context, and then reopen the EKO in read mode below.

n3fit/src/evolven3fit_new/evolve.py
AleCandido, line 111:
```suggestion
    with eko.EKO.edit(eko_path) as eko_op:
        x_grid_obj = eko.interpolation.XGrid(x_grid)
        eko.io.manipulate.xgrid_reshape(eko_op, targetgrid=x_grid_obj, inputgrid=x_grid_obj)
    
    info = info_file.build(theory, op, 1, info_update={})
    info["NumMembers"] = "REPLACE_NREP"
    info["ErrorType"] = "replicas"
    info["XMin"] = float(x_grid[0])
    info["XMax"] = float(x_grid[-1])

    with eko.EKO.read(eko_path) as eko_op:
        info["AlphaS_Qs"] = eko_op.mu2grid.tolist()
        dump_info_file(usr_path, info)
        for replica in initial_PDFs_dict.keys():
            evolved_block = evolve_exportgrid(initial_PDFs_dict[replica], eko_op, x_grid)
```

n3fit/src/evolven3fit_new/evolve.py
```diff
@@ -165,16 +165,16 @@ def evolve_exportgrid(exportgrid, eko, x_grid):
...
 
     def ev_pdf(pid, x, Q2):
         return x * evolved_pdf[Q2]["pdfs"][pid][targetgrid.index(x)]
 
     block = genpdf.generate_block(
         ev_pdf,
         xgrid=targetgrid,
-        Q2grid=list(eko["Q2grid"]),
-        pids=br.flavor_basis_pids,
+        Q2grid=list(eko.mu2grid),
```
AleCandido, line 176: Later on I would consider:
```suggestion
        mu2grid=list(eko.mu2grid),
```

For consistency with:
- new EKO notation
- physical meaning
- Python naming (PEP 8)

But of course this would follow from the matching change in the function in `genpdf`. But actually that is already done in a PR https://github.com/NNPDF/eko/pull/220, so consider this more as a warning for the future...

n3fit/src/evolven3fit_new/eko_utils.py
andreab1997: I would say that in general the user is not going to use it. The only real use case is if he wants to change the solution of the evolution or something like that and this is actually possible. The only two things that at the moment cannot be changed are the q-grid and the x-grid but I would say that they are usually not changed when one does the postfit evolution.

n3fit/src/evolven3fit_new/eko_utils.py
andreab1997: [b12102b](https://github.com/NNPDF/nnpdf/pull/1694/commits/b12102b6b6128bee8eef9574a606e2fce8b5a35c)

n3fit/src/evolven3fit_new/evolve.py
andreab1997: Yes this would be cleaner

n3fit/src/evolven3fit_new/evolve.py
andreab1997: [72d523b](https://github.com/NNPDF/nnpdf/pull/1694/commits/72d523bbc1ab9d3c0bbcca5646c6079a60a8af33)

conda-recipe/meta.yaml
andreab1997: [e00d11d](https://github.com/NNPDF/nnpdf/pull/1694/commits/e00d11d53d05c7c42cdcc5afd0d98a7d4f29a714)

AleCandido (approved):

PR #1696, opened 2023-03-20
read NN output size for hyperscan from runcard

RoyStegeman:
For fitted v15 we want to do a hyperscan in different basis than the usual eight output PDFs. This PR reads the output size from the network size as set in the runcard where previously the output size was hardcoded to be eight.


scarlehoff (approved):
Thanks for this.

wrt the test I'm a bit confused, but just rebase on top of the eko 0.12.whatever one and then versions are fixed and it should pass.

PR #1697, opened 2023-03-21
Positivity constraints for c and cbar

enocera:
This PR contains the implementation of the positivity constraints `POSF2C_CCE` and `POSF2C_CCP` that correspond to the CC DIS structure functions F2c,W+-. An extensive set of fits that include these positivity constraints is available at this [link](https://www.wiki.ed.ac.uk/display/nnpdfwiki/List+of+fits+for+the+intrinsic+charm+asymmetry+paper).

Note: there are two associated PRs:
- in APFEL: https://github.com/scarrazza/apfel/pull/41
- in APFELcomb: https://github.com/NNPDF/apfelcomb/pull/87

Theories 200 (mc=1.51 GeV), 229 (mc=1.38 GeV), and 230 (mc=1.64 GeV) have been updated on the server.
cc @RoyStegeman @scarlehoff in case this is needed for hyperopt


enocera:
I wonder whether it may be good to modify the basis in the validphys reports by default [here](https://github.com/NNPDF/apfelcomb/pull/87) as follows:
- `flavour` -> `CCBAR_ASYMM_FLAVOUR`
- `evolution` -> `CCBAR_ASYMM`

And I also wonder whether it may be good to add some extra partonic luminosities [here](https://github.com/NNPDF/nnpdf/blob/5baa716fdf3b67b63ebb7329994a1e08a98edba8/validphys2/src/validphys/comparefittemplates/comparecard.yaml#L70), namely `ccbar`, `scbar`, `csbar`.
cc @Zaharid

scarlehoff:
We should add this also to pinefarm through yadism so that we have it also in th 400 @AleCandido 

> I wonder whether it may be good to modify the basis in the validphys reports by default...

While that's a discussion for another time, since we are playing more and more with the fitting basis I wonder whether we shouldn't have all "three": 1. whatever the fit basis was during the fit 2. evolution 3. flavour

validphys2/src/validphys/comparefittemplates/comparecard.yaml
```diff
@@ -52,9 +52,10 @@ PDFnormalize:
     Normtitle: Ratio
 
 Basespecs:
-  - basis: flavour
+<<<<<<< HEAD
```
AleCandido: Leftover conflict resolution markup

Zaharid:
@enocera I have no strong opinions as to what these defaults should be like. I suppose it would be sensible to add the corresponding basis if the charm is to be parametrized separately by default in some sense. I would be more conservative with the default luminosity plots: They are expensive to compute and looked at less often.

RoyStegeman (approved):
Looks good to me and I don't see why we should wait to merge.

enocera:
Thanks @RoyStegeman .

PR #1699, opened 2023-03-22
[New commondata format] Jets

comane:
PR for the implementation of existing Jet datasets into new CommonData format.


buildmaster/ATLAS_2JET_7TEV_R06/filter.py
```diff
@@ -0,0 +1,329 @@
+# Filter for ATLAS_2JET_7TEV_R06
+"""
+Created on Mar  2023
+
+@author: Mark N. Costantini
+"""
+
+import yaml
```
Zaharid, line 8: As a general note, I'd prefer using the ruamel.yaml library (or `from reportengine.compat import yaml`).

It doesn't matter very much if you want parse into a simple dict, but it is better to use only one yaml library in the whole codebase, and also the capability to know about things like line numbers and comments is useful for things like this https://validobj.readthedocs.io/en/latest/examples.html#yaml-line-numbers

buildmaster/ATLAS_2JET_7TEV_R06/filter.py
```diff
@@ -0,0 +1,329 @@
...
+import numpy as np
+import pandas as pd
+from scipy.linalg import block_diag, cholesky
+
+
+from validphys.covmats import dataset_inputs_covmat_from_systematics
+from validphys.loader import Loader
+import warnings
+warnings.filterwarnings('ignore')
```
Zaharid, line 17: Even in these simpler scripts it is better to have this scoped as a context manager (see warnings.catch_warnings) in the place where you want to ignore them, along with a comment as to why.

buildmaster/ATLAS_2JET_7TEV_R06/filter.py
```diff
@@ -0,0 +1,329 @@
...
+
+
+
+def filter_ATLAS_2JET_7TEV_R06(scenario=0,kin_data=True):
+    """
+    Parameters
+    ----------
+
+    scenario : 0, 1, 2
```
Zaharid, line 26: Could we have `Literal["nominal", "stronger", "weaker"]` as input instead? Would make the code a little bit nicer.

enocera:
@Zaharid That's a good question. I think that this should be part of https://github.com/NNPDF/nnpdf/pull/1693.

buildmaster/ATLAS_2JET_7TEV_R06/filter.py
```diff
@@ -0,0 +1,329 @@
...
+            input = yaml.safe_load(file)
+        
+        ystar = range_str_to_floats(input['dependent_variables'][0]['qualifiers'][8]['value'])
+        sqrts = float(input['dependent_variables'][0]['qualifiers'][7]['value'])
+
+        # measurements of several dijet mass values
+        values = input['dependent_variables'][0]['values']
+
+        # loop over different diijet mass bins
+        for i,value in enumerate(values):
```
Zaharid, line 61: A very slight improvement would be something like
```python
m12_values = input['independent_variables'][0]['values']
for value, m12 in zip(values, m12_values)
```

so you don't need the i index. But is is really a minor thing.

buildmaster/ATLAS_2JET_7TEV_R06/filter.py
```diff
@@ -0,0 +1,329 @@
...
+    # keys errors and value
+    card = card['dependent_variables'][scenario]['values']
+    df_nan = HEP_table_to_df(heptable,scenario)
+    
+    for i,dat in enumerate(card):
+        cv = dat['value']
+    
+        for j, err in enumerate(dat['errors']):
+
+            if (scenario == 0 and j == 67) or (scenario == 1 and j == 57) or (scenario == 2 and j == 69):
```
Zaharid, line 277: I think there should be some comments here.

buildmaster/ATLAS_2JET_7TEV_R06/filter.py
```diff
@@ -0,0 +1,329 @@
...
+
+
+from validphys.covmats import dataset_inputs_covmat_from_systematics
+from validphys.loader import Loader
+import warnings
+warnings.filterwarnings('ignore')
+
+
+
+def filter_ATLAS_2JET_7TEV_R06(scenario=0,kin_data=True):
```
Zaharid, line 21: It might be a good idea (for readability, and perhaps reusability) to break this down into several functions, for example by slitting offs the bits that read data, process the kinematics and compute the covmats.

Zaharid: Also I feel the kindata processing could as well be a separate function altogether. The boolean flag to only enable it on the first pass feels a bit unnatural.

buildmaster/ATLAS_2JET_7TEV_R06/filter.py
```diff
@@ -0,0 +1,329 @@
...
+    filter_ATLAS_2JET_7TEV_R06(scenario=0,kin_data=True)
+    filter_ATLAS_2JET_7TEV_R06(scenario=1, kin_data=False)
+    filter_ATLAS_2JET_7TEV_R06(scenario=2, kin_data=False)
+    setname = "ATLAS_2JET_7TEV_R06"
+    l = Loader()
+    cd = l.check_commondata(setname = setname).load_commondata_instance()
+    dataset_input = l.check_dataset(setname,theoryid=200)
+    from validphys.commondataparser import parse_commondata
+    dat_file = '/Users/markcostantini/codes/nnpdfgit/nnpdf/buildmaster/results/DATA_ATLAS_2JET_7TEV_R06.dat'
+    sys_file = '/Users/markcostantini/codes/nnpdfgit/nnpdf/buildmaster/results/systypes/SYSTYPE_ATLAS_2JET_7TEV_R06_DEFAULT.dat'
```
Zaharid, line 307: These should be generalized.

buildmaster/ATLAS_2JET_7TEV_R06/filter.py
```diff
@@ -0,0 +1,329 @@
...
+        with open("uncertainties.yaml",'w') as file:
+            yaml.dump(uncertainties_yaml,file, sort_keys=False)
+    elif scenario==1:
+        with open("uncertainties_stronger.yaml",'w') as file:
+            yaml.dump(uncertainties_yaml,file, sort_keys=False)
+    elif scenario==2:
+        with open("uncertainties_weaker.yaml",'w') as file:
+            yaml.dump(uncertainties_yaml,file, sort_keys=False)
+
+    return covmat
```
Zaharid, line 136: It doesn't look like we are using the return value.

buildmaster/ATLAS_2JET_7TEV_R06/filter.py
Zaharid: Relatedly, I wonder if the processing of the hepdata raw files could be done with validobj (like we already do for the commondata new format). May or may not be simpler, but very likely more reusable.

Radonirinaunimi:
> @Zaharid That's a good question. I think that this should be part of #1693.

It must be indeed part of a PR in which the changes in #1693 are introduced. AFAIC, adding this feature will just amount to moving the implementation in #1500 and ask people to test it. However, I am not sure if the utils in #1693 are already pointing to the correct branch (based on the exchanges that seems to be not yet the case).

buildmaster/ATLAS_2JET_7TEV_R06/filter_bugged.py
```diff
@@ -0,0 +1,53 @@
...
+    - compute artificial uncertainties and save them to
+      .yaml file
+    """
+    inp = dict(
+     dataset_input={'dataset': 'ATLAS_2JET_7TEV_R06'},
+     theoryid=200,
+     use_cuts="internal"
+     )
+
+    covmat = API.covmat_from_systematics(**inp)
```
Zaharid, line 27: Maybe instead of having artificial anonymous systematics for everything we can do more accurately than this by following what `validphys.covmats.construct_covmat` does, i.e reading off
```
d = Loader().check_commondata("ATLAS_2JET_7TEV_R06").load()
cd.systematic_errors()
cd.stat_errors
```

ideally we would want to keep things like the atlas lumi correlation.

?

PR #1701, opened 2023-03-22
Update recommended pylintrc

Zaharid:
Pylint has deprecated a ton of options and broken some vim plugins in the process. This removes the deprecated configurations it complains about.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -51,83 +51,20 @@ confidence=
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
 disable=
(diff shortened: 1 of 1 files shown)


scarlehoff (approved):
What happened to your pylint-bot btw?

scarlehoff:
If this doesn't branch directly from master, yes.

Zaharid:
FWIW the hard part was get it to act on the diff only, rather than the whole file, which is what made it somewhat useful. In  practice, I ran it on both files and only displayed the messages on the diffed lines.

PR #1703, opened 2023-03-22
Remove replica selector functionality

Zaharid:
This was a long forgotten experiment. I thought it was a good idea, but nobody is going to work on it now. The replica plots in there had been broken for a while and nobody had noticed.


scarlehoff (approved):
If you want to use `black` so what is left in the file is standardized it might be a good moment. Otherwise lgtm.

Issue #1691, opened 2023-03-10
Document new commondata format


Zaharid:
The proposal for the new commondata format lives in the heads of some of us, and various partial implementations across pull requests, some of which are different. Ideally we would have some RFC style document discussing both high level ideas such as motivation (do we still want it to be "for experimentalists"?) and complete implementation details (i.e. something with as much information as the validobj stubs).

I believe this would make it easier to discuss changes and to actually port to the new format.

Zaharid:
Examples of prior art include 


  - https://docs.nnpdf.science/data/exp-data-files.html?highlight=commondata
  - https://docs.nnpdf.science/data/th-data-files.html#fk-table-compression
  - https://docs.nnpdf.science/data/plotting_format.html?highlight=plotting+format 

But we can always go better.

scarlehoff:
I'll write a bit about the theory.

cschwan:
In commit b5e26ea7c5e525a047e6fe56163f39ca88cb3530 I added a preliminary document.

RoyStegeman:
Closed by #1708

PR #1708, opened 2023-03-30
Docs for new commondata format

scarlehoff:
I've added a first draft of the new commondata format. This deals with #1691.

At the moment I've only added a description that follows the implementation of @t7phy in #1684. I believe the format is now more or less final (we are at a point where we are deciding whether it should be written `definition` or `definitions` so  really the details).  

Since I was at it I've changed what @cschwan had added about the naming convention to `rst`. I've added in the `new-commondata.rst` file the sentence:

```
In all cases one can reconstruct the name of the folder by separating the observable name on the last ``_``
```

which si what we discussed yesterday.

Please feel free to add new documents to this PR, there are many things missing (a TODO below, feel free to add more items by editing this comment).

TODO:

- [ ] Format of the theory definition
- [ ] Motivation etc
- [ ] Tools to facilitate the hepdata integration
- [ ] Integrating it with the rest of the documentation
- [ ] Utilities and implementation of new files


doc/sphinx/source/data/dataset-naming-convention.rst
```diff
@@ -3,58 +3,57 @@ NNPDF's dataset naming convention
...
-                       | <experiment> "_" <process> "_" <energy> "_" <variant>
+  <valid dataset name> ::= <experiment> "_" <process>
+                        | <experiment> "_" <process> "_" <energy>
+                        | <experiment> "_" <process> "_" <variant>
+                        | <experiment> "_" <process> "_" <energy> "_" <variant>
 
-<experiment> ::= "ATLAS" | "BCDMS" | "CHORUS" | "CMS" | "E605" | "E866"
-               | "E906" | "EMC" | "HERA" | "LHCB" | "NMC" | "NNPDF" | "NUTEV"
+  <experiment> ::= "ATLAS" | "BCDMS" | "CHORUS" | "CMS" | "E605" | "E866"
+                | "E906" | "EMC" | "HERA" | "LHCB" | "NMC" | "NNPDF" | "NUTEV"
```
t7phy, line 14: It would be better to drop HERA and add H1 and ZEUS, just as we don't have LHC but ATLAS and CMS

enocera: @t7phy NO. The inclusive DIS data are for H1 and ZEUS combined, so HERA is better.

enocera: We should then have also H1 and ZEUS for the DIS+jet data.

t7phy: fine, but we should indeed have H1 and ZEUS also.

Zaharid:
How many versions and version comments can there be? Is it a complete changelog or do we forget about everything but the last one?

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,155 @@
+Organization following the naming convention
+--------------------------------------------
+
+All dataset in the new data format follow the exact same naming convention::
+
+    <experiment>_<process>_<energy>{_<extras>}_<observable>
+
+The data is contained in folders, each containing one single hepdata publication.
```
Zaharid, line 8: I assume we can do other things that are not hepdata?

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,155 @@
+Organization following the naming convention
```
Zaharid, line 1: Is the convention enforced mechanically? Is it needed to process the names?

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,155 @@
...
+
+The data is contained in folders, each containing one single hepdata publication. 
+In all cases one can reconstruct the name of the folder by separating the observable name on the last ``_``, i.e., the folder will always be named::
+
+    <experiment>_<process>_<energy>{_<extras>}
+
+
+Where all observables contained in one hepdata entry are separated by their observable name.
+
+Metadata Format
```
Zaharid, line 16: At some point before here we should discuss what is required (metadata, kinematics) and what is optional (everything else).

Zaharid:
If I remember correctly now, we were supposed to keep all the old version metadata files and add a full new one when needed.

doc/sphinx/source/data/new-commondata.rst
scarlehoff: Of course vp doesn't know where the data come from but I think the idea is that you open a folder and it comes from one single hepdata publication.

In the odd case in which it doesn't we hope the person implement it add comments or something explaining what happened there.

scarlehoff:
> How many versions and version comments can there be? Is it a complete changelog or do we forget about everything but the last one?

The versions are for actual fixes and the comment should address whatever the change was. Since we have version control I think we can forget about everything but the last one.

> We should discuss things like versions vs variants? 

The version is a fix of whatever thing. The variant is a variation of the dataset (for instance, a variation of the dataset in which a particular uncertainty is also included).

I'll add some lines about that.

doc/sphinx/source/data/new-commondata.rst
scarlehoff: I don't agree with the everything else. I think hepdata is required and then if it is empty one can add "NA" or whatever. Same for version. Or maybe you are referring to that with `metadata`.

The reason is that I would like to avoid a situation in which a dataset is implemented (or amended) with that information missing even if the cost is a slightly more cumbersome implementation for say a positivity dataset.

doc/sphinx/source/data/new-commondata.rst
scarlehoff: > Is it needed to process the names?

Yes. The folder in which the data is will always be `name.rsplit("_", 1)[0]` and the choice of the observable `name.rsplit("_", 1)[1]

doc/sphinx/source/data/new-commondata.rst
Zaharid: The comment was more in the direction that these docs should be rather general.

scarlehoff:
I've added some information on `variants` and `theory`.

Btw, please feel (@ everyone) free to modify this branch to complete this document.

Zaharid:
@enocera @scarlehoff At some point one goal of this effort was mainly that it should be understandable and hopefully implementable by external people. I think it works pretty well for that (in principle, less clear to me in practice), with the exception of the theory section. It requires external names which index into external tool chains. The keys themselves "fktables" are NNPDF jargon. (And there are the ugly double lists...). I believe it that section would look intimidating to an external person trying to understand what is going on.

With that, I think that if the strategic goal is to publicize the format, we'd be better off if this section was in a separate `nnpdf_theory.yaml` or similar.

scarlehoff:
> With that, I think that if the strategic goal is to publicize the format, we'd be better off if this section was in a separate nnpdf_theory.yaml or similar.

I guess in practice this is already the case with the `yamldb` folder in the new theories. We might want to leave it as part of the theory and never move it to the data.

On the flip side, people wanting to implement new data to use them in a fit will have to deal with that section anyway so I'm not sure whether having it separate is really an advantage at all.

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,205 @@
...
+---------------
+
+The new data format is center around the file ``metadata.yaml`` contained inside the set folder.
```
felixhekhorn, line 19:
```suggestion
The new data format is centered around the file ``metadata.yaml`` contained inside the set folder.
```

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,205 @@
...
+This ``metadata.yaml`` file contains first portion of general information which might be shared by several sets and a list of ``implemented_observables`` which define the separate observables.
```
felixhekhorn, line 20:
```suggestion
This ``metadata.yaml`` file contains a first portion of general information which might be shared by several sets and a list of ``implemented_observables`` which define the separate observables.
```

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,205 @@
...
+        experiment: "EXPERIMENT_NAME"
+
+    implemented_observables:
+    - observable_name: "OBS"
+        observable:
+            description: "Description of the observable"
+            label: "Latex label for the observable"
+            units: "[u]"
+        ndata: n_of_datapoints
+        tables: (optional) [n, j, k] # corresponding tables in the hepdata entry
```
felixhekhorn, line 50:
```suggestion
        tables: [n, j, k] # (optional) corresponding tables in the hepdata entry
```
similar below

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,205 @@
...
+        plot_x: variable_1
+        figure_by:
+            - variable_2
+        kinematic_coverage:
+            variable_1: var_1
+            variable_2: var_2
+
+        kinematics:
+            variables:
+                var_1: {description: "Description of var", label: latex, units: "u"}
```
felixhekhorn, line 64:
```suggestion
                var_1: {description: "Description of var", label: "latex", units: "u"}
```
to be consistent with the above - also do I need the `[` inside units?

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,205 @@
...
+The syntax of the ``variants`` is:
+
+Theory
+~~~~~~
+
+The theory field defines how predictions for the dataset are to be computed.
+It includes two entries:
+
+- ``FK_tables``: this is a list of lists which defines the fktables to be loaded. The outermost list are the operands (in case an operation is needed to recover the observable, more on that below). The innermost list are the grids that are to be concatenated in order to form the operands.
+- ``operaton``: operation to be applied in order to compute the observable
```
felixhekhorn, line 110: we definitely need some cross link here ... I don't know if and where the operations are explained, though

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,205 @@
...
+                  - Wp_contribution
+                  - Wm_total
+                - - total_xs
+            operation: 'ratio'
+
+In this case the ``fktables`` for the Z, W+ and W- contributions will be concatenated (the dataset might include predictions for all three contributions).
+After that, the final observable will be computed by taking the ratio of the concatenation of all those observables and the total cross section (``total_xs``).
+
+
+..  code-block:: yaml
```
felixhekhorn, line 127: this paragraph needs to move back above the theory section

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,205 @@
...
+Therefore, we have a list of ``bins`` (of the same size as the previous lists) and for each entry we have the information of all the variables.
+
+..  code-block:: yaml
+
+    bins:
+        - var:
+            min: 0
+            max: 1
+            mid: 0.5
+        var_2:
```
felixhekhorn, line 202:
```suggestion
        - var:
            min: 0
            max: 1
            mid: 0.5
          var_2:
```
same as above

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,205 @@
...
+
+..  code-block:: yaml
+
+    data_central:
+        - val1
+        - val2
+        - val3
+
+Uncertainties:
+--------------
```
felixhekhorn, line 160:
```suggestion
Uncertainties
-------------
```

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,205 @@
...
+
+When loading this dataset with no variant only the ``uncertainties.yaml`` file will be read.
+Instead, when choosing ``variant: name_of_the_variant``, both ``uncertainties.yaml`` and  ``extra_uncertainties.yaml`` will be loaded.
+Note that if we want to substitute the default set of uncertainties we just need to not include it in the variant (as done in ``another_variant``).
+
+
+Data
+----
+
+The format of the data is a ``yaml`` file with an entry ```data_central``` which is a list for all values for all bins.
```
felixhekhorn, line 150: I guess we should stress somehow more that this is the crucial part

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,205 @@
...
+    bins:
+        - stat:
+        error_name:
+        error_name_2:
+
+Kinematics:
+-----------
+The kinematics file follow a convention very similar to the uncertainties file, where the ``definitions`` field is skipped since that information is already contained in the parent ``metadata.yaml`` file.
+
+Therefore, we have a list of ``bins`` (of the same size as the previous lists) and for each entry we have the information of all the variables.
```
felixhekhorn, line 193:
```suggestion
Therefore, we have a list of ``bins`` (of the same size as the list for `data_central`) and for each entry we have the information of all the variables.
```

felixhekhorn (commented):
I added some comments

scarlehoff:
> I added some comments

I've applied the suggestions. Feel free to add the other changes to the docs directly!

doc/sphinx/source/data/new-commondata.rst
felixhekhorn: @scarlehoff  how about the units?

scarlehoff:
Adding here a few notes RE the latest changes to the commondata format (/cc @Radonirinaunimi )

I will also update the documentation, but just in case I don't finish today.

Plotting:
    The information in PLOTTING_<DATASET> and corresponding PLOTTINGTYPE_<TYPE> should be under a `plotting:` section in the metadata.yaml for each dataset.
    The nnpdf31_process at the top of the metadata should match the old names

Kinematics:
    It will the order of the kinematics from `kinematics_coverage`. That means that if a dataset has more than 3, only the 3 defined in kinematics_coverage will be used

Datasets_names.yaml
    There will be a datasets_names.yaml file which will define the mapping between the old and the new datasets.

scarlehoff:
cc @t7phy @Radonirinaunimi 
An update with respect to the discussion of yesterday:

`kinematic_coverage` should not go inside the `plotting` dictionary but rather be at the same level. This is used to define which set of kinematics to use and is not part of the `plotting`.

In the old plotting file you will see the variable `x` which is usually (k1, k2, k3 or idat). Please change `x` to `plot_x: <variable>`.

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,221 @@
...
+        # Plotting information
+        plotting:
+            dataset_label: "Label to be used in reports"
+            plot_x: variable_1
+            figure_by:
+                - variable_2
+            kinematic_coverage:
+                variable_1: var_1
+                variable_2: var_2
+                variable_3: var_3
```
scarlehoff, line 66:
```suggestion
        kinematic_coverage:
            variable_1: var_1
            variable_2: var_2
            variable_3: var_3
```

Radonirinaunimi:
> cc @t7phy @Radonirinaunimi An update with respect to the discussion of yesterday:
> 
> `kinematic_coverage` should not go inside the `plotting` dictionary but rather be at the same level. This is used to define which set of kinematics to use and is not part of the `plotting`.
> 
> In the old plotting file you will see the variable `x` which is usually (k1, k2, k3 or idat). Please change `x` to `plot_x: <variable>`.

Thanks for the heads up @scarlehoff!

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,218 @@
...
+    implemented_observables:
+    - observable_name: "OBS"
+        observable:
+            description: "Description of the observable"
+            label: "Latex label for the observable"
+            units: "[u]"
+        ndata: n_of_datapoints
+        tables: [n, j, k] # (optional) corresponding tables in the hepdata entry
+        npoints: [n, j, k] # (optional) number of points per table
+
```
scarlehoff, line 56:
```suggestion
        process_type: INC # for instance, INC, JET, DIJET, etc
```

scarlehoff:
For the people implementing new commondata:

the `process_type` key can usually be found in the old commondata files (it's the second column, e.g):
https://github.com/NNPDF/nnpdf/blob/master/nnpdfcpp/data/commondata/DATA_ATLASTTBARTOT8TEV.dat

if you are implementing a new one, so you don't have a reference, the best thing you can do is have a look at the list and use the one that looks the closest. This is used for some internal vp operations (for instance for internal cuts) so it is a required key.

https://github.com/NNPDF/nnpdf/blob/e005a42cfbb2bc049c5e49556d425952c39e2326/validphys2/src/validphys/plotoptions/kintransforms.py#L35

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -0,0 +1,232 @@
...
+Uncertainties
+-------------
+
+The uncertainties are (also) ``.yaml`` files. 
+Note that in the ``metadata.yaml`` the ``data_uncertainties`` entry is given as a list. 
+When using more than one uncertainty file they will be concatenated. 
+This allows the user the flexibility of creating variants where only a subset of the uncertainties are modified.
+
+The format of the uncertainty files is of two fields, a ``definitions`` field that contains metadata about all the uncertainties (their name, their treatment (``ADD`` or ``MULT``) and their type) and a second field ``bins`` which is a list of mappings with as many entries as the `data_central` with the named uncertainties.
```
scarlehoff, line 183:
```suggestion
The format of the uncertainty files is of two fields, a ``definitions`` field that contains metadata about all the uncertainties (their name, their treatment (``ADD`` or ``MULT``) and their type) and a second field ``bins`` which is a list of mappings with as many entries as the `data_central` with the named uncertainties.

Note that, regardless of their treatment type, the uncertainties should always be written as absolute values and not relative to the data values.
```

doc/sphinx/source/data/new-commondata.rst
```diff
@@ -182,6 +182,8 @@ This allows the user the flexibility of creating variants where only a subset of
 
 The format of the uncertainty files is of two fields, a ``definitions`` field that contains metadata about all the uncertainties (their name, their treatment (``ADD`` or ``MULT``) and their type) and a second field ``bins`` which is a list of mappings with as many entries as the `data_central` with the named uncertainties.
 
+Note that, regardless of their treatment type, the uncertainties should always be written as absolute values and not relative to the data values.
```
Radonirinaunimi, line 185: This statement, I think, is not fully correct.

scarlehoff:
This is not perfect, but since we are merging the new commondata format today I wanted to have the documentation in a state in which it can be merged.

I've removed some stuff that was plainly wrong or referred to the old format and added an explanation to every entry but plotting.
The plotting metadata hasn't really changed that much so I've left a reference to the old plotting metadata with a warning that this information is outdated.

The changes in the last commit I need for the docs to compile correctly in my laptop with py3.12.

doc/sphinx/source/data/commondata.rst
```diff
@@ -0,0 +1,355 @@
...
+Due to historical reasons and for backwards compatibility the special energy ``NOTFIXED`` is used
+for datasets where more than one center of mass energy is used.
+When in doubt, it is preferable to utilize two different folders.
+The ``<extras>`` string is free and can be used to disambiguate.
+
+The data downloaded or parsed from hepdata or other sources is kept in the
+``<setname>/<rawdata>`` folder and it is not installed with the rest of the code.
+Each folder must contain a ``<setname>/metadata.yaml`` file which will define
+all datasets implemented within the folder and that will be described below.
+Only ``.yaml`` file are allowed to be installed together with the ``nnpdf`` code.
```
felixhekhorn, line 45:
```suggestion
Only ``.yaml`` files are allowed to be installed together with the ``nnpdf`` code.
```

doc/sphinx/source/data/commondata.rst
```diff
@@ -0,0 +1,355 @@
...
+      - observable_metadata_1
+      - observable_metadata_2
+
+
+The header of the ``metadata.yaml`` file contains information shared among different datasets.
+
+Setname
+~~~~~~~
+
+Correspond to the name of the set and must be equal to the folder. It acts a s a sanity check.
```
felixhekhorn, line 102:
```suggestion
Correspond to the name of the set and must be equal to the folder. It acts as a sanity check.
```

doc/sphinx/source/data/commondata.rst
```diff
@@ -0,0 +1,355 @@
...
+
+Versioning
+~~~~~~~~~~
+
+The initial version of a dataset should be set to ``version: 1``.
+Any change on a dataset should be *always* accompanied of a version bump and a ``version_comment`` explaining the update.
```
felixhekhorn, line 108: Do you want to suggest a default `version_comment` for `version: 1`?

doc/sphinx/source/data/commondata.rst
```diff
@@ -0,0 +1,355 @@
...
+..  code-block:: yaml
+
+    data_central:
+        - val1
+        - val2
+        - val3
+
+``data_uncertainties``
+~~~~~~~~~~~~~~~~~~~~~~
+A list of ``.yaml`` file containing the uncertainty information for the measurement.
```
felixhekhorn, line 258:
```suggestion
A list of ``.yaml`` files containing the uncertainty information for the measurement.
```

doc/sphinx/source/data/commondata.rst
```diff
@@ -0,0 +1,355 @@
...
+=======================
+Experimental data files
+=======================
+
+Data made available by experimental collaborations comes in a variety of
+formats. For use in a fitting code, this data must be converted into a common
+format that contains all the required information for use in PDF fitting.
+Existing formats commonly used by the community, such as in `HepData <https://www.hepdata.net/>`_,
+are generally unsuitable. Principally as they often do not fully describe the
+breakdown of systematic uncertainties. Therefore over several years an NNPDF
```
Radonirinaunimi, line 12: We should also mention that one of the main reasons is that they do not have common format. The breakdown of the systematics are always (at least most of the time) included in the raw data.

doc/sphinx/source/data/commondata.rst
```diff
@@ -0,0 +1,355 @@
...
+This allows the user the flexibility of creating variants
+where only a subset of the uncertainties are modified.
+
+The format of the uncertainty files is of two fields, a ``definitions`` field that contains
+metadata about all the uncertainties: name, treatment (``ADD`` or ``MULT``) and type
+and a second field ``bins`` which is a list of mappings with ``ndata`` entries
+with the named uncertainties.
+
+Note that, regardless of their treatment, uncertainties should always be written as absolute values
+and not relative to the data values. If the data should be updated, the uncertainties should be too.
```
felixhekhorn, line 269:
```suggestion
and not relative to the data values. If the data is updated, the uncertainties have to be too.
```

doc/sphinx/source/data/commondata.rst
```diff
@@ -0,0 +1,355 @@
...
+            treatment:
+            type:
+        error_name_2:
+            description:
+            treatment:
+            type:
+    bins:
+        - stat:
+          error_name:
+          error_name_2:
```
felixhekhorn, line 289:
```suggestion
    definitions:
        stat:
            description: statistical error
            treatment: ADD
            type:
        error_name:
            description: an additive uncertainty
            treatment: ADD
            type:
        error_name_2:
            description: an multiplicative uncertainty
            treatment: MULT
            type:
    bins:
        - stat: 1.0
          error_name: 2.0
          error_name_2: 3.0
```
what is `type`? it is not mentioned above ...

doc/sphinx/source/data/commondata.rst
```diff
@@ -0,0 +1,355 @@
...
+
+Observable specific information
+===============================
+
+Within a ``metadata.yaml`` we can find one or more implemented datasets.
+These correspond to different observables of a single measurement.
+For instance, the LHCB publication of Z rapidity measurements at 13 TeV
+(``setname: LHCB_Z0_13TEV``) contains two observables: Z decay into two electrons
+and Z decay into 2 muons.
+This setname contain two datasets: ``LHCB_Z0_13TEV_DIELECTRON-Y`` and ``LHCB_Z0_13TEV_DIMUON-Y``.
```
Radonirinaunimi, line 134: We should mention in this Section how positivity and integrability are particular in that most of the fields below are not required and should not be defined.

doc/sphinx/source/data/commondata.rst
```diff
@@ -0,0 +1,355 @@
...
+Note that if we want to substitute the default set of uncertainties we just need to not include it in the variant (as done in ``another_variant``).
+
+``theory``
+~~~~~~~~~~
+
+The theory field defines how predictions for the dataset are to be computed.
+It includes two entries:
+
+- ``FK_tables``: this is a list of lists which defines the FK Tables to be loaded. The outermost list are the operands (in case an operation is needed to recover the observable, more on that below). The innermost list are the grids that are to be concatenated in order to form the operands.
+- ``operaton``: operation to be applied in order to compute the observable
```
felixhekhorn, line 332:
```suggestion
- ``operation``: operation to be applied in order to compute the observable
```
"vp currently supports 'ratio', 'asy', ...." "The identity transformation is indicated by None." or similar

doc/sphinx/source/data/commondata.rst
```diff
@@ -0,0 +1,355 @@
...
+In this case the ``fktables`` for the Z, W+ and W- contributions will be concatenated (the dataset might include predictions for all three contributions).
+After that, the final observable will be computed by taking the ratio of the concatenation of all those observables and the total cross section (``total_xs``).
+
+``plotting``
+~~~~~~~~~~~~
+
+The ``plotting`` section defines the plotting style inside ``validphys``
+and is described in detail in :ref:`plotting-format`.
+
+Note that name of the variables need to be the same in the plotting and kinematics.
```
felixhekhorn, line 355:
```suggestion
Note that the names of the variables need to be the same in the plotting and kinematics.
```

Radonirinaunimi (commented):
Thanks a lot @scarlehoff for this extensive documentation. I haven't read it fully but I only have two main comments regarding the:
- `filter`: so far, there is no mention of how the commondata are generated from input hepdata. A (couple of) sentence(s) should suffice.
- `positivity` and `integrability` which mainly affects the **Observable specific information** section.

doc/sphinx/source/data/commondata.rst
scarlehoff: We can leave that for next iterations of the documentation. I think we should merge this asap so that the current docs are not heavily outdated.

doc/sphinx/source/data/commondata.rst
scarlehoff: The type. A lumi uncertainty, calorimeter, etc.

(i agree someone should define it better...)

doc/sphinx/source/data/commondata.rst
scarlehoff: unless someone gets annoyed due to our use of `often` I think it is fine...

scarlehoff:
> filter: so far, there is no mention of how the commondata are generated from input hepdata. A (couple of) sentence(s) should suffice.

I agree but I haven't added any filters so his should be added by people who did it...

RE positivity and integrability, that should be added for whoever ends up in charge of updating the commondata docs (we can assign tasks in the next code meeting).

PR #1710, opened 2023-04-03
Add a `pyproject.toml`

scarlehoff:
Any other options we should add?

(I guess  eventually this file will contain the entire installation rules, for now I'd just like to add the `isort` `black` options.
This is a first small step towards having pre-commit hooks*)

*which I hate

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[tool.black]
+line-length = 100


Zaharid:
How is this supposed to be used? Is it going to be touching every file, or every edited file? 

Can isort be made to only group imports in srdlib/external/internal, but not touch any other ordering?

Personally I stand by what I said here https://github.com/NNPDF/nnpdf/pull/1349#issuecomment-893343497 and more generally note that each of these things comes with tradeoffs costs that may be higher than telling every new contributor once "Please group imports in <this> way".

RoyStegeman:
Note that this does not automate anything - people still have to run `isort` and `black` manually. And as such I would only use it for touched files.

What this mainly does is provide standard settings for those tools such that people don't [run  these with different settings ](https://github.com/NNPDF/nnpdf/commit/4c709ef4bdc38b719bfa0933c62407719f61e08f)

pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[tool.black]
+line-length = 100
```
Zaharid, line 2: Could we add the option to not touch the string quotes (whatever the equivalent of `let g:black_skip_string_normalization=1` is). This reduces unneeded disruption.

scarlehoff: You mean ' to "?
But this I think we want for new code (as @RoyStegeman said, this is just the file that black and isort use for their config options when you run)

Zaharid: I think it is likely enough that this gets executed in old files (where `git add -uip` would give you a point of pointless changes to sort trough) and also I don't particularly love the setting in new code, even if I'll use it if it's set already: it is the opposite choice of the Python repr, people do like to take issue with that https://github.com/grantjenks/blue, and personally I was used to https://stackoverflow.com/a/56190, from before people decided that having two string quotes was intolerable.

scarlehoff: We started using `black` upon your suggestion.

That said, I don't have any strong opinion about style as long as it is consistent. Feel free to add any options to this `.toml` file.

Honestly, what would be great is to arrive to a set of options we are happy with and then run it for the entire codebase, skip the particular commit in the blame and continue from there.

RoyStegeman: Do those preferences really affect your efficiency though? I don't mean this to argue, but genuinely wondering.

If I had to do things manually I might decide to make it a bit more fancy as well (indeed for example https://stackoverflow.com/a/56190), but 
1. that doesn't really work with multiple people, unless we're going to put way more emphasis on formatting guidlines, so we'll have to end up accepting different formats anyway (if we don't use a tool, that is)
2. More than getting to choose my own formatting I like not having to spend mental energy on it at all (within reason, e.g. `pdfbases.py` would probably require some care).

RoyStegeman: > Honestly, what would be great is to arrive to a set of options we are happy with and then run it for the entire codebase, skip the particular commit in the blame and continue from there.

Is this reasonable considering the different PRs that are open at any given moment?

scarlehoff: I was expressing a desire but I realise it probably won't happen.

Zaharid: When I am working on a personal project I enable the option to not touch the string marks. When I am working on a project that does change the quotes, then I do that. But that is not the main point I was making.

The argument is that for this project, in this context, it is better to have the option because it reduces the diffs when reformatting stuff, accidentally or otherwise, and because enabling the option is arguably of little value overall.

IMO the way to deal with string quotes in this project is not to add review comments expressing the desire to Marie Kondo the world in regards to string quotes (which is what has happened so far without particular issues).

RoyStegeman: While I don't necessarily agree I would be happy to add `skip-string-normalization = 1` (assuming that's correct) to the `.toml` file. The main benefit of black for me would be a universal way of enforcing line-length and indentation anyway,  the quotations don't bother me as much. 

Having said that - consistent quotations do spark joy :)

RoyStegeman: To come back to this, would both of  you be ok with `skip-string-normalization = 1`? It still  won't enforce anything but with this at least there would be consistency in output if we ask people to run black.

RoyStegeman, line 2:
```suggestion
line-length = 100
skip-string-normalization = 1
```

I checked that this is the correct key

RoyStegeman (approved):
Ok for me this is fine.

pyproject.toml
```diff
@@ -0,0 +1,11 @@
...
+skip-string-normalization = 1
+
+[tool.isort]
+atomic = true
+line_length = 100
+multi_line_output = 3
+skip_gitignore = true
+known_first_party = ["validphys", "eko", "n3fit", "reportengine"]
+force_sort_within_sections = true
```
RoyStegeman, line 11:
```suggestion
profile = "black"
force_sort_within_sections = true
```
See: https://pycqa.github.io/isort/docs/configuration/black_compatibility.html

Not sure what exactly this does, or if it clashes with some of the other settings. It didn't complain though...

pyproject.toml
```diff
@@ -0,0 +1,11 @@
+[tool.black]
+line-length = 100
+skip-string-normalization = 1
+
+[tool.isort]
+atomic = true
+line_length = 100
+multi_line_output = 3
```
RoyStegeman, line 8:
```suggestion
profile = "black"
```
Ok so `profile="black"` fixes a number of settings including `multi_line_output` (which just happens to be the same as the one explicitly set in this case): https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#custom-configuration

pyproject.toml
```diff
@@ -0,0 +1,11 @@
+[tool.black]
+line-length = 100
+skip-string-normalization = 1
+
+[tool.isort]
+atomic = true
+line_length = 100
+profile = "black"
```
RoyStegeman, line 8:
```suggestion
profile = "black" # https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#custom-configuration
```

Zaharid (approved):

PR #1711, opened 2023-04-03
Expand possible dataset configurations in the training report

### LINKED ISSUE #1707: Include chi2 of excluded datasets in training reports
People in the mailing list have manifested their wish to have the chi2 of excluded datasets in the training reports. I think it could be a nice addition (but I would have it as a separate table).

An example of such a table is given by @enocera here https://vp.nnpdf.science/o7dD198iRsSv_UFS6bwPAw==

Note that this is not trivial since it can only be done in some cases (for instance, if the theory choices are different in both reports, such a table cannot be constructed in an unambiguous way).

RoyStegeman:
This resolves #1707 

I'm not really happy about the way it's implemented now. The problem is that the chi2 info in the reports is generated by collecting over the fitcontexts so there's not an easy way to insert a set of preferred datasets in the vp workflow. 

I guess it's probably somehow possible to edit the reportengine graph itself? But this seems like it would be a pain. Maybe @Zaharid has an idea?

### DIFF
--- a/validphys2/src/validphys/comparefittemplates/comparecard.yaml
+++ b/validphys2/src/validphys/comparefittemplates/comparecard.yaml
@@ -25,6 +25,9 @@
 #   theoryid:
 #     from_: theory
 #   speclabel: "Reference Fit"
+#
+# data_theory_dataset:
+#   dataset_join: the dataset_join setting
 
 pdfs:
   - from_: current
--- a/validphys2/src/validphys/comparefittemplates/report.md
+++ b/validphys2/src/validphys/comparefittemplates/report.md
@@ -78,6 +78,7 @@ Training-validation
 -------------------
 {@fits plot_training_validation@}
 
+{@with data_theory_dataset@}
 {@with DataGroups@}
 $\chi^2$ by {@processed_metadata_group@}
 ----------------------------------------
@@ -98,6 +99,7 @@ $\phi$ by {@processed_metadata_group@}
 --------------------------------------
 {@plot_fits_groups_data_phi@}
 {@endwith@}
+{@endwith@} <!-- ends data_theory_dataset -->
 
 Dataset plots
 -------------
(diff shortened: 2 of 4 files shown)


Zaharid:
I guess we should discuss this in some kind of call. As it is, it is not quite what I had in mind. I don't think we should be modifying how `fitinputcontext` works, but instead adding generalizations 

https://github.com/NNPDF/nnpdf/blob/278310410398b36b95e45246992c8ed07db6d6a6/validphys2/src/validphys/config.py#L877

in the way described in the comment in #1707. Roughly: Check that you have two dataspecs, compute the set difference of the names, copy the settings from the fit that contains the dataset but swap the PDF in each of the two dataspecs.

RoyStegeman:
I agree, that was also what I originally had in mind but that would probably take a bit longer than I had initially anticipated so I just opened this PR to start a discussion. 

In the end what I would like to do is much the same as what is already done for `produce_matched_datasets_from_dataspecs`, where now the datasets and corresponding cuts (and whatever other information) needs to be copied. Assuming we also want to apply the selected dataset configuration to the data-theory plots I suppose the existing `produce_matched_datasets_from_dataspecs` should be updated rather than create a new function?

I'm not sure a call is necessary since I believe we're on the same page, but I'd be happy to if you think it would be useful.

Zaharid:
> I agree, that was also what I originally had in mind but that would probably take a bit longer than I had initially anticipated so I just opened this PR to start a discussion.
> 

I am not expecting this to be particularly easy. Hence the cost part of the cost benefit analysis that we all seem to be too happy to let other people ignore.

> In the end what I would like to do is much the same as what is already done for `produce_matched_datasets_from_dataspecs`, where now the datasets and corresponding cuts (and whatever other information) needs to be copied. Assuming we also want to apply the selected dataset configuration to the data-theory plots I suppose the existing `produce_matched_datasets_from_dataspecs` should be updated rather than create a new function?

I think it ends up being substantially different because of the copying configuration thing. There may be some clever refactoring regarding the slicing of the datasets.

In my mind "union" would result in looping over three namespaces, each of which have different `dataspecs` (different dataset, different PDF, different other settings as appropriate). 


Also I think it is a non objective to have a fully customizable datasets for comparison. At that point, people are better off editing the templates.

PR #1715, opened 2023-04-17
Allow categorical variables in smpdf plots

Zaharid:
Fix crash in plot_smpdf when the variable in the color axis is categorical, such as for the W and Z total cross sections. Encode the categorical variables as integers and use an appropriate color map to display them.

The implementation feels a bit untidy, with some variables acting at a distance, but there doesn't seem to be an easy better way considering that we'd like the color bar to be the same for all figures.

Also, the fact that np.unique sorts the output may not be what we want. See https://github.com/numpy/numpy/issues/8621. But don't want to make it more complicated.


Zaharid:
See result here https://vp.nnpdf.science/kqbWu8OWQum14sMYRegLAg==/#atlas_wz_tot_13tev

validphys2/src/validphys/dataplots.py
```diff
@@ -855,14 +854,33 @@ def plot_smpdf(pdf, dataset, obs_pdf_correlations, mark_threshold:float=0.9):
...
+    categorical = not np.issubdtype(plotting_var.dtype, np.number)
+    if categorical:
+        # Categorical
+        keys, values = np.unique(plotting_var, return_inverse=True)
+        plotting_var = values
+        nunique = len(keys)
+        if nunique <= len(cm.Set2.colors):
+            cmap = mcolors.ListedColormap(cm.Set2.colors[:nunique])
+        else:
+            cmap = cm.viridis.resample(nunique)
```
RoyStegeman, line 866: I guess this will hardly ever be used, but perhaps something like `gist_ncar` would make it easier to match plot and label than `vidris`. See for some continuous colorschemes: https://matplotlib.org/stable/_images/sphx_glr_colormaps_021.png

validphys2/src/validphys/dataplots.py
```diff
@@ -855,14 +854,33 @@ def plot_smpdf(pdf, dataset, obs_pdf_correlations, mark_threshold:float=0.9):
...
+        bins = np.linspace(0, nunique, nunique + 1)
+        norm = mcolors.BoundaryNorm(bins, nunique)
+
     else:
-        norm = mcolors.Normalize(vmin, vmax)
-    #http://stackoverflow.com/a/11558629/1007990
-    sm = cm.ScalarMappable(cmap=cm.viridis, norm=norm)
+        cmap = cm.viridis
+        #TODO: vmin vmax should be global or by figure?
+        vmin,vmax = min(plotting_var), max(plotting_var)
```
scarlehoff, line 873:
```suggestion
        vmin, vmax = min(plotting_var), max(plotting_var)
```

scarlehoff (commented):
Can this be merged?

I guess it would be good to add another smpdf test in which the categorical branch is used, maybe using one of the datasets in your report @Zaharid ? `ATLAS_WZ_TOT_13TEV` for instance

RoyStegeman (changes requested):
There are different figures for only W and only Z but both figures contain labels for W and Z.

There are double parentheses due to group_label and the title both adding them.

validphys2/src/validphys/dataplots.py
Zaharid: As you say it doesn't matter, but I think all prefer all the many colors being similarly different to having some that are accidentally very close.

validphys2/src/validphys/dataplots.py
```diff
@@ -891,54 +891,82 @@ def plot_smpdf(pdf, dataset, obs_pdf_correlations, mark_threshold: float = 0.9):
 
     plotting_var = info.get_xcol(table)
 
-    # TODO: vmin vmax should be global or by figure?
-    vmin, vmax = min(plotting_var), max(plotting_var)
-    if info.x_scale == 'log':
-        norm = mcolors.LogNorm(vmin, vmax)
+    categorical = not np.issubdtype(plotting_var.dtype, np.number)
+    if categorical:
+        # Categorical
```
RoyStegeman, line 896: Maybe explain that this means discrete colouring, the comment alone is not very enlightening.

validphys2/src/validphys/dataplots.py
```diff
@@ -891,54 +891,82 @@ def plot_smpdf(pdf, dataset, obs_pdf_correlations, mark_threshold: float = 0.9):
 
     plotting_var = info.get_xcol(table)
```
RoyStegeman, line 892: Maybe this can be moved under the `else` statement

RoyStegeman (approved):
Was this just waiting on a review or is there another reason why this hasn't been merged?

Could you run `black` and `isort`?

RoyStegeman (changes requested):
Oh right, the tests...

Zaharid:
@scarlehoff @RoyStegeman I don't know what the current test failure with the evolution is, but don't believe it matters for this. Please merge this.

RoyStegeman (approved):
I'm happy with this PR. I think the error is because our tests are not compatible with the eko version and I believe this is fixed in #1754. @scarlehoff can you confirm?

PR #1716, opened 2023-04-17
Make pineparser handle optimized fktables

andreab1997:
Starting from the next release of `pineappl` (and already for the last pre-releases, e.g. `0.6.a5`) the fktable optimizer will perform much better and will remove zero dimensions form the fktable. This means that, for compound datasets, the x_grids of the various `raw_fktables` will not be the same anymore and so `pineappl_reader` will crush. 

In order to solve this problem, zero entries have been added to the raw_fktables in order to match the bigger of the dimensions between the various compound fktables. 

cc: @cschwan

### DIFF
--- a/validphys2/src/validphys/pineparser.py
+++ b/validphys2/src/validphys/pineparser.py
@@ -294,11 +294,12 @@ def pineappl_reader(fkspec):
                 ndata += apfelcomb["shifts"][i]
 
         # Add empty points to ensure that all fktables share the same x-grid upon convolution
-        missing_x_points = np.setdiff1d(xgrid, p.x_grid())
-        missing_x_points_index = [list(xgrid).index(miss) for miss in missing_x_points]
-        for miss_index in missing_x_points_index:
+        missing_x_points = np.setdiff1d(xgrid, p.x_grid(), assume_unique=True)
+        for x_point in missing_x_points:
+            miss_index = list(xgrid).index(x_point)
             raw_fktable = np.insert(raw_fktable, miss_index, 0., axis=2)
-            raw_fktable = np.insert(raw_fktable, miss_index, 0., axis=3)
+            if hadronic:
+                raw_fktable = np.insert(raw_fktable, miss_index, 0., axis=3)
         # Check conversion factors and remove the x* from the fktable
         raw_fktable *= fkspec.metadata.get("conversion_factor", 1.0) / xdivision
 


cschwan:
Let me add some details that might be helpful at some point:
- this is an optimization in PineAPPL and concerns both grids and FK tables.
- it's there since `v0.6.0-alpha.0`, so directly after `v0.5.9`. Commit <https://github.com/NNPDF/pineappl/commit/e2948c463f5cf74101cf0c300f7d7dc3c6f35370> adds the optimization.
- the advantage, as @andreab1997 said, should be that at some point not only should all FK tables use the same 40x50x50 (Q^2, x1, x2) grid values (the standard grid values used when the grids have been generated with Pinefarm), but each FK table should in fact use a much smaller subset of these grid values. Loosely speaking the optimization throws away all grid-values where the FK table is zero. This should reduce the size of the FK table tensor for each dataset. Depending on how the convolution at fit-level is done this might be beneficial, @scarlehoff can comment on that.
- The disadvantage is that for compound observables the tensors don't 'match', in the sense that two FK tables used in compound observable may have different x1/x2 limits, and therefore they use different x1/x2 subsets. This means that adding two FK tables, for instance, doesn't naively work anymore, and that the tensors have to be enlarged with zeros so that the FK tables' limits agree again. This is what this patch does.

RoyStegeman:
Should the version in `meta.yaml` be updated?

validphys2/src/validphys/pineparser.py
```diff
@@ -254,7 +253,10 @@ def pineappl_reader(fkspec):
             "pineappl_reader is not prepared to read a hadronic fktable with no protons!"
         )
     Q0 = np.sqrt(pine_rep.muf2())
-    xgrid = pine_rep.x_grid()
+    x_grids = [pine.x_grid() for pine in pines]
+    xgrids_shapes = [grid.shape[-1] for grid in x_grids]
+    max_xgrid_shape_index = xgrids_shapes.index(max(xgrids_shapes))
+    xgrid = x_grids[max_xgrid_shape_index]
```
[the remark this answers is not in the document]
andreab1997: Mmm wait, why the union?

RoyStegeman:
I made it a draft for now - once the new `pineappl` version is released we can proceed to merge (once approved).

validphys2/src/validphys/results.py
```diff
@@ -98,7 +98,7 @@ def central_value(self):
 
     @property
     def std_error(self):
-        return np.sqrt(np.diag(self.covmat))
+        return np.diag(self.sqrtcovmat)
```
scarlehoff, line 102: This change doesn't seem connected to the rest. Why the change?

andreab1997: Yes, as I wrote in a comment, it is just a bug that I found and corrected. Since it is small I believe we can keep it in this PR but if you prefer I can move it in another PR.

scarlehoff: if it is a bug better a PR and merge it quickly so it gets fixed (specially given this pr is now a draft)

RoyStegeman: Actually I don't think this is correct. Originally I assumed `self.sqrtcovmat` and `np.sqrt(self.covmat)` would be the same thing, but they are not (as you clearly know). However, for the `std_error` I assume we just want the square root of the variances.

scarlehoff: > Yes we only want the sqrt of the variances

Then the right order is to first take the diag, then you get the variances, and do the `sqrt` to that:

`np.sqrt(np.diag(self.covmat))`

andreab1997: Right but the problem is that self.covmat is not working which is the reason of the bug fix

scarlehoff: oh, but then we should fix `self.covmat` ? Or that one simply doesn't exist and this was always wrong?

scarlehoff: (also the two lines are not the same thing, but more than that I want to understand what the bug actually is)

andreab1997: I believe the problem is that `self.covmat` is initialized in the class init and that particular class in the code is only used twice. In both cases the covariance matrix is taken from here https://github.com/NNPDF/nnpdf/blob/4170be2630f98420f72037a3099428aed60f2e2e/validphys2/src/validphys/config.py#L838 which I believe, in that case, return the function instead of the object produced by the function (but I don't know why). Note that I discovered this bug trying to produce a report using the function `plot_fancy` which in turn uses https://github.com/NNPDF/nnpdf/blob/4170be2630f98420f72037a3099428aed60f2e2e/validphys2/src/validphys/results.py#L504 which uses again `covariance_matrix`.

RoyStegeman: Apologies, I somehow completely missed your discussion. Could you maybe share the runcard that results in the error?

RoyStegeman: I get no errors when trying [this runcard](https://github.com/NNPDF/nnpdf/blob/master/validphys2/examples/dataspecs.yaml) using this branch but undoing the change at line 101 of `results.py`

andreab1997: Mmm then I might have done some mistake myself. Let me revert this commit.

validphys2/src/validphys/pineparser.py
```diff
@@ -292,6 +294,13 @@ def pineappl_reader(fkspec):
             if apfelcomb.get("shifts") is not None:
                 ndata += apfelcomb["shifts"][i]
 
+        # Here I need to provide the additional dimension to match the other FKs
+        # First I need to find where holes are
+        missing_x_points = np.setdiff1d(xgrid, p.x_grid())
+        missing_x_points_index = [list(xgrid).index(miss) for miss in missing_x_points]
+        for miss_index in missing_x_points_index:
+            raw_fktable = np.insert(raw_fktable, miss_index, 0., axis=2)
+            raw_fktable = np.insert(raw_fktable, miss_index, 0., axis=3)
```
scarlehoff, line 303:
```suggestion
        missing_x_points = np.setdiff1d(xgrid, p.x_grid(), assume_unique=True)
        for x_point in missing_x_points:
            miss_index = list(xgrid).index(x_point)
            raw_fktable = np.insert(raw_fktable, miss_index, 0., axis=2)
            if hadronic:
                raw_fktable = np.insert(raw_fktable, miss_index, 0., axis=3)
```

A few considerations here, which I'm writing as a suggestion to order my own thoughts but please read through the changes to make sure I didn't break them.

- `assume_unique` ensures that the output is ordered, pineappl fktables are always ordered and the xgrids should be unique so this ensures that the result is ordered which is needed for:
- inserting things into an array makes me nervous because the results get shifted. `p.x_grid` is guaranteed to be a subset of `xgrid` so I think the only possibility for trouble is if the missing points are not ordered. So by looping over a list of the ordered points in `xgrid` that are not in `p.x_grid` you are guaranteed that they get inserted in the right order.
- since boths loops are essentially the same one I've joined them
- Finally, I'm guessing that axis=3 would only be active for double-hadronic observables.

validphys2/src/validphys/pineparser.py
```diff
@@ -254,7 +253,10 @@ def pineappl_reader(fkspec):
             "pineappl_reader is not prepared to read a hadronic fktable with no protons!"
         )
     Q0 = np.sqrt(pine_rep.muf2())
-    xgrid = pine_rep.x_grid()
+    x_grids = [pine.x_grid() for pine in pines]
+    xgrid = np.array([])
+    for grid in x_grids:
+        xgrid = np.union1d(xgrid, grid)
```
scarlehoff, line 259:
```suggestion
    xgrid = np.array([])
    for pine in pines:
        xgrid = np.union1d(xgrid, pine.x_grid())
```

I don't think you need the intermediate list.

scarlehoff (commented):
I know I've commented every single line but, essentially, it looks good to me x)

validphys2/src/validphys/pineparser.py
andreab1997: So I agree with points 1, 2 and 3. For point 4 I am not sure, let me check again.

andreab1997: Yes you are right. I believe I can commit your suggestion safely. Thanks!

scarlehoff (approved):
I think this could be merged, right?

It should work fine with current grids (they all have the same values of x) but also with future grids where some might have optimised out some values.

(and you have been using this branch already for some of the fits you have showed on Friday, right?)

andreab1997:
> I think this could be merged, right?
> 
> It should work fine with current grids (they all have the same values of x) but also with future grids where some might have optimised out some values.
> 
> (and you have been using this branch already for some of the fits you have showed on Friday, right?)

Yes you are right, I am already using this (and also @niclaurenti and @giacomomagni I believe). So I agree that we can merge this. We were waiting because we should also change the pineappl version in che conda recipe but at the moment there are no proper versions (only pre-releases). However we can merge this now and change the pineappl version later.

RoyStegeman (approved):

PR #1717, opened 2023-04-19
Validphys script to download hepdata tables

Radonirinaunimi:
As far as I understood from the report of the last CM, the script to download the HepData tables should be implemented in a separate PR and not part of https://github.com/NNPDF/nnpdf/pull/1693 as mentioned in https://github.com/NNPDF/nnpdf/pull/1699.

The following is a slight improvement on the script initially introduced in https://github.com/NNPDF/nnpdf/pull/1500. It is working fine for some of the datasets I tried in https://github.com/NNPDF/nnpdf/pull/1684 and https://github.com/NNPDF/nnpdf/pull/1699 but people should try it.


Zaharid:
FWIW a minimum non-scripting usage of requests involves

```
response = requests.get(...)
response.raise_for_status()
```

Otherwise the request may fail with a status code and write the wrong thing to the file.

validphys2/src/validphys/scripts/get_hepdata.py
```diff
@@ -0,0 +1,187 @@
...
+        Parameters
+        ----------
+            file: bytes
+                HepData table in bytes
+            table_id: int
+                Denote the `table_id`-th table
+        """
+        self.folder.mkdir(exist_ok=True)
+        ins_id = self.url.split("/")[-1]
+        filename = f"HEPData-{ins_id}-v{self.version}-Table_{table_id}"
```
Zaharid, line 97: It doesn't look like table_id=None works particularly well here. Why is it in the function signature?

validphys2/setup.py
```diff
@@ -33,6 +33,7 @@
                         'vp-nextfitruncard = validphys.scripts.vp_nextfitruncard:main',
                         'vp-hyperoptplot = validphys.scripts.vp_hyperoptplot:main',
                         'vp-deltachi2 = validphys.scripts.vp_deltachi2:main',
+                        'get_hepdata = validphys.scripts.get_hepdata:main',
```
Zaharid, line 36: I am not sure I love this name. Certainly not the underscore.

validphys2/src/validphys/scripts/get_hepdata.py
```diff
@@ -0,0 +1,187 @@
...
+            if table_id in table_numbers:
+                table_dict = requests.get(table["@id"]).json()
+                # Loop over the different table formats. The following surely
+                # could be done in a much more efficient way in order to get
+                # rid of the recursive loops.
+                for url_tab in table_dict["distribution"]:
+                    if "YAML" in url_tab["description"]:
+                        get_reads_yaml = requests.get(url_tab["contentUrl"])
+                        self.write_tables(get_reads_yaml.content, table_id)
+                log.info(f"Table {table_id} downloaded and stored properly.")
```
Zaharid, line 140: I don't understand this: Surely the table is only written when self.write_tables is called?

Radonirinaunimi:
> FWIW a minimum non-scripting usage of requests involves
> 
> ```
> response = requests.get(...)
> response.raise_for_status()
> ```
> 
> Otherwise the request may fail with a status code and write the wrong thing to the file.
> 
> In practice we'd have some reasonable error handling for that as well.

Agree! The error handling is the main part that is still missing here. I was hoping that people will try, document the errors, and we can address them all.

validphys2/setup.py
Radonirinaunimi: Ah, right! For sure people can propose a better name, but it is correct that at least in order be consistent with the previous ones the underscore should be `-`.

AleCandido: `vp-data` or `vp-hepdata` seem quite natural

Radonirinaunimi: This sounds like a reasonable suggestion. But perhaps for more clarity we should go for `vp-getdata` (see [1ded03f](https://github.com/NNPDF/nnpdf/pull/1717/commits/1ded03f9bc616b3199765b84644e94e738b34971))?

Radonirinaunimi:
Closing as no longer required.

PR #1719, opened 2023-04-19
multiclosure internal dataset loader should use t0_covmat_from_systematics

comane:
Bug in the current implementation. The `internal_multiclosure_dataset_loader` function should take the t0 covmat from a single dataset rather than from data. In the previous implementation collect commands like the following were crushing because of shape incompatibility since t0 covmat was being constructed for the whole list of DataSetSpec's .

`datasets_expected_bias_variance = collect( "expected_dataset_bias_variance", ("data",))`


RoyStegeman (approved):
Thanks, you're right. This bug was introduced in 907cec1c4e7bbc8c15c0ddf128df522e5e8a56fd

Issue #1718, opened 2023-04-19
Validphys should not use a matplotlib  gui


Zaharid:
Currently we initiate figures with

```
import maptplib.pyplot as plt

fig, ax = plt.subplots()
```

instead we should do

```
from matplotlib.figure import Figure

def simple_subplots():
    fig = Figure()
    ax = fig.subplots()
    return fig, ax
```

and not import pyplot anywhere. The reason is that (as @AleCandido pointed out) pyplot maintains a  global state that makes it misbehave in multithreaded applications such when executed under dask. One sees warnings like

```
/home/zah/nngit/nnpdf/validphys2/src/validphys/plotutils.py:307: UserWarning: Starting a Matplotlib GUI outside of the main thread will likely fail.
```

as well as initialize some gui bindings that we don't need and have to work around (which is more complicated in multiporcessing  applications such as how we like to run dask).

This is hinted at e.g. in

https://matplotlib.org/stable/gallery/user_interfaces/web_application_server_sgskip.html

AFAICT a big search and replace should have few side effects.

The reportengine example should be tweaked as well, and figure.py should not need calling plt.close.

scarlehoff:
Closed by #1723

PR #1723, opened 2023-04-28
nnpdf without pyplot imports

comane:
The idea of the PR is to implement what is asked in

https://github.com/NNPDF/nnpdf/issues/1718

pyplot should not be imported anywhere in order to avoid a global state that makes it misbehave in multithreaded applications such when executed under dask

### DIFF
--- a/validphys2/src/validphys/closuretest/multiclosure_output.py
+++ b/validphys2/src/validphys/closuretest/multiclosure_output.py
@@ -782,7 +782,7 @@ def plot_experiments_sqrt_ratio_bootstrap_distribution(
     # generate title, so appending string is fine.
     for sqrt_ratio_sample, exp in zip(
         experiments_bootstrap_sqrt_ratio, experiments_data + ["Total"]
-    ): 
+    ):
         fig, ax = plotutils.subplots()
         ax.hist(sqrt_ratio_sample, bins=20, density=True)
         mean = np.mean(sqrt_ratio_sample)
--- a/validphys2/src/validphys/closuretest/multiclosure_output.py
+++ b/validphys2/src/validphys/closuretest/multiclosure_output.py
@@ -8,7 +8,7 @@ data.
 """
 import numpy as np
 import pandas as pd
-import matplotlib.pyplot as plt
+from matplotlib.figure import Figure
 import scipy.special as special
 import scipy.stats
 
@@ -26,7 +26,9 @@ def plot_dataset_fits_bias_variance(fits_dataset_bias_variance, dataset):
 
     """
     biases, variances, _ = fits_dataset_bias_variance
-    fig, ax = plt.subplots()
+    fig = Figure()
+    ax = fig.subplots()
+    
     ax.plot(biases, "*", label=f"bias, std. dev. = {np.std(biases):.2f}")
     ax.axhline(
         np.mean(biases), label=f"bias, mean = {np.mean(biases):.2f}", linestyle="-"
@@ -316,7 +318,8 @@ def plot_dataset_xi(dataset_xi, dataset):
     (0.68). In the legend include the mean across eigenvectors.
 
     """
-    fig, ax = plt.subplots()
+    fig = Figure()
+    ax = fig.subplots()
     ax.plot(
         dataset_xi,
         "*",
@@ -344,7 +347,8 @@ def plot_dataset_xi_histogram(dataset_xi, dataset):
     of the distribution.
 
     """
-    fig, ax = plt.subplots()
+    fig = Figure()
+    ax = fig.subplots()
     ax.hist(
         dataset_xi,
         label=(
@@ -396,7 +400,8 @@ def plot_data_central_diff_histogram(experiments_replica_central_diff):
         for sigma, central_diff
         in experiments_replica_central_diff
     ])
-    fig, ax = plt.subplots()
+    fig = Figure()
+    ax = fig.subplots()
     ax.hist(
         scaled_diffs, bins=50, density=True, label="Central prediction distribution"
     )
@@ -782,7 +787,8 @@ def plot_experiments_sqrt_ratio_bootstrap_distribution(
     for sqrt_ratio_sample, exp in zip(
         experiments_bootstrap_sqrt_ratio, experiments_data + ["Total"]
     ):
-        fig, ax = plt.subplots()
+        fig = Figure()
+        ax = fig.subplots()
         ax.hist(sqrt_ratio_sample, bins=20, density=True)
         mean = np.mean(sqrt_ratio_sample)
         std = np.std(sqrt_ratio_sample)
@@ -846,7 +852,8 @@ def plot_bias_variance_distributions(
             experiments_fits_bias_replicas_variance_samples,
             group_dataset_inputs_by_experiment
         ):
-        fig, ax = plt.subplots()
+        fig = Figure()
+        ax = fig.subplots()
         labels = [
             "fits bias distribution",
             "replicas variance distribution",
@@ -865,7 +872,8 @@ def plot_bias_variance_distributions(
         experiments_fits_bias_replicas_variance_samples,
         axis=0
     )
-    fig, ax = plt.subplots()
+    fig = Figure()
+    ax = fig.subplots()
     ax.hist(
         [total_bias, total_var],
         density=True,
--- a/validphys2/src/validphys/closuretest/multiclosure_output.py
+++ b/validphys2/src/validphys/closuretest/multiclosure_output.py
@@ -8,13 +8,13 @@ data.
 """
 import numpy as np
 import pandas as pd
-from matplotlib.figure import Figure
 import scipy.special as special
 import scipy.stats
 
 from reportengine.figure import figure, figuregen
 from reportengine.table import table
 
+from validphys import plotutils
 
 @figure
 def plot_dataset_fits_bias_variance(fits_dataset_bias_variance, dataset):
@@ -26,8 +26,7 @@ def plot_dataset_fits_bias_variance(fits_dataset_bias_variance, dataset):
 
     """
     biases, variances, _ = fits_dataset_bias_variance
-    fig = Figure()
-    ax = fig.subplots()
+    fig, ax = plotutils.subplots()
     
     ax.plot(biases, "*", label=f"bias, std. dev. = {np.std(biases):.2f}")
     ax.axhline(
@@ -318,8 +317,7 @@ def plot_dataset_xi(dataset_xi, dataset):
     (0.68). In the legend include the mean across eigenvectors.
 
     """
-    fig = Figure()
-    ax = fig.subplots()
+    fig, ax = plotutils.subplots()
     ax.plot(
         dataset_xi,
         "*",
@@ -347,8 +345,7 @@ def plot_dataset_xi_histogram(dataset_xi, dataset):
     of the distribution.
 
     """
-    fig = Figure()
-    ax = fig.subplots()
+    fig, ax = plotutils.subplots()
     ax.hist(
         dataset_xi,
         label=(
@@ -400,8 +397,7 @@ def plot_data_central_diff_histogram(experiments_replica_central_diff):
         for sigma, central_diff
         in experiments_replica_central_diff
     ])
-    fig = Figure()
-    ax = fig.subplots()
+    fig, ax = plotutils.subplots()
     ax.hist(
         scaled_diffs, bins=50, density=True, label="Central prediction distribution"
     )
@@ -786,9 +782,8 @@ def plot_experiments_sqrt_ratio_bootstrap_distribution(
     # generate title, so appending string is fine.
     for sqrt_ratio_sample, exp in zip(
         experiments_bootstrap_sqrt_ratio, experiments_data + ["Total"]
-    ):
-        fig = Figure()
-        ax = fig.subplots()
+    ): 
+        fig, ax = plotutils.subplots()
         ax.hist(sqrt_ratio_sample, bins=20, density=True)
         mean = np.mean(sqrt_ratio_sample)
         std = np.std(sqrt_ratio_sample)
@@ -852,8 +847,7 @@ def plot_bias_variance_distributions(
             experiments_fits_bias_replicas_variance_samples,
             group_dataset_inputs_by_experiment
         ):
-        fig = Figure()
-        ax = fig.subplots()
+        fig, ax = plotutils.subplots()
         labels = [
             "fits bias distribution",
             "replicas variance distribution",
@@ -872,8 +866,7 @@ def plot_bias_variance_distributions(
         experiments_fits_bias_replicas_variance_samples,
         axis=0
     )
-    fig = Figure()
-    ax = fig.subplots()
+    fig, ax = plotutils.subplots()
     ax.hist(
         [total_bias, total_var],
         density=True,
--- a/validphys2/src/validphys/closuretest/multiclosure_preprocessing.py
+++ b/validphys2/src/validphys/closuretest/multiclosure_preprocessing.py
@@ -18,8 +18,8 @@ from validphys.closuretest.closure_checks import (
     check_fits_have_same_basis,
     check_fits_underlying_law_match,
 )
-from validphys.plotutils import plot_horizontal_errorbars, subplots
-
+from validphys.plotutils import plot_horizontal_errorbars
+from validphys import plotutils
 
 
 def _next_multiclosure_preprocessing_table(fits_pdf, fits_preprocessing_table):
@@ -180,7 +180,7 @@ def plot_next_multiclosure_alpha_preprocessing_range_width(
         next_flavour_range = next_multiclosure_alpha_preprocessing_table.loc[:, flavour]
         next_flavour_range_vals = next_flavour_range.to_numpy()
         diffs = next_flavour_range_vals[:, 1] - next_flavour_range_vals[:, 0]
-        fig, ax = subplots()
+        fig, ax = plotutils.subplots()
         ax.hist(diffs, label="Next range width")
         prev_lims = first_prev_ranges.loc[flavour].to_numpy().squeeze()
         ax.axvline(
--- a/validphys2/src/validphys/closuretest/multiclosure_preprocessing.py
+++ b/validphys2/src/validphys/closuretest/multiclosure_preprocessing.py
@@ -7,7 +7,7 @@ closure fits with the previous effective exponents, to see if there is a big
 dependence on the level 1 shift.
 
 """
-import matplotlib.pyplot as plt
+from matplotlib.figure import Figure
 import pandas as pd
 
 from reportengine import collect
@@ -180,7 +180,8 @@ def plot_next_multiclosure_alpha_preprocessing_range_width(
         next_flavour_range = next_multiclosure_alpha_preprocessing_table.loc[:, flavour]
         next_flavour_range_vals = next_flavour_range.to_numpy()
         diffs = next_flavour_range_vals[:, 1] - next_flavour_range_vals[:, 0]
-        fig, ax = plt.subplots()
+        fig = Figure()
+        ax = fig.subplots()
         ax.hist(diffs, label="Next range width")
         prev_lims = first_prev_ranges.loc[flavour].to_numpy().squeeze()
         ax.axvline(
--- a/validphys2/src/validphys/closuretest/multiclosure_preprocessing.py
+++ b/validphys2/src/validphys/closuretest/multiclosure_preprocessing.py
@@ -7,7 +7,6 @@ closure fits with the previous effective exponents, to see if there is a big
 dependence on the level 1 shift.
 
 """
-from matplotlib.figure import Figure
 import pandas as pd
 
 from reportengine import collect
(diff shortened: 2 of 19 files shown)


scarlehoff:
(running the fit bot, not worried about the fit but that way we have the comparefits report... at some point we need to really implement a "run many vp reports bot" without the fit part which in many cases doesn't really matter)

Zaharid:
Could be. Or plotutils in vp. Also I’d call it subplots or simple–subplots

scarlehoff:
Hi @comane when this is ready for review, could you rebase on top of master ? (to run the bot)

(if and only if there are no conflicts, it's just a matter of convenience, If you try to rebase and find conflicts let it be and I'll review it by running the runcards in my own computer)

validphys2/src/validphys/closuretest/multiclosure_output.py
```diff
@@ -781,8 +782,8 @@ def plot_experiments_sqrt_ratio_bootstrap_distribution(
     # generate title, so appending string is fine.
     for sqrt_ratio_sample, exp in zip(
         experiments_bootstrap_sqrt_ratio, experiments_data + ["Total"]
-    ):
-        fig, ax = plt.subplots()
+    ):
```
scarlehoff, line 785:
```suggestion
    ):
```

validphys2/src/validphys/closuretest/multiclosure_preprocessing.py
```diff
@@ -19,7 +18,8 @@
     check_fits_have_same_basis,
     check_fits_underlying_law_match,
 )
-from validphys.plotutils import plot_horizontal_errorbars
+from validphys.plotutils import plot_horizontal_errorbars, subplots
+
```
scarlehoff, line 22:
```suggestion
```

validphys2/src/validphys/dataplots.py
```diff
@@ -556,9 +554,9 @@ def plot_groups_data_phi_spider(groups_data, groups_data_phi, processed_metadata
     """Plot the phi of all groups of datasets as a spider plot."""
     phi = [exp_phi for (exp_phi, _npoints) in groups_data_phi]
     xticks = [group.name for group in groups_data]
-
-    fig = plt.figure()
-    ax = fig.add_subplot(projection='polar')
+
```
scarlehoff, line 557:
```suggestion
```

validphys2/src/validphys/dataplots.py
```diff
@@ -727,12 +723,11 @@ def plot_training_validation(fit, replica_data, replica_filters=None):
...
-        figsize=(
-            max(plt.rcParams.get("figure.figsize")),
-            max(plt.rcParams.get("figure.figsize")),
-        )
-    )
+    fig, ax = plotutils.subplots(figsize=(
+            max(mpl.rcParams.get("figure.figsize")),
+            max(mpl.rcParams.get("figure.figsize")),
+        ))
+
```
scarlehoff, line 730: Could you run black on the changes? This one will probably look a bit more like in the previous version:

```suggestion
    fig, ax = plotutils.subplots(
        figsize=(
            max(mpl.rcParams.get("figure.figsize")),
            max(mpl.rcParams.get("figure.figsize")),
        )
    )
```

validphys2/src/validphys/deltachi2.py
```diff
@@ -7,7 +7,7 @@
 import warnings
 from collections import namedtuple
 
-import matplotlib.pyplot as plt
+from matplotlib.figure import Figure
```
scarlehoff, line 10: Why do you need `Figure` here?

validphys2/src/validphys/mc_gen.py
```diff
@@ -91,8 +93,9 @@ def art_data_moments(art_rep_generation, color="green"):
     _, _, normart_replicas, _ = art_rep_generation
 
     artrep_array = np.asarray(normart_replicas)
-
-    fig, axes = plt.subplots(nrows=3, figsize=(10, 12))
+    
+    fig = Figure(figsize=(10, 12))
+    axes = [fig.add_subplot(3, 1, i+1) for i in range(3)]
```
scarlehoff, line 98: Why does this file uses `Figure` directly instead of relying in `plotutils`?

(I remember previous pain when dealing with this file so I guess it does something strange that plotutils didn't account for? Should it?)

validphys2/src/validphys/plotutils.py
```diff
@@ -25,13 +26,59 @@
 
 log = logging.getLogger(__name__)
 
+def subplots(figsize=None,nrows=1, ncols =1,sharex=False, sharey=False):
```
scarlehoff, line 29: Please run `black` so these new functions all follow the "sanctioned" style.

validphys2/src/validphys/plotutils.py
```diff
@@ -25,13 +26,59 @@
...
+    """
+    Use matplotlib.figure.Figure() objects and not import pyplot anywhere. 
+    The reason is that pyplot maintains a global state that makes it misbehave 
+    in multithreaded applications such when executed under dask parallel mode.
```
scarlehoff, line 33:
```suggestion
    Use matplotlib.figure.Figure() objects to avoid importing ``pyplot`` anywhere. 
    The reason is that ``pyplot`` maintains a global state that makes it misbehave 
    in multithreaded applications such as when executed under dask parallel mode.
```

Add also an explanation of what the function does (same for the one above). I guess something like "Wrapper to generate subplots using matplotlib.figure...." would be enough. Same for the function below.

validphys2/src/validphys/plotutils.py
```diff
@@ -25,13 +26,59 @@
...
+    Returns
+    -------
+    tuple
+        fig, ax = (matplotlib.figure.Figure, fig.subplots)
+    """
+    fig = Figure(figsize=figsize)
+    ax = fig.subplots(nrows=nrows, ncols = ncols, sharex=sharex, sharey=sharey)
+    return fig, ax
+
+def add_subplot(figsize=None,projection=None):
```
scarlehoff, line 53: I find the name of this function confusing because it not only adds a subplot but also creates a figure.
I naively expected it (when I saw it called before) to take a `Fig` as an input and add a subplot to that input figure.

validphys2/src/validphys/plotutils.py
```diff
@@ -459,10 +510,9 @@ def kde_plot(a, height=0.05, ax=None, label=None, color=None, max_marks=100000):
     -------
 
     >>> import numpy as np
-    >>> import matplotlib.pyplot as plt
     >>> dist = np.random.normal(size=100)
-    >>> kde_plot(dist)
-    >>> plt.show()
+    >>> ax = kde_plot(dist)
+
```
scarlehoff, line 515: But I would like it to be shown in the example!

scarlehoff (commented):
I like this. I think the idea of having the plotutils.subplots helper makes it also a very clean change to the codebase.

I left a bunch of comments, mostly (all?) style-related. As far as I can see this doesn't have any undesirable side effects on the actual plots (famous last words?) which is also good.

validphys2/src/validphys/dataplots.py
```diff
@@ -1159,7 +1155,7 @@ def plot_xq2(
 
     """
 
-    w,h = plt.rcParams["figure.figsize"]
+    w,h = mpl.rcParams["figure.figsize"]
```
AleCandido, line 1158: This `rcParams` are another global, I wonder if there's a better way (or even just a way to make them more local...).

validphys2/src/validphys/plotutils.py
AleCandido, line 29: Consider adding `**kwargs` and pass them down directly to `fig.subplots()`.

In particular, you could treat this way all the arguments, since the defaults are already present in [`Figure.subplots()`](https://matplotlib.org/stable/api/figure_api.html#matplotlib.figure.Figure.subplots) and you do not manipulate them at all.

```suggestion
def subplots(figsize=None, **kwargs):
```
and
```py
ax = fig.subplots(**kwargs)
```

AleCandido (approved):
@scarlehoff made a much more thorough review, so, please, answer to that before merging.

From my point of view, this PR is perfectly fine, rather minimal and achieving its goal (I appreciate it).
I just left a couple of minor points.

validphys2/src/validphys/plotutils.py
scarlehoff: I think this is a good idea. It makes the change even more transparent.

validphys2/src/validphys/mc_gen.py
comane: Yes, the main reason is that list comprehension. So the easiest thing to do was to use Figure

validphys2/src/validphys/dataplots.py
comane: Ciao @AleCandido, thank you for the comment!

validphys2/src/validphys/plotutils.py
```diff
@@ -13,45 +13,120 @@
...
+    ----------
+    figsize : tuple
+            default (6.4, 4.8), same as mpl.rcParams["figure.figsize"]
+            but avoids global state
+    Returns
+    -------
+    tuple of sizes for figure
+
+    """
+    return figsize
```
comane, line 95: Hi @scarlehoff , @AleCandido . Thank you both for the comments.
I added this simple function. However, not sure whether this is the best way. A hard coded variable `default_figsize = (6.4,4.8)` would probably do as well.

scarlehoff: > A hard coded variable default_figsize = (6.4,4.8) would probably do as well.

Yes. It also helps (having a `DEFAULT_FIGSIZE = whatever`) for having a place where to change our default everywhere.

That said, I guess the problem with global states is when something gets modified. Reading constants like `mpl.rcParams` I would naively guess it is fine.

AleCandido: > That said, I guess the problem with global states is when something gets modified. Reading constants like `mpl.rcParams` I would naively guess it is fine.

Definitely, but `mpl.rcParams` is also aware of user config files, potentially the distribution default configurations and so on.

If we have a preferred size, then having written somewhere is not a bad idea.

As @scarlehoff pointed out, they are usually written uppercase, when they are located top-level in a module, and not supposed to be modified.
For this reason they are often called *constants*, even if Python has no proper constant at all (even if it is top-level and upper case, you can always modify it, even though you shouldn't, most likely).

validphys2/src/validphys/plotutils.py
```diff
@@ -26,6 +26,8 @@
 
 log = logging.getLogger(__name__)
 
+DEFAULT_FIGSIZE = (6.4, 4.8)
```
AleCandido, line 29: For me, it is sufficiently clear as it is (the variable name is clear enough), but just to inform you: Sphinx has also a syntax for documenting these variables, i.e. just place a docstring right after:

```py
MY_VAR = ...
"""My docstring.

Bla bla bla...

"""
```

If you really want to add docs, I'd suggest to just limit to a single line:
```py
MY_VAR = ...
"""My docstring."""
```

AleCandido: Btw: of course any string literal is a docstring, the triple quotes are not required.

validphys2/src/validphys/plotutils.py
RoyStegeman: Why is this needed? `Reportengine` takes a `default_style` argument which is set in vp (our settings are in `small.mplstyle`), does that not suffice?

scarlehoff: I completely forgot about those files. Apologies for that.

You're right, the default should be there.

validphys2/src/validphys/plotutils.py
Zaharid: Please remove this. This is set in the matplotlib style and should be there.

Zaharid:
I really wish this had kept small with a targeted change. Now it is reformatting a bunch of stuff, making it hard to review and confusing the history, and moving things that belong to the existing styling mechanism.

validphys2/src/validphys/plotutils.py
RoyStegeman: I'm not entirely sure what you mean by "would be possible to use the file content for direct usage?".

What currently happens is that vp (through reportengine) overwrites the global state of matplotlib using the `.mplstyle` file, i.e. it overwrites `rcParams`. This gives us the default settings for our plots. Of course, these can be read and overwritten locally if deemed necessary but based of the name of this function it seems you want to set another default which is why I asked.

validphys2/src/validphys/plotutils.py
```diff
@@ -13,45 +13,107 @@
...
+                default is None
+
+    Returns
+    -------
+    tuple
+        fig, ax = (matplotlib.figure.Figure, fig.add_subplot)
+    """
+    fig = Figure(figsize=figsize)
+    ax = fig.add_subplot(projection=projection, **kwargs)
+    return fig, ax
```
scarlehoff, line 82: I think it might be worth to revert the changes back to this commit: https://github.com/NNPDF/nnpdf/pull/1723/commits/8f78b4891ed46fb483beb0f326d240c121558a3b

And then, for both functions, pass the arguments through subplots as `**kwargs`. This should leave the default as given in the style file.

I would still like `black` to be run for these two functions but I'd avoid running it for files where the only change was `plt.subplots` -> `plotutils.subplots`

(as a separate note, I didn't realize we had files that were mixing 8 and 4 spaces for the indentation... I'm still convinced we should do a commit with a run of black for the entire code, but it should not be this PR)

scarlehoff: Yes. I know (and I can volunteer to eventually do it). But there are some other (like mixing 4/8 spaces) that I'd like to fix for my own sanity like this function [here](https://github.com/NNPDF/nnpdf/blob/f02b49a8e6eb785af6a56dc3195133abff571046/validphys2/src/validphys/plotutils.py#L51)

validphys2/src/validphys/plotutils.py
AleCandido: `rcParams` can be modified by `reportengine`, but also by other factors, since it is coming through `matplotlib`.

So, since a default style is set, it would be less surprising if all the values would come from there, rather than potentially pass through overridable matplotlib settings.
However, if the goal is to make it overridable, and just be consistent in the use of overridable properties, then reading `rcParams` is the correct course of action.

validphys2/src/validphys/plotutils.py
AleCandido: But if you know there are specific points in which you don't want Black to act, why don't you just mark them explicitly [`# fmt: on/off`](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html)?

scarlehoff: The problem is the same, one person has to go through the codebase to do so. There are files I've never opened there!

AleCandido: True, but it can also be faced incrementally: if you already know that there are parts of `pdfbases.py` that should not be touched by Black, you could start marking them.

This would be useful even for other people working on it (e.g. I have Black running on save, even though if I notice that it is massively reformatting a file beyond my changes I manually skip it)

scarlehoff:
@RoyStegeman are you ok with me merging this after the last test passes?

validphys2/src/validphys/closuretest/multiclosure_output.py
scarlehoff, line 785:
```suggestion
    ):
```

validphys2/src/validphys/dataplots.py
scarlehoff, line 557:
```suggestion
```

validphys2/src/validphys/dataplots.py
```diff
@@ -600,8 +598,7 @@ def plot_datasets_chi2_spider(groups_data, groups_chi2):
         xticks = [dataset.name for dataset in group]
         dschi2 = [dsres.central_result/dsres.ndata for dsres in group_res]
 
-    fig = plt.figure(figsize=(4,4))
-    ax = fig.add_subplot(projection='polar')
+    fig, ax = plotutils.add_subplot(figsize=(4,4), projection='polar')
```
scarlehoff, line 601:
```suggestion
    fig, ax = plotutils.add_subplot(figsize=(4, 4), projection='polar')
```

validphys2/src/validphys/dataplots.py
```diff
@@ -627,8 +624,7 @@ def _plot_chi2s_spider_df(df, size=6):
     data = df.iloc[:, df.columns.get_level_values(1)==chilabel].T.values
     fitnames = df.columns.get_level_values(0).unique()
     expnames = list(df.index.get_level_values(0))
-    fig = plt.figure(figsize=(size,size))
-    ax = fig.add_subplot(projection="polar")
+    fig, ax = plotutils.add_subplot(figsize=(size,size),projection='polar')
```
scarlehoff, line 627:
```suggestion
    fig, ax = plotutils.add_subplot(figsize=(size, size), projection='polar')
```

RoyStegeman:
If you reviewed it I'm fine with merging. I didn't properly review it though but just left those comments after quickly glossing over the diffs.

validphys2/src/validphys/dataplots.py
scarlehoff, line 730:
```suggestion
    fig, ax = plotutils.subplots(
        figsize=(
            max(mpl.rcParams.get("figure.figsize")),
            max(mpl.rcParams.get("figure.figsize")),
        )
    )
```

scarlehoff, line 730:
```suggestion
    fig, ax = plotutils.subplots(
        figsize=(
            max(mpl.rcParams.get("figure.figsize")),
            max(mpl.rcParams.get("figure.figsize")),
        )
    )
```

validphys2/src/validphys/dataplots.py
```diff
@@ -878,9 +873,9 @@ def plot_smpdf(pdf, dataset, obs_pdf_correlations, mark_threshold:float=0.9):
         title = "%s %s\n[%s]" % (info.dataset_label, '(%s)'%label if label else '' ,pdf.label)
 
         #Start plotting
-        w,h = plt.rcParams["figure.figsize"]
+        w,h = mpl.rcParams["figure.figsize"]
```
scarlehoff, line 876:
```suggestion
        w, h = mpl.rcParams["figure.figsize"]
```

validphys2/src/validphys/dataplots.py
```diff
@@ -878,9 +873,9 @@ def plot_smpdf(pdf, dataset, obs_pdf_correlations, mark_threshold:float=0.9):
...
         h*=2.5
-        fig,axes = plt.subplots(nrows=nf ,sharex=True, figsize=(w,h), sharey=True)
+        fig, axes = plotutils.subplots(nrows=nf ,sharex=True, figsize=(w,h), sharey=True)
```
scarlehoff, line 878:
```suggestion
        fig, axes = plotutils.subplots(nrows=nf, sharex=True, figsize=(w, h), sharey=True)
```

validphys2/src/validphys/dataplots.py
```diff
@@ -932,7 +927,8 @@ def plot_positivity(pdfs, positivity_predictions_for_pdfs, posdataset, pos_use_k
     kinematic variable (if pos_use_kin==True).
 
     """
-    fig, ax = plt.subplots()
+    fig, ax = plotutils.subplots()
+
```
scarlehoff, line 931:
```suggestion
```

validphys2/src/validphys/paramfits/plots.py
```diff
@@ -15,7 +14,10 @@
 from reportengine.checks import make_argcheck, CheckError, check_positive
 from reportengine.figure import figure, figuregen
 
-from validphys.plotutils import plot_horizontal_errorbars, barplot, kde_plot, marker_iter_plot
+from validphys.plotutils import (plot_horizontal_errorbars, 
+                                    barplot, kde_plot, 
+                                    marker_iter_plot)
+from validphys import plotutils
```
scarlehoff, line 20:
```suggestion
from validphys import plotutils
from validphys.plotutils import barplot, kde_plot, marker_iter_plot, plot_horizontal_errorbars
```

validphys2/src/validphys/plotutils.py
```diff
@@ -43,11 +100,11 @@ def ax_or_newfig(f):
     def _f(*args, **kwargs):
         noax = 'ax' not in kwargs or kwargs['ax'] is None
         if noax:
-                plt.figure()
-                kwargs['ax'] = plt.gca()
+                fig = Figure()
+                kwargs['ax'] = fig.add_subplot(1,1,1)
```
scarlehoff, line 104:
```suggestion
                kwargs['ax'] = fig.add_subplot(1, 1, 1)
```

scarlehoff, line 104:
```suggestion
                kwargs['ax'] = fig.add_subplot(1, 1, 1)
```

validphys2/src/validphys/theorycovariance/tests.py
```diff
@@ -680,7 +681,10 @@ def projector_eigenvalue_ratio(theory_shift_test):
     ratio = ratio[mask]
     xvals = np.arange(1, len(evals) + 1, 1)
     # Plotting
-    fig, (ax1, ax2) = plt.subplots(2, figsize=(5, 5))
+    fig = Figure(figsize=(5, 5))
+    ax1 = fig.add_subplot(2,1,1)
+    ax2 = fig.add_subplots(2,1,2)
```
scarlehoff, line 686:
```suggestion
    fig = Figure(figsize=(5, 5))
    ax1 = fig.add_subplot(2, 1, 1)
    ax2 = fig.add_subplots(2, 1, 2)
```

scarlehoff (approved):
Yes. I will actually leave the bot running overnight (also to make sure my changes don't break anything) and merge it tomorrow as I don't think we need it in master tonight.

PR #1725, opened 2023-05-03
Change the fitbot solver to libmamba

scarlehoff:
Testing the fitbot


RoyStegeman:
Was this due to a memory problem? If so, do you know why it did work before but no longer? Asking because of #1732.

scarlehoff:
No, it was just hitting the 6 hours limit during evolution because it took 3 hours to install nnpdf and about 3 to fit. The installationbwas very slow but was working. I think the memory limit of github was 7 GB so in principle it was staying below that.

PR #1726, opened 2023-05-04
Hyperopt loss

APJansen:
Improving hyperoptimization, experimenting with different hyperoptimization loss functions

## Tasks done in this PR
- [x] Added a new `HyperLoss`class with buit-in methods that can automatically perform statistics over replicas and then folds. The user can select statistics via `replica_statistic` and `fold_statistic` in runcard `kfold`.
- [x] Added new a statistics over replicas, $\varphi^2_{k}$, that (in addition to $\chi^{2}$) can be selected via `loss_type` option in runcard `kfold`.
- [x] Addressed #1894. Here, regardless the selected metrics to minimise by `Hyperopt` ($\varphi^2$ or $\chi^{2}$), we also print in `tries.json` (specifically within `kfold_meta` entry) a matrix (folds x replicas) of calculated $\chi^{2}$ values named `hyper_losses_chi2` and a vector (folds) of $\varphi^2$ values named `hyper_losses_phi2`.

## Description
The implemented `HyperLoss` is instantiated within `ModelTrainer` and later on used in `ModelTrainer.hyperparametrizable`. The user must pass three paramaters that are set in the runcard:
- `loss_type`: The type of loss to be used. Options are `chi2` or  `phi2`.
- `replica_statistic`: The statistics over replicas to be used within each fold. For `loss_type = chi2`, it can assume the usual statistics: `average`, `best_worst` and `std`. **Note: `replica_statistic` is inactive if `loss_type = phi2` as $\varphi^2_{k}$ is by definition a statistics over replicas.**
- `fold_statistic`: The statistics over folds. Options are: `average`, `best_worst` and `std`.
- As discussed by @juanrojochacon, the calculation of any $\varphi^2$ statistics over folds is done with the **reciprocal** of the chosen function. For example, for `loss_type: phi2` and `fold_statistic: average`, the figure of merit to be minimised is actually:
$$\Huge \left( \frac{1}{n_\text{fold}} \sum_{k=1}^{n_\text{fold}} \varphi^2_{k} \right)^{-1} $$

The current implementation of $\varphi^2_{k}$ is based on `validphys` functions. It is evaluated using only experimental data within the hold out fold (as expected).

## Runcard examples
- Default run: `hyper_loss` is set as the $\chi^2$ averaged over replicas and then over folds.
```
kfold:
      loss_type: chi2
      replica_statistic: average
      fold_statistic: average
      penalties:
        - saturation
        - patience
        - integrability
...
```
- Setting $\varphi^2_{k}$ within each $k$-fold and then set `hyper_loss` as the **inverse** of the max value of $[\varphi^2_{1}, \varphi^2_{2}, ..., \varphi^2_{k}]$:
```
kfold:
      loss_type: phi2
      fold_statistic: best_worst
      penalties:
        - saturation
        - patience
        - integrability
...
```
## Notes
It must be merged after #1788 as the current `hyperopt_loss` branch has been created from `trvl-mask-layers`.

### DIFF
--- a/doc/sphinx/source/n3fit/hyperopt.rst
+++ b/doc/sphinx/source/n3fit/hyperopt.rst
@@ -380,6 +380,10 @@ The figure of merit will be the difference between the :math:`\chi2` of the seco
 
 New hyperoptimization metrics with fold and replica statistics
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The hyperopt measures discussed above are all based on performing a single replica fit per fold.
+However one may also wish to run the hyperoptimization algorithm on fits consisting of many
+replicas per fold. This is a feasible option in ``n3fit``, since it has been optimised to
+efficiently run many replica fits in parallel on GPU.
 The combination of :ref:`k-folding <hyperkfolding-label>` and multi-replica experiments
 opens several possibilities for the choice of figure of merit. The simplest option would be to minimize
 the average of :math:`\chi^2` across both replica and k folds, *i.e.*,
@@ -408,7 +412,7 @@ In NNPDF, this hyperoptimisation metrics is selected via the following generic r
 
 By combining the ``average``, ``best_worst``, and ``std`` figures of merit discussed in :ref:`hyperkfolding-label`,
 several alternatives may arise. For example, one approach could involve minimizing
-the maximum value of the set of averaged-over-replicas :math:`\chi2`,
+the maximum value of the set of averaged-over-replicas :math:`\chi^2`,
 
 .. math::
     L_{2} = {\rm max} \left ( \left< \chi^2_{1} \right>_{\rm rep}, \left< \chi^2_{2} \right>_{\rm rep}, ..., \left< \chi^2_{n_{\rm fold}} \right>_{\rm rep}\right),
--- a/doc/sphinx/source/n3fit/hyperopt.rst
+++ b/doc/sphinx/source/n3fit/hyperopt.rst
@@ -384,6 +384,7 @@ The hyperopt measures discussed above are all based on performing a single repli
 However one may also wish to run the hyperoptimization algorithm on fits consisting of many
 replicas per fold. This is a feasible option in ``n3fit``, since it has been optimised to
 efficiently run many replica fits in parallel on GPU.
+
 The combination of :ref:`k-folding <hyperkfolding-label>` and multi-replica experiments
 opens several possibilities for the choice of figure of merit. The simplest option would be to minimize
 the average of :math:`\chi^2` across both replica and k folds, *i.e.*,
--- a/doc/sphinx/source/n3fit/hyperopt.rst
+++ b/doc/sphinx/source/n3fit/hyperopt.rst
@@ -34,6 +34,8 @@ The desired features of this figure of merit can be summarized as:
 3. Be reliable even when the number of points is not very large.
 
 
+.. _hyperkfolding-label:
+
 K-folding cross-validation
 --------------------------
 A good compromise between all previous points is the usage of the cross-validation technique
@@ -320,6 +322,10 @@ Changing the hyperoptimization target
 -----------------------------------
 
 Beyond the usual :math:`\chi2`-based optimization figures above, it is possible to utilize other measures as the target for hyperoptimization.
+
+Future tests
+~~~~~~~~~~~~
+
 One possibility is to use a :ref:`future test<futuretests>`-based metric for which the goal is not to get the minimum :math:`\chi2` but to get the same :math:`\chi2` (with PDF errors considered) for different datasets. The idea is that this way we select models of which the prediction is stable upon variations in the dataset.
 In order to obtain the PDF errors used in the figure of merit it is necessary to run multiple replicas, luckily ``n3fit`` provides such a possibility also during hyperoptimization.
 
@@ -371,6 +377,94 @@ The figure of merit will be the difference between the :math:`\chi2` of the seco
 .. math::
    L_{\rm hyperopt} = \chi^{2}_{(1) \rm pdferr} - \chi^{2}_{(2)}
 
+
+New hyperoptimization metrics with fold and replica statistics
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The combination of :ref:`k-folding <hyperkfolding-label>` and multi-replica experiments
+opens several the possibilities for the choice of figure of merit. The simplest option would be to minimize
+the average of :math:`\chi2` across both replica and k folds, *i.e.*,
+
+.. math::
+    L_{1} = \frac{1}{n_{\rm fold}} \sum_{k=1}^{n_{\rm fold}} \left< \chi^2_{k} \right>_{\rm rep}.
+
+In NNPDF, this hyperoptimisation metrics is selected via the following generic runcard:
+
+.. code-block:: yaml
+
+        dataset_inputs:
+        ...
+
+        kfold:
+          loss_type: chi2
+          replica_statistic: average
+          fold_statistic: average
+          partitions:
+          - datasets:
+            ...
+          - datasets:
+            ...
+
+        parallel_models: true
+        same_trvl_per_replica: true
+
+By combining the ``average``, ``best_worst``, and ``std`` figures of merit discussed in :ref:`hyperkfolding-label`,
+several alternatives may arise. For example, one approach could involve minimizing
+the maximum value of the set of averaged-over-replicas :math:`\chi2`,
+
+.. math::
+    L_{2} = {\rm max} \left ( \left< \chi^2_{1} \right>_{\rm rep}, \left< \chi^2_{2} \right>_{\rm rep}, ..., \left< \chi^2_{n_{\rm fold}} \right>_{\rm rep}\right),
+
+with correspond runcard ``kfold`` settings:
+
+.. code-block:: yaml
+
+        dataset_inputs:
+        ...
+
+        kfold:
+          loss_type: chi2
+          replica_statistic: average
+          fold_statistic: best_worst
+          partitions:
+          - datasets:
+            ...
+          - datasets:
+            ...
+
+An alternative metric that is
+sensitive to higher moments of the probability distribution
+has been defined in `NNPDF3.0 <https://link.springer.com/article/10.1007/JHEP04(2015)040>`_ [see Eq. (4.6) therein],
+namely, the :math:`\varphi` estimator. In the context of hyperopt, :math:`\varphi^{2}` can be calculated for each k-fold as
+
+.. math::
+  \varphi_{k}^2 = \langle \chi^2_k  [ \mathcal{T}[f_{\rm fit}], \mathcal{D} ] \rangle_{\rm rep} - \chi^2_k [ \langle \mathcal{T}[f_{\rm fit}] \rangle_{\rm rep}, \mathcal{D} ],
+
+where the first term represents the usual averaged-over-replicas :math:`\left< \chi^2_{k} \right>_{\rm rep}`
+calculated based on the dataset used in the fit (:math:`\mathcal{D}`) and
+the theory predictions from each fitted PDF (:math:`f_{\rm fit}`) replica.
+The second term involves the calculation of :math:`\chi2` but now with respect to the theory predictions from the central PDF.
+
+On the basis of :math:`\varphi`, we define the loss as
+
+.. math::
+    L_{3} = \left (\frac{1}{n_{\rm fold}} \sum_{k=1}^{n_{\rm fold}} \varphi_{k}^2 \right)^{-1},
+
+which selects hyperparameters that favor the most conservative extrapolation.
+In NNPDF, this figure of merit is chosen using the following settings:
+
+.. code-block:: yaml
+
+        kfold:
+          loss_type: phi2
+          fold_statistic: average
+            ...
+
+Alternatively, it is currently also possible to combine :math:`L_1` (which is only sensitive to the first moment)
+with :math:`L_3` (which provides information on the second moment).
+For example, one might minimize :math:`L_1` while monitoring the values of :math:`L_3` for a final model selection.
+The optimal approach for this combination is still under development.
+All the above options are implemented in the :class:`~n3fit.hyper_optimization.rewards.HyperLoss` class
+which is instantiated and monitored within :meth:`~n3fit.model_trainer.ModelTrainer.hyperparametrizable` method.
 
 Restarting hyperoptimization runs
 ---------------------------------
--- a/doc/sphinx/source/n3fit/hyperopt.rst
+++ b/doc/sphinx/source/n3fit/hyperopt.rst
@@ -381,7 +381,7 @@ The figure of merit will be the difference between the :math:`\chi2` of the seco
 New hyperoptimization metrics with fold and replica statistics
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The hyperopt measures discussed above are all based on performing a single replica fit per fold.
-However one may also wish to run the hyperoptimization algorithm on fits consisting of many
+However, one may also wish to run the hyperoptimization algorithm on fits consisting of many
 replicas per fold. This is a feasible option in ``n3fit``, since it has been optimised to
 efficiently run many replica fits in parallel on GPU.
 
--- a/doc/sphinx/source/n3fit/hyperopt.rst
+++ b/doc/sphinx/source/n3fit/hyperopt.rst
@@ -381,7 +381,7 @@ The figure of merit will be the difference between the :math:`\chi2` of the seco
 New hyperoptimization metrics with fold and replica statistics
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The combination of :ref:`k-folding <hyperkfolding-label>` and multi-replica experiments
-opens several the possibilities for the choice of figure of merit. The simplest option would be to minimize
+opens several possibilities for the choice of figure of merit. The simplest option would be to minimize
 the average of :math:`\chi2` across both replica and k folds, *i.e.*,
 
 .. math::
--- a/doc/sphinx/source/n3fit/hyperopt.rst
+++ b/doc/sphinx/source/n3fit/hyperopt.rst
@@ -382,7 +382,7 @@ New hyperoptimization metrics with fold and replica statistics
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The combination of :ref:`k-folding <hyperkfolding-label>` and multi-replica experiments
 opens several possibilities for the choice of figure of merit. The simplest option would be to minimize
-the average of :math:`\chi2` across both replica and k folds, *i.e.*,
+the average of :math:`\chi^2` across both replica and k folds, *i.e.*,
 
 .. math::
     L_{1} = \frac{1}{n_{\rm fold}} \sum_{k=1}^{n_{\rm fold}} \left< \chi^2_{k} \right>_{\rm rep}.
--- a/doc/sphinx/source/n3fit/hyperopt.rst
+++ b/doc/sphinx/source/n3fit/hyperopt.rst
@@ -405,7 +405,6 @@ In NNPDF, this hyperoptimisation metrics is selected via the following generic r
             ...
 
         parallel_models: true
-        same_trvl_per_replica: true
 
 By combining the ``average``, ``best_worst``, and ``std`` figures of merit discussed in :ref:`hyperkfolding-label`,
 several alternatives may arise. For example, one approach could involve minimizing
--- a/n3fit/src/n3fit/checks.py
+++ b/n3fit/src/n3fit/checks.py
@@ -9,6 +9,7 @@ import numpy as np
 
 from n3fit.hyper_optimization import penalties as penalties_module
 from n3fit.hyper_optimization import rewards as rewards_module
+from n3fit.hyper_optimization.rewards import HyperLoss
 from reportengine.checks import CheckError, make_argcheck
 from validphys.core import PDF
 from validphys.pdfbases import check_basis
@@ -255,15 +256,33 @@ def check_kfold_options(kfold):
             raise CheckError(
                 f"The penalty '{penalty}' is not recognized, ensure it is implemented in hyper_optimization/penalties.py"
             )
-    loss_target = kfold.get("target")
-    if loss_target is not None:
-        if not hasattr(rewards_module, loss_target):
+
+    loss_type = kfold.get("loss_type")
+    if loss_type is not None:
+        if loss_type not in HyperLoss().implemented_losses:
             raise CheckError(
-                f"The hyperoptimization target '{loss_target}' loss is not recognized, "
-                "ensure it is implemented in hyper_optimization/rewards.py"
+                f"Loss type '{loss_type}' is not recognized, "
+                "ensure it is implemented in the HyperLoss class in hyper_optimization/rewards.py."
+                "Options so far are 'chi2' or 'phi2'."
             )
+    replica_statistic = kfold.get("replica_statistic")
+    if replica_statistic is not None:
+        if replica_statistic not in HyperLoss().implemented_stats:
+            raise CheckError(
+                f"The replica statistic '{replica_statistic}' is not recognized, "
+                "ensure it is implemented in the HyperLoss class in hyper_optimization/rewards.py"
+            )
+    fold_statistic = kfold.get("fold_statistic")
+    if fold_statistic is not None:
+        if fold_statistic not in HyperLoss().implemented_stats:
+            raise CheckError(
+                f"The fold statistic '{fold_statistic}' is not recognized, "
+                "ensure it is implemented in the HyperLoss class in hyper_optimization/rewards.py"
+            )
+
     partitions = kfold["partitions"]
     # Check specific errors for specific targets
+    loss_target = kfold.get("fold_statistic")  # TODO: haven't updated this
     if loss_target == "fit_future_tests":
         if len(partitions) == 1:
             raise CheckError("Cannot use target 'fit_future_tests' with just one partition")
--- a/n3fit/src/n3fit/checks.py
+++ b/n3fit/src/n3fit/checks.py
@@ -9,7 +9,7 @@ import numpy as np
 
 from n3fit.hyper_optimization import penalties as penalties_module
 from n3fit.hyper_optimization import rewards as rewards_module
-from n3fit.hyper_optimization.rewards import HyperLoss
+from n3fit.hyper_optimization.rewards import IMPLEMENTED_LOSSES, IMPLEMENTED_STATS
 from reportengine.checks import CheckError, make_argcheck
 from validphys.core import PDF
 from validphys.pdfbases import check_basis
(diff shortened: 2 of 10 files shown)


APJansen:
# Description of what happens with hyperoptimization

For the benefit of my future self and my colleagues @goord and @Cmurilochem I describe here exactly what happens during hyperoptimization. It mainly describes how it works in master, so far this branch only changes it on two points, as indicated.

## Single Trial
A "trial" is the entire evaluation of one set of hyperparameters. One trial in the code is one execution of the function [`ModelTrainer.hyperparametrizable`](https://github.com/NNPDF/nnpdf/blob/3dd32b0de65a6f656dc729369d02581bcae8bbdd/n3fit/src/n3fit/model_trainer.py#L793).

### Folds

Trials are scored by their `hyper_loss`,  a validation loss computed using k-folding. 
This means that the datasets are partitioned into k (=4) parts. This is done manually to ensure that all folds cover the same kinematic range. It is configured in the runcard.
 
[For each fold](https://github.com/NNPDF/nnpdf/blob/3dd32b0de65a6f656dc729369d02581bcae8bbdd/n3fit/src/n3fit/model_trainer.py#L859), the datasets in that fold are left out completely, and the model is trained as normal on the remaining k-1 folds. 
Construction of the models for each fold happens [here](https://github.com/NNPDF/nnpdf/blob/3dd32b0de65a6f656dc729369d02581bcae8bbdd/n3fit/src/n3fit/model_trainer.py#L886). The training and validation model both use the datasets in the k-1 remaining folds, and have a different randomized per-dataset mask. The experimental model uses the held out fold, without any mask. It is this model that determines the hyper_loss.

[The model is trained](https://github.com/NNPDF/nnpdf/blob/3dd32b0de65a6f656dc729369d02581bcae8bbdd/n3fit/src/n3fit/model_trainer.py#L926) on the k-1 folds (masked with the training mask).

**Early exit 1**
It is said to have `passed` if any replica has a best epoch. This will only be false if for all replicas and for all epochs, the 
 validation chi2 is worse than at the previous epoch. (technically, if `passes` [here](https://github.com/NNPDF/nnpdf/blob/3dd32b0de65a6f656dc729369d02581bcae8bbdd/n3fit/src/n3fit/stopping.py#L554) is always false). If this fails, the trial is [exited early](https://github.com/NNPDF/nnpdf/blob/3dd32b0de65a6f656dc729369d02581bcae8bbdd/n3fit/src/n3fit/model_trainer.py#L930) and [given an infinite loss](https://github.com/NNPDF/nnpdf/blob/3dd32b0de65a6f656dc729369d02581bcae8bbdd/n3fit/src/n3fit/model_trainer.py#L977).

Then the [`hyper_losses`](https://github.com/NNPDF/nnpdf/blob/3dd32b0de65a6f656dc729369d02581bcae8bbdd/n3fit/src/n3fit/model_trainer.py#L943C30-L943C30), per replica are computed, as the loss on the held out fold, and penalties added (this branch fixes this to be per replica, it was being overcounted).

**Early exit 2**
If the *mean* per replica is above a threshold set in the runcard (the `threshold` key under `kfold`), the trial is again [exited early](https://github.com/NNPDF/nnpdf/blob/3dd32b0de65a6f656dc729369d02581bcae8bbdd/n3fit/src/n3fit/model_trainer.py#L956C8-L956C8) and given an infinite loss.

### Loss
We get `l_hyper`, an array of losses of shape (folds, replicas).
This is passed into a [`HyperLoss`](https://github.com/NNPDF/nnpdf/blob/3dd32b0de65a6f656dc729369d02581bcae8bbdd/n3fit/src/n3fit/hyper_optimization/rewards.py#L36) object (implemented in this branch), which computes the specified statistics first over replicas, then over folds.
This is then put into the `loss` key of the returned dictionary.

## Managing trials

Results are saved in `runcard/nnfit/replica_1/tries.json`, for all replicas (other replica folders are empty). It is a list of dictionaries, one for each trial.  The entire dictionary returned by hyperparametrizable is under the `"result"` key.

The trials start from the function `hyper_scan_wrapper`, called in [performfit](https://github.com/NNPDF/nnpdf/blob/3dd32b0de65a6f656dc729369d02581bcae8bbdd/n3fit/src/n3fit/performfit.py#L228). The best parameters are printed, and it seems retraining a model with those parameters is to be done manually.

Inside `hyper_scan_wrapper`, the hyperopt library's [fmin](https://github.com/NNPDF/nnpdf/blob/3dd32b0de65a6f656dc729369d02581bcae8bbdd/n3fit/src/n3fit/hyper_optimization/hyper_scan.py#L123) function is called.
The hyperoptimization algorithm used is a Tree structured Parzen Estimator (TPE), section 4 of [this paper](https://proceedings.neurips.cc/paper_files/paper/2011/file/86e8f7ab32cfd12577bc2619bc635690-Paper.pdf). 
It can handle hierarchical hyperparameters, i.e. the number of units needs to be chosen for each layer, and the number of layers can vary too. I didn't read in any detail, but it does use previous results to suggest new trials.

## possible issue
 on a test with 2 trials varying only the nodes per layer, it tested (27,12,8) and then (30,36,8), but the best parameters printed were the default (25,20,8). I'm not sure what exactly happens in this final call to [`space_eval`](https://github.com/NNPDF/nnpdf/blob/3dd32b0de65a6f656dc729369d02581bcae8bbdd/n3fit/src/n3fit/hyper_optimization/hyper_scan.py#L131).

# Some thoughts on improvements

## Parallelizing trials, checkpointing

Trials can be run in parallel using MongoDB, as described [here](http://hyperopt.github.io/hyperopt/scaleout/mongodb/#parallelizing-evaluations-during-search-via-mongodb). 
It seems this could also be used to enable continuing from a previous run as discussed in issue #1800.

## K-folding

When using K-folding, we now have 3 classes of data at every fold:
1. training data: in the k-1 folds that ARE used, seen during training
2. validation data: in the k-1 folds that ARE used, not seen during training
3. "hyper-validation" (not sure what to call it) data: in the held out fold

Both 2 and 3 are not seen during training. The difference in how they are used is that 2. is used for the "early exit 1" above, while 3 is used to compute the hyper loss. If that was all, I'd say 2 is wasted as this condition is very unlikely to trigger even with one replica. 
But actually 2 is also used for early stopping. So we do want to keep all 3, as combining 2 and 3 would introduce a bias.

We also discussed a long time ago that perhaps K-folding won't be necessary anymore. The point of it is to have a more accurate validation loss, but if we run with say 100 replicas, it is already more accurate. We would still need 3 groups of data, to avoid the bias of joining 2 and 3, however they could be split randomly per dataset, (per replica). The advantage would be that it'd require only a single run per trial (slightly bigger, but still nearly 4x speedup), and simplify the code. Not sure though what the effect on the learning would be.

## Code improvements
- It would be nice to move as much as possible of `ModelTrainer.hyperparametrizable` into one of the hyperopt modules
- models are recreated from scratch every fold, PDFs only need to be re-initialized. In fact the PDFs could be generated only once per entire hyperopt run (and copied and reinitialized), and the observables only once per fold, and copied and re-initialized for each trial.
- It seems that the best hyperparameters are just printed, and not saved anywhere?
- Maybe we want to add (optionally) rerunning a training run without K-folding using the best parameters.

Cmurilochem:
This PR also adds a new hyperoptimization metrics to the newly implemented [`HyperLoss` class](https://github.com/NNPDF/nnpdf/blob/hyperopt_loss/n3fit/src/n3fit/hyper_optimization/rewards.py#L36C7-L36C16) , $`\varphi^{2}`$ [Eq.(4.6) of the [NNPDF3.0 paper](https://link.springer.com/article/10.1007/JHEP04(2015)040) and #1849 ]. $`\varphi^{2}`$ is defined for each $k$-fold as
$$\Huge \varphi_{\chi^2_k}^2 = \langle \chi^2_k  [ \mathcal{T}[f_{\rm fit}], \mathcal{D} ] \rangle_{\rm rep} - \chi^2_k [ \langle \mathcal{T}[f_{\rm fit}] \rangle_{\rm rep}, \mathcal{D} ] $$

where the first term represents our usual averaged-over-replicas hyper loss, $`\chi^2_k`$, that is calculated based on the dataset used in the fit ($`\mathcal{D}`$) and the theory predictions from each fitted PDF ($f_{\rm fit}$) replica. The second term of the above equation would involve the calculation of the hyper loss but now using the theory predictions from the averaged-over-replicas PDF.

APJansen:
We talked again about refactoring this, and we decided to keep the changes minimal for now, just to allow implementation of the phi2 loss. We can refactor the code about hyperoptimization inside model_trainer later, once we have things running.

We made some assumptions listed below, are these correct @scarlehoff?
And it's still missing the computation of the phi2 itself, that's still quite tricky.
For now we assumed that there is a function `compute_phi2` in `vpinterface.py` that takes n3pdfs and a dictionary of data and covmat, but we didn't implement anything. We extract these from `self.experimental["output"]`, but still need to convert these to the arguments that `result` expects, so `DataSetSpec` etc.
Is that right? You mentioned in #1849 that you would come back to it, I think some help here would be useful if you know more.

# Assumptions 

Previously `HyperLoss` was implemented with the assumption that we can first compute all losses per replica and per fold, and at the end compute first a statistic over the replicas, and then over the folds.
This is not true for phi2, as the chi2 of the central pdf cannot be obtained by any statistic over the chi2's of individual pdfs.

So now we assume instead that first the losses are computed per fold, so already having some kind of aggregation of the replicas. This is done in the training loop, as it is quit early if it is too high.
Then we assume that after a trial, we just take some statistic over the fold, as before. We assume this is done for chi2 as well as phi2.

And then there are also penalties, which currently are all computed per replica. We assumed for now that this is general enough, and that we always want to take an average over the per-replica penalties, and add this to the loss.

Cmurilochem:
## Start implementation of $\varphi^{2}$
Continuing on #1849, I am proceeding with a provisory implementation of [compute_phi2](https://github.com/NNPDF/nnpdf/blob/b6911aec2246d9fcb46b2ace11a0a3b1ffee1ba1/n3fit/src/n3fit/vpinterface.py#L343) in order to calculate $\varphi^{2}_k$ with each $k$ fold.

### Implementation ideas (also in #1849)

- With the help of @scarlehoff I saw that `validphys` pass an `experiments_data` variable to `performfit`. It contains a list of [DataGroupSpec](https://github.com/NNPDF/nnpdf/blob/b6911aec2246d9fcb46b2ace11a0a3b1ffee1ba1/validphys2/src/validphys/core.py#L540) objects for each group exp data that could be directly used by `validphys.results` functions to obtain $\varphi$.
- I am now passing `experiments_data` as argument to instantiate `ModelTrainer` inside `performfit`.
- With `self.experiments_data` and `self.exp_info` attributes of `ModelTrainer`, I create inside `Model.parametrizable`  a list of tuples defining `validphys.core.DataGroupSpec` representations of each group experimental data set and the associated covariant matrices.
- This list and the `N3PDF` object are them used by [HyperLoss.compute_loss](https://github.com/NNPDF/nnpdf/blob/b6911aec2246d9fcb46b2ace11a0a3b1ffee1ba1/n3fit/src/n3fit/hyper_optimization/rewards.py#L68C9-L68C9) and passed as argument to `compute_phi2` in `vpinterface.py`. This allows us to calculate the sum of $\varphi^2$ by resorting to the built-in `validphys.results` functions: `results`, `abs_chi2_data` and `phi_data`.

## Questions and ToDos

- the current calculation of $\varphi^2_{k}$ for each $k$ fold uses all input experimental data sets. Would the correct calculation include only the experimental data of the held out fold only ?
- I am now using the multi-replica `N3PDF` objects to obtain $\varphi^2_{k}$ as expected by [results](https://github.com/NNPDF/nnpdf/blob/b6911aec2246d9fcb46b2ace11a0a3b1ffee1ba1/validphys2/src/validphys/results.py#L463). I noticed that there was some discussion about [_set_central_value](https://github.com/NNPDF/nnpdf/blob/b6911aec2246d9fcb46b2ace11a0a3b1ffee1ba1/n3fit/src/n3fit/hyper_optimization/rewards.py#L138). This has also been discussed by @scarlehoff above. It is still not very clear to me how this is done by [abs_chi2_data](https://github.com/NNPDF/nnpdf/blob/b53389a70bddeea5e17d411adbed80b233055452/validphys2/src/validphys/results.py#L542) and the possible difference between this approach and setting the central pdf as `PDF_0` by `_set_central_value`.
- Need to do more extensive testings

@scarlehoff, @APJansen and @goord, any thoughts on this ?

Cmurilochem:
## Question regarding added tests

@scarlehoff, I added new tests into (test_hyperopt.py)[https://github.com/NNPDF/nnpdf/blob/hyperopt_loss/n3fit/src/n3fit/tests/test_hyperopt.py#L40].

To test the calculation of $\varphi$, I am generating mocked [pdf_models](https://github.com/NNPDF/nnpdf/blob/8f5acc43dce4ebfe8929ddc33a4c8da86a7d07d2/n3fit/src/n3fit/tests/test_hyperopt.py#L13) by providing random seeds; see [here](https://github.com/NNPDF/nnpdf/blob/8f5acc43dce4ebfe8929ddc33a4c8da86a7d07d2/n3fit/src/n3fit/tests/test_hyperopt.py#L54).  However, even when providing the same seeds to `pdfNN_layer_generator`, I am not able to obtain the same test result when I run it on my local computer and in the CI/CD. Do you think that this would be due to the different TensorFlow versions ? If not, how could I warrant that the weights/pre-processing exponents of my `pdf_models` are exactly the same ?

scarlehoff:
> even when providing the same seeds ...

Make sure the seeds are also fixed for numpy. Note that when the fit is set to `debug` we do all this https://github.com/NNPDF/nnpdf/blob/e470c7e73490f3d4725dde6813aba2a2de018b69/n3fit/src/n3fit/backends/keras_backend/internal_state.py#L106 to ensure reproducibility. You might need some of these steps in your test as well.

_however_ while the tests should be robust in linux (they empirically seem to be) I'm not so sure about mac m1 (i.e., if you are locally running on a mac and then trying to test in the ci... not sure what will happen)

Cmurilochem:
> > even when providing the same seeds ...
> 
> Make sure the seeds are also fixed for numpy. Note that when the fit is set to `debug` we do all this
> 
> https://github.com/NNPDF/nnpdf/blob/e470c7e73490f3d4725dde6813aba2a2de018b69/n3fit/src/n3fit/backends/keras_backend/internal_state.py#L106
> 
> to ensure reproducibility. You might need some of these steps in your test as well.
> _however_ while the tests should be robust in linux (they empirically seem to be) I'm not so sure about mac m1 (i.e., if you are locally running on a mac and then trying to test in the ci... not sure what will happen)

Thanks @scarlehoff. I have added a custom `set_initial_state` to be used together with `generate_pdf` in `test_hyperopt.py`. After some unsuccessful trials, I decided to add only property-based tests for $\varphi$ for the moment. 
I will come back later to it. For now, I will try first to address the problem of calculating $\varphi^{2}$ only with experimental data of the held out set.

n3fit/src/n3fit/vpinterface.py
```diff
@@ -331,3 +333,57 @@ def compute_arclength(self, q0=1.65, basis="evolution", flavours=None):
...
+            res = results(datasetspec, n3pdf, covmat, sqrt_covmat(covmat))
+
+            # calculate standard chi2 (all_chi2) and chi2 using PDF central values (central_chi2)
+            chi2 = abs_chi2_data(res)
+
+            # calculate phi and store phi**2
+            phi, _ = phi_data(chi2)
+            sum_phi2 += phi**2
+
+    return sum_phi2
```
RoyStegeman, line 389:
```suggestion
            # calculate phi and store phi**2
            phi, ndat = phi_data(chi2)
            sum_phi += np.sqrt(ndat)*phi
            ndat_tot += ndat

    return sum_phi/np.sqrt(ndat_tot)
```
Where you initialise `ndat_tot=0` outside the loop. 

Since it's a secondary measure and only the hierarchy in phi value between different trials is what's used in hyperopt it doesn't really make a difference whether we use phi or phi2. However, looking at the paper where it says "an indicator which measures the standard deviation over the replica sample in units of the data uncertainty." we may want to go with phi.

n3fit/src/n3fit/hyper_optimization/hyper_scan.py
```diff
@@ -23,6 +24,11 @@
 
 log = logging.getLogger(__name__)
 
+# Hyperopt uses these strings for a passed and failed run
+# it also has statusses "new", "running" and "suspended", but we don't use them
```
RoyStegeman, line 28:
```suggestion
# it also has statuses "new", "running" and "suspended", but we don't use them
```

n3fit/src/n3fit/hyper_optimization/rewards.py
```diff
@@ -24,30 +24,300 @@
...
+        Parameters
+        ----------
+            fold_losses: np.ndarray
+                Input array.
+            axis: int, optional
+                Axis along which the mean is computed. Default is 0.
+
+        Returns
+        -------
+            np.ndarray: The average along the specified axis.
```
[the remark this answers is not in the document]
Cmurilochem: Thanks for spotting these ones. I made a separate commit in which I correct all of them at once.

n3fit/src/n3fit/hyper_optimization/rewards.py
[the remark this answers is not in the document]
Cmurilochem: Thanks @RoyStegeman. Done!

n3fit/src/n3fit/vpinterface.py
Cmurilochem: I am not sure if I understood well but I will accept anyway. Will this affect later on our measure of inverse phi2 in the model selection ? I am asking because this is the function that will calculate all phi2 per replica and fold as output in the `tries.json` file.

n3fit/src/n3fit/hyper_optimization/rewards.py
```diff
@@ -24,30 +24,300 @@
...
+        self.reduce_over_folds = self._parse_statistic(fold_statistic, "fold_statistic")
+
+        self.phi2_vector = []
+        self.chi2_matrix = []
+
+        self.penalties = {}
+
+    def compute_loss(
+        self,
+        penalties: Dict[str, np.ndarray],
```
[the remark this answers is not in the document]
Cmurilochem: @RoyStegeman. They are essentially the same. The only thing is that they are converted to a dictionary here:
https://github.com/NNPDF/nnpdf/blob/d5b25b43e9eb9cdc190cb0ca769357a74f1ed5da/n3fit/src/n3fit/model_trainer.py#L996-L999
before entering `compute_loss`; this will later on make it easier handling these penalties in the overall statistics. Just in case, I updated the docstring and example of `compute_loss` to clarify it.

n3fit/src/n3fit/vpinterface.py
RoyStegeman: I agree the comment was a bit confusing as I was talking about one thing but the change mainly addresses something I don't mention :sweat_smile:. The two things I change here are: 
1. compute phi instead of phi2 
2. change how phi2 is computed, see also this function: https://github.com/NNPDF/nnpdf/blob/162380804ca964da6de4c8ce47acddc36f25543b/validphys2/src/validphys/results.py#L678

The first one doesn't make much difference for the hyperopt we do since the absolute value of the phi measure is irrelevant and we just want to order trials by their value of phi (which is the same order if we were to use phi2). 

The second one actually changes the computation of phi.

Do you agree?

n3fit/src/n3fit/model_trainer.py
```diff
@@ -981,20 +1047,32 @@ def hyperparametrizable(self, params):
...
                 "status": passed,
-                "loss": self._hyper_loss(
-                    fold_losses=l_hyper, pdfs_per_fold=pdfs_per_fold, experimental_models=exp_models
-                ),
+                "loss": final_hyper_loss,
                 "validation_loss": np.average(l_valid),
                 "experimental_loss": np.average(l_exper),
                 "kfold_meta": {
                     "validation_losses": l_valid,
+                    "validation_losses_phi2": np.array(trvl_phi2_per_fold),
```
RoyStegeman, line 1068: This is called `validation_losses_phi2` but I think it's computed over both validation and training datasets?

RoyStegeman (commented):
Can you add documentation for the different losses/statistics that can be optimised?

n3fit/src/n3fit/vpinterface.py
Cmurilochem: Thanks @RoyStegeman. It it now clear from the docs in `total_phi_data_from_experiments`. Just changed then the name of `compute_phi2` to `compute_phi`

RoyStegeman: Thanks!

n3fit/src/n3fit/hyper_optimization/hyper_scan.py
```diff
@@ -139,6 +145,19 @@ def hyper_scan_wrapper(replica_path_set, model_trainer, hyperscanner, max_evals=
...
+    """
+
+    def wrapped(*args, **kwargs):
+        results_dict = hyperparametrizable(*args, **kwargs)
+        results_dict["status"] = HYPEROPT_STATUSSES[results_dict["status"]]
+        return results_dict
+
+    return wrapped
+
+
```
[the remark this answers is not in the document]
APJansen: Fixed in [6d785be](https://github.com/NNPDF/nnpdf/pull/1726/commits/6d785bed6be894d84061a4015e37a71f6bb24f36) (as well as comment above)

n3fit/src/n3fit/vpinterface.py
Cmurilochem: Done @RoyStegeman!

n3fit/src/n3fit/model_trainer.py
Cmurilochem: Yes, @RoyStegeman you are right. Do you suggest any better provisory name for it (until we actually do that if we have time) ?

RoyStegeman: I don't mind the `trvl` as you also use it in  `trvl_phi2_per_fold`, at least that doesn't risk causing the same confusion as specifying validation while it is both validation and training

Cmurilochem: Done! Thanks!

Cmurilochem:
> Can you add documentation for the different losses/statistics that can be optimised and how the runcard can be changed to use them?

Done. Just added docs in [f88fbb5](https://github.com/NNPDF/nnpdf/pull/1726/commits/f88fbb5414bdfce24ea2b21ea53025c11310a6af)

doc/sphinx/source/n3fit/hyperopt.rst
```diff
@@ -372,6 +378,94 @@ The figure of merit will be the difference between the :math:`\chi2` of the seco
    L_{\rm hyperopt} = \chi^{2}_{(1) \rm pdferr} - \chi^{2}_{(2)}
 
 
+New hyperoptimization metrics with fold and replica statistics
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The combination of :ref:`k-folding <hyperkfolding-label>` and multi-replica experiments
+opens several the possibilities for the choice of figure of merit. The simplest option would be to minimize
+the average of :math:`\chi2` across both replica and k folds, *i.e.*,
```
RoyStegeman, line 385:
```suggestion
the average of :math:`\chi^2` across both replica and k folds, *i.e.*,
```

doc/sphinx/source/n3fit/hyperopt.rst
```diff
@@ -372,6 +378,94 @@ The figure of merit will be the difference between the :math:`\chi2` of the seco
    L_{\rm hyperopt} = \chi^{2}_{(1) \rm pdferr} - \chi^{2}_{(2)}
 
 
+New hyperoptimization metrics with fold and replica statistics
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The combination of :ref:`k-folding <hyperkfolding-label>` and multi-replica experiments
```
RoyStegeman, line 383: Maybe provide a brief intro in what you mean here
> The combination of :ref:`k-folding <hyperkfolding-label>` and multi-replica experiments

Perhaps something like the following:
> The hyperopt measures discussed above are all based on performing a single replica fit per fold. However one may also wish to run the hyperoptimization algorithm on fits consisting of many replicas per fold. This is a feasible option in ``n3fit``, since it has been optimised to efficiently run many replica fits in parallel on GPU.

doc/sphinx/source/n3fit/hyperopt.rst
```diff
@@ -372,6 +378,94 @@ The figure of merit will be the difference between the :math:`\chi2` of the seco
...
+opens several the possibilities for the choice of figure of merit. The simplest option would be to minimize
```
RoyStegeman, line 384:
```suggestion
opens several possibilities for the choice of figure of merit. The simplest option would be to minimize
```

doc/sphinx/source/n3fit/hyperopt.rst
```diff
@@ -372,6 +378,94 @@ The figure of merit will be the difference between the :math:`\chi2` of the seco
...
+          replica_statistic: average
+          fold_statistic: average
+          partitions:
+          - datasets:
+            ...
+          - datasets:
+            ...
+
+        parallel_models: true
+        same_trvl_per_replica: true
```
RoyStegeman, line 408:
```suggestion
```

RoyStegeman (approved):
Thanks, the docs look good

doc/sphinx/source/n3fit/hyperopt.rst
Cmurilochem: Done. Thanks!

doc/sphinx/source/n3fit/hyperopt.rst
```diff
@@ -372,6 +378,97 @@ The figure of merit will be the difference between the :math:`\chi2` of the seco
    L_{\rm hyperopt} = \chi^{2}_{(1) \rm pdferr} - \chi^{2}_{(2)}
 
 
+New hyperoptimization metrics with fold and replica statistics
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The hyperopt measures discussed above are all based on performing a single replica fit per fold.
+However one may also wish to run the hyperoptimization algorithm on fits consisting of many
+replicas per fold. This is a feasible option in ``n3fit``, since it has been optimised to
+efficiently run many replica fits in parallel on GPU.
```
RoyStegeman, line 386:
```suggestion
efficiently run many replica fits in parallel on GPU.

```

doc/sphinx/source/n3fit/hyperopt.rst
```diff
@@ -372,6 +378,97 @@ The figure of merit will be the difference between the :math:`\chi2` of the seco
    L_{\rm hyperopt} = \chi^{2}_{(1) \rm pdferr} - \chi^{2}_{(2)}
 
 
+New hyperoptimization metrics with fold and replica statistics
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The hyperopt measures discussed above are all based on performing a single replica fit per fold.
+However one may also wish to run the hyperoptimization algorithm on fits consisting of many
```
RoyStegeman, line 384:
```suggestion
However, one may also wish to run the hyperoptimization algorithm on fits consisting of many
```

n3fit/src/n3fit/checks.py
```diff
@@ -255,15 +256,33 @@ def check_kfold_options(kfold):
...
+                "ensure it is implemented in the HyperLoss class in hyper_optimization/rewards.py"
             )
+    fold_statistic = kfold.get("fold_statistic")
+    if fold_statistic is not None:
+        if fold_statistic not in HyperLoss().implemented_stats:
+            raise CheckError(
+                f"The fold statistic '{fold_statistic}' is not recognized, "
+                "ensure it is implemented in the HyperLoss class in hyper_optimization/rewards.py"
+            )
+
```
RoyStegeman, line 282: Please do these checks without initializing an instance of `HyperLoss`, since it triggers the warnings in `_parse_statistics` and `_parse_loss`

PR #1728, opened 2023-05-05
Combine t0 tables

Zaharid:
Use the "namespaces_" trick to properly set the t0 pdf parameters.

### DIFF
--- a/validphys2/src/validphys/comparefittemplates/comparecard.yaml
+++ b/validphys2/src/validphys/comparefittemplates/comparecard.yaml
@@ -128,14 +128,17 @@ dataspecs:
       from_: reference
     speclabel:
       from_: reference
-      
-t0_info:
-  - use_t0: True
+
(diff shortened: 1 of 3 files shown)


Zaharid:
Note this is untested

Zaharid:
@RoyStegeman I recognize this is not as obvious as it could. Ideally `from_: fit` would bind late and you would be able to write `{@t0_info dataspecs_t0_chi2_info_table @}`, with `fit` being resolved for each dataspec. In practice making that work is "here be dragons" territory (both because of the code and coming up with sensible semantics in edge cases such as what happens when `fit` is needed to resolve  a namespace). 

In essence `namespaces_` is a hack that allows combining namespaces so as to make these things work somewhat elegantly.  The hack part being mostly the implementation.

RoyStegeman:
This doesn't have the expected behaviour.
image

PR #1731, opened 2023-05-11
Fixing logs of evolven3fit_new

andreab1997:
The logs of `evolven3fit_new` are broken. This is fixing it. 

@giacomomagni


scarlehoff:
What do you mean by broken?

andreab1997:
> What do you mean by broken?

It is not logging the information but just a  constant string

scarlehoff:
Ah, ok. But the evolution does work. It is just the content of the log which is not useful, right?

andreab1997:
> Ah, ok. But the evolution does work. It is just the content of the log which is not useful, right?

Yes exactly, the evolution is fine. It is just a logging problem

giacomomagni (approved):

giacomomagni (approved):

PR #1733, opened 2023-05-15
FTDY maxTau cuts in runcard

andreab1997:
We want to add the possibility of specifying the `maxTau` cut for FTDY in the runcard.


andreab1997:
Sorry for the large amount of commits but this has been rebased on #1716 otherwise I could not test it with the new fktables.

RoyStegeman:
Ah, but `allow_optfk` has not been rebased in a while

andreab1997:
> Ah, but `allow_optfk` has not been rebased in a while

Ok let me rebase both on master then

RoyStegeman:
The history of the pr also works with merge it just doesn't update automatically (that's why I changed the base branch). The "problem" is that some of the old commits in master are overwritten by this PR, I put problem in parentheses because it doesn't affect the diffs and I don't care that the authors of the commits change.

scarlehoff:
It' easier to review if the PR only contains the changes from the PR. Right now the diff includes changes from `allow_optfk`.

validphys2/src/validphys/config.py
```diff
@@ -1421,16 +1421,19 @@ def produce_defaults(
...
     ):
-        """Produce default values for filters taking into account both the
-        values of ``q2min`` and ` `w2min`` defined at namespace
+        """Produce default values for filters taking into account the
+        values of ``q2min``, ` `w2min``, ``maxTau``, ``max_M`` and ``min_M`` defined at namespace
         level and those inside a ``filter_defaults`` mapping.
         """
         from validphys.filters import default_filter_settings_input
-
+
```
RoyStegeman, line 1436: trailing whitespaces

andreab1997:
Is there anything left here? Can this be merged?

RoyStegeman:
The description is 
> We want to add the possibility of specifying the `maxTau` cut for FTDY in the runcard. 

but then you also include `min_M` and `max_M`, though not `maxY` and `max_rapidity`. What is the reason for these choices? Also, I'm a little uncomfortable with the exception for `ATLAS_DY_2D_8TEV_LOWMASS` which is of course something people are not going to be aware of when they change `max_M` in the runcard. 

P.S. if I am not mistaken @enocera wouldn't have gotten a notification for your tag if it's added in an edit

andreab1997:
> The description is
> 
> > We want to add the possibility of specifying the `maxTau` cut for FTDY in the runcard.
> 
> but then you also include `min_M` and `max_M`, though not `maxY` and `max_rapidity`. What is the reason for these choices? Also, I'm a little uncomfortable with the exception for `ATLAS_DY_2D_8TEV_LOWMASS` which is of course something people are not going to be aware of when they change `max_M` in the runcard.
> 
> P.S. if I am not mistaken @enocera wouldn't have gotten a notification for your tag if it's added in an edit

Yes you are right, we should probably choose now what we want to do. Probably the most sensible thing is to allow only the FTDY cut, do you agree?

RoyStegeman:
I would indeed have no problem with adding the `maxTau` cut for FTDY in a runcard. That is assuming there is no reason to think that we'll want different values for different datasets in the future and end up having to implement an exception such as is done for `ATLAS_DY_2D_8TEV_LOWMASS` in this PR.

RoyStegeman (approved):
For me this is fine, but before merging I'd indeed like @enocera to approve as well.

PR #1734, opened 2023-05-17
Refactoring model creation code

APJansen:
## Aim
I am refactoring the code that creates the PDF model (and perhaps the full one as well). The main objective is to make the code and the resulting architecture more readable, without changing anything in the behavior. 
If anything I expect the changes to make it slightly faster and smaller in memory, but either way this will likely be negligible.

It is still a work in progress, but here is a summary of the changes I made so far, roughly from big to small:
- I changed the structure of the sum rule code. Before there was a function `msr_impose` that creates a function that takes a pdf function and returns a normalized pdf function. Now it is a tensorflow model that takes as input the unnormalized pdf on the grid and on the integration grid, and the integration grid itself, and outputs the normalized pdf. This makes the code and the resulting model plots much easier to read.
- I grouped all the actual neural layers into a neural network model, for the same reason
- All of the smaller steps I put into named lambda layers, shared wherever possible between replicas and grids

With these changes, it's possible to create much more understandable plots of the architecture using `tf.keras.utils.plot_model`. These are from the basic runcard:
- A single replica PDF
pdf_model
- The NN block
NN_model
- The MSR block
msr_model


RoyStegeman:
Thanks @APJansen! What is the status of this PR, do you want to merge it or are you still working on it? Could you please have a look at the failed test before though?

Perhaps you already figured out the answers to the questions in your comments above, but if not:

> Do nothing, x -> x I assume this is never used?

Indeed, there is no use-case for this

> **Wouldn't it be better to just remove `inp` from `pdfNN_layer_generator` as a parameter and base it on `scaler`, saying if scalar is None, add logs, otherwise not?**

Indeed, having the `inp` parameter does not make much sense

> **I assume the scaling happens before the model because it's not possible to use scipy functions inside a tensorflow layer?**

Indeed it happens before the model. It also needs to be taken into account when doing the integration, but I guess by now you understand very well how this was done.

> After the scaling, inside the PDF model, the grid is mapped from the interval [0, 1] to [-1, 1]. **Why not include this in the scaler?**

To be honest I don't remember, though if I had to guess it would simply be because it made it easier to test the impact of the domain. There is no longer any reason not to include it in the scaler at this point.

> **Rather than concatenating scaler(x) to x and then splitting it again, wouldn't it be simpler to have separate inputs?**
You'd have 5 inputs: the scaled and the original grid, the scaled and the original integration grid, and the scaled x=1 grid (not the original as that's used for the preprocessing prefactor, but x=1 only goes into the NN.
For the case of adding logs instead, you could have 4 inputs to keep it as similar as possible, (x, log(x)) for both grids as input to the NN and the original x for both grids as input to the preprocessing prefactor.

I'm not sure I understand your proposal. The scalar is not concatenated to x, while indeed log(x) is, but your proposal does keep (x,logx) together (which I think is sensible). At the moment I think the model generation script is rather easy to understand, how would your proposal improve that?

> **Unrelated: The FkRotation docstring is not up to date, it says the input is 8-dimensional but it's actually 9-dimensional, as you can see clearly from the model plot. I don't know enough to fix this but it'd be good if someone could.**

You're right, this was recently changed but the docstring has not been updated.

> **Do you want me to add these plots as an option in the logging, and if so can you point me to where and how to best do this?**

I don't think that will be necessary, some people may find them useful but I don't think this is something we need to support, others may disagree though. Also, I believe tensorboard has a similar feature. I haven't used it in a while so I'm not sure if it's as clear, but did you have a look a that?

APJansen:
Thanks for the comments @RoyStegeman! I think it's almost ready to merge, of course there is always more that can be done but I wanted some feedback first. I put the status at the top of the first post.

> I'm not sure I understand your proposal. The scalar is not concatenated to x, while indeed log(x) is, but your proposal does keep (x,logx) together (which I think is sensible). At the moment I think the model generation script is rather easy to understand, how would your proposal improve that?

Actually the scaler does concatenate to x, after scaling the input consists of pairs (scaler(x), x), which inside the PDF model are split again. The proposal was to have a separate scaled x input, but I agree it doesn't make much of a difference.

> I don't think that will be necessary, some people may find them useful but I don't think this is something we need to support, others may disagree though. Also, I believe tensorboard has a similar feature. I haven't used it in a while so I'm not sure if it's as clear, but did you have a look a that?

Tensorboard does in principle show a graph of the model as well, but currently it gives an error. And even if that's fixed I'm not sure how readable it will be, I can try to have a look. 
For me having these plots was super useful to understand what's going on. All I need to integrate this is two dependencies (pydot and graphviz) and to know how to add an option in the runcard. Alternatively (or additionally) perhaps they could be added to the documentation.

scarlehoff:
If the model structure has changed, you will need to modify also the weights here:

[n3fit/runcards/examples/developing_weights.h5](https://github.com/NNPDF/nnpdf/blob/b3595a62538486a23d25255937d03c71828adb6a/n3fit/runcards/examples/developing_weights.h5)

These are used by the [fitbot workflow](https://github.com/NNPDF/nnpdf/blob/master/.github/workflows/fitbot.yml)  (which uses the developing.yml runcard in the same folder).

scarlehoff:
> Locally it still fails, but I think that's a Mac issue as this test is also skipped for the mac in the CI

Yes, the numbers are slightly different in Mac/Linux so we prioritised it being an exact regression to working in both systems.

> "If the results are known to need a change, it is necessary to flag something to regenerate regression"

Since the tests have passed, then the results do not need a change! By the results we mean the actual results of the regression test. The weights are just an starting point to speed it up.
If you could also convert the weight for the fitbot that would be really useful, then we can run the whole fit with the bot and have a report to see whether something changes.

scarlehoff:
> are you sure you're comfortable with me changing these weights "manually"?

The fact that the tests are passing means that the network, initiated with those weights, and after ~1000 epochs, finishes in the exact same place as with the old weights. Maybe they are not completely correct, but if so we will see it with the comparison that the fitbot generates.

APJansen:
@scarlehoff I modified the develop weights as well and the fitbot passed :) Or at least it ran, I'm not able to open the report.

Radonirinaunimi:
> @scarlehoff I modified the develop weights as well and the fitbot passed :) Or at least it ran, I'm not able to open the report.

Please check the slack channel to view the report.

n3fit/src/n3fit/tests/test_xops.py
```diff
@@ -0,0 +1,33 @@
...
+    custom_indices = [0, 1, 7]
+    x_div = xDivide(div_list=custom_indices)
+    test_input = np.array([1, 2, 3], dtype=np.float32).reshape((1, 3, 1))
+    test_output = x_div(test_input)
+
+    expected_output = np.ones(shape=(1, 3, 14))
+    for i in custom_indices:
+        expected_output[:, :, i] = 1 / test_input[:, :, 0]
+
+    assert np.allclose(test_output, expected_output)
```
scarlehoff, line 32: Same as above.

n3fit/src/n3fit/layers/preprocessing.py
```diff
@@ -5,11 +5,11 @@
...
 class Preprocessing(MetaLayer):
     """
-    Applies preprocessing to the PDF.
+    Computes prefactor for the PDF.
 
     This layer generates a factor (1-x)^beta*x^(1-alpha) where both beta and alpha
-    are model paramters that can be trained. If feature scaling is used, the preprocessing
-    factor is x^(1-alpha).
+    are model paramters that can be trained. If feature scaling is used, the
+    prefactor is x^(1-alpha).
```
scarlehoff, line 12: Please leave "preprocessing" in the documentation since this is the notation for this quantity that we have been using in papers [example here](https://arxiv.org/pdf/2109.02653.pdf).

n3fit/src/n3fit/model_gen.py
```diff
@@ -396,29 +399,29 @@ def output_layer(ilayer):
...
+    out: int = 14,
+    seed: int = None,
+    dropout: float = 0.0,
     regularizer=None,
     regularizer_args=None,
-    impose_sumrule=None,
+    impose_sumrule: str = None,
     scaler=None,
-    parallel_models=1,
+    parallel_models: int = 1,
```
scarlehoff, line 415: Same as above, please complete the annotations.

scarlehoff (commented):
> @scarlehoff I modified the develop weights as well and the fitbot passed :) Or at least it ran, I'm not able to open the report.

It looks fine as far as I can see.

RE the PR

I'll have a look at the changes to the code when I have some time. I've had a very quick look and left a few comments. In general I think they are ok and make the code actually easier to read.
You said though in some other comment that they were not final, what is the actual status? (i.e., what do you want to do still to make them final?)

I would suggest to leave the actual changes to the code as they are (or with small modifications) and just add the documentation that @RoyStegeman asked for (I agree with him, the reporting works best as documentation explaining that you can do this or that since it is only used for debugging, and very specific debugging). The reason for that is that this PR is already nearing the size of what can be humanly looked at in the time we have for that.

Some things that I would like to see tested (maybe you did already!) of which the last point is probably the most important one.
 - [ ] PDF fitting with the photon. Sadly the photon fit is not part of the regression test or the fitbot yet since we need to create a testing theory for that (it's WIP #1738) but in the meantime you would need to manually perform a fit with the photon from master and from this branch and check that the results are the same. Since the regression tests are passing I would expect the results to be even numerically the same.
 - [ ] PDF fitting with feature scaling: since you did the report I guess you did already, but better safe than sorry.
 - [ ] PDF fitting in flavour basis, since the normal fit is working this one should as well but again... better safe than sorry!
 - [ ] What is the memory usage of the fit before and after the changes? And the speed of the fit? Some of the convoluted and complicated things that you see in the model is to scratch a few minutes of every fit... which the moment you start producing many hundred of replicas become very relevant. This point in particular I would like to see it well tested in Linux and not mac, since we usually run using Linux.

APJansen:
On the plots, I do think it'd be most convenient to have an option in the runcard, for the plots and also the summaries (which as I understand are meant to be optional but not actually implemented), and perhaps other things. Perhaps joined into a group of development options that all default to false. 
But an example in the docs is also nice, I'll look into that, thanks for the script @RoyStegeman

scarlehoff:
> On the photon, I should stress that I haven't tested that, and I can't because it's not working for me. There's a numpy version conflict arising from fiatlux, through eko and numba. I had to, at Tanjona's suggestion, remove the eko import from photon.py to make it run without the photon

If you perform a clean installation using conda it should work (if it doesn't then it is something that we have to fix asap!). At least in linux it should...

> On other checks, the only ones I've done apart from the CI and the fitbot is to run the basic runcard, with either feature scaling or logs. I'm not sure what exactly is missing from this combination of tests?

I would suggest running all runcards in the example folder, that would cover all settings that we use I think. For the flavour basis you can use (this one)[https://github.com/NNPDF/nnpdf/blob/master/n3fit/runcards/reproduce_nnpdf40/NNPDF40_lo_as_01180.yml]. The full NNPDF40 runcard is [this one](https://github.com/NNPDF/nnpdf/blob/master/n3fit/runcards/reproduce_nnpdf40/NNPDF40_nnlo_as_01180_1000.yml) you might want to use it -maybe with fewer datasets- for benchmarking.

> On speed and memory, my assumption was that the bulk of the time and memory is taken after the pdf model

It is. But when running 100-1000 replicas even 1 minute difference (even if the rest takes 10 times more) adds up to many hours and not everyone has the same resources to run the code.
I'm a bit worried because the fitbot took a bit more than 3 hours in this PR to run n3fit while it usually stays below the 3 hours mark, but it is within the variance.

> I had an idea that I think will improve it: to pass all of the grids through the model once, rather than a separate input, integration and potentially x=1 grid

I would leave that for a future PR though!

> On the plots [...]

The documentation would still be appreciated though. If it is something that ends up being used by many people then we can also add it (but I would leave that for a separate future PR as well).

Radonirinaunimi:
> If you perform a clean installation using conda it should work (if it doesn't then it is something that we have to fix asap!). At least in linux it should...

It does indeed on linux but not on M1/2 macs. With the new photon dependencies, the documentation here (https://docs.nnpdf.science/get-started/installation.html#installation-from-source-on-m1-m2-macs) should be updated.

APJansen:
Checking the runcards in examples, with 2 replicas, locally on my Mac:

- [x] Basic                                   chi2 0.9, 0.8
- [ ] Basic_qed                           error: ID 522 not found in database.
- [x] Basic_parallel                    chi2 1.6, 1.2
- [x] Basic_hyperopt                 
- [x] Basic_feature_scaling      chi2 1.8, 2.8
- [x] developing                         chi2 1.2, 1.2
- [ ] Fit_with_theory_covmat    in progress (downloading)
- [x] DIS_diagonal                      chi2 1.3, 3.7

And these I probably shouldn't run locally
- [ ] NNPDF40_nnlo_as_01180_1000. 
- [ ] NNPDF40_lo_as_01180

scarlehoff:
>  Basic_qed: error: ID 522 not found in database.

You need to reinstall the code with cmake. The installation of the database from `nnpdfcpp/data/theory.db` to `share/NNPDF/data` is still associated to the C++ code. We should fix that (but someone has to do it...)

> And these I probably shouldn't run locally

The leading order (`NNPDF40_lo_as_01180`) one is the one that is different to what you've already run. You can just remove a bunch of datasets to check that it runs well.

I mentioned `NNPDF40_nnlo_as_01180_1000` to be used for benchmarking since this is the actual way we run the code, but the content of the runcard itself (albeit with a smaller NN and dataset) is similar to what the fitbot already does.

RoyStegeman:
For `NNPDF40_nnlo_as_01180_1000` could you please run a larger sample size such that we can see if the distribution has remained the same - usually we run 100 replicas. This could also be used to see how long it takes compared to the master branch,  though this would require a corresponding 100 replica fit using master on the same machine. This is quite straightforward since the times are stored in the `.json` files inside each replica folder.

APJansen:
> You need to reinstall the code with cmake. The installation of the database from nnpdfcpp/data/theory.db to share/NNPDF/data is still associated to the C++ code. Apologies for that, it should be fixed at some point (but someone has to do it...)

I'm afraid this will end up taking me a whole day. Perhaps someone else can run this one?

RoyStegeman:
All you'd need to do is `make -j && make install` in your build folder while having a theorydb that contains theory 522. If this branch doesn't you could even cheat and just copy it from master.

scarlehoff:
It's also enough if you copy `nnpdfcpp/data/theory.db` to `share/NNPDF/data`

APJansen:
The QED runcard gives me an error from pineapple, after applying juan's suggestion, from the line `pines = [FkTable.read(i) for i in fkspec.fkpath]`, `pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: NonUniqueGrids`.
Seems like the necessary fk tables are still not present?

scarlehoff:
Do `conda update -n $your_environment --alll`. You might just have an outdated version of `pineappl`.

APJansen:
Indeed that didn't fix it, there was no update of pineapple

andreab1997:
> Do `conda update -n $your_environment --alll`. You might just have an outdated version of `pineappl`.
> 
> Edit: on second thought... maybe there is no updated version of pineappl in conda @niclaurenti @andreab1997 @cschwan is the version of pineappl ready for optimized grids available? If not I think a release is now needed.

I believe there is no proper version yet but there are plenty (working) pre-releases. For the moment one can just install one of those

APJansen:
NNPDF40_lo_as_01180 ran fine (I limited the number of epochs rather than deleting datasets, as doing that I created inconsistent configurations)

niclaurenti:
> The QED runcard gives me an error from pineapple, after applying juan's suggestion, from the line `pines = [FkTable.read(i) for i in fkspec.fkpath]`, `pyo3_runtime.PanicException: called `Result::unwrap()`on an`Err` value: NonUniqueGrids`.
> Seems like the necessary fk tables are still not present?

Hi @APJansen , that error should appear if one tries to use the optimized fktables without the PR #1716 , that was merged. Are you sure you are running with current master?

APJansen:
@niclaurenti Well I ran it on the model_refactor branch, but it is up to date with master from a few days ago. I can try it on the latest master but I didn't change anything that should affect this.

niclaurenti:
> @niclaurenti Well I ran it on the model_refactor branch, but it is up to date with master from a few days ago. I can try it on the latest master but I didn't change anything that should affect this.

Which pineappl version are you using?

cschwan:
> Edit: on second thought... maybe there is no updated version of pineappl in conda @niclaurenti @andreab1997 @cschwan is the version of pineappl ready for optimized grids available? If not I think a release is now needed.

I've just released `v0.6.0`! For the conda package we have to wait for a bot to create the pull request, but then it should a mattero pressing the merge button.

APJansen:
@RoyStegeman Actually you can, I realise I didn't upload that one. I just need a way to load the model, is it possible in the same way you did the PDF model, all the info being in the validphys api object?

(You can see there are a few identity layers here as well. If the changes I did here are appreciated I can perhaps have another go later, in a new PR)
full_model

APJansen:
Thanks @cschwan, updated to 0.6.0 and that fixed the error (was using 0.5.9 before @niclaurenti). 
I now get an error from [this line](https://github.com/NNPDF/nnpdf/blob/5b5d058c4e1d114b03666b91570818fe41ef798c/validphys2/src/validphys/photon/compute.py#LL98C65-L98C65) in fiatlux. I'm not sure what's happening here, but it seems to me that `tmp_file` should be `tmp`, as the former is out of context already. Indeed changing that allows it to progress further (waiting for it to complete now).

scarlehoff:
> I now get an error from [this line](https://github.com/NNPDF/nnpdf/blob/5b5d058c4e1d114b03666b91570818fe41ef798c/validphys2/src/validphys/photon/compute.py#LL98C65-L98C65) in fiatlux.

@niclaurenti could you have a look and submit a quick fix, thanks! (I'm surprised that was not caught before)

APJansen:
Ok now it runs into the EKO issue, I'll try to run it on snellius instead

n3fit/src/n3fit/layers/preprocessing.py
APJansen: Oh I thought I had reverted all those changes. I turned preprocessing into preprocessing factor if it wasn't already there, just to distinguish it from things like feature scaling (only in documentation), hope that's ok.

n3fit/src/n3fit/tests/test_xops.py
```diff
@@ -0,0 +1,33 @@
...
+    x_div = xDivide()
+    test_input = np.array([1, 2, 3], dtype=np.float32).reshape((1, 3, 1))
+    test_output = x_div(test_input)
+
+    expected_output = np.ones(shape=(1, 3, 14))
+    default_indices = [3, 4, 5, 6]
+    for i in default_indices:
+        expected_output[:, :, i] = 1 / test_input[:, :, 0]
+
+    assert np.allclose(test_output, expected_output)
```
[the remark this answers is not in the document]
APJansen: Done, had to set the rtol value back to 1e-5, which is default for `allclose`. `assert_allclose` has 1e-7 as a default which was just failing.

n3fit/src/n3fit/model_gen.py
APJansen: Done except for the photons argument [here](https://github.com/NNPDF/nnpdf/blob/38a1ccbcf1cbfa41540e6ceef13c9d96acdc117f/n3fit/src/n3fit/model_gen.py#L416) , which I didn't find documented anywhere

n3fit/src/n3fit/layers/preprocessing.py
scarlehoff: Sure, preprocessing factor is fine. But the word preprocessing should remain.

Indeed, in the physics (as in, when we discuss the physics in the paper) both this factor and the feature scaling are forms of preprocessing.

n3fit/src/n3fit/model_gen.py
scarlehoff: @niclaurenti what is the object type for the photon?

RoyStegeman:
I get a `bad label format` error when trying to produce the plot for the full model including observables or running my script with this branch. Just to be clear, at which line did you intercept the model for that plot? 

While placeholder inputs are used for the inputs, this is not the case for the outputs so I think if you wanted to have a simple function to produce that figure it would require restructuring the functionality in the `ModelTrainer` class. Currently the observable layers are build while loading the data, which you'd probably not want since that requires a lot of time and memory, and the construction of the layers happens in the same function as the training of the fit so that would have to be separated. I.e. you need a lot more of `n3fit` to get that figure than just the PDF model.

Personally I would just keep it with the PDF model only, especially  since I  think the most informative part of these plots would be the PDF model anyway, though you can probably judge that better. Also note that now the figure seems quite manageable, but it will blow up for the full set of datasets.

APJansen:
@RoyStegeman It's [here](https://github.com/NNPDF/nnpdf/blob/38a1ccbcf1cbfa41540e6ceef13c9d96acdc117f/n3fit/src/n3fit/model_trainer.py#L486), the plot above is from the basic runcard. (Note you need to change [this](https://github.com/NNPDF/nnpdf/blob/38a1ccbcf1cbfa41540e6ceef13c9d96acdc117f/n3fit/src/n3fit/model_trainer.py#L180) back to true if using it directly.)

Perhaps we can leave it out in the documentation, but still consider adding it later in the code itself. I don't see any downside if it's in some utility module that gets called based on a flag in the runcard that defaults to false.

scarlehoff:
I don't want to be a killjoy, but I think we should first finish the refactoring done in this PR (i.e, making sure that everything work and most importantly ensuring that this doesn't harm performance in anyway) before getting to far into the diagrams.

As said before, I think the best thing at this stage would be just an explanation on the docs. And I wouldn't worry about the specifics of the code since it might be that after benchmarking some changes are needed and thus any such script might need to change.

Btw, beyond the many-replica fit that @RoyStegeman suggested to get a distribution, you could start by timing how long per epoch and per replica a fit take, to see whether there's a noticeable change already there. Also, let me run the fitbot again just to check whether it was an outlier.

APJansen:
Yes I agree, I'll add the script to the docs and remove the plotting inside the code at least for now. 
Having some issues also on snellius with numpy version conflicts, so don't expect any further results on timing this week.

n3fit/src/n3fit/model_gen.py
```diff
@@ -453,7 +455,7 @@ def pdfNN_layer_generator(
     A function is constructed that joins all those layers. The function takes a
     tensor as the input and applies all layers for NN in order.
 
-    4. Create a preprocessing layer (that takes as input the same tensor x as the NN)
+    4. Create a prefactor layer (that takes as input the same tensor x as the NN)
```
RoyStegeman, line 458: same

APJansen:
I incorporated Roy's comments but didn't push it yet, not sure if this will reset the fitbot.

On the timing of the fitbot, comparing to [the last time it was run](https://github.com/NNPDF/nnpdf/actions/runs/5002219066/jobs/8961893220) on the add photon branch, the current run is actually a lot faster, especially comparing the n3fit part, which here finished in 2:14 and in the add photon branch finished took 2:56. I would be very surprised if that were the actual difference, there must be a huge variance? Or is it somehow doing more there?

scarlehoff:
> I incorporated Roy's comments but didn't push it yet, not sure if this will reset the fitbot.

No, it doesn't get reset (although to see it you will need to go to either the actions tab or to the commit!)

> there must be a huge variance? 

There is, since it depends on what computer you get. It is not very reliable as a measurement, that's why a proper benchmark is needed.

n3fit/src/n3fit/model_gen.py
niclaurenti: Yes exactly. It is the object created in the `validphys.photon.compute.Photon` class

niclaurenti:
> I incorporated Roy's comments but didn't push it yet, not sure if this will reset the fitbot.
> 
> On the timing of the fitbot, comparing to [the last time it was run](https://github.com/NNPDF/nnpdf/actions/runs/5002219066/jobs/8961893220) on the add photon branch, the current run is actually a lot faster, especially comparing the n3fit part, which here finished in 2:14 and in the add photon branch finished took 2:56. I would be very surprised if that were the actual difference, there must be a huge variance? Or is it somehow doing more there?

In the photon branch it calls `fiatlux` to generate the photon PDF for every point in the xgrid. Being the grid very large (196 points) it takes a lot of time.

APJansen:
I removed the plotting from the code and added it to the documentation. Well, only the pdf one, as all 3 would take up too much space and joining them together makes it too small to read.

niclaurenti:
> Thanks @cschwan, updated to 0.6.0 and that fixed the error (was using 0.5.9 before @niclaurenti).
> I now get an error from [this line](https://github.com/NNPDF/nnpdf/blob/5b5d058c4e1d114b03666b91570818fe41ef798c/validphys2/src/validphys/photon/compute.py#LL98C65-L98C65) in fiatlux. I'm not sure what's happening here, but it seems to me that `tmp_file` should be `tmp`, as the former is out of context already. Indeed changing that allows it to progress further (waiting for it to complete now).

Hi @APJansen , the fix you suggested has been merged in master so if you rebase/merge now everything should work

doc/sphinx/source/n3fit/methodology.rst
```diff
@@ -126,6 +126,55 @@ provide a faster convergence to the solution.
...
+           'plot_nn': nn_model,
+           'plot_msr': msr_model
+           }
+
+   for name, model in models_to_plot.items():
+       plot_model(model, to_file=f"./{name}.png", show_shapes=True)
+
+
+This will produce for instance the plot of the PDF model below, and can also be used to plot the
+neural network model, the momentum sum rule model, and the model containing everything that is illustrated above.
```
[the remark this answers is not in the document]
RoyStegeman: I see, that wasn't really clear from the text to me. Yes, personally I would just not mention it.

Radonirinaunimi:
> I'd suggest moving to libmamba as the solver https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community as it seems to be much faster.

@goord did try this but I am not sure how it went as I had to leave.

APJansen:
Amazingly `mamba install --only-deps nnpdf` solved the environment in 2 minutes. We thought however that this would include eko and fiatlux but it doesn't

APJansen:
I ran a first test of the NNPDF40_nnlo_as_01180_1000 runcard on Snellius, on the CPU. It takes 6 hours to load all of the data.. I had it run for 30 hours on the master branch, with 100 replicas, and it got 38 of them done in the 24 hours remaining after data loading, so 1.5 replica per hour. The run in this refactored branch is still going, but so far averaging about 1.55 per hour.

scarlehoff:
> It takes 6 hours to load all of the data...

That's quite weird. How are you running it?

APJansen:
I managed to get everything installed and collected it into the script below (`anaconda` is a script that loads a module on snellius). The actual fit is called with `n3fit NNPDF40_nnlo_as_01180_1000.yml 1 -r 100`. I do get a tensorflow error `Could not load dynamic library 'libcudart.so.11.0`, but that doesn't seem related.

```
ENV=$1
source ~/.bashrc
anaconda
conda create -n "$ENV" python=3.9 --yes
conda activate "$ENV"
mamba install -vv --yes --only-deps nnpdf
mamba install --yes swig pkg-config make
cd ~/nnpdfgit/nnpdf
mkdir "build_$ENV"
cd "build_$ENV"
cmake .. -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX
make
make install
cd ~/nnpdfgit/testBenchmark/
./submit.sh "$ENV"
```

APJansen:
Here's my conda list, python version is 3.9.16.
[condalist.txt](https://github.com/NNPDF/nnpdf/files/11720322/condalist.txt)

scarlehoff:
> The actual fit is called with n3fit NNPDF40_nnlo_as_01180_1000.yml 1 -r 100

In order to benchmark I think you should run the replicas one-by-one, i.e., 

```
n3fit NNPDF40_nnlo_as_01180_1000.yml ${job_idx}
```

where `job_idx` goes from 0 to 1 with each replica being run separately. This gives you a better distribution of the timing (otherwise your benchmark will depend a lot on the specific node each of the two fit runs, by distributing the replicas you can average some random effects), it's also more gentle in the memory.

You can also read the time it takes to run the fit directly from the `NNPDF40_nnlo_as_01180_1000/nnfit/replica_N/NNPDF40_nnlo_as_01180_1000.json` files.

> Could not load dynamic library 'libcudart.so.11.0, but that doesn't seem related.

Seems to be cuda related, unless you are running in a GPU that doesn't really matter. You can use `export CUDA_VISIBLE_DEVICES=""` to ensure the GPU are not seen.

APJansen:
But wouldn't that way of running load the data anew every time? One replica would take ~7 hours, 6 of which are data loading..

scarlehoff:
> But wouldn't that way of running load the data anew every time?

Yes.

> One replica would take ~7 hours, 6 of which are data loading..

It should not. I believe it took 6 hours because while generating 100 replicas you probably ended up in swap and at some point it got super slow. Loading the data (depends on your disk) but should not take that long.

APJansen:
Do you always run it like this, one replica at a time, or just for the timing? Just wondering because of the data loading issue. If you do run it with 100 or 1000 replicas in a single command, any estimate on how long the loading takes for you?

On the timings, the nodes are all the same, so that isn't an issue. The only possible difference is that because I didn't reserve a whole node, performance might be affected differently on both runs, but that seems unlikely. Snellius is down today and tomorrow, so I can't do anything atm.

scarlehoff:
> Do you always run it like this, one replica at a time, or just for the timing? Just wondering because of the data loading issue. If you do run it with 100 or 1000 replicas in a single command, any estimate on how long the loading takes for you?

It greatly depends on the computers that you have. But it rarely takes more than 1 hour to fit one replica of which the data loading can be 10 or 20 minutes. In my laptop it takes about 4 minutes but I have an ssd.

> the nodes are all the same, so that isn't an issue. The only possible difference is that because I didn't reserve a whole node, performance might be affected differently on both runs, but that seems unlikely.

Depends on how is it configure, but a fit is heavy in memory so not having the whole node reserver can affect the run. In any case, running many will average those effects out as well. I've never reserved whole nodes for a fit, not worth it. In the cluster in Milan I usually run them with 4 or 8 cores btw.

APJansen:
The CI is now failing, after merging master back in, because the qed test was added.
I can see where it's coming from, it's because I've set `ph_replica=None` [here](https://github.com/NNPDF/nnpdf/blob/2b8c806e731725ed3140ce91a591588f67f56d1c/n3fit/src/n3fit/msr.py#L97), but I'm having a hard time fixing it in a clean way without understanding what's going on with the photon.

From what I understand, rather than including the photon on an equal level with the other partons, it's added after the fact, and rather than integrating it along with the others, integrals are provided separately, with all replicas concatenated together into one tensor of integrals. Then `ph_replica` is just the replica number in general, and this is needed in the normalization layer to pick out the right component of this all replica tensor that the layer keeps as an attribute.
Is that correct? And can I find anywhere documented why it's done in this way?

To stay closest to both this new photon functionality and my refactor, what I think I'll try is to avoid making the sumrule model dependent on the replica (and thus avoid needing a separate copy of this whole construction per replica), but rather add one extra argument to it, namely the replica number, and pass `i_replica` to it  that to it in the loop [here](https://github.com/NNPDF/nnpdf/blob/2b8c806e731725ed3140ce91a591588f67f56d1c/n3fit/src/n3fit/model_gen.py#L679).
Does that sound ok?

niclaurenti:
> The CI is now failing, after merging master back in, because the qed test was added. I can see where it's coming from, it's because I've set `ph_replica=None` [here](https://github.com/NNPDF/nnpdf/blob/2b8c806e731725ed3140ce91a591588f67f56d1c/n3fit/src/n3fit/msr.py#L97), but I'm having a hard time fixing it in a clean way without understanding what's going on with the photon.
> 
> From what I understand, rather than including the photon on an equal level with the other partons, it's added after the fact, and rather than integrating it along with the others, integrals are provided separately, with all replicas concatenated together into one tensor of integrals. Then `ph_replica` is just the replica number in general, and this is needed in the normalization layer to pick out the right component of this all replica tensor that the layer keeps as an attribute. Is that correct? And can I find anywhere documented why it's done in this way?
> 
> To stay closest to both this new photon functionality and my refactor, what I think I'll try is to avoid making the sumrule model dependent on the replica (and thus avoid needing a separate copy of this whole construction per replica), but rather add one extra argument to it, namely the replica number, and pass `i_replica` to it that to it in the loop [here](https://github.com/NNPDF/nnpdf/blob/2b8c806e731725ed3140ce91a591588f67f56d1c/n3fit/src/n3fit/model_gen.py#L679). Does that sound ok?

Hi @APJansen , the photon is computed before the fit starting from a PDF set (that can contain or not a photon) that I called `luxset`. 
Then `ph_photon` is needed since for every replica to be fitted a different photon will be computed (i-th replica will compute a photon starting from the i-th replica of the `luxset`. Since the `validphys.Photon` object will have a list containing all the photons (plural if you are using parallel mode, if not it will be only one). In practice, `ph_replica` will be equal to the index `i` in

https://github.com/NNPDF/nnpdf/blob/7eb0c4eb67604a7d0b80f9a8ac075aab994904cc/n3fit/src/n3fit/model_gen.py#L593

So, if you are not using parallel mode it will be always 0

scarlehoff:
> (and thus avoid needing a separate copy of this whole construction per replica), but rather add one extra argument to it, namely the replica number, and pass i_replica to it that to it in the loop [here](https://github.com/NNPDF/nnpdf/blob/2b8c806e731725ed3140ce91a591588f67f56d1c/n3fit/src/n3fit/model_gen.py#L679).
Does that sound ok?

Personally I find it more confusing that simply having a different sum rule per replica (which is truly what is happening). Having a separate copy of the sum rule should not be much of an overhead anyway, shouldn't it?

RoyStegeman:
For the sumrules I don't really have a preference. Other than that, is there anything else you plan to do in this PR?

APJansen:
No I've been stuck waiting for the tests for weeks. The regression test with this one runcard that was still missing is actually done, I just can't get the vp-comparefits to work, talking to Tanjona tomorrow. Anything else I'll do in a separate PR.

That's also why I would still want to fix the issue with the photon as I mentioned above, it stays closest to what there is now. I do think that the more differently the photon is treated the more confusing it is, I assume there are good reasons why it can't be treated exactly the same. To make it more similar I would remove the photon integral as an attribute of the MSR model, and just pass it as an input. But I can do that separately.

APJansen:
I have the comparison between master and this branch here: https://vp.nnpdf.science/vEkkEc1jQIiHACiGX5ilTA==/#pdf-plots
The results are virtually identical, and the timings as well:
Timings for master:
Wall time, mean: 3011.522996826172, std: 468.57613438655545
CPU time, mean: 9045.533652957289, std: 1523.929561609167
Timings for model_refactor:
Wall time, mean: 3047.954376196861, std: 409.96760923276776
CPU time, mean: 9163.189698291579, std: 1351.249681278691

So I think apart from the new test which I still have to fix it's ready to be merged.

RoyStegeman:
Thanks for doing this check @APJansen. Could you repeat it for feature scaling and a flavor basis fit as well?

APJansen:
@RoyStegeman What exactly? Which runcards and how many replicas? I better fix this test first and then I'll do them. I'm writing a script that just takes a runcard and two branches and does the comparison, so hopefully it will go smoother in the future.

APJansen:
Yes they ran independently, and the times above are the "Total" times, I just checked that the "replica_fitted" times are identical, up to 2 or 3 decimals.

RoyStegeman:
Could you try these two runcards?

[feature_scaling.txt](https://github.com/NNPDF/nnpdf/files/11834936/feature_scaling.txt)
[flavor_basis.txt](https://github.com/NNPDF/nnpdf/files/11834938/flavor_basis.txt)

APJansen:
@RoyStegeman Sure, you mean a full 100 replica fit for both right?

RoyStegeman:
Indeed, and if you could then compare them in a vp-report to `NNPDF40_nnlo_as_01180_1000_refactor` that would be great.

APJansen:
Why to `NNPDF40_nnlo_as_01180_1000_refactor` rather than to the same runcard run with the master branch?

RoyStegeman:
If resources are not a problem that would indeed be even better.

APJansen:
Ok I've fixed the issue by passing the photon integrals as inputs to the MSR_Normalization layer. I think that decouples it more, makes it clearer that this is coming from somewhere external, and treats it a bit more closely to the other partons.

@RoyStegeman resources may become a problem, I'll submit the jobs for those 2 runcard on the refactor branch first. Next week I'm on vacation, I hope they're done when I'm back and I'll first do the comparison to  NNPDF40_nnlo_as_01180_1000_refactor (wait times are still very long on Snellius..).

n3fit/src/n3fit/model_gen.py
```diff
@@ -9,9 +9,13 @@
 
 
 """
+from typing import List, Callable
 from dataclasses import dataclass
 
 import numpy as np
+from n3fit.msr import generate_msr_model_and_grid
+from n3fit.layers import DIS, DY, ObsRotation, losses
+from n3fit.layers.observable import is_unique
```
RoyStegeman, line 18:
```suggestion
```

n3fit/src/n3fit/model_gen.py
```diff
@@ -544,135 +543,206 @@ def pdfNN_layer_generator(
...
     else:
         sumrule_layer = lambda x: x
+    # The photon is treated separately, need to get its integrals to normalize the pdf
+    if photons:
+        photon_integrals = photons.integral
+    else:
+        photon_integrals = [0.0 for _ in range(len(seed))]
+    import tensorflow as tf
+    photon_integrals = [tf.constant(photon_integrals[i_replica])
+                            for i_replica in range(len(seed))]
```
RoyStegeman, line 616: Please move this down to where it's actually passed to the `sumrules_layer` to keep relevant bits of code together. 

Though I don't really have a preference regarding taking the specific replica inside the `Photon` class or not (this is something you can discuss with @scarlehoff), I don't really like this solution and would rather see the if-statement inside the loop where it's passed to the `sumrules_layer`.

n3fit/src/n3fit/model_gen.py
RoyStegeman, line 1: Is the following correct? `seed` is a list of `int`s so this does nothing. 

```
elif isinstance(seed, int):
    seed = parallel_models * [seed]
```

I think it should be 
```
elif isinstance(seed, list):
    # you can check if all elements in the list are int if you like
    seed = parallel_models * seed
```

@scarlehoff ?

n3fit/src/n3fit/model_trainer.py
```diff
@@ -472,6 +472,12 @@ def _model_generation(self, xinput, pdf_models, partition, partition_idx):
 
         if self.print_summary:
             training.summary()
+            pdf_model = training.get_layer("PDF_0")
+            pdf_model.summary()
+            nn_model = pdf_model.get_layer("NN_0")
+            nn_model.summary()
```
RoyStegeman, line 478: If you'd loop over multiple seeds (in the other comment) the name here would be replica-dependent.

RoyStegeman (commented):
This looks mostly okay already. 

I don't really understand the multi-replica case yet which, I believe, will become important soon, but that's not solely related to this PR.

n3fit/src/n3fit/backends/keras_backend/operations.py
```diff
@@ -134,6 +141,8 @@ def numpy_to_input(numpy_array, no_reshape=False, name=None):
         batched_array = np.expand_dims(numpy_array, 0)
         batch_size = 1
         shape = numpy_array.shape
+    if custom_shape is not None:
+        shape = custom_shape
```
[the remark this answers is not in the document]
APJansen: The reason was that without this, several layers get passed multiple shapes, e.g. (1, None, 1) for the grid, (1, 2000, 1) for the integration grid and (1, 1, 1) for the x=1 input. Because of this, in the model.summary the shapes are not displayed, but rather something like [?]. I apply it to the latter two inputs to make them consistent. It was explained at one of the two points it was used, but not very clearly, so I'll add it to the docstring.

On how to clean up the `numpy_to_input` function, I see no harm in doing this always, replacing whatever the number of gridpoints is with None. 
Also, the other boolean argument `no_reshape` is not used anywhere as far as I can tell (e.g. barring any reportengine magic). This means that everything passed to it is without batch dimension, so we can get rid of both the `custom_shape` and the `no_reshape` arguments and the if statement entirely, and set the shape to that of the input with the 0 index replaced by None. Do you agree? (I'll have to test ofcourse, still/again don't have a working local installation so it's cumbersome..)

n3fit/src/n3fit/backends/keras_backend/operations.py
```diff
@@ -125,6 +130,8 @@ def numpy_to_input(numpy_array, no_reshape=False, name=None):
             if true, don't add batch dimension, take the first dimension of the array as the batch
         name: bool
             name to give to the layer
+        custom_shape: tuple
+            To specify a more general shape with None values
```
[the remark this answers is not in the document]
APJansen: See other comment on these lines.

n3fit/src/n3fit/layers/x_operations.py
```diff
@@ -1,14 +1,15 @@
 """
     This module contains layers acting on the x-grid input of the NN
 
-    The three operations included are:
+    The two operations included are:
         - ``xDivide``
```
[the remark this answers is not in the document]
APJansen: I would keep the classes in this module, but join it with `layers/msr_normalization.py`, and perhaps also with `msr.py`. Currently there's these 3 modules that all deal with the same thing. But agreed not in this PR.

n3fit/src/n3fit/model_gen.py
APJansen: I think the intent here is to use the same seed for all replicas if only one seed is given. This will be changed in #1661 so it's probably best discussed there.

n3fit/src/n3fit/model_gen.py
APJansen: Oh first of all I didn't mean to leave the import statement there ofcourse, sorry about that, I can just use the existing `op.numpy_to_tensor`.
My aim with putting it there was to first define everything and then keep the final loop as clean as possible. But no matter what you'll have some somewhat confusing code somewhere due to the photon. I applied your suggestion, I agree it's a bit cleaner.

RoyStegeman:
Thanks! That seems reasonable. 

> Not sure what you want to see from this? 

I just want to see if the result is as expected. This PR changes the model construction quite a bit and because these scenarios I asked for are not tested in the CI I wanted to see the results, otherwise we'd be merging this based only on the hopes we didn't miss any typos.

APJansen:
@RoyStegeman Ok great, having some trouble with the other one, the fits didn't run because NNPDF40_nnlo_as_0118_flav cannot be found or downloaded. This is given in the flavor_basis runcard you posted above as the value for t0pdfset under datacuts. Is there a typo maybe?

RoyStegeman:
Looks like the server is currently offline. The t0pdf isn't too important so feel free to change it to a regular nnpdf40 (or equivalent) fit.

scarlehoff:
I can access the server. Can't you?

RoyStegeman:
> I can access the server. Can't you?

Now I can but somehow not a few minutes ago.

It seems that, that PDF indeed does not exist on the server so you could try 210701-n3fit-meth-013 instead (which is actually the fit originally produced with that runcard).

APJansen:
Ok that works, running now

n3fit/src/n3fit/backends/keras_backend/operations.py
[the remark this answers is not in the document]
APJansen: This is implemented in [this commit](https://github.com/NNPDF/nnpdf/pull/1734/commits/80ef5d72a5bf1c2476ba1f0bd1374b2d7239c700)

APJansen:
@RoyStegeman Are you satisfied with how I addressed your comments? Still waiting on the results of the flavor runcard.
Meanwhile I'm trying to merge master again, I ran the basic runcard as a quick check before pushing, but now I get `Error Processing filter rules: Could not find dataset LHCBWMU8TEV`. Tried vp-setupfit, but same thing.

RoyStegeman:
Thanks, I'll have a look at the changes today. 

The error is due to a recent PR #1760 and you should be able to solve it by simply re-running `make && make install`.

n3fit/src/n3fit/backends/keras_backend/operations.py
```diff
@@ -126,26 +124,17 @@ def numpy_to_input(
...
-        batch_size = numpy_array.shape[0]
-        shape = numpy_array.shape[1:]
-    else:
-        batched_array = np.expand_dims(numpy_array, 0)
-        batch_size = 1
-        shape = numpy_array.shape
-    if custom_shape is not None:
-        shape = custom_shape
+    batched_array = np.expand_dims(numpy_array, 0)
+    batch_size = 1
```
RoyStegeman, line 131: Since `batch_size` doesn't depend on any condition you may as well skip defining it and just pass `batch_size=1` to Input`. 

P.S. the docstring of this function can be changed from "By default it adds..." to "It adds ..."

n3fit/src/n3fit/backends/keras_backend/operations.py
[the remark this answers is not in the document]
RoyStegeman: Ah, I hadn't realised it affects the model.summary, but I agree with everything you say here. There is no reportengine magic happening in this part of the code since reportengine isn't even aware of the existence of this function. I do understand that what reportengine does may seem a bit like black magic but it works a lot like [pytest fixtures ](https://docs.pytest.org/en/6.2.x/fixture.html), if you're familiar with that.

n3fit/src/n3fit/layers/x_operations.py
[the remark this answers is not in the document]
RoyStegeman: I thought I ended up deleting this comment xd, for now the organisation of the functions in modules also doesn't' really bother me.

APJansen:
Here's the comparison between NNPDF40_nnlo_as_01180_1000_refactor and flavor_basis: https://vp.nnpdf.science/Nf81M-MBQUKfHPbrhwvfQA==/

scarlehoff:
I'll have a final look at this between today and tomorrow (before the CM on wednesday for sure) since I think it is ready to be merged.

Just one final request, could you run `black` and `isort` on the files you modified? I'm not sure whether the `pyproject.toml` file was already there when you started the branch but if not here are the options we use: https://github.com/NNPDF/nnpdf/blob/master/pyproject.toml 
Thanks!

scarlehoff (commented):
lgtm, just a few small comments

One question, did you by any chance test the "one flavour per network" fit when doing the changes? (nobody has used that for ages so I'm wondering whether it even works...)

RE the comparisons that have been done, some comments:

- [x] standard fit (master vs this branch): https://vp.nnpdf.science/vEkkEc1jQIiHACiGX5ilTA==/ 

perfect agreement, nothing to say

- feature scaling
https://vp.nnpdf.science/Z5lMkJABTDOvsZ51cAGy4g==/
Would it be possible to do the comparison with the _same_ fit done in master and this branch?

- flavour basis
 https://vp.nnpdf.science/Nf81M-MBQUKfHPbrhwvfQA==/

Same, I would like to see the comparison against the same fit (i.e., same runcard) but in this branch (which you already have) and master.

The hope is that the agreement is perfect as in the standard fit! (flavour basis is particularly important because it gets routinely used, so I'd like to check that nothing has really changed)

-------

other than that this LGTM. As soon as we have those two reports linked in the PR we could merge :D Thank you very much for your work!

n3fit/src/n3fit/backends/keras_backend/operations.py
```diff
@@ -107,36 +106,31 @@ def numpy_to_tensor(ival, **kwargs):
...
-        shape = numpy_array.shape[1:]
-    else:
-        batched_array = np.expand_dims(numpy_array, 0)
-        batch_size = 1
-        shape = numpy_array.shape
-    input_layer = Input(batch_size=batch_size, shape=shape, name=name)
+    batched_array = np.expand_dims(numpy_array, 0)
+    shape = list(numpy_array.shape)
+    # set the number of gridpoints to None, otherwise shapes don't show in model.summary
+    shape[0] = None
```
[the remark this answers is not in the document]
APJansen: I think it's because the regular grid has shape None in this axis while the integration grid has shape 1_000, and the x=1 grid shape 1. It doesn't merge them together as None

n3fit/src/n3fit/scaler.py
```diff
@@ -0,0 +1,63 @@
...
+        scaler = PchipInterpolator(map_from, map_to)
+    except ValueError:
+        raise ValueError("interpolation_points is larger than the number of unique input x-values")
+
+    def _scaler(x):
+        x_scaled = scaler(np.log(x))
+        x_scaled = 2 * x_scaled - 1
+        return np.concatenate([x_scaled, x], axis=-1)
+
+    return _scaler
```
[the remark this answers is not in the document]
APJansen: Thanks for spotting, no idea why I removed them.

n3fit/src/n3fit/backends/keras_backend/operations.py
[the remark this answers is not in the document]
APJansen: Ok I'll run those checks.

I never tested the one flavour per network, actually I was wondering if that was still being used, if not I wanted to merge the `generate_dense_per_flavour_network` and `generate_nn` functions. If you're already wondering if it still works it might make sense to remove?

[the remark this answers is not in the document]
scarlehoff: No, if it doesn't work it should actually be fixed! While it is not something we actively use, it used to be the way in which the PDF fit was done so it's good to have it if someone would like to reproduce the old ways.
(if you didn't try it don't worry though, I was just wondering)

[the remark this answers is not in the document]
APJansen: But if it hasn't been used in a long time doesn't it make sense to refer people to an older release? In this case it doesn't matter much, but I can imagine the need to always be backward compatible with everything will force you to make suboptimal choices for new additions, and overcomplicate the code. Not an issue here but just wondering what the policy is.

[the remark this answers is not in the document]
scarlehoff: No, I mean if people want to use the newer features (data, theory, whatever) with the older strategy.

And, in any case, we want the architecture to be quite flexible since it is the type of thing that make sense to change for a side project (for example, there's currently people trying with functional forms, in the past we have fitted using a quantum computer). I guess is the kind of thing that might look inelegant but it has been useful in the past so I'd like to keep it. Btw, I just tried it out and seems to work fine at least with the basic_runcard, so the refactor didn't break that :)

Policy-wise the idea is to keep everything compatible one version below, so, to first approximation, nothing that was used for the 4.0 paper should disappear until 5.0 (this is also one of the reasons we are keeping all the C++ around for the time being...)

n3fit/src/n3fit/backends/keras_backend/operations.py
```diff
@@ -107,36 +106,29 @@ def numpy_to_tensor(ival, **kwargs):
...
 # f(x: tensor) -> y: tensor
 def batchit(x, batch_dimension=0, **kwarg):
-    """ Add a batch dimension to tensor x """
+    """Add a batch dimension to tensor x"""
     return tf.expand_dims(x, batch_dimension, **kwarg)
 
 
 # layer generation
-def numpy_to_input(numpy_array, no_reshape=False, name=None):
+def numpy_to_input(numpy_array: np.ndarray, name: str = None):
```
Radonirinaunimi, line 114:
```suggestion
def numpy_to_input(numpy_array: npt.NDArray, name: str | None = None):
```
If type checking is to be adopted, why not use the [numpy.typing](https://numpy.org/devdocs/reference/typing.html)? A pickier point concerns the assignment of  `None` to `str`.

n3fit/src/n3fit/layers/x_operations.py
```diff
@@ -1,54 +1,55 @@
...
     -----------
         output_dim: int
             dimension of the pdf
         div_list: list
-            list of indices to be divided by X (by default [3,4,5]; [v, v3, v8]
+            list of indices to be divided by X (by default [3, 4, 5, 6]; [v, v3, v8, v15]
     """
 
-    def __init__(self, output_dim=BASIS_SIZE, div_list=None, **kwargs):
+    def __init__(self, output_dim: int = BASIS_SIZE, div_list: List = None, **kwargs):
```
Radonirinaunimi, line 37:
```suggestion
    def __init__(self, output_dim: int = BASIS_SIZE, div_list: List | None = None, **kwargs):
```
Same as above and everywhere else below! A very minor point and hence could be discarded. If you want, however, you could also use `typing.Union`.

n3fit/src/n3fit/model_gen.py
```diff
@@ -544,135 +540,202 @@ def pdfNN_layer_generator(
...
+
+        return pdf_unnormalized
+
+    # Finally compute the normalized PDFs for each replica
+    pdf_models = []
+    for i_replica, (preprocessing_factor, nn) in enumerate(
+        zip(preprocessing_factor_replicas, nn_replicas)
+    ):
+        pdf_unnormalized = compute_unnormalized_pdf(pdf_input, nn, preprocessing_factor)
+        pdf_integration_grid = compute_unnormalized_pdf(integrator_input, nn, preprocessing_factor)
```
Radonirinaunimi, line 672: This was not an issue before as the `inegrator_input` was not passed explicitly, but now it appears that it is unbound.

n3fit/src/n3fit/scaler.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Parameters
+    ----------
+    input_list : list of numpy.ndarray
+        The list of input data arrays.
+    interpolation_points : int, optional
+
+    Returns
+    -------
+    _scaler : function
```
Radonirinaunimi, line 19:
```suggestion
    interpolation_points : int, optional
           Number of interpolation points

    Returns
    -------
    _scaler : typing.Callable
```

n3fit/src/n3fit/scaler.py
```diff
@@ -0,0 +1,72 @@
...
+    if selected_points[0] != 0:
+        selected_points = [0] + selected_points
+    map_from = map_from_complete[selected_points]
+    map_from = np.log(map_from)
+    map_to = map_to_complete[selected_points]
+
+    try:
+        scaler = PchipInterpolator(map_from, map_to)
+    except ValueError:
+        raise ValueError("interpolation_points is larger than the number of unique input x-values")
```
Radonirinaunimi, line 65: This was already present previously, but is this actually a valid exception? At least the error message should be passed to print...

Radonirinaunimi (commented):
LGTM! I think that these changes are really relevant in that they make the codes cleaner, more explicit and transparent.

I just left various minor comments here and there (most of which are styling).

n3fit/src/n3fit/model_gen.py
APJansen: What do you mean exactly? It is undefined when `impose_sumrule` is false, but I don't think that ever happens?

n3fit/src/n3fit/scaler.py
scarlehoff, line 65:
```suggestion
    except ValueError as e
        raise ValueError("interpolation_points is larger than the number of unique input x-values") from e
```

This would be better. Empirically, the reason why that error happens is that, but probably a good idea to raise from the original error just in case.

n3fit/src/n3fit/model_gen.py
scarlehoff: Oh, good catch!
Yes, fits can be run without imposing any sumrule at all. By putting `sum_rules: False` in the runcard.

One example: https://vp.nnpdf.science/qXweGQ_NSHux7muMmUESCw==/

n3fit/src/n3fit/backends/keras_backend/operations.py
APJansen: Thanks I didn't know about `|`. Actually I see this only works from python 3.10, and another way to express this is `Optional[str]`, so I'll replace everything with that.
I can use numpy.typing as well, didn't know about that.
Maybe we should have a policy on type hints, so that we give them consistently in any new or modified code (if we want them at all, I think they're useful)?

n3fit/src/n3fit/model_gen.py
APJansen: Ok, I've added an if statement, made it a bit more verbose than necessary just for clarity.

APJansen:
The final 2 comparisons of this branch vs master on the feature scaling and flavor basis runcards are done:
[feature scaling](https://vp.nnpdf.science/zE33cch8RvSWHG269WnbjQ==/)
[flavor basis](https://vp.nnpdf.science/Blu9d0uBQ8qxbgU-YBxvNw==/)

RoyStegeman:
Thanks @APJansen. While the feature scaling seems fine, the flavor basis looks a bit odd (better agreement than expected based on random sampling). Are you sure you are comparing the two different branches? Under "Code versions" in the report the versions are shown to be "inconsistent" so perhaps something went wrong?

APJansen:
Hm... I'm not sure... have you seen this before? I did have some trouble running vp-comparefit, eventually fixed by remaning and reuploading the master branch run.

RoyStegeman:
"inconsistent" in this caes just means that not all replicas were produced with the same commit. 

The large agreement between two fits I'm basing on the distance plots. They are the distance between two central values normalised wrt the uncertainty in std of the central values, for more details see appendix B of https://arxiv.org/pdf/1410.8849.pdf. Of course, if two fits are done independently (i.e. different seeds, though in this case I assume even the different model should have effectively the same impact) these distances should be of order 1. While this is the case for feature scaling, for the flavor basis fit the agreement looks better than expected if the two fits consisted of independent samples.

APJansen:
Hmm, I see it's also inconsistent in the previous report on the flavor basis where I compared it to something else. Perhaps I was developing while the jobs were in the queue. I guess I'll have to rerun it.

scarlehoff:
If you are rerunning, make sure you use the latest commit from this branch. I think this is basically done now so it would be great to link here the reports with what would be basically the last commit before merging.

APJansen:
Here's the new report: https://vp.nnpdf.science/1pUg_vIYTkGmuHVMVSjQkw==/
The inconsistency of the code version is gone, but the results are still very similar (identical even?)
I have no idea how, I'm using the same scripts as before, printing the environment and the last commit, which are the last commits of either branch and the corresponding conda environments.

scarlehoff:
Perfect agreement also for the flavour basis :D 

Great! If @RoyStegeman is also happy with the tests I think this is basically done?

RoyStegeman:
> If @RoyStegeman is also happy with the tests

I'm not really... Somehow flavor basis seems to be exactly the same reply-by-replica but not feature scaling, which I don't understand. Either I'd expect the model to be exactly the same given the same seed before and after the changes in this PR, or I'd expect them to differ (which I would've thought more likely). Unless I'm missing something, these results hint at an inconsistency.

APJansen:
I did a quick sanity check locally, running both the flavor_basis and feature_scaling runcards that @RoyStegeman posted here locally on both branches and just comparing the validation chi2 after 100 epochs. Indeed they are different for the feature_scaling runcard, already in the first digit, whereas for flavor_basis they are the same up to the 5th decimal.

I don't understand why, I tried copying either the model architecture and seeds or the basis from flavor_basis into feature_scaling and running that on both branches, but they're still not identical. Conversely starting with the flavor_basis runcard and copying both the model architecture, seeds and basis from feature_scaling, the results still differ.

I'll try some more comparisons

APJansen:
I'm still confused. Doing both of 1. turning off feature scaling and 2. changing the basis to the flavor basis makes the difference between the two branches very small (~5 decimals in validation chi2 after 100 epochs), but either separately doesn't.
But the scaled grids (i.e. the output of [_xgrid_generation](https://github.com/NNPDF/nnpdf/blob/43951795a0d06e6e8a67a3fbc204f9559ec906ab/n3fit/src/n3fit/model_trainer.py#L311)) are identical between branches.
Also, rather than changing the basis to the flavorbasis just setting the smallx limits to [1, 1] as they are in that runcard is not sufficient (it gets to two decimals).

scarlehoff:
I'm very confused. At some point you had a fit which is this one:

 standard fit (master vs this branch): https://vp.nnpdf.science/vEkkEc1jQIiHACiGX5ilTA==/

in the evolution basis, and that's exactly the same.

So in principle, doing 1 but not 2 should make the difference also very small.

The three comparisons that we want are:

- standard fit: (which unless I'm mistaken is this one https://vp.nnpdf.science/vEkkEc1jQIiHACiGX5ilTA==/)
this is with evolution basis and with the (x, logx) input

- feature scaling: this can be in the evolution basis, no problem with that

- flavour basis (this can be without feature scaling)
-----

What are the comparisons we actually have?

APJansen:
Right so we have:
| name | report | runcard | report diff | 100 epoch diff |
| -- | -- | -- | -- | -- |
| feature scaling | [report](https://vp.nnpdf.science/zE33cch8RvSWHG269WnbjQ==/) | [feature_scaling.txt](https://github.com/NNPDF/nnpdf/files/11834936/feature_scaling.txt) | small | large |
| flavor basis | [report](https://vp.nnpdf.science/1pUg_vIYTkGmuHVMVSjQkw==/)  | [flavor_basis.txt](https://github.com/NNPDF/nnpdf/files/11834938/flavor_basis.txt) | very small | very small |
| standard | [report](https://vp.nnpdf.science/vEkkEc1jQIiHACiGX5ilTA==/) | NNPDF40_nnlo_as_01180_1000.yml | very small | large |

So I'm also confused, I just checked the standard runcard, both with the latest commits and the commits in that report, and the difference in the 100 epoch validation chi2 is not small in either case.

scarlehoff:
Well, rather than small. The reports make it look like the results are _exactly equal_.

I don't understand how you can get the same results if the differences are large after 100 epochs.

APJansen:
Just did the same check, the standard runcard after 100 epochs, but on snellius rather than on my mac, and now the results are identical.

RoyStegeman:
And what if you do this check on snellius with the feature scaling runcard?

APJansen:
Yes that was it, feature scaling has a slight difference, but just turning that off leaving everything else the same they become identical.

RoyStegeman:
Do you know in which part of the code this difference originates? I thought everything would be deterministic, but maybe I'm forgetting something.

APJansen:
I think I found the reason: I have changed the scaler to include the scaling from [0, 1] to [-1, 1], originally this was done with a lambda layer inside the model. This causes a difference of ~10^-8 in the scaled x using this runcard. That difference is amplified I assume by training. So this explains why it only happens with feature scaling, and I think it's harmless.

RoyStegeman:
That indeed sounds like it's harmless. Do you understand what causes the difference of ~10^-8?  I'd expect it wouldn't matter if we shift by `2*x-1` outside the model or inside a model layer.

scarlehoff:
> ~10^-8 in the scaled x using this runcard.

Do you mean  a difference of `10^-8` (which is basically entering the numerical precision zone) or a difference around `x=1e-8`?

APJansen:
To be precise, a maximum absolute difference of `6e-8`, mean absolute difference of `2e-8` and mean (signed) difference of `1e-9`. 
In one case it's done inside a keras layer, and in the other by numpy. I would guess that they just have slightly different numerical errors.

scarlehoff:
Then that's perfectly fine.

RoyStegeman:
Great, well done for finding the difference

RoyStegeman (approved):
Having checked that the results of this branch and master are numerically the same, I'd say this can be merged. I believe `black` and `isort` have both been run on the changed files?

APJansen:
Great, thanks! Indeed I've run black and isort.

Issue #1735, opened 2023-05-17
Create a test theory for QED and reorganize tests


scarlehoff:
As mentioned in these two comments

https://github.com/NNPDF/nnpdf/pull/1643/files#r1146321927
https://github.com/NNPDF/nnpdf/pull/1643/files#r1155696692

the tests for the photon should be reorganized and modified so that we are testing something a bit closer to the actual usage. Similar to what is done for other things (like the new cfactors).

This is related #1729 since also Theory 399 needs to be updated to have a small eko to test the info file. I'll coordinate with @niclaurenti to create the new theories and update the tests.

niclaurenti:
@scarlehoff I didn't understand how the `eko.tar` of the low precision theory should be done. Moreover, do we want to produce a low precision `eko_photon.tar` as well?

scarlehoff:
>  do we want to produce a low precision eko_photon.tar as well 

Yes.

> I didn't understand how the eko.tar of the low precision theory should be done

My idea with this is to create an eko with a very small number of points (in Q but also x). The final PDF doesn't need to be sensible, it just needs to look like a PDF from the point of view of the code.

PR #1738, opened 2023-05-22
QED tests

niclaurenti:
PR made to address issue #1735 , i.e. to create a low precision QED theory (similar to theory 399) that will be used to test the code and in particualr the photon module of `validphys`.


scarlehoff:
For the small eko for Theory 399 I've used `evolven3fit_new --q-points 10 produce_eko -i 0.00001 --x-grid-points 10 399 eko.tar`

This produces an eko that weights only ~70 MB which I think it is ok for the tests. I'm going to open a PR with tests for `evolven3fit_new`, you can rebase on top of that branch and use the same test.

RoyStegeman:
@niclaurenti going forward, could you try rebasing a bit more often to avoid the merge commit?

n3fit/src/n3fit/tests/test_evolven3fit.py
```diff
@@ -169,3 +169,29 @@ def test_perform_evolution(tmp_path, fitname):
     info = check_lhapdf_info(tmp_info)
     for datpath in tmp_nnfit.glob("replica_*/*.dat"):
         check_lhapdf_dat(datpath, info)
+
+
+@pytest.mark.parametrize("fitname", ["Basic_runcard_qed_3replicas_lowprec_398"])
```
scarlehoff, line 174: Instead of adding a complete new test you can put `Basic_runcard_qed_3replicas_lowprec_398` in the parametrize list of the test above. The test itself should do the same.

n3fit/src/n3fit/tests/test_fit.py
```diff
@@ -153,3 +154,50 @@ def test_weirdbasis(tmp_path, timing=30):
...
+
+@pytest.mark.darwin
+def test_performfit_qed(tmp_path):
+    auxiliary_performfit_qed(tmp_path, replica=2, timing=False, rel_error=1e-1)
+
+
+@pytest.mark.linux
+@pytest.mark.parametrize("replica", [1, 2])
+def test_performfit_and_timing_qed(tmp_path, replica):
+    auxiliary_performfit_qed(tmp_path, replica=replica, timing=True)
```
scarlehoff, line 203: Same here, rather than adding a complete new function, you could add a "parametrize" instead and pass the runcard as an argument of the runner.

niclaurenti:
@scarlehoff @RoyStegeman Since the call to `fiatlux` takes a lot should I make that if debug: true the photon is computed on a smaller number of points and reinterpolated?

scarlehoff:
Is the number of points not exposed in the runcard? One could want to run with `debug: true` while having the correct photon.

n3fit/runcards/examples/Basic_runcard_qed.yml
```diff
@@ -95,7 +95,7 @@ datacuts:
 
 ############################################################
 theory:
-  theoryid: 522     # database id
+  theoryid: 523     # database id
```
scarlehoff, line 98: What happened with theory 522 ?

scarlehoff:
I think a better solution is to expose the integration precision to the runcard.

https://github.com/NNPDF/nnpdf/blob/6bfc220a766777e31df9ce242cfd4c584b078a9a/validphys2/src/validphys/photon/compute.py#L22

Then it becomes fast and you can test that things are working (we don't care about correctness in the test after all).

niclaurenti:
@scarlehoff @RoyStegeman where does the `test_evolven3fit.py` is supposed to download the theories 399 and 398? Or they are downloaded somewhere else?

scarlehoff:
I haven't looked at the error, but the easiest solution is to add `API.theoryid(theoryid=THEORYID_399)` at some point at the beginning of the test so that vp downloads it.

niclaurenti:
Should I add some other tests?

validphys2/src/validphys/photon/compute.py
```diff
@@ -59,6 +58,13 @@ def __init__(self, theoryid, lux_params, replicas):
         # This is going to be changed in favor of a bool em_running
         # in the runcard
         fiatlux_runcard["mproton"] = theory["MP"]
+        if "eps_base" in lux_params:
+            fiatlux_runcard["eps_base"] = lux_params["eps_base"]
+            log.warning(f"Using fiatlux parameter eps_base from runcard")
+        else:
+            fiatlux_runcard["eps_base"] = 1e-5 # precision on final integration of double integral.
```
RoyStegeman, line 65: I suppose this comment applies to the whole `if-else` block so can go on top

n3fit/src/n3fit/tests/regressions/quickcard_qed_2.json
```diff
@@ -0,0 +1,95 @@
...
+    }
+  },
+  "version": {
+    "keras": "2.11.0",
+    "tensorflow": "2.11.0, mkl=False",
+    "numpy": "1.22.4",
+    "nnpdf": "4.0.6.846+g0a926fdf1",
+    "validphys": "4.0.6.846+g0a926fdf1"
+  }
+}
```
RoyStegeman, line 95: Whiteline

n3fit/src/n3fit/tests/test_evolven3fit.py
```diff
@@ -145,11 +146,14 @@ def test_eko_utils(tmp_path):
...
 
-@pytest.mark.parametrize("fitname", ["Basic_runcard_3replicas_lowprec_399"])
+@pytest.mark.parametrize("fitname", ["Basic_runcard_3replicas_lowprec_399", "Basic_runcard_qed_3replicas_lowprec_398"])
 def test_perform_evolution(tmp_path, fitname):
     """Test that evolven3fit_new is able to utilize the current eko in the respective theory.
     In addition checks that the generated .info files are correct
     """
+    theory = API.theoryid(theoryid=int(fitname[-3:]))
+    # check that nothing went wrong
+    assert isinstance(theory, TheoryIDSpec)
```
RoyStegeman, line 156: The output of the function is always a `TheoyrIDSpec`.  Also, this doesn't really test something specific for evolevn3fit - in fact, tests of the `check_theoryid` function exist elsewhere

n3fit/src/n3fit/tests/test_evolven3fit.py
```diff
@@ -145,11 +146,14 @@ def test_eko_utils(tmp_path):
     assert_allclose(list(eko_op.operator_card.raw["mugrid"]), op_card_dict["mugrid"])
 
 
-@pytest.mark.parametrize("fitname", ["Basic_runcard_3replicas_lowprec_399"])
+@pytest.mark.parametrize("fitname", ["Basic_runcard_3replicas_lowprec_399", "Basic_runcard_qed_3replicas_lowprec_398"])
 def test_perform_evolution(tmp_path, fitname):
     """Test that evolven3fit_new is able to utilize the current eko in the respective theory.
     In addition checks that the generated .info files are correct
     """
+    theory = API.theoryid(theoryid=int(fitname[-3:]))
```
RoyStegeman, line 154: I think it's a bit nicer to get the theoryid from the fit runcard instead

RoyStegeman (commented):
Thanks for adding this test

validphys2/src/validphys/tests/photon/test_compute.py
```diff
@@ -9,82 +9,53 @@
...
+from validphys.photon.compute import Alpha, Photon, FIATLUX_DEFAULT
 
 from ..conftest import PDF
 
 TEST_THEORY = API.theoryid(theoryid=398)
 
 FIATLUX_RUNCARD = {
     "luxset": PDFset(PDF),
-    "additional_errors": PDFset("LUXqed17_plus_PDF4LHC15_nnlo_100"),
+    "additional_errors": API.pdf(pdf='LUXqed17_plus_PDF4LHC15_nnlo_100'),
```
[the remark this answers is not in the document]
niclaurenti: Because the test was failing because that PDF wasn't installed

validphys2/src/validphys/tests/photon/test_compute.py
```diff
@@ -145,15 +129,15 @@ def test_photon():
         lux.PlugAlphaQED(alpha.alpha_em, alpha.qref)
         lux.InsertInelasticSplitQ(
             [
-                theory["mb"],
-                1e100,
+                theory["kbThr"] * theory["mb"],
+                theory["ktThr"] * theory["mt"] if theory["MaxNfPdf"] == 6 else 1e100,
```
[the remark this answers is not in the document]
RoyStegeman: Oh right, it's passed to fiatlux so `np.inf` is not an option.

validphys2/src/validphys/tests/photon/test_compute.py
```diff
@@ -10,14 +10,17 @@
...
 from validphys.photon import structure_functions as sf
 from validphys.photon.compute import Alpha, Photon, FIATLUX_DEFAULT
+from validphys.loader import FallbackLoader
 
 from ..conftest import PDF
 
 TEST_THEORY = API.theoryid(theoryid=398)
 
+# check if "LUXqed17_plus_PDF4LHC15_nnlo_100" is installed
+FallbackLoader().check_pdf("LUXqed17_plus_PDF4LHC15_nnlo_100")
```
RoyStegeman, line 20: This also returns the object

RoyStegeman:
I think this is done now?

niclaurenti:
> I think this is done now?

yes let's wait for the test to pass and then we can merge

RoyStegeman (approved):
Thanks - looks good to me and tests seem to pass.

PR #1745, opened 2023-05-26
Allow added cuts

Zaharid:
This PR adds the ability to have additional cut definitions that get added to the internal ones. Internal cuts must be used for this to have any effect. This is useful to make comparisons as a function of some cut, without having to copy and modify all of the rules.

The PR also adds support for zero sized datasets (after cuts), at least enough so we can compute chi2 tables and such. I'd assume other things would crash still.

TODO: Add tests and docs.


Zaharid:
Example (note  CMSDY1D12  being different) https://vp.nnpdf.science/oyqljo77QVKLwnOzh6YTPw==/

validphys2/src/validphys/convolution.py
```diff
@@ -295,6 +295,10 @@ def _gv_hadron_predictions(loaded_fk, gv1func, gv2func=None):
     else:
         gv2 = gv1
 
+    # TODO: Ideally we would return before computing the grid
```
RoyStegeman, line 298: Computation of which grid do you mean?

RoyStegeman (commented):
I'm wondering if there should be a warning in case all points of a given dataset are cut, otherwise this seems fine.

Zaharid:
Please review and merge

Zaharid:
@scarlehoff would you please into the new and exciting errors with the the new grids?

scarlehoff:
it looks like some theory failed to download? I've resubmitted one of them.

scarlehoff:
It looks like is not able to find things that should be in theory 398, but I don't understand why is only in this branch. Could you try rebasing or merging master?

It could be that there's something going on in your changes to `convolution.py`, which is the only thing that could interfere somehow as far as I can see.

doc/sphinx/source/vp/filters.rst
```diff
@@ -0,0 +1,393 @@
...
+
+It is thus useful to define these default values somewhere. These values
+can be found within ``validphys.cuts`` inside the ``defaults.yaml``
+file. One can overwrite these values and this is discussed later.
+
+Filters
+-------
+
+In ``validphys 2`` the default filter rules used can be found in the
+``validphys.cuts`` module within the ``filter.yaml`` file. This file is
```
scarlehoff, line 53: If you could add a link to the file in the repo it would be nice, so someone reading the docs can immediately check these out.

validphys2/src/validphys/covmats.py
```diff
@@ -589,7 +589,7 @@ def sqrt_covmat(covariance_matrix):
     dimensions = covariance_matrix.shape
 
     if covariance_matrix.size == 0:
-        raise ValueError("Attempting the decomposition of an empty matrix.")
+        return np.zeros((0,0))
```
scarlehoff, line 592: Why do we want to return an empty array instead of just crashing?

validphys2/src/validphys/tests/test_filter_rules.py
```diff
@@ -93,3 +94,36 @@ def test_good_rules():
...
+            {
+                "speclabel": "fewer data",
+                "added_filter_rules": [
+                    {"dataset": "ATLAS1JET11", "rule": "p_T2 < 1000**2", "reson": "pt cut"}
+                ],
+            },
+            {
+                "speclabel": "empty data",
+                "added_filter_rules": [
+                    {"dataset": "ATLAS1JET11", "rule": "eta < 0", "reson": "empty data"}
```
scarlehoff, line 120:
```suggestion
                    {"dataset": "ATLAS1JET11", "rule": "eta < 0", "reason": "empty data"}
```

scarlehoff (approved):
I took the liberty of adding the rules part of the runcard to your comment.

Given that the tests are ok now in both master and the PRs rebasing to master should be enough for this to pass.

Issue #1729, opened 2023-05-08
Test for lhapdf info files.


scarlehoff:
We have had a few problems with the creation of lhapdf `.info` files. These bugs were flying under the radar for a while since for most of the study we do we don't go fully through `lhapdf` so the fact that were broken didn't affect us.

@cschwan @andreab1997 
could you list here some of the errors / bugs to have an idea of what to test for exactly?

cschwan:
1. the first element of `AlphaS_Qs`, squared, should be equal to `QMin`
2. the number of elements of `AlphaS_Qs` should be equal to the number of elements of `AlphaS_Vals`
3. `AlphaS_OrderQCD` should correspond to `x` in `NxLO`, so `0` for LO, `1` for NLO, ...
4. check that `Particle` is set to an integer in any case!

Conditions 1-3 weren't fulfilled for the QED fit.

andreab1997:
> 1. the first element of `AlphaS_Qs`, squared, should be equal to `QMin`
> 2. the number of elements of `AlphaS_Qs` should be equal to the number of elements of `AlphaS_Vals`
> 3. `AlphaS_OrderQCD` should correspond to `x` in `NxLO`, so `0` for LO, `1` for NLO, ...
> 4. check that `Particle` is set to an integer in any case!
> 
> Conditions 1-3 weren't fulfilled for the QED fit.

Indeed, this is a full list of what already happened. If you want a full list of what could happen in the future I am not sure I can provide one but I can try.

RoyStegeman:
I think the `lhapdf` people are planning to do something similar: https://gitlab.com/hepcedar/lhapdf/-/merge_requests/64 (though we may not want to wait on that). 

Also, I don't think the QED fit bugs would have been found since it was done only locally by Niccolo, right?

PR #1746, opened 2023-05-26
Tests for `evolven3fit_new` and lhapdf `.info` files.

scarlehoff:
I've added a tests that takes an existing fit (with theory 399) and performs the convolution with the eko internal to the theory.

After which the LHAPDF `.info` file is loaded and a few quick check are performed.
Feel free to suggest further things to check (I've added the ones suggested in #1729 by @cschwan)

Note (@niclaurenti) that I've added the tests with 
```
@pytest.mark.parametrize("fitname", ["Basic_runcard_3replicas_lowprec_399"])
```
once theory 398 is ready, you can just create a small fit equivalent to `Basic_runcard_3replicas_lowprec_399` and add it there and it should work ootb.


scarlehoff:
@cschwan @RoyStegeman do you know if the order of the flavours in the `.info` file and the `.dat` files needs to be the same? (because it isn't right now with eko evolved fits, but I don't know whether this is a bug or a feature at this point)

cc @AleCandido @andreab1997

AleCandido:
I guess it would be better if the order were monotonic...

AleCandido:
https://gitlab.com/hepcedar/lhapdf/-/blob/main/src/Info.cc#L18-65

The only entries of `.info` known to the LHAPDF library are:
- [`"Format"`](https://gitlab.com/hepcedar/lhapdf/-/blob/main/src/Factories.cc#L66)
- [`"AlphaS_OrderQCD"`](https://gitlab.com/hepcedar/lhapdf/-/blob/main/src/LHAGlue.cc#L358)
- [`"MinLHAPDFVersion"`](https://gitlab.com/hepcedar/lhapdf/-/blob/main/src/PDF.cc#L20)
- [`"Interpolator"`](https://gitlab.com/hepcedar/lhapdf/-/blob/main/src/GridPDF.cc#L38)
- [`"ErrorConfLevel"`](https://gitlab.com/hepcedar/lhapdf/-/blob/main/src/PDFSet.cc#L71)

apparently...
https://gitlab.com/search?group_id=6283098&language%5B%5D=C%2B%2B&nav_source=navbar&project_id=19453268&repository_ref=main&scope=blobs&search=get_entry+path%3Asrc%2F&search_code=true

(I would have expected more for $\alpha_s$...)

andreab1997 (commented):
Everything seems good to me. Before approving this, let me add just a few commits

felixhekhorn:
Can you add a test where you actually read the PDF using LHAPDF? something along the lines of
```py
import lhapdf # or vp equivalent
pdf = lhapdf.mkPDF("Blub",0)
for q2 in [10,100]: # make them in two different FNS
  assert type(pdf.xfxQ2(pid, x, Q2)) == float # check LHAPDF can actually read for some chosen pid and x
  assert type(pdf.alphasQ2(Q2)) == float # check LHAPDF can actually read
```

AleCandido:
> Can you add a test where you actually read the PDF using LHAPDF? something along the lines of

Then you would have to install LHAPDF in the CI...

RoyStegeman:
> Then you would have to install LHAPDF in the CI...

That's already being done (through conda)

n3fit/src/evolven3fit_new/evolve.py
```diff
@@ -94,23 +89,38 @@ def evolve_fit(
...
+
+        # Modify the info file with the fit-specific info
+        info = info_file.build(theory, op, 1, info_update={})
+        info["NumMembers"] = "REPLACE_NREP"
+        info["ErrorType"] = "replicas"
+        info["XMin"] = float(x_grid[0])
+        info["XMax"] = float(x_grid[-1])
+        # Save the PIDs in the info file in the same order as in the evolution
+        info["Flavors"] = basis_rotation.flavor_basis_pids
+        info["NumFlavors"] = theory.couplings.max_num_flavs
```
scarlehoff, line 117:
```suggestion
        info["NumFlavors"] = theory.heavy.num_flavs_max_pdf
```

n3fit/src/evolven3fit_new/eko_utils.py
```diff
@@ -59,19 +60,29 @@ def construct_eko_cards(
...
+    if theory["MaxNfPdf"] < 6:
+        thresholds["t"] = np.inf
+
     # construct operator card
     q2_grid = utils.generate_q2grid(
         theory["Q0"],
         q_fin,
         q_points,
-        {theory["mb"]: theory["kbThr"], theory["mt"]: theory["ktThr"]},
+        {theory["mb"]: thresholds["b"], theory["mt"]: thresholds["t"]},
```
AleCandido, line 80: It seems to me that here it is hard-coded the assumption that the fit is done in Nf4

n3fit/src/evolven3fit_new/evolve.py
```diff
@@ -114,6 +114,7 @@ def evolve_fit(
         info["XMax"] = float(x_grid[-1])
         # Save the PIDs in the info file in the same order as in the evolution
         info["Flavors"] = basis_rotation.flavor_basis_pids
+        info["NumFlavors"] = theory.couplings.max_num_flavs
```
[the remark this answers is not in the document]
scarlehoff: I'm currently overwriting it from eko because the eko one is wrong. When that property is removed it will already be fixed so I think it's fine.

(also, this comment answers my question on the eko PR, feel free to commit the suggestion you left there)

n3fit/src/evolven3fit_new/eko_utils.py
scarlehoff: I have the same feeling, but I'm not dealing with that (and the associated testing) in this PR x)

scarlehoff:
RE the explicit LHAPDF test (cc @felixhekhorn) I'm not doing postfit or creating an actual PDF fit. I'm looking at the `.dat` and `.info` files in the `nnfit` folder.

I'm not going to do it here since this PR is already growing a bit too much, but I'm planning to do a full fit (setupfit -> fit (2 replicas) -> eko -> postfit) to check that the entire fitting pipeline runs since at the moment each bit is tested by itself.

andreab1997 (approved):

PR #1747, opened 2023-05-26
Fixes to prepare the code for pandas 2.0

### LINKED ISSUE #1722: Use Pandas 2.0
I tried updating to pandas 2.0. So far I have seen an error here:

https://github.com/NNPDF/nnpdf/blob/b5ee4ba61b25e596b3403326f9fb1ef5e861d071/validphys2/src/validphys/tableloader.py#L84

Prescient docstring.

```
    def set_actual_column_level0(df, new_levels):
        """Set the first level of the index to new_levels. Note:
        This is a separate function mostly because it breaks
        in every patch update of pandas."""
        cols = df.columns
>       cols.set_levels(new_levels, inplace=True, level=0)
E       TypeError: MultiIndex.set_levels() got an unexpected keyword argument 'inplace'
```

scarlehoff:
Fixes #1722, at least such that the test pass in my computer. 

Note, right now we are "blocked" in pandas < 2.0 due to `eko`. But we might as well fix these small things as we find them. In this case they removed a keyword and change some defaults https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.groupby.html

Since pandas 2.0 changed a bunch of defaults we will have to look very careful at many reports when we decide to make the change... so for now I'm happy that we are blocked at 1.X.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -41,7 +41,7 @@ requirements:
         - matplotlib >=3.3.0
         - blessings >=1.7
         - scipy >=0.19.1
-        - pandas
+        - pandas <2 # In principle the code runs with pandas 2 but the results (plots, tables, formats) are untested
         - requests
         - prompt_toolkit
         - validobj
--- a/validphys2/src/validphys/tableloader.py
+++ b/validphys2/src/validphys/tableloader.py
@@ -94,7 +94,7 @@ def set_actual_column_level0(df, new_levels):
     This is a separate function mostly because it breaks
     in every patch update of pandas."""
     cols = df.columns
-    cols.set_levels(new_levels, inplace=True, level=0)
+    df.columns = cols.set_levels(new_levels, level=0)
 
 
 # TODO: Find a better place for this function
@@ -130,7 +130,7 @@ def combine_pseudoreplica_tables(
     # The idea is: Set to inf the nans of the valid curves, so that we select
     # the minimum (which is not infinite).  Leave the bad nans as nans, so we
     # write nan always for those.
-    total_chis = total_chis.groupby(axis=1, level=1).apply(fixup_min_points)
+    total_chis = total_chis.groupby(axis=1, level=1, group_keys=False).apply(fixup_min_points)
 
     # Note, asarray is needed because it ignores NANs otherwise.
     argmin = lambda x: pd.Series(np.argmin(np.asarray(x), axis=1), index=x.index)


felixhekhorn:
Since we decoupled [banana from eko](https://github.com/NNPDF/eko/pull/226) there should be no dependency to `pandas` left from the eko side - see here: https://github.com/NNPDF/eko/blob/master/pyproject.toml
Keep in mind that here we need the "box" extras (for evolven3fit_new), but *not* the "mark" extras (only used for internal benchmarking)

felixhekhorn:
> We need banana for the evolution with eko.

Hmm why? now, not any longer ... you need `genpdf` which is in `ekobox`, no? cc @andreab1997

scarlehoff:
> now, not any longer

Or maybe is because QED? I don't know. But it is still in the conda recipe

https://github.com/NNPDF/nnpdf/blob/84c26d598e00e25154944758ac7817abb9618520/conda-recipe/meta.yaml#L55

But as I said, doesn't matter at the moment!

felixhekhorn:
If and when you merge or not I don't mind :stuck_out_tongue: but I want to take eko out of the game :innocent: maybe the conda recipe is outdated ...

RoyStegeman (commented):
How did you ensure this covers the complete set of changes? Did you go through the list of changes listed here: https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html?

scarlehoff:
This does not. This just ensures that at least the code runs in both 2.0 and 1.X at least as far as the tests are concerned.

If we want to update to pandas 2.X we need to actually check the reports one by one since some changes might be only in their effects while the code still runs. Maybe we actually want to fix pandas<2 in the conda recipe I guess.

RoyStegeman:
That's indeed what I expected, but in that case I'd rather not make these changes yet. These are rather a minor effort and worse case we may in the future move to 2.0 without thinking about it (because the tests pass) but some behaviour changes nonetheless.

I suppose another way to prevent this is indeed to set an upper bound in the dependencies, so that would be fine with me as well.

scarlehoff:
> I suppose another way to prevent this is indeed to set an upper bound in the dependencies, so that would be fine with me as well.

I'll do this. The reason is that I would like to move (locally) to pandas 2.0 as soon as possible so that I can fix things organically when/if I find them.

RoyStegeman:
One way is indeed to try it locally for a while, though it's not a huge list of changes that would have to be checked either. We wouldn't need to check for the deprecated or removed functions/inputs since that would result in an error and we'd notice. The things we need to be careful about is changes to defaults, but that's not such a long list to check.

RoyStegeman (approved):

PR #1749, opened 2023-06-04
Fix calling of fiatlux

niclaurenti:
Opening this quick PR to fix the bug that @APJansen found.


scarlehoff (approved):
And that's why we need the small theory + eko for qed as well as soon as possible :P

niclaurenti:
> And that's why we need the small theory + eko for qed as well as soon as possible :P

Yes I will start producing it today, or at most tomorrow

PR #1753, opened 2023-06-08
Fix custom_group processing

Zaharid:
Because we use if instead of elif, the code would search for the custom_group key in the runcard.


validphys2/src/validphys/config.py
```diff
@@ -1580,10 +1580,11 @@ def produce_group_dataset_inputs_by_metadata(
         res = defaultdict(list)
         for dsinput in data_input:
             # special case of custom group, take the grouping from the dataset input
+            log.critical(processed_metadata_group)
```
RoyStegeman, line 1583:
```suggestion
```
Was this meant to be here?

RoyStegeman (commented):
Thanks for spotting this

RoyStegeman (approved):
Thanks for fixing this!

Before this fix the `if processed_metadata_group=="custom_group"` statement was rendered useless

PR #1754, opened 2023-06-09
Enable nf=3 with `evolven3fit_new`.

scarlehoff:
This PR enables the possibility of running `evolven3fit_new` with a fit in `nf=3`.

In order to make it more general now the operator receives a card where every time there is a threshold the Q_threshold is passed for both `nf` and `nf+1` in order to avoid the hack of simply repeating one value.

The problem at the bottom threshold still remains.

Testing a fit evolved with apfel and eko for theory 200/400, they agree with a relative precision of `1e-3` for all points in x and for Q != mb (for nf=5 the agreement is a bit worse when you are close to mb).
I've done the same test for theory 211 (perturbative charm, fit done in nf=3), there we again have a problem at the bottom threshold but as far as I can see the same problem doesn't appear at the charm threshold not sure why.

-> The problem at Q=mb for nf=5 is being investigated. Might be connected to this issue: https://github.com/NNPDF/eko/issues/173

-> The problem at Q=mb for nf=4 is due to eko which at the moment returns an object after evolution which is pdf(q), and pdf(q, nf=4) get's overwritten by pdf(q, nf=5). (same thing happens at nf=3 in theory).

This is still a draft because I would like to:

- [x] Organize better the Q-grids in a separate file.
- [x] Decide whether to set to 0.0 from `n3fit` the exportgrid for `nf > nf0`.


n3fit/src/evolven3fit_new/eko_utils.py
```diff
@@ -58,46 +57,74 @@ def construct_eko_cards(
...
+    mu0 = theory["Q0"]
+    if "nf0" not in theory:
+        if mu0 < theory["mc"] * thresholds["c"]:
+            theory["nf0"] = 3
+        elif mu0 < theory["mb"] * thresholds["b"]:
+            theory["nf0"] = 4
+        elif mu0 < theory["mt"] * thresholds["t"]:
+            theory["nf0"] = 5
+        else:
+            theory["nf0"] = 6
```
felixhekhorn, line 81: this should be the same as `int(nf_default(mu0**2, atlas))` with the `atlas` from below (you can first use an atlas with a random origin)

PS: @AleCandido you see: for outsiders we need to preserve `nf_default`

scarlehoff: If I need to create a fake atlas I prefer to have this explicit here. All things considered it is not very ugly and it makes very explicit what nf0 means.

scarlehoff:
This is ready for review. Some notes:

1. The agreement between an apfel and eko evolved PDF is mostly at the level of a relative difference of `1e-3` but for some points it can be up to `1e-1`.
2. There is no agreement at the quark thresholds, at both side of the threshold. This seems to be a genuine difference https://github.com/NNPDF/eko/issues/173 
3. At every threshold we have two instances of the threshold Q: (Q, nf-1) and (Q, nf). At the moment `eko` cannot deal at evolution with two values of nf for the same Q so I've added a `-epsilon` at `nf-1`.

> Organize better the Q-grids in a separate file.

I think they actually fit well in the `utils.py` module which has only those grids and functions around them. So I've added the one for `nf=3` there.

> Decide whether to set to 0.0 from n3fit the exportgrid for nf > nf0.

Right now there is no `nf0` in the theory and would need to be derived. So I'll pospone this decision to #1756

scarlehoff:
> At every threshold we have two instances of the threshold Q: (Q, nf-1) and (Q, nf). At the moment eko cannot deal (at evolution time) with two values of nf for the same Q so I've added a -epsilon for nf-1.

@giacomomagni @andreab1997 @niclaurenti once this PR is merged it will be necessary to recompute any eko due to the `-epsilon`. So whether to merge this change or not depends on whether you rather prefer to keep the same eko (but have at `nf-1` the result of `nf`). And this depends on whether you believe eko is giving the right results around the thresholds (@felixhekhorn @AleCandido) because if it isn't then it doesn't matter (that much) how wrong they are.

scarlehoff:
I've opened an issue in eko https://github.com/NNPDF/eko/issues/283

I will remove the hack of `- epsilon` since the current bug is just a "reimplementation" of the old behaviour, and that `-epsilon` forces a recomputation of the eko.

scarlehoff:
I've reverted the hack. Please @andreab1997 have a look at this when you have time, as far as I am concerned this can be merged.

n3fit/src/evolven3fit_new/evolve.py
```diff
@@ -145,7 +145,7 @@ def load_fit(usr_path):
     """
     nnfitpath = usr_path / "nnfit"
     pdf_dict = {}
-    for yaml_file in nnfitpath.glob("replica_*/*.exportgrid"):
+    for yaml_file in nnfitpath.glob(f"replica_*/{usr_path.name}.exportgrid"):
```
andreab1997, line 148: Thanks for this, it is surely more consistent now!

n3fit/src/evolven3fit_new/utils.py
```diff
@@ -123,15 +178,22 @@ def get_theoryID_from_runcard(usr_path):
...
+def generate_q2grid(Q0, Qfin, Q_points, match_dict, nf0=None):
     """Generate the q2grid used in the final evolved pdfs or use the default grid if Qfin or Q_points is
     not provided.
 
     match_dict contains the couples (mass : factor) where factor is the number to be multiplied to mass
     in order to obtain the relative matching scale.
     """
     if Qfin is None and Q_points is None:
-        return DEFAULT_Q2GRID
+        if nf0 == 4:
```
andreab1997, line 189: Since now it is becoming more difficult and we have two different defaults, I am wondering if we should just remove all the defaults and just compute the q2grid with Q0, Qfin and Q_points

n3fit/src/n3fit/tests/test_evolven3fit.py
```diff
@@ -23,11 +23,6 @@ def assert_sorted(arr, title):
         raise ValueError(f"The values of {title} are not sorted!")
 
 
-def check_consecutive_members(grid, value):
```
andreab1997, line 26: Why did you remove this?

andreab1997 (approved):
Just a few comments but it looks good to me

n3fit/src/n3fit/tests/test_evolven3fit.py
scarlehoff: Because now there are no consecutive members on the grids since they are separated by `nf`. So, the grid in Q itself contains only 4.92, but when a point belong to two different `nfs` the single point in Q will be added to each separated block.

n3fit/src/evolven3fit_new/utils.py
scarlehoff: Given a theory I think it makes sense to have a default or at least a recipe. Since the theories are a bit messy right now I think I prefer to have two defaults for the ~400 theories we have rather than put my hands there (for now).

n3fit/src/n3fit/tests/test_evolven3fit.py
andreab1997: I see, fine then!

n3fit/src/evolven3fit_new/eko_utils.py
giacomomagni: In which case `nf0` will not be present in the theory card? Maybe the old ones?

scarlehoff: In the nnpdf theory database it's not present (although the information is there because we have mu0 and the thresholds.

n3fit/src/evolven3fit_new/eko_utils.py
```diff
@@ -58,46 +57,75 @@ def construct_eko_cards(
...
+        mu0,
         q_fin,
         q_points,
-        {theory["mb"]: thresholds["b"], theory["mt"]: thresholds["t"]},
+        {
+            theory["mc"]: thresholds["c"],
+            theory["mb"]: thresholds["b"],
+            theory["mt"]: thresholds["t"],
+        },
+        theory["nf0"],
```
giacomomagni, line 100: I understand this work, so in principle is fine. 
When you call `generate_q2grid` with both Q2s given `nf0` is ignored. On contrary by default from the cli here we have Q2s to None so you actually want to use `nf0`, it's a bit confusing...

scarlehoff: Yes, the `nf0` is acting a bit like a secret parameter to generate the q2grid from the defaults.
I guess another option is _not_ to call `generate_q2grid` when there is nothing to generate (just a default being used).

giacomomagni: maybe that's more clean.

AleCandido: Just to point out: `nf0` parameter is going to disappear, or better to move. Now that we are propagating evolution points from EKO, everything should be an evolution point (a point on a lane, rather than on the ladder), and also fitting scale will be an evolution point.

So `nf0` should be *always* used, and never `None` (despite being always 4, at least for the time being - but remember that is 3 for perturbative charm). You should be allowed to discard the `nf` information only at the latest possible time (i.e. when generating the LHAPDF grid file).

`nf0` is an EKO extension to the NNPDF theory database, and inferred from the comparison of `Q0` with `mc * kcThr`, but it would be better to always have it explicit, and infer as little as possible.

scarlehoff:
@niclaurenti @andreab1997 check that this version is compatible with QED also for 0.13.5 and if so please feel free to merge, there's still the `nf`-`qgrid` reorganization open but I won't be back until late next week and I think having all these fixes in master would be best.

Issue #1755, opened 2023-06-09
make_level1_data sorts data in the wrong order


Zaharid:
When calling something like

```
make_level1_data(data, level0_commondata_wc, 0, experiments_index, sep_mult)
```

where all of data, level0_commondata_wc and experiments_index are in coherent "runcard" order, I expect make_level1_data to return the new commondatas in the same order, rather than ordered alphabetically by name.

Reproducer:

```python
from validphys.api import API
from validphys.pseudodata import make_level1_data

inp = {
    "use_cuts": "internal",
    "theoryid": 200,
    "dataset_inputs": [{"dataset": "NMC"}, {"dataset": "ATLASTTBARTOT8TEV"}],
    "fakepdf": "NNPDF40_nnlo_as_01180"
}

level0_commondata_wc = API.level0_commondata_wc(**inp)
experiments_index = API.experiments_index(**inp)
data = API.data(**inp)

make_level1_data(
    level0_commondata_wc=level0_commondata_wc, 
    data=data, 
    experiments_index=experiments_index, 
    filterseed=0, 
    sep_mult=False
)
```

This prints:

```
[CommonData(setname='ATLASTTBARTOT8TEV', ndata=1, commondataproc='INC', nkin=3, nsys=3),
 CommonData(setname='NMC', ndata=204, commondataproc='DIS_NCE', nkin=3, nsys=16)]
```

so in the wrong order wrt everything else.

comane:
https://github.com/NNPDF/nnpdf/pull/1757

Zaharid:
Closed by https://github.com/NNPDF/nnpdf/pull/1765

PR #1757, opened 2023-06-12
correct sorting of level1 data

comane:
PR due to issue https://github.com/NNPDF/nnpdf/issues/1755


validphys2/src/validphys/pseudodata.py
```diff
@@ -369,15 +369,21 @@ def make_level1_data(data, level0_commondata_wc, filterseed, experiments_index,
 
     indexed_level1_data = indexed_make_replica(experiments_index, level1_data)
 
+    dataset_order = [cd.setname for cd in level0_commondata_wc]
```
Zaharid, line 372: not that it matters much, but we should be doing this with a dict rather than a list, to avoid quadratic behavior.

validphys2/src/validphys/pseudodata.py
```diff
@@ -369,15 +369,21 @@ def make_level1_data(data, level0_commondata_wc, filterseed, experiments_index,
...
+
     # ===== create commondata instances with central values given by pseudo_data =====#
     level1_commondata_dict = {c.setname: c for c in level0_commondata_wc}
     level1_commondata_instances_wc = []
 
+    # note that groupby sorts alphabetically
```
Zaharid, line 378: We could add `sort=False` to groupby. AFAICT the behaviour is not documented, as keeping the order, so I would not touch the rest in any way.

comane: `sort=False` does not seem to solve the problem

comane:
Apart from the ordering of the level1 commondata list, the following bug is fixed in the present PR.
The bug is due to `experiments_index` and can be visualised from the snippet below

```
from validphys.api import API
from validphys.pseudodata import make_level1_data


inp = {
    "fit": "210623_mnc_disonly_linear_1000",
    "dataset_inputs": {"from_": "fit"},
    "theoryid":200,
    "use_cuts": "internal",
    "fakepdf": "NNPDF40_nnlo_as_01180"
}

data = API.data(**inp)
experiments_index =  API.experiments_index(**inp) # API.data_index(**inp)
level0_commondata_wc = API.level0_commondata_wc(**inp)



replica = make_level1_data(
    data=data, 
    level0_commondata_wc=level0_commondata_wc,
    filterseed=2,
    experiments_index=experiments_index,
    sep_mult=True
)




print("Ratio level 1/ level 0: all exps")
print(replica[4].central_values / level0_commondata_wc[4].central_values)
```

Gives:

```
Ratio level 1/ level 0: all exps
entry
1        4.586865
2        4.472092
3        4.856960
4        4.761477
5        4.783476
          ...    
250    205.996344
251    199.764029
252    201.027175
253    221.003710
254    207.598561
Name: data, Length: 248, dtype: float64
```

comane:
> @comane the diff is unreadable now unfortunately. Could you please make it so only your changes show up?

Hi @Zaharid, not sure on how to do that. The diffs appears this way because I had to rebase the branch.
The changed files, however, are:  `filters.py`, `pseudodata.py` and `results.py`

RoyStegeman:
When you opened this PR the additional commits were not yet in master. Perhaps changing the base branch to something else and reverting it to master corrects the diffs in this interface? Just a guess though.

> The diffs appears this way because I had to rebase the branch.

Maybe I'm just misinterpreting what you mean, but there is a genuine difference between `git rebase` and `git merge`. It looks like you only merged this branch.

PR #1760, opened 2023-06-21
Separate DY_CC and DY_NC from DY in nnpdf31_process

andreab1997:


enocera:
Note: We should split data sets that include both NC and CC DY. These are:
- `ATLASWZRAP36PB`
- `ATLASWZRAP11CC `
- `ATLASWZRAP11CF`
- `ATLAS_WZ_TOT_13TEV`
- `LHCBWZMU7TEV`
- `LHCBWZMU8TEV`

andreab1997:
> Note: We should split data sets that include both NC and CC DY. These are:
> 
> * `ATLASWZRAP36PB`
> * `ATLASWZRAP11CC `
> * `ATLASWZRAP11CF`
> * `ATLAS_WZ_TOT_13TEV`
> * `LHCBWZMU7TEV`
> * `LHCBWZMU8TEV`

Yes sure. Note that for these datasets I put a comment "TO SPLIT" below the `nnpdf31_process` key

Zaharid:
I guess the point of calling the thing `nnpdf31_process` was that it was clearly made up and would be changed eventually, including in fundamental ways.  I suppose the intention at the time was that a new key would be added. This change could be fine, but also a pain when you are inevitably asked to compare old with new.

scarlehoff:
I agree with @Zaharid, since we are even moving to a new data format, maybe it is a good time to change it?

Otherwise (not sure if this is a sequential or concurrent comment) I'd suggest having the `nnpdf_metadata` being one of the things that can be overwritten by variants in the new metadata. That would allow for easy comparisons with the old ways.

andreab1997:
@enocera (and maybe @scarlehoff ) just a question: looking at the yamldb files of these datasets 
ATLASWZRAP36PB
ATLASWZRAP11CC 
ATLAS_WZ_TOT_13TEV
LHCBWZMU7TEV
LHCBWZMU8TEV

I noticed that, while for ATLASWZRAP36PB and ATLAS_WZ_TOT_13TEV is clear that we have 3 fks linked to W+, W- and Z, for the others we have something like this

`conversion_factor: 1000.0`
`operands:`
`- - ATLAS_WP_7TEV`
  `- ATLAS_WM_7TEV`
  `- ATLAS_DY_7TEV_CC`
`operation: 'null'`
`target_dataset: ATLASWZRAP11CC`

Should I conclude that `- ATLAS_DY_7TEV_CC` is the one for the Z? (it feels a bit weird for the CC at the end)

enocera:
> Should I conclude that `- ATLAS_DY_7TEV_CC` is the one for the Z? (it feels a bit weird for the CC at the end)

Yes, DY tout court stands for NC DY. In this case CC means central selection (contrasted to CF, which stands for forward selection). So nothing to do with the process, but with the measured range in rapidity.

enocera:
@andreab1997 The data sets have been split and tested with a version of `theory 400` which is now updated and available to download. The data sets have been split according to the following names:
- 'ATLASWZRAP36PB' becomes `ATLASWRAP36PB` and `ATLASZRAP36PB`
- 'ATLASWZRAP11CC' becomes  'ATLASWRAP11CC' and  'ATLASZRAP11CC'
- 'ATLAS_WZ_TOT_13TEV' becomes `ATLAS_W_TOT_13TEV' and `ATLAS_Z_TOT_13TEV'
- 'LHCBWZMU7TEV' becomes 'LHCBWMU7TEV' and 'LHCBZMU7TEV'
- 'LHCBWZMU8TEV' becomes 'LHCBZWU8TEV' and 'LHCBZMU8TEV'
The usage of these data sets is the same as their parent counterparts (in particular, note that the `NRM` K-factors must be used with LHCB).

scarlehoff:
@enocera could you also update the "yaml files" for theory 400 in https://github.com/NNPDF/theories_slim ? (now the repository is much smaller so it will be nicer to use than the one with all the grids inside!)

enocera:
@scarlehoff Thanks for pointing this out. That's the kind of things @andreab1997 will take care of.

andreab1997:
Just to be clear I will

- [ ] Update the yaml files in all the MHOU theories (so from 401 to 409 and from 420 to 428)
- [ ] Update the "yaml files" for theory 400 in https://github.com/NNPDF/theories_slim

If there is anything else, please add to this comment. Also, @giacomomagni I think you need to update your theories.

andreab1997:
Sorry @enocera, what about `ATLASWZRAP36PB`?

andreab1997:
Sorry @enocera, what about `ATLASWRAP11CF`?

enocera:
> Sorry @enocera, what about `ATLASWRAP11CF`?

It only contains Z production, so it's DY_NC. No splitting needed.

andreab1997:
@scarlehoff Do you understand why that `evolven3fit` test is failing?

scarlehoff:
You need to rebase since the version of the repository you branched off was for the old version of eko.

scarlehoff (approved):
This looks sane to me. Have you already used these datasets in fits? Because if so I think we can assume this is working so please feel free to merge it.

andreab1997:
> This looks sane to me. Have you already used these datasets in fits? Because if so I think we can assume this is working so please feel free to merge it.

Yes, I am uploading them in the wiki right now. I am merging this then.

PR #1764, opened 2023-06-26
Separate photon's elastic, inelastic and msbar components

niclaurenti:
This PR aim is the following: we want to provide PDF sets in which, instead of the total photon, there is only its elastic or inelastic component.
However, in the sum rules and in the FKtables contraction we still want to have the total component. So, the elastic/inelastic part must be called only at the very last moment, when we register the photon in the lhapdf grid.

At the moment I still don't know how we should treat the additional errors, since they are supposed to be the errors of the total pdf and not of a component of it.

### DIFF
--- a/n3fit/runcards/examples/Basic_runcard_qed.yml
+++ b/n3fit/runcards/examples/Basic_runcard_qed.yml
@@ -170,3 +170,4 @@ fiatlux:
   additional_errors: true # should be set to true only for the last iteration
   luxseed: 1234567890
   eps_base: 1e-2
+  component: total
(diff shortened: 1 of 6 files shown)


scarlehoff:
I would change the strategy a bit. I would make it so the output of the photon is always three components and the flag you give decides whether to sum them together or return only one of them.

So basically in here:

https://github.com/NNPDF/nnpdf/blob/f57975266f3c6e0df12445429485ff621bf2aa82/n3fit/src/n3fit/layers/rotations.py#L120

You would have separate components `_ph_elastic`, `_ph_inelastic` and then either in `register_photon` or in `call` you can do `ph = _ph_elastic + _ph_inelastic` or `ph = _ph_elastic` or whatever you need.

That way you avoid having to pass the flag one level down.

In addition, I would do `total = True` or `separate_components = False` or something like that (so, have a default where the default is "the entire photon") and don't pass anything for the fktables / msbar, only when you actually needed it behaving differently.

PR #1765, opened 2023-06-27
correct sorting and indexing of level1 data

comane:
This PR responds to https://github.com/NNPDF/nnpdf/issues/1755

See also https://github.com/NNPDF/nnpdf/pull/1757

### DIFF
--- a/validphys2/src/validphys/results.py
+++ b/validphys2/src/validphys/results.py
@@ -172,8 +172,6 @@ def data_index(data):
 
     """
     tuples = []
-
-
     for ds in data.datasets:
         
         experiment = get_info(ds).experiment
--- a/validphys2/src/validphys/results.py
+++ b/validphys2/src/validphys/results.py
@@ -173,7 +173,6 @@ def data_index(data):
     """
     tuples = []
     for ds in data.datasets:
-        
         experiment = get_info(ds).experiment
 
         for i in ds.cuts.load():
--- a/validphys2/src/validphys/results.py
+++ b/validphys2/src/validphys/results.py
@@ -26,6 +26,7 @@ from validphys.checks import (
 )
 from validphys.convolution import PredictionsRequireCutsError, predictions
 from validphys.core import PDF, DataGroupSpec, DataSetSpec, Stats
+from validphys.plotoptions import get_info
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


RoyStegeman:
To keep some consistency with the implementation of other groupings I believe you should obtain the same index list if you group all datasets into a single metadata group, i.e. grouping by `ALL` instead of `experiment`. Then, maybe instead of  always reverting the ordering you could do it only if the grouping is `ALL` (this is just a suggestion, I haven't really looked into it too much). 

This is labelled a bug, where does this bug manifest itself? Are the covmat or data to which the index is applied not ordered in the same way as the index?

Zaharid:
@RoyStegeman the bug exists in the snippets demostrated by @comane  and myself in #1755. AFAICT the covmats are built group by group and thus not affected but have not looked in detail.

validphys2/src/validphys/results.py
```diff
@@ -151,6 +152,38 @@ def from_convolution(cls, pdf, posset):
         return cls(stats)
 
 
+def data_index(data):
+    """
+    Given a core.DataGroupSpec instance, return pd.MultiIndex
+    with the following levels:
+    
+    1. experiment
```
Zaharid, line 160: This is not correct, as the grouping can be arbitrary.

Zaharid:
The problem with the implementation in this PR is that now the correlations taken into account for level 1 are restricted to be by experiment (thus excluding exotic covariance matrices).

validphys2/src/validphys/results.py
```diff
@@ -151,6 +152,38 @@ def from_convolution(cls, pdf, posset):
...
+    data: core.DataGroupSpec
+
+    Returns
+    -------
+    pd.MultiIndex
+
+    """
+    tuples = []
+
+
```
scarlehoff, line 176:
```suggestion
```

validphys2/src/validphys/results.py
```diff
@@ -151,6 +152,38 @@ def from_convolution(cls, pdf, posset):
...
+    for ds in data.datasets:
+
```
scarlehoff, line 178:
```suggestion
```

scarlehoff (approved):

scarlehoff:
If you rebase on top of master the tests should pass (I'm tempted to merge even, but maybe it's better to rebase, ensure they pass, then merge)

PR #1767, opened 2023-07-03
Change also luxset with vp-nextfitruncard

niclaurenti:
This PR's aim is to change the also `luxset` (if present) with `vp-nextfitruncard`, and set it to the input fit.

In this case there's no need to change also the `luxseed` since it is used only by the last iteration of the fit, i.e. when `additional_errors` is set to `True` (operation that has to be done manually since one in principle could do as many iterations as he wishes)


RoyStegeman:
What about the seed?

RoyStegeman (approved):
You're right - the seed is not something that's included in the loss function of previous iterations so it doesn't need to change.

PR #1768, opened 2023-07-05
Alpha variation but same thcovmat

andreab1997:
We want to add the possibility of doing  thcovmat fits with theories computed with alpha=0.117 and alpha=0.119 (using however the usual thcovmat computed with alpha=0.118). This PR will contain:

- [x] The new list of theoryids for the construction of the thcovmat
- [x] The updated datasets containing the four new theories (528=NLO,alpha=0.117, 529=NLO,alpha=0.119, 530=NNLO,alpha=0.117, 531=NNLO,alpha=0.119)
- [x] Putting the new theories in the server


RoyStegeman:
Why not extend `pointprescriptions.yaml` by adding the alphas mappings? Would make it a bit more readable.

andreab1997:
> Why not extend `pointprescriptions.yaml` by adding the alphas mappings? Would make it a bit more readable.

Maybe I did not understand but in this case I just want to have exactly the same thcovmat just used together another theoryid. So the point prescription is really the same and the fact that I am changing alpha has 0 effect on the thcovmat.

RoyStegeman:
Oh nvm, I thought you wanted to vary alphas itself as opposed to R/F scales at different central alphas.

andreab1997:
> Oh nvm, I thought you wanted to vary alphas itself as opposed to R/F scales at different central alphas.

Nono, this is just what I have been asked to do: a fit done with varied reference values of alpha (0.117 and 0.119) but with the exact same thcovmat that we use with alpha=0.118.

RoyStegeman (approved):
I'm not going to check if the theory content is all correct...

Issue #1769, opened 2023-07-05
Problem when running fits with MHOU


enocera:
Dear @andreab1997, I'm trying to perform fits with the theory cov matrix. When I run `n3fit` I get the following error (for each replica; this is replica 10, for illustration purposes):
```
[INFO]: Creating replica output folder in /data/theorie/enocera/nnpdfgit/fits/230704-ern-009a/nnfit/replica_10
[WARNING]: Using q2min from runcard
[WARNING]: Using w2min from runcard
[WARNING]: Using maxTau from runcard
[INFO]: All requirements processed and checked successfully. Executing actions.
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: NonUniqueGrids', src/fk_table.rs:47:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Traceback (most recent call last):
  File "/data/theorie/enocera/miniconda3/envs/nnpdf-new-new/bin/n3fit", line 33, in <module>
    sys.exit(load_entry_point('n3fit', 'console_scripts', 'n3fit')())
  File "/data/theorie/enocera/nnpdfgit/nnpdf/n3fit/src/n3fit/scripts/n3fit_exec.py", line 298, in main
    a.main()
  File "/data/theorie/enocera/miniconda3/envs/nnpdf-new-new/lib/python3.9/site-packages/reportengine/app.py", line 395, in main
    self.run()
  File "/data/theorie/enocera/nnpdfgit/nnpdf/n3fit/src/n3fit/scripts/n3fit_exec.py", line 286, in run
    super().run()
  File "/data/theorie/enocera/nnpdfgit/nnpdf/validphys2/src/validphys/app.py", line 152, in run
    super().run()
  File "/data/theorie/enocera/miniconda3/envs/nnpdf-new-new/lib/python3.9/site-packages/reportengine/app.py", line 380, in run
    rb.execute_sequential()
  File "/data/theorie/enocera/miniconda3/envs/nnpdf-new-new/lib/python3.9/site-packages/reportengine/resourcebuilder.py", line 166, in execute_sequential
    result = self.get_result(callspec.function,
  File "/data/theorie/enocera/miniconda3/envs/nnpdf-new-new/lib/python3.9/site-packages/reportengine/resourcebuilder.py", line 175, in get_result
    fres =  function(**kwdict)
[...]

PR #1771, opened 2023-07-10
remove mention that vp-setupfit doesn't always have to be the first step

RoyStegeman:
Since the note did not clarify what is meant by "a standard fit" this has caused some confusion recently: https://github.com/NNPDF/nnpdf/issues/1769. 

We can clarify what a standard fit is (or rather, what isn't), but I imagine exceptions won't be updated if they are added so we may as well instruct people to just always run `vp-setupfit` before running `n3fit`. I would be hard-pressed to think of a scenario, as the last sentence of the note suggests, in which it is necessary to run `vp-setupfit` after `n3fit`. 

This of course shouldn't stop people who know what they're doing from running it in a different order, but there's no need to mention that explicitly in the docs.

### DIFF
--- a/doc/sphinx/source/tutorials/run-fit.md
+++ b/doc/sphinx/source/tutorials/run-fit.md
@@ -122,15 +122,6 @@ following the points presented above you can proceed with a fit.
     downloaded automatically. Alternatively they can be obtained with the
     ``vp-get`` tool.
 
-    ```eval_rst
-    .. note::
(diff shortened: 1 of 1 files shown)


scarlehoff (approved):

Zaharid (approved):

Zaharid:
+1. I always thought this caused more problems than it solved.

PR #1772, opened 2023-07-10
Implement FkRotation as subclass of Rotation by rewriting using a rotation tensor

APJansen:
Implements an old TODO by @scarlehoff to write the `FkRotation` layer as a subclass of `Rotation`.

Unlike for `FlavourToEvolution` the rotation matrix is fixed, so I thought it made sense to define it right in the class.

Unrelated to this, the docstring isn't accurate, referring to 8 dimensions while actually there are 9. This may be a good occasion to fix it, but I'm not sure what to write as I don't know where the +1 is from, any suggestion?

I've verified that this gives the same output using the script below, and also checked that the new way is about 3-5 times faster.
```python
rot = FkRotation()
test_x = tf.random.normal(shape=(1, 20, 9))

out1 = rot(text_x)
out2 = rot.call_old(test_x)

print(out1 - out2)
```

### DIFF
--- a/n3fit/src/n3fit/layers/rotations.py
+++ b/n3fit/src/n3fit/layers/rotations.py
@@ -58,11 +58,11 @@ class FlavourToEvolution(Rotation):
 
 class FkRotation(Rotation):
     """
-    Applies a transformation from the dimension-8 evolution basis
+    Applies a transformation from the dimension-9 evolution basis
     to the dimension-14 evolution basis used by the fktables.
 
     The input to this layer is a `pdf_raw` variable which is expected to have
-    a shape (1,  None, 8), and it is then rotated to an output (1, None, 14)
+    a shape (1,  None, 9), and it is then rotated to an output (1, None, 14)
     """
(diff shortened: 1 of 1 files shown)


scarlehoff:
> Unlike for FlavourToEvolution the rotation matrix is fixed, so I thought it made sense to define it right in the class.

Yes, indeed. And in any case, it is better than the combination "by hand" that we had right now imho. Thanks!

> Unrelated to this, the docstring isn't accurate, referring to 8 dimensions while actually there are 9. This may be a good occasion to fix it, but I'm not sure what to write as I don't know where the +1 is from, any suggestion?

The +1 is coming form the addition of V15 to the fit, in order to allow for the charm-anticharm asymmetry a few months ago. During that change it was anticipated that some "8s" would need to be changed by hand, I'm happy that it's only docs for now :P https://github.com/NNPDF/nnpdf/pull/1629#discussion_r1021238955 so please update the docstr!

scarlehoff (approved):

PR #1777, opened 2023-07-17
Refactor preprocessing

APJansen:
This PR simplifies the preprocessing layer, improving readability without changing any results.  I think these changes are harmless and uncontroversial and can easily be merged.

## Additional changes proposed
**However my goal was a slightly bigger change, that may cause issues so I put the last commit in a [different branch](https://github.com/NNPDF/nnpdf/commit/a44a6b4c8edddad11f0b7b1b3f8b19c816d8b64b).**
What this does is unite all the preprocessing factors into a single alpha vector and a single beta vector, rather than having tons of scalars. This layer is the only place in the model where flavors are treated individually. Each parameter has flavor-dependent min and max values, that go into both the initializer and a constraint, but this can be done as a vector as well.

The reason this may be controversial is that it doesn't allow setting weights as trainable on a flavor-by-flavor basis, only all alphas and/or all betas. I don't see why this would be necessary, but I think I did see a runcard that does this.

## Timing
I did some timing tests as well, creating a model with only the preprocessing layer and training it on random targets for 10_000 epochs:

branch | time (s) |
---- | ---- | 
master | 38 |
refactor_preprocessing| 36 |
prepro_join_weights | 24 | 

The timing script is something like:
```python
    input = Input(shape=(200, 1), batch_size=1)
    prepro = Preprocessing(flav_info=flav_info, seed=0)
    output = prepro(input)
    model = Model(inputs=input, outputs=output)
    test_x = tf.random.uniform(shape=(1, 200, 1), seed=42)
    test_y = tf.random.uniform(shape=(1, 200, 8), seed=43)

    model.compile(loss="mse", optimizer='adam')

    start = datetime.now()
    model.fit(test_x, test_y, epochs=10_000)
    end = datetime.now()
    diff = end - start

```

## checks
- refactor_preprocessing: Regression test passes, so identical results (**note though, even when specifying a seed, the initialization depends on the tensorflow version, the regression test passes on Snellius but not on my laptop**)
- prepro_join_weights: initialization is different when the weights are a vector, so the regression test needs to be updated. But I checked that manually setting the weights to what they were in master, the results are still the same. Other regression tests and the structure of saved models will need to be changed as well. I'll do that only if these changes are approved.


Radonirinaunimi:
This will also remove the possibility of having a fixed-functional form (the one we briefly talked about at some point) that @peterkrack is working on.

APJansen:
@Radonirinaunimi Why exactly? You mean only the changes in the separate prepro_join_weights branch?

APJansen:
Ok, so two good reasons to keep this option. Keeping this option and having the weights as vectors will probably make things more complicated rather than simpler, so let's discard the prepro_join_weights branch then.
Are the changes in this branch ok?

n3fit/src/n3fit/tests/test_preprocessing.py
```diff
@@ -0,0 +1,29 @@
...
+import numpy as np
+
+
+def test_preprocessing():
+    """Regression test"""
+    # taken from basic runcard
+    flav_info = [
+        {'fl': 'sng', 'smallx': [1.05, 1.19], 'largex': [1.47, 2.7]},  # 'trainable': False},
+        {'fl': 'g',   'smallx': [0.94, 1.25], 'largex': [0.11, 5.87]},  # 'trainable': False},
+        {'fl': 'v',   'smallx': [0.54, 0.75], 'largex': [1.15, 2.76]},  # 'trainable': False},
```
RoyStegeman, line 12:
```suggestion
        {'fl': 'sng', 'smallx': [1.05, 1.19], 'largex': [1.47, 2.7], 'trainable': False},
        {'fl': 'g',   'smallx': [0.94, 1.25], 'largex': [0.11, 5.87], 'trainable': False},
        {'fl': 'v',   'smallx': [0.54, 0.75], 'largex': [1.15, 2.76], 'trainable': False},
```
Doesn't really change anything, but since we want to keep supporting setting trainable per flavor this should never give an error.

n3fit/src/n3fit/tests/test_preprocessing.py
```diff
@@ -0,0 +1,29 @@
+from n3fit.layers import Preprocessing
+import tensorflow as tf
```
RoyStegeman, line 2:
```suggestion
```

n3fit/src/n3fit/tests/test_preprocessing.py
```diff
@@ -0,0 +1,29 @@
...
+        {'fl': 'cp',  'smallx': [0.12, 1.19], 'largex': [1.83, 6.7]},
+    ]
+    prepro = Preprocessing(flav_info=flav_info, seed=0)
+    test_x = tf.random.uniform(shape=(1, 4, 1), seed=42)
+    test_prefactors = tf.constant([[
+        [4.28409985e-04, 2.33196355e-02, 2.19706148e-02, 3.12583987e-04, 1.52197943e-04, 6.52114686e-05, 1.36401606e-04, 1.18874144e-04],
+        [5.75779974e-02, 2.65057504e-01, 2.13224307e-01, 3.90783511e-02, 3.33565399e-02, 1.79548096e-02, 3.96815017e-02, 2.73224674e-02],
+        [1.78151987e-02, 1.46395922e-01, 1.27500281e-01, 1.30333444e-02, 9.50834434e-03, 5.17322961e-03, 1.01000341e-02, 7.87989516e-03],
+        [2.80656964e-02, 1.83944702e-01, 1.56266689e-01, 2.01078728e-02, 1.55397197e-02, 8.49795807e-03, 1.71364844e-02, 1.28620844e-02],
+                                  ]])
```
RoyStegeman, line 26:
```suggestion
    test_prefactors = [
        [4.28409985e-04, 2.33196355e-02, 2.19706148e-02, 3.12583987e-04, 1.52197943e-04, 6.52114686e-05, 1.36401606e-04, 1.18874144e-04],
        [5.75779974e-02, 2.65057504e-01, 2.13224307e-01, 3.90783511e-02, 3.33565399e-02, 1.79548096e-02, 3.96815017e-02, 2.73224674e-02],
        [1.78151987e-02, 1.46395922e-01, 1.27500281e-01, 1.30333444e-02, 9.50834434e-03, 5.17322961e-03, 1.01000341e-02, 7.87989516e-03],
        [2.80656964e-02, 1.83944702e-01, 1.56266689e-01, 2.01078728e-02, 1.55397197e-02, 8.49795807e-03, 1.71364844e-02, 1.28620844e-02],
                                  ]
```

RoyStegeman (approved):
Thanks @APJansen, this looks good! For as far as I'm concerned this can be merged after you've addressed a few minor points.

n3fit/src/n3fit/layers/preprocessing.py
```diff
@@ -34,28 +34,36 @@ class Preprocessing(MetaLayer):
...
+        self.betas = []
         super().__init__(**kwargs)
 
-    def generate_weight(self, weight_name, kind, dictionary, set_to_zero=False):
+    def generate_weight(self,
+        name: str,
+        kind: str,
+        dictionary: dict,
+        set_to_zero: bool = False
+        ):
```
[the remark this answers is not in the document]
scarlehoff: Also `isort` (which at least for the tests would change the order of the imports I think).

APJansen:
@RoyStegeman As you saw I was having some issues with the test, thanks for fixing it. Did you understand what was going wrong? It was passing for me locally but not in the CI after incorporating your comment.

n3fit/src/n3fit/tests/test_preprocessing.py
```diff
@@ -0,0 +1,65 @@
...
+            2.6867521e-01,
+            6.4237528e-02,
+            5.9941418e-02,
+            3.0898115e-02,
+            7.7342905e-02,
+            4.8169162e-02,
+        ],
+    ]
+    prefactors = prepro(test_x)
+    assert np.allclose(test_prefactors, prefactors)
```
scarlehoff, line 65:
```suggestion
    np.testing.assert_allclose(test_prefactors, prefactors)
```

n3fit/src/n3fit/tests/test_preprocessing.py
```diff
@@ -0,0 +1,65 @@
...
+        {'fl': 'sng', 'smallx': [1.05, 1.19], 'largex': [1.47, 2.7], 'trainable': False},
+        {'fl': 'g', 'smallx': [0.94, 1.25], 'largex': [0.11, 5.87], 'trainable': False},
+        {'fl': 'v', 'smallx': [0.54, 0.75], 'largex': [1.15, 2.76], 'trainable': False},
+        {'fl': 'v3', 'smallx': [0.21, 0.57], 'largex': [1.35, 3.08]},
+        {'fl': 'v8', 'smallx': [0.52, 0.76], 'largex': [0.77, 3.56]},
+        {'fl': 't3', 'smallx': [-0.37, 1.52], 'largex': [1.74, 3.39]},
+        {'fl': 't8', 'smallx': [0.56, 1.29], 'largex': [1.45, 3.03]},
+        {'fl': 'cp', 'smallx': [0.12, 1.19], 'largex': [1.83, 6.7]},
+    ]
+    prepro = Preprocessing(flav_info=flav_info, seed=0)
```
scarlehoff, line 19:
```suggestion
    prepro = Preprocessing(flav_info=flav_info, seed=1)
```

you never know when someone wrote `if seed:` in their code :P

scarlehoff (approved):
Please check that I didn't made any typos in the suggestion before accepting any of them ¡!

RE the differences, by changing to using `np.testing` we'll have extra information on the failures, it might have just been numerics.

APJansen:
@RoyStegeman It was passing locally for me before your change, and failing after. I think it's because of tensorflow, I had this issue before. The tensorflow version I have locally doesn't handle seeds properly. i.e. initializing a layer with say `seed=42` will be consistent if you run it multiple times, but if you do it twice, initialize two layers with the same seed, they will actually be different.

Now I changed the seed at Juan's suggestion, which I agree is better, but it will break the tests of course, and if I update the numbers it will probably again pass for me locally but fail in the CI, so could you change them again? Sorry about that.

RoyStegeman:
No problem!

PR #1778, opened 2023-07-17
Add produce_eko_photon to evolven3fit_new

niclaurenti:
PR for adding `produce_eko_photon` in `evolven3fit_new`

### DIFF
--- a/n3fit/src/evolven3fit_new/eko_utils.py
+++ b/n3fit/src/evolven3fit_new/eko_utils.py
@@ -39,6 +39,8 @@ def construct_eko_cards(
     x_grid,
     op_card_dict: Optional[Dict[str, Any]] = None,
     theory_card_dict: Optional[Dict[str, Any]] = None,
+    q_gamma = None,
+    is_eko_photon = False
 ):
     """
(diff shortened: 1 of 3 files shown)


andreab1997 (changes requested):
It looks ok but please instead of branching in `construct_eko_cards`, just make another function for the photon. At the end in `construct_eko_cards` you are doing a lot of different things so it is probably better to have another function. All the things that instead are in common you can put in yet another function to be called in both cases.

niclaurenti:
> It looks ok but please instead of branching in `construct_eko_cards`, just make another function for the photon. At the end in `construct_eko_cards` you are doing a lot of different things so it is probably better to have another function. All the things that instead are in common you can put in yet another function to be called in both cases.

Actually the branching `produce_eko_photon` happens only in three points. I did it like this exactly because there were common blocks of code that I had to copy paste, and I didn't want to add a lot of functions for those blocks

scarlehoff:
I agree with @andreab1997 I think it is better to have different functions (and then lift some of the common functionality to wrappers) rather than so many branches.

andreab1997 (approved):
Good for me!

PR #1779, opened 2023-07-18
Refactor xintegrator

APJansen:
Adds a regression test for `xIntegrator` and refactors it to use a tensor product on the x-axis rather than replicating the grid tensor.

I expected it to be faster but the difference is negligible (few % faster).

The main point though is in preparation for other changes. I think @Radonirinaunimi 's inclusion of the nuclear pdfs will benefit from this, as well as a possible optimization for multi-replica fits I'm working on.
I'm trying to put changes that will be beneficial regardless in their own small PRs, I think that'll be easier to review. (Not a lot, I think just 2 more)

### DIFF
--- a/n3fit/src/n3fit/backends/keras_backend/operations.py
+++ b/n3fit/src/n3fit/backends/keras_backend/operations.py
@@ -196,18 +196,6 @@ def tensor_ones_like(*args, **kwargs):
     return K.ones_like(*args, **kwargs)
 
 
-@tf.function
-def many_replication(grid, replications, axis=0, **kwargs):
-    """
-    Generates a tensor with one extra dimension:
-        a repetition of "grid" n times along the given axis
-    from keras documentation:
-    If x has shape (s1, s2, s3) and axis is 1, the output will have shape (s1, s2 * rep, s3)
-    see full `docs <https://www.tensorflow.org/api_docs/python/tf/keras/backend/repeat_elements>`_
-    """
-    return K.repeat_elements(grid, rep=replications, axis=axis, **kwargs)
-
-
 # Property operations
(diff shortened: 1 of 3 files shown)


Radonirinaunimi:
This looks good to me (and it removes some extra-operations)!

But I don't really understand how this would make easier the inclusion of nPDFs?

> The main point though is in preparation for other changes. I think @Radonirinaunimi 's inclusion of the nuclear pdfs

In general, I am for atomic changes, but in order to not lose track of the bigger picture/aim shouldn't these be included in the respective PRs -- and especially if they are really needed?

scarlehoff:
Please don't merge this yet.

APJansen:
@Radonirinaunimi The reason it helps is that now as long as the x axis stays the same it works for pdfs of any shape, so you can add an A axis.

APJansen:
@scarlehoff Is it ok to merge now? I did some tests running for a few epochs. The numerical noise is different so you get very small differences that grow every epoch. I don't think anything short of a full multi replica fit so you can see the distribution will give more checks than the unit test.

scarlehoff (commented):
> The numerical noise is different so you get very small differences that grow every epoch

That's strange, from the changes to the code I'd expect the changes to be identical.

Are you sure that the changes are due to this? Do you have an idea why could it be?

(in any case, the report by the bot should be done in a few hours, we can merge then)

scarlehoff:
Seems exactly the same to me https://vp.nnpdf.science/3G193e67SqS3DlDHqwUEeg==/#Scales0_Normalize_Basespecs0_PDFscalespecs0_Distspecs0_plot_pdfdistances maybe you saw some differences because a different version of some library ?

Issue #1773, opened 2023-07-11
Improve the installation of `nnpdf` packages


scarlehoff:
Currently both `validphys` and `n3fit` are completely free of C++ dependence. In addition, the C++ part of the code is no longer used (some people are still using `evolven3fit` but it is no longer a requirement).

The only external dependency that is not available as a python package is `lhapdf`, which for the moment would need to be installed by the user. However, in some of the systems that we use to run there is either `lhapdf` available by default or an official package for your favourite linux (or macos) flavour so I think we can sidestep the issue for the moment.

So the idea would be to write a `pyproject.toml` file such that it:

- installs `n3fit`
- copies the data and theory db into the package data
- installs all dependencies other than `lhapdf` and `pandoc`
- installs `validphys` 
- then `vp-get` will create the `share/NNPDF/` structure as needed

All `libnnpdf`, `evolven3fit`, etc, can be skipped. If anyone needs any part of it they can still use conda (for the time being we will have to maintain both) but this gives an extra option to people.

Some notes:
1. in principle the list of dependencies is not very different from conda https://github.com/NNPDF/nnpdf/blob/master/conda-recipe/meta.yaml only the C++-specific ones need to be dropped: `swig`, `apfel`, `libarchive`, etc.

2.  python is not allowed to write to `share` during installation, the data and profile should go as a package-data (both in the conda installation and the pip package) and `share` would be for things that are "vp-downloaded". This is also more consistent with how other programs do it and also ensures that installation and data are always in sync as we intended.

3. since `validphys` needs `pandoc` we might want to start by having a package that only works for `n3fit` (which is what needs to run in clusters and might benefit the most from an unmanaged pip installation), validphys would still be installed but would crash badly when creating reports if pandoc is not available (or with a nice error).

cc @APJansen I've added you here since you already worked quite a bit on things adjacent to that when you had to install for MacOS so I'd be grateful if you could chime in (or even take this task if you wish since you already have some experience with the installation!)

A minimal requirements.txt file with packages needed in order for the fit to start running

```
# n3fit
tensorflow
psutil
# evolven3fit
eko
# validphys
pineappl
fiatlux
reportengine
validobj
prompt_toolkit
## hyperopt
hyperopt
seaborn
```

Only LHAPDF is missing from this list.

Zaharid:
[...]

PR #1781, opened 2023-07-18
Refactor msr

APJansen:
Adds a regression test for the `MSR_Normalization` layer.
Mainly improving readability by rewriting the indices.
Also makes it compatible with higher rank pdfs by changing a `flatten(y)` to `y[0]` and allowing for an `out_shape` rather than only a 1d `output_dim`.

Awaiting further tests as per discussion [here](https://github.com/NNPDF/nnpdf/pull/1780#issuecomment-1640063925).


n3fit/src/n3fit/layers/msr_normalization.py
```diff
@@ -1,11 +1,18 @@
...
+    'photon': 0,
+    'sigma': 1,
+    'g': 2,
+    'v': 3,
+    'v3': 4,
+    'v8': 5,
+    'v15': 6,
+    'v35': 7,
+    'v24': 8,
+}
```
RoyStegeman, line 15: It doesn't affect the result, but the order of `v35` and `v24` is reversed.

Since we rely on hardcoding the basis of the fktable anyway, it may be best to have a `dict` like this one stored in a single location which all modules that need it can then import (which at the moment I think are just `x_operations.py`, the current module and `writer.py`).

RoyStegeman (commented):
Not completely done yet with my review, but I ran out of time. Anyway, here are a few comments

n3fit/src/n3fit/layers/msr_normalization.py
APJansen: I agree, do you want me to do that in this commit? If so I'm not sure where to put it.

RoyStegeman: That's also what I'm not too sure about, perhaps we should add an `n3fit/utils.py`? @scarlehoff do you have any ideas?

scarlehoff: The information should be here https://github.com/NNPDF/nnpdf/blob/9cf5f9ecc83ec59fff804f859b531198139021fe/validphys2/src/validphys/pdfbases.py#L695

incidentally, the name of that function, `fitbasis_to_NN31IC`, is no longer correct.

Some things that we said (a very long time ago) someone should do are:
1. Create a function that generate the PDF output -> fktable rotation and use that on the fit (half of the work @APJansen  already did with  https://github.com/NNPDF/nnpdf/pull/1772)
2. Define in that module what the "final fitbasis" actually is so that both `fitbasis_to_NN31IC` and any other function can use that information.

Maybe `fitbasis_to_fitoutput` would be more correct. Or even skip that step and have just `fitbasis_to_fkbasis` since the `fitoutput->fkbasis` step is known and constant. That would also require (small) changes to the QED fit or the integration.

n3fit/src/n3fit/layers/msr_normalization.py
```diff
@@ -27,19 +33,23 @@ def __init__(self, output_dim=14, mode="ALL", **kwargs):
...
 
         super().__init__(**kwargs)
 
-    def call(self, pdf_integrated, photon_integral):
+    def build(self, input_shape):
+        out_shape = input_shape[1:]
+        self._out_scatter = lambda pdf_integrated: op.scatter_to_one(
+            pdf_integrated, indices=self.indices, output_shape=out_shape
+        )
+        super().build(input_shape)
```
scarlehoff, line 50: Since you moved this to `build`, I'd suggest moving the indices part also here, that way you avoid filling up a `self.indices` variable in the middle.

Also, is there a reason you prefer the `lambda`? Because you could also avoid the call to `build` and just do `scatter_to_one` later (I mean, once you are changing the entire layer there's no reason to constrain yourself to the previous one as long as input and output are the same)

n3fit/src/n3fit/layers/msr_normalization.py
```diff
@@ -51,30 +61,26 @@ def call(self, pdf_integrated, photon_integral):
...
-            )
+            n_ag = [(1.0 - y[IDX['sigma']] - y[IDX['photon']]) / y[IDX['g']]]
             norm_constants += n_ag
 
         if self._vsr_enabled:
-            n_av = [3.0 / y[V_IDX[0][0]]] * len(V_IDX)
-            n_av3 = [1.0 / y[V3_IDX[0][0]]] * len(V3_IDX)
-            n_av8 = [3.0 / y[V8_IDX[0][0]]] * len(V8_IDX)
-            n_av15 = [3.0 / y[V15_IDX[0][0]]] * len(V15_IDX)
+            n_av = [3.0 / y[IDX['v']]] * 3
```
scarlehoff, line 80: Same as before, here I'd rather separate "V" and "V3815" (or something)
so that you can have
```
V = ['v', 'v35', 'v24']
V38  = ['v3', 'v8', 'v15]
```
since the `V` list might actually not be `3` forever.

n3fit/src/n3fit/msr.py
```diff
@@ -86,13 +86,17 @@ def generate_msr_model_and_grid(
...
+        lambda pdf_photon: op.concatenate([pdf_photon[1], pdf_photon[0][:, 1:]], axis=1),
+        name="join_photon",
+    )([pdf_integrated, photon_integral])
 
     # 6. Compute the normalization factor
     # For now set the photon component to None
-    normalization_factor = MSR_Normalization(output_dim, mode, name="msr_weights")(
-        pdf_integrated, photon_integral
-    )
+    normalization_factor = MSR_Normalization(mode, name="msr_weights")(pdf_integrated)
```
scarlehoff, line 99: Why this change? I feel the previous one was actually much clearer since it makes the difference between the PDF integration and the photon explicit in the MSR.

(and they are really different objects, while the PDF comes from the NN, the photon comes from an analytical computation and is constant throughout the fit, so the separation in the `call` is good)

scarlehoff (commented):
I wrote a lot because I was writing as I went but basically the summary is:

1. Feel free to change the whole `msr_normalization.py` layer (as long as the input-outputs are the same) and the more things you can put as constants at the top of the module the better.
2. I'd strongly prefer to leave the input fitted (integrated) PDF and photon separated.

n3fit/src/n3fit/layers/msr_normalization.py
APJansen: Good point, moved it to call.

I actually had a bigger rewrite of this layer, removing all of the indexing in the call and replacing it with tensor operations. But I had to do so many tricks to get the broadcasting to work correctly that in the end it was harder to read than what there is now, so I discarded it.

n3fit/src/n3fit/msr.py
APJansen: Maybe I misunderstand (I never understood why the photon can't be treated the same) but they're not really different objects right? They might arise differently but they are all integrals right? I thought a dedicated layer that inserts the photon integral in the first component makes it more transparent, and simplifies the MSR layer.
But if you still prefer the old way I'll change it back.

scarlehoff: The photon is analytically computed before the fit 
https://github.com/NNPDF/nnpdf/blob/9cf5f9ecc83ec59fff804f859b531198139021fe/validphys2/src/validphys/photon/compute.py#L125C13-L125C13

Computing the integral in the same way of the others would require too many points and the photon contribution is not large so we don't need the same level of precision.

So, while they are all "the result of integrals", the photon is a constant that gets injected into the calculation while the others are quantities that are computed and change from epoch to epoch.

n3fit/src/n3fit/layers/msr_normalization.py
APJansen: I don't understand what exactly you propose, could you make the change?

scarlehoff: Here you assume that the same normalization (that of V) is going to be always applied to V, V35 and V24 (you changed the `len(V_IDX)` to just `3`) but this is not necessarily true.

One way around this problem is, since at this stage the content of `pdf_integrated` should be the same for V, V35 and V24, to treat them separately.
i.e., if vsr_enabled, compute all of them separately even if some are the same.

I proposed organizing them in the ones that go with V and the ones that don't, but on second thought is not necessarily better, while separating all of them would also work in the future.

(Please check that this is indeed the case and `pdf_integrated[3] == pdf_integrated[7] == pdf_integrated[8]`)

APJansen: Actually `pdf_integrated[7] == pdf_integrated[8]`, but `pdf_integrated[3]` is different. I was worried for a while but there is a good reason, 3 is divided by x in msr.py while the others are not. I assume that is intentional?

How about what I have now? I factored out all the physics to the top, apart from the msr numerator, which would get messy.

scarlehoff: >  I assume that is intentional?

Since it was not being used, it was not included in the division by x. If it were to be used it should be also included!

> How about what I have now?

I like this new version a lot! It makes it very clear what is going to happen just by looking at the top of the file.
Btw, the definitions of these normalization constants are here https://arxiv.org/pdf/2109.02653.pdf (eq. 3.8) in case you are interested on what they are / why they are there / etc

n3fit/src/n3fit/msr.py
APJansen: Ok, I changed it back to what it was.

Any idea why the CI is failing? Thought it was some fluke but it happend twice in a row now, just says the operation was cancelled when solving the environment..

n3fit/src/n3fit/layers/msr_normalization.py
APJansen: This is probably better left to another PR?

APJansen:
@scarlehoff @RoyStegeman Tests are failing, at the stage of resolving the environment. Also on master now after merging the preprocessing and rotation PRs, though I assume only because they reran and something external has happened, I can't imagine why the code changes would break the conda environment when I didn't add or remove any dependencies. Not sure what to do about this?

scarlehoff:
The problem with macos seems to be different (and real, but stochastic?) 
linux / python 3.9 seems to be something wrong with conda, not sure what triggered it but has been there for a while and in many branches (e.g. https://github.com/NNPDF/nnpdf/pull/1785) 

(this is a time as good as any other to promote #1773)

scarlehoff:
I see the CI has come back to life (yey!) 

Is this ready for review? (I'm going to be on holiday starting next week and I'd like to review the changes because of being a control-freak but also don't want to block the progress in the GPU/multireplica stuff ¡!)

APJansen:
Ah great :) Yes it's ready. You wouldn't really block progress, but if it's possible to review before you go that'd be great :)

I'll reply to #1773 there

n3fit/src/n3fit/layers/msr_normalization.py
```diff
@@ -1,22 +1,51 @@
 from n3fit.backends import MetaLayer
```
scarlehoff, line 1:
```suggestion
"""
    Definition of the imposition of the Momentum Sum Rule and Valence Sum Rules to in the PDF fit.
    
    In the module level constants ``{MSR/VSR}_COMPONENTS`` the flavours affected by the MSR and VSR are defined.
    For the Valence Sum Rule instead `VSR_DENOMINATOR` defines the integral of which flavour are used
    to compute the normalization. Note that for a Nf=4 fit  `v35=v24=v`.
    If the number of flavours were to be changed in the future, this would need to be updated accordingly.
"""
from n3fit.backends import MetaLayer
```

n3fit/src/n3fit/layers/msr_normalization.py
```diff
@@ -51,30 +83,35 @@ def call(self, pdf_integrated, photon_integral):
...
+            the integrated photon PDF
 
         Returns
         -------
         normalization_factor: Tensor(14)
             The normalization factors per flavour.
         """
-        y = op.flatten(pdf_integrated)
+        y = pdf_integrated[0]  # get rid of the batch dimension
+        photon_integral = photon_integral[0]  # get rid of the batch dimension
```
scarlehoff, line 97: Does the photon integral need a batch dimension? You could as well remove the `[[` in below (or is it because of symmetry arguments?)

(you can also do directly here photon_integral[0][0] btw)

n3fit/src/n3fit/layers/msr_normalization.py
```diff
@@ -51,30 +83,35 @@ def call(self, pdf_integrated, photon_integral):
...
+        divisors = op.gather(y, self.divisor_indices, axis=0)
+        norm_constants = norm_constants / divisors
+
+        # Fill in the rest of the flavours with 1
+        norm_constants = op.scatter_to_one(
+            norm_constants, indices=self.indices, output_shape=y.shape
+        )
 
-        return self._out_scatter(norm_constants)
+        return norm_constants
```
scarlehoff, line 117: Please don't use the same name for all these variables, specially the last one where things have been reorganized

n3fit/src/n3fit/layers/msr_normalization.py
```diff
@@ -1,22 +1,51 @@
...
+    'v3': 4,
+    'v8': 5,
+    'v15': 6,
+    'v24': 7,
+    'v35': 8,
+}
+MSR_COMPONENTS = ['g']
+MSR_DENOMINATORS = {
+    'g': 'g',
+}
```
scarlehoff, line 18:
```suggestion
MSR_DENOMINATORS = { 'g': 'g' }
```

n3fit/src/n3fit/layers/msr_normalization.py
```diff
@@ -27,20 +56,23 @@ def __init__(self, output_dim=14, mode="ALL", **kwargs):
...
-            idx += V_IDX + V3_IDX + V8_IDX + V15_IDX
-
-        self._out_scatter = op.as_layer(
-            op.scatter_to_one, op_kwargs={"indices": idx, "output_dim": output_dim}
-        )
+            self.divisor_indices += [IDX[VSR_DENOMINATORS[c]] for c in VSR_COMPONENTS]
+            self.indices += [IDX[c] for c in VSR_COMPONENTS]
+            self.vsr_factors = op.numpy_to_tensor([VSR_CONSTANTS[c] for c in VSR_COMPONENTS])
+        # Need this extra dimension for the scatter_to_one operation
+        self.indices = [[i] for i in self.indices]
```
scarlehoff, line 69: It would be better to remove `self.` from the previous `self.indices` since you are doing this at the end (`indices` being shape `(n,)` and `self.indices` shape `(n,1)`)

n3fit/src/n3fit/layers/msr_normalization.py
```diff
@@ -51,30 +83,35 @@ def call(self, pdf_integrated, photon_integral):
...
         norm_constants = []
 
         if self._msr_enabled:
-            n_ag = [(1.0 - y[GLUON_IDX[0][0] - 1] - photon_integral[0]) / y[GLUON_IDX[0][0]]] * len(
-                GLUON_IDX
-            )
-            norm_constants += n_ag
+            norm_constants += [
+                op.batchit(1.0 - y[IDX['sigma']] - photon_integral[0], batch_dimension=0)
+            ]
```
scarlehoff, line 103: Why do you need to "batch this" while it wasn't necessary before?

scarlehoff (approved):
Beyond the stylstic nitpicking, I think this is good to go.

> Ah great :) Yes it's ready. You wouldn't really block progress, but if it's possible to review before you go that'd be great :)

Better safe than sorry!

n3fit/src/n3fit/layers/msr_normalization.py
APJansen: This is because before the vsr factors were a list of scalars, which I've turned into a rank 1 tensor. So to be able to concatenate that with the msr factor, that needs to be a rank 1 tensor too.

n3fit/src/n3fit/layers/msr_normalization.py
APJansen: I suppose this was more relevant when I actually joined them together before being passed to this layer, but still having the same shapes makes it easier to change things I think, even if it's a bit superfluous now.

APJansen:
Conda broke again... I'll rerun the tests once in a while and merge it if it passes, thanks for the review :)

PR #1782, opened 2023-07-24
Multi Replica PDF

APJansen:
# Question

This will be some work, so before continuing past this I'd like to confirm that you agree that once finished this will be a beneficial change. 

# Idea

The idea of this PR is to refactor the tensorflow model from taking a list of single-replica pdfs into taking a single multiple replica pdf, a single pdf whose output has an extra axis representing the replica. This is much faster on the GPU, see tests below. 

The main ingredient to make this possible is a `MultiDense` layer, (see [here](https://github.com/NNPDF/nnpdf/blob/ef0da118da5cd0114465d93c2a3f368a7257ab48/n3fit/src/n3fit/backends/keras_backend/multi_dense.py#L6)) which is essentially just a dense layer where the weights have one extra dimension, with size the number of replicas. For the first layer, which takes x's as input, this is exactly it. For deeper layers, the input already has a replica axis, and so the right index of the input has to be multiplied by the corresponding axis of the weights.

# Development Strategy

To integrate this into the code, many small changes are necessary.
To make it as simple as possible to review and test, I aim to make small, independent changes that ideally are beneficial, or at least not detrimental, on their own. Wherever it's sensible I'll first create a unit test that covers the changes I want to make, and make sure it still passes after, and wherever possible I'll try to have the outputs be identical up to numerical errors. I'll put all of these on their own branch and with their own PR (maybe I should create a special label for those PRs?).

Once those small changes are merged, the actual implementation should be easily managable to review.

This PR itself for now is a placeholder, where I just added the commit so that I can create a draft PR and so you can check out the `MultiDense` layer.

I expect that as a final result you'll still want single replica pdf. I will add code that, once all computations are done, just splits the multi replica pdf into single ones, so the saving and any interaction with validphys will remain unchanged.

# Performance

## Timing

These are the timing tests I did on a 1/4 node on Snellius, with one GPU. I'm reporting the average seconds per epoch that is printed in debug mode.
runcard | replicas | multi_replica_pdf_test | trvl-mask-layers | master |
--- | --- | --- | --- | --- |
Basic | 200 | 0.12 | 1.2 | 2.3 |
NNPDF40_nnlo_as_01180_1000 | 200 | out of memory | out of memory | - |
NNPDF40_nnlo_as_01180_1000 | 100 | 0.76 | 1.12 | - |

## Memory

Memory also appears to be significantly reduced.
I checked the peak cpu memory usage using libmemprofile, on the basic runcard with 200 replicas, and found 3.5Gb versus 16.5 for the trvl-mask-layers branch.

# Status

I have a [test branch](https://github.com/NNPDF/nnpdf/tree/multi_replica_pdf-test) where this is working up to the end of the model training, which is what I used to obtain the timings above. 

branch | finished | tested | merged | comments |
| ----- | -------- | ------ | -------  | ----------- |
|[refactor_xintegrator](https://github.com/NNPDF/nnpdf/pull/1779#issue-1809420566)| X | unit | X | |
|[refactor_msr](https://github.com/NNPDF/nnpdf/pull/1781#issue-1809989517)| X | unit | X | |
|[refactor_preprocessing](https://github.com/NNPDF/nnpdf/pull/1777#issue-1807542208)| X | unit | X | |
|[refactor_rotations](https://github.com/NNPDF/nnpdf/pull/1780#issue-1809627910)| X | unit | X | |
|[refactor_stopping](https://github.com/NNPDF/nnpdf/pull/1792) | X | unit | X| |
|[multi-dense-logistics](https://github.com/NNPDF/nnpdf/pull/1818) ||||currently working on this|
|[multi_replica_pdf-test](https://github.com/NNPDF/nnpdf/tree/multi_replica_pdf-test)  |  |  |  | This is my test branch, which has the 4 above, and trvl-mask-layers,  merged into it and has the code that will eventually go into this PR |

### DIFF
--- a/n3fit/src/n3fit/backends/keras_backend/base_layers.py
+++ b/n3fit/src/n3fit/backends/keras_backend/base_layers.py
@@ -17,14 +17,21 @@
     The names of the layer and the activation function are the ones to be used in the n3fit runcard.
 """
 
-from tensorflow.keras.layers import Lambda, LSTM, Dropout, Concatenate
-from tensorflow.keras.layers import concatenate, Input # pylint: disable=unused-import
+from tensorflow import expand_dims, math, nn
+from tensorflow.keras.layers import (  # pylint: disable=unused-import
+    Dropout,
+    Input,
+    Lambda,
+    concatenate,
+)
 from tensorflow.keras.layers import Dense as KerasDense
-from tensorflow import expand_dims
+from tensorflow.keras.layers import LSTM, Concatenate  # pylint: disable=unused-import
 from tensorflow.keras.regularizers import l1_l2
-from tensorflow import nn, math
+from tensorflow.keras.initializers import GlorotNormal
 
 from n3fit.backends import MetaLayer
+from n3fit.backends.keras_backend.multi_dense import MultiDense
+
 
 # Custom activation functions
 def square_activation(x):
@@ -123,6 +130,19 @@ layers = {
             "units": 5,
             "activation": "sigmoid",
             "kernel_regularizer": None
+        },
+    ),
+    "multidense": (
+        MultiDense,
+        {
+            "input_shape": (1,),
+            "initializer_class": GlorotNormal,
+            "units": 5,
+            "activation": "sigmoid",
+            "kernel_regularizer": None,
+            "replicas": 1,
+            "seed": 42,
+            "replica_input": True,
         },
     ),
     "dense_per_flavour": (
--- a/n3fit/src/n3fit/backends/keras_backend/multi_dense.py
+++ b/n3fit/src/n3fit/backends/keras_backend/multi_dense.py
@@ -1,0 +1,139 @@
+import tensorflow as tf
+from tensorflow.keras.initializers import Initializer
+from tensorflow.keras.layers import Dense
+
+
+class MultiDense(Dense):
+    """
+    Dense layer for multiple replicas at the same time.
+
+    The kernel initializer is set using the custom arguments `initializer_class` and
+    `seed`. The `seed` is incremented by 1 for each replica.
+
+    Inputs to this layer may contain multiple replicas, for the later layers.
+    In this case, the `replica_input` argument should be set to `True`, which is the default.
+    For the first layer, there are no replicas yet, and so the `replica_input` argument
+    should be set to `False`.
+
+    Example
+    -------
+
+    >>> from tensorflow.keras import Sequential
+    >>> from tensorflow.keras.layers import Dense
+    >>> from tensorflow.keras.initializers import GlorotUniform
+    >>> import tensorflow as tf
+    >>> replicas = 2
+    >>> multi_dense_model = Sequential([
+    >>>     MultiDense(units=8, replicas=replicas, seed=42, replica_input=False, initializer_class=GlorotUniform),
+    >>>     MultiDense(units=4, replicas=replicas, seed=52, initializer_class=GlorotUniform),
+    >>>     ])
+    >>> single_models = [
+    >>>     Sequential([
+    >>>         Dense(units=8, kernel_initializer=GlorotUniform(seed=42 + r)),
+    >>>         Dense(units=4, kernel_initializer=GlorotUniform(seed=52 + r)),
+    >>>         ])
+    >>>     for r in range(replicas)
+    >>>     ]
+    >>> gridsize, features = 100, 2
+    >>> multi_dense_model.build(input_shape=(None, gridsize, features))
+    >>> for single_model in single_models:
+    >>>     single_model.build(input_shape=(None, gridsize, features))
+    >>> test_input = tf.random.uniform(shape=(1, gridsize, features))
+    >>> multi_dense_output = multi_dense_model(test_input)
+    >>> single_dense_output = tf.stack([single_model(test_input) for single_model in single_models], axis=-1)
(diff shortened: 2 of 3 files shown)


APJansen:
> Yes to both. But note that this is not due to interactions with validphys (which we could modify at will) but rather because each replica is independent of all the others. I.e., training (+data, trvlsplit, stopping, lagrange multipliers, etc) should be independent.
Well also as a practical consideration, doing everything per replica is very baked in, different folders to save to etc, it would require a lot of changes.

And I imagine that there are users that want to be able to evaluate a single PDF without having to evaluate all replicas. I actually know nothing about what happens with these PDFs once trained, can you say something about that or link to something?

> Regarding the photon or the hyperopt penalties (again, I guess this was your plan already, but writing it here to make sure we are all on the same page), I'd suggest leaving that for after the standard multireplica fit in GPU is well tested and merged. The photon might not even be suitable for GPU paralellization since a non negligible amount of time is spent calculating the photon with [fiatlux](https://github.com/scarrazza/fiatlux.git) so the best thing would be to make sure that the QED fit is not broken when running it in the "normal 1-replica way".

Yes this whole branch only makes sense to merge after the trvl-mask-layers branch is merged. The hyperopt penalties can trivially be parallelized across replicas, the photon I'm not sure, if not there just needs to be an interface extracting single replicas from the joined model.

RoyStegeman:
> I actually know nothing about what happens with these PDFs once trained, can you say something about that or link to something?

In particle physics we collide protons, i.e. bound states of quarks and gluons, with another proton or lepton. However, in perturbative QCD we can only calculate Feynman diagrams with individual incoming quarks/gluons, not with incoming hadrons. However, each Feynman diagram essentially needs to be weighted by the probability of finding the corresponding incoming states inside the proton in order to make a connection between the proton-lepton collision and the quark/gluon-electron Feynman diagram. These weights are what the PDFs provide, if you will. There are different "factorization" arguments for different processes that provide the theoretical underpinning of this factorizing of the quark from the proton (there are no formal proofs for all processes though). 

For a general introduction to QCD/collider physics any set of lecture notes on the topic will do. For a more specific discussion of what NNPDF does you could have a look here: https://arxiv.org/pdf/2008.12305.pdf (see equation 1 in these notes for the factorization equation I explained above). Perhaps you don't want to read the entire thing, but up to section 2.2 might be useful.

niclaurenti:
> Regarding the photon or the hyperopt penalties (again, I guess this was your plan already, but writing it here to make sure we are all on the same page), I'd suggest leaving that for after the standard multireplica fit in GPU is well tested and merged. The photon might not even be suitable for GPU paralellization since a non negligible amount of time is spent calculating the photon with [fiatlux](https://github.com/scarrazza/fiatlux.git) so the best thing would be to make sure that the QED fit is not broken when running it in the "normal 1-replica way".

Dear @APJansen and @scarlehoff ,
just to let you know that at the moment the QED fit can handle the multireplica fits with the limitation that all the photons will be computed sequentially. It means that for a 100 replicas fit, to compute all the photons it will take roughly 30 min * 100 = forever.
This is because when I implemented it we where not using the parallel fits so I didn't bother to parallelize the photons computation.
Obviously, if it is needed I'm happy to help to speed up that part of the code

APJansen:
@RoyStegeman Thanks, this I knew, sorry for not being clear. (My background is in theoretical physics as well, though mostly on black hole physics, but I did take master courses on QFT and particle physics so I know the basics)
I meant on a practical level, does this code have users outside of the collaboration, or do people only use the model outputs that you provide for example?

APJansen:
@scarlehoff Can you comment on [this](https://github.com/NNPDF/nnpdf/blob/5492542e1d5eb55033aee1de9364c42e3f556fe4/n3fit/src/n3fit/stopping.py#L244)? 

Looking at this again, I'm trying to rewrite everything in terms of `pdf_model` which is a single model consisting of a stack of `pdf_models` before actually merging the pdfs inside. Wondering what to do here.

Also, later in the same class, replicas are set to non-trainable, but this only takes effect when the model is recompiled, which as far as I can see is not happening here. (And this won't be possible any longer I think once all replicas are a single model)

scarlehoff:
Hi @APJansen, feel free to modify the strategy as you wish.
The important part is that the replicas are 100% independent from each other.
This I achieved back in the day with the trainable=false and maybe that either triggered a recompilation or it was recompiled manually. Or maybe I was just storing the epoch number and rewinding back to the right epoch at the end. I don't remember (I stopped the development of the multigpu once JR got the grant for doing it in Amsterdam so I cannot even promise I tested it...).
In any case, this was easy in principle since the model was a concatenation of models so you could treat them independently and the only unsolved thing at the time was the trvl split.


In any case, for your situation one possible strategy might be this (talking without actually having put my hands in the code to test what the problems/issues might be):

1. Ensure there's no crosstalking between the nodes corresponding to different replicas  
2. Record the state of the _part_ of the network that should've stopped by itself (say, weights 10 to 20) 
3. Set the chi2 of that part of the network to a constant 0 (so that it doesn't affect the loss landscape anymore)

And then continue training until all replicas have triggered the stopping condition. At that point you readd all weights back since you got them at their best and then you have a network in which each replica has been trained independently.

APJansen:
Yes that was my plan, except that I hadn't thought of step 3, thanks!
I'll look into it!

APJansen:
Closing this, all of this has been done.

Issue #10, opened 2018-01-31
validphys.loader.download_file should fail for partial downloads


Zaharid:
Right now if the download is cancelled in the middle, the user will get a file with the requested name, but corrupted. Subsequent attempts to use the file will fail at runtime. Instead we should download to a temporary file (but in the same folder, rather than `/tmp`) and move (rename) the file, only after the download has succeeded.

PR #1788, opened 2023-08-09
Parallel replicas with varying tr-vl masks

goord:
This is a continuation of the pull request #1661 that implements the parallel replicas with ``same-trvl-seed=false`` support. The branch has been migrated to this original repo and the latest master was merged.


validphys2/src/validphys/utils.py
```diff
@@ -5,14 +5,52 @@
...
+        return False
+
+
+def immute(obj: Any):
+    # So that we don't infinitely recurse since frozenset and tuples
+    # are Sequences.
+    if is_hashable(obj):
+        return obj
+    elif isinstance(obj, Mapping):
+        return frozenset([(immute(k), immute(v)) for k, v in obj.items()])
```
Zaharid, line 34: The dict keys are already hashable.

```suggestion
        return frozenset([k, immute(v)) for k, v in obj.items()])
```

validphys2/src/validphys/utils.py
```diff
@@ -5,14 +5,52 @@
...
+    elif isinstance(obj, Sequence):
+        return frozenset([immute(i) for i in obj])
```
Zaharid, line 36:
```suggestion
        return tuple(immute(i) for i in obj)
```

It looks like a bad idea to destroy the order using a frozenset, which doesn't keep order. Functions  dealing with a sequence typically assign meaning to it. I think you may want a tuple here instead.

validphys2/src/validphys/utils.py
```diff
@@ -5,14 +5,52 @@
...
+def freeze_args(func):
+    """Transform mutable dictionary
+    Into immutable
+    Useful to be compatible with cache
+    """
+    @functools.wraps(func)
+    def wrapped(*args, **kwargs):
+        args = tuple([immute(arg) for arg in args])
+        kwargs = {k: immute(v) for k, v in kwargs.items()}
+        return func(*args, **kwargs)
```
Zaharid, line 50: I feel this is rather dangerous as a general decorator, as it changes the meaning of the inputs in unexpected ways.

It may make more sense to do this closer to the runcard input. In which case the freezer becomes

```
def freeze(obj):
    if isinstance(obj, list):
        return tuple(freeze(ele) for ele in obj)
    if isinstance(obj, dict):
        return ((k, freeze(v)) for k, v in obj.items())
    if isinstance(obj, set):
        return frozenset(obj)
    return obj
```

validphys2/src/validphys/utils.py
```diff
@@ -5,14 +5,52 @@
...
+        hash(obj)
+        return True
+    except:
+        return False
+
+
+def immute(obj: Any):
+    # So that we don't infinitely recurse since frozenset and tuples
+    # are Sequences.
+    if is_hashable(obj):
```
Zaharid, line 31: Computing hashes of potentially deeply nested objects can be expensive. Given that you are failing anyway, you can as well do nothing and deal with effectively the same error.

validphys2/src/validphys/utils.py
goord: You're right. I still suffer from breaking tests though, I may have to reintroduce the frozendict anyway...

validphys2/src/validphys/utils.py
goord: Right, but where in the code should the freezer be called?

Zaharid: An input to a vp action is either:

  - Some input coming from the runcard, which is not hashable in general.
  - The ouput of a parsing or a production rule (like a `PDF` object) which really needs to be hashable already.

Hence you really only need to make sure to make the inputs hashable, which would not change the semantics too much. And the best place to do that is close to where you are parsing the input. It is something I had around for reportengine "1.0" which is why I had the above function lying around.

validphys2/src/validphys/utils.py
goord: Indeed we could skip the hashing check. As a side note: I have no idea why tests are currently failing in this branch...

validphys2/src/validphys/utils.py
Radonirinaunimi: FYI, I am looking into this now!

scarlehoff:
Looking through all the `cuts` classes the `lru_cache`s that you added should be fine. Below I've added a small script that tests that they are actually loaded independently.

1. `load_with_cuts`, the `cuts` variable will only be understood as being the same when it is really the same (either it comes from the same path or are internal to the same dataset or are defined by the same similaritycuts, so it should be fine)
2. `dataset_t0_predictions` for a given dataset and a given pdf set these should be exactly the same.
3. `loaded_commondata_with_cuts` (same as (1.))
4. `produce_rules`, it's only used with `use_cuts=internal` (and then they are what they are) or when `filter_rules` is given. And if `filter_rules` are given then, again, they will be defining the cut and if they are the same the rules should be the same.

So it _seems_ safe to me. We might want to do a comparefits report when we know we have different cuts to make sure that stays the same. Several examples here (that use different kind of cuts)

https://vp.nnpdf.science/sK8VI_HITF-hrg2Cx-_lPQ==/
https://vp.nnpdf.science/GMXWE_wgS-Os5pbbsNA4gA==/
https://vp.nnpdf.science/rm4Vqm7zQwW8m6v5O1c_4A==/

However, some coments.

(2.) Suggests to me that the `lru_cache` could be put instead at the level of the covariance matrix since this is the only thing the t0 predictions are used for and for a given list of datasets, cuts an a given `t0set` the covariance matrix should be the same. This might be trickier to implement though and if this already gives a good speed up maybe not worth it.


```
from validphys.api import API
from validphys.convolution import central_predictions
dsname = "NMCPD_dw_ite"
ds = API.dataset(dataset_input={"dataset": dsname}, theoryid=400, use_cuts="fromfit", fit="230718-jcmwfaser-001")
ds2 = API.dataset(dataset_input={"dataset": dsname}, theoryid=400, use_cuts="fromfit", fit="230725-jcm-nnpdf40-002")
pdf = API.pdf(pdf="NNPDF40_nnlo_as_01180")
cv = central_predictions(ds, pdf)
cv2 = central_predictions(ds2, pdf)
print(cv.shape[0] != cv.shape[1])
> True
```

I've manually changed the cuts in both fits to make sure they were different.

RE the tests, I think you are seeing the problem that was fixed here https://github.com/NNPDF/nnpdf/pull/1805

goord:
It looks like something is still wrong. I did a test with a single epoch, 10 parallel replicas, and it seems to me that the difference between baseline (sequential replicas) and the parallel case is growing with the replica number:
`
[gijstest@int5 1_epoch_10_replicas]$ grep -i "Best fit for replica" master/seq-cpu/nnpdf-cpu-10.log 
[INFO]: Best fit for replica #1, chi2=12155.363 (tr=15396.849, vl=3370.149)
[INFO]: Best fit for replica #2, chi2=52.119 (tr=56.119, vl=78.265)
[INFO]: Best fit for replica #3, chi2=44.637 (tr=49.594, vl=67.921)
[INFO]: Best fit for replica #4, chi2=1759.277 (tr=1103.458, vl=3076.298)
[INFO]: Best fit for replica #5, chi2=181.414 (tr=190.806, vl=254.887)
[INFO]: Best fit for replica #6, chi2=880.913 (tr=1119.586, vl=281.592)
[INFO]: Best fit for replica #7, chi2=9869.230 (tr=12839.809, vl=1296.612)
[INFO]: Best fit for replica #8, chi2=136.152 (tr=144.090, vl=194.078)
[INFO]: Best fit for replica #9, chi2=351.785 (tr=424.573, vl=156.781)
[INFO]: Best fit for replica #10, chi2=70873.164 (tr=14250.999, vl=225323.688)
[gijstest@int5 1_epoch_10_replicas]$ grep -i "Best fit for replica" trvl-mask/seq-cpu/nnpdf-cpu-10.log 
[INFO]: Best fit for replica #1, chi2=12155.471 (tr=15396.991, vl=3370.150)
[INFO]: Best fit for replica #2, chi2=52.119 (tr=56.119, vl=78.265)
[INFO]: Best fit for replica #3, chi2=44.637 (tr=49.594, vl=67.921)
[INFO]: Best fit for replica #4, chi2=1759.277 (tr=1103.458, vl=3076.298)
[INFO]: Best fit for replica #5, chi2=181.414 (tr=190.806, vl=254.887)
[INFO]: Best fit for replica #6, chi2=880.913 (tr=1119.586, vl=281.592)
[INFO]: Best fit for replica #7, chi2=9869.230 (tr=12839.809, vl=1296.612)
[INFO]: Best fit for replica #8, chi2=136.152 (tr=144.090, vl=194.078)
[INFO]: Best fit for replica #9, chi2=351.789 (tr=424.577, vl=156.781)
[INFO]: Best fit for replica #10, chi2=70875.047 (tr=14250.945, vl=225331.047)
[gijstest@int5 1_epoch_10_replicas]$ grep -i "Best fit for replica" trvl-mask/par-cpu/nnpdf-cpu-10.log 
[INFO]: Best fit for replica #1, chi2=12155.734 (tr=15381.599, vl=3383.293)
[INFO]: Best fit for replica #2, chi2=52.516 (tr=56.810, vl=78.048)
[INFO]: Best fit for replica #3, chi2=43.484 (tr=47.829, vl=69.136)
[INFO]: Best fit for replica #4, chi2=1679.792 (tr=1526.910, vl=1633.075)
[INFO]: Best fit for replica #5, chi2=200.910 (tr=216.563, vl=258.565)
[INFO]: Best fit for replica #6, chi2=2451.295 (tr=3158.492, vl=608.226)
[INFO]: Best fit for replica #7, chi2=2967.282 (tr=3846.834, vl=671.785)
[INFO]: Best fit for replica #8, chi2=1429.978 (tr=1877.560, vl=221.299)
[INFO]: Best fit for replica #9, chi2=553.733 (tr=642.818, vl=245.547)
[INFO]: Best fit for replica #10, chi2=123811.469 (tr=144675.531, vl=52737.551)
`
I expect some roundoff errors, but certainly not a systematic increase per replica. BTW, for the sequential mode the chi2 is identical to the master branch

Radonirinaunimi:
As already mentioned, I don't expect the numbers at a given epoch to be fully consistent/close. In the parallel replicas case, even a very small difference at the start of the fit could lead to fluctuations during the training. It is not that all the changes here kept the numerical values to be exactly the same. So, the proper way to compare the results are by running proper fits. At the very least for now we should compare: master sequential vs trvl-mask sequential and master sequential vs trvl-mask parallel. 

Does this make sense?

RoyStegeman:
Well I agree with @goord that the pattern is curios. If all the seeds are the same at the start of each replica then the result should be the same for the sequential and parallel cases. It may be that the seeds are not the same because e.g. the optimizer/nn initialization/trvl split/whatever seed doesn't get reset for the parallel case(?) but then I still don't understand why the not only the first but also the second and third replicas have quite good agreement while this agreement has clearly deteriorated by replicas 9 and 10. If the seeds were not the same I'd expect the deterioration at replica 2 to be equivalent to all later differences... 

Also the edit that it's a problem for the NNPDF4.0 runcard but not for the basic runcard is odd. Is `debug` always False when doing the comparison for the basic runcard?

scarlehoff:
In this case this could point however to a bug on how the fit is being stopped. The first replica is stopped correctly, the second one is a bit farther away from the optimal point and so on and so forth.
For instance, more than the numerical difference, I'd be worried that for replica 8 chi_tr < chi_vl in the sequential case but chi_tr > chi_vl in the parallel case.

The seed is the other thing that comes to mind, as @RoyStegeman said, it would make sense that it is the same for the 1st replica but then it is different for the rest. However, I fail to see why should it be more different for 5 than for 2 (and the first few replicas are very close...)

The basic_runcard has `same_trvl_per_replica: True`, did you remove that option? Because if you didn't that might be pointing to the source of the bug.

goord:
Well my first suspicion was that somewhere along the losses or observables, tensors get somehow accumulated along the replica dimension, but then again the basic runcard seems to reproduce the sequential fit ..

RoyStegeman:
But are all the important settings the same for the basic runcard and the NNPDF4.0 runcard? I.e. the changes between the basic runcard you use are limited to choice of dataset, seeds (same seeds just different value),  and preprocessing, while keys such as `debug` and `same_trvl_per_replica` are the same in the two cases? If they are all the same, do you again find the divergence of results if all you do is add more datasets to the basic runcard?

I'd like to understand what the important difference is between the runcards that causes the different behaviour.

goord:
Both have dis and Dy datasets, positivity constraints and same_trvl_per_replica set to false to test the new functionality. The nnpdf4.0 has integrability layers though, which the basic runcard doesn't have

goord:
Regarding the single-epoch reproducibility test:
* On macos, both sequential and parallel modes seem to produce the same result (modulo small numerical noise)
* On Linux, also the master branch (with same tr/vl split for all replicas) gives a different result for the single-epoch sequential and parallel fit with 10 replicas.

goord:
Over the weekend I did an extensive set of regression tests between master (rev. db8b790c1bacf52f71cc3a7fabf0fd9674cbf476) and trvl-mask-layers branch (latest rev.). To my untrained eye, results look more or less identical:

| runcard  | comparison (100 replicas)                  | link                                              |
|----------|--------------------------------------------|---------------------------------------------------|
| NNPDF4.0 | master vs. trvl-mask-layers sequential CPU | https://vp.nnpdf.science/I6rdrAFsQ6Gah3fJ1Vf2rQ== |
| NNPDF4.0 | master vs. trvl-mask-layers parallel CPU   | https://vp.nnpdf.science/IX_nDzOGSfaclCah74K-rQ== |
| NNPDF4.0 | master vs. trvl-mask-layers parallel GPU   | https://vp.nnpdf.science/3JOhq50GRqizFGwZPdRfUQ== |

results for feature scaling and flavour basis will be uploaded soon

goord:
TODO: the observation data rotation is currently not working correctly, because it is applying a masked rotation matrix after the masking layer. This should change to applying the unmasked rotation to the observation data before the masking.

scarlehoff:
When you try to run DIS_diagonal_l2reg_example.yml it might be that not all datasets declared there exist anymore. You might need to remove some of them to test the runcard (as mentioned earlier today... it would be good to have all examples tested...)

goord:
Refactored the LossInvCovmat to handle the diagonal case in an optimized way. I get no significant numerical changes after 4000 epochs.

I was running the black tool, but I get a whole bunch of files I didn't touch that need reformatting... Should I ignore those?

RoyStegeman:
> I was running the black tool, but I get a whole bunch of files I didn't touch that need reformatting... Should I ignore those? 

Yes please limit the changes with black to the files you've made changes to

APJansen:
This is an easy way to do so: `black $(git diff --name-only master | grep .py)`

RoyStegeman:
You could also consider using [pre-commit](https://pre-commit.com/) to run black (and whatever else you may want) before each commit. Just so you don't need to remember to do so yourself

n3fit/src/n3fit/performfit.py
```diff
@@ -159,31 +159,31 @@ def performfit(
...
+        # all_experiments = copy.deepcopy(replica_experiments[0])
+        # n_experiments dicts
+        # for i_exp in range(len(all_experiments)):
+        #    training_data = []
+        #    validation_data = []
+        #    for i_rep in range(n_models):
+        #        training_data.append(replica_experiments[i_rep][i_exp]['expdata'])
+        #        validation_data.append(replica_experiments[i_rep][i_exp]['expdata_vl'])
+        #    all_experiments[i_exp]['expdata'] = np.concatenate(training_data, axis=0)
+        #    all_experiments[i_exp]['expdata_vl'] = np.concatenate(validation_data, axis=0)
```
APJansen, line 171:
```suggestion
```

This can probably be removed?

n3fit/src/n3fit/checks.py
```diff
@@ -362,11 +362,11 @@ def check_consistent_parallel(parameters, parallel_models, same_trvl_per_replica
...
-    if not same_trvl_per_replica:
-        raise CheckError(
-            "Replicas cannot be run in parallel with different training/validation "
-            " masks, please set `same_trvl_per_replica` to True in the runcard"
-        )
+    #    if not same_trvl_per_replica:
+    #        raise CheckError(
+    #            "Replicas cannot be run in parallel with different training/validation "
+    #            " masks, please set `same_trvl_per_replica` to True in the runcard"
+    #        )
```
APJansen, line 369:
```suggestion
```

This as well

goord:
Rebased squashed commit: 251d06c5bd581ad12e1b8b3c25a197f6df7b3f6c

scarlehoff:
I think the rebase / squash went wrong in some way...

In particular, I think it would be much easier to squash then rebase. You might want to squash groups of commits together if you want to keep some part of the work organised in separated commit.

Btw, the point about squashing is that it can make the rebase much easier in these cases because you need to fix the conflicts only once (even if they are bigger), but if the rebase works without doing anything there's no need to squash.

scarlehoff:
Yes. Another possibility is to cherry-pick the changes you need on top of current master. That's in my experience more work, but sometimes a bit easier to do.

goord:
It seems the rebase itself went ok, it's just the squash I did afterward that messed up everything. So I will do a hard push of the rebased commits

scarlehoff:
ok! Keep a local copy of the state of the repository just in case !!!

APJansen:
You need to rebase on master again, there was an issue with conda in the CI.

n3fit/src/n3fit/model_trainer.py
```diff
@@ -268,7 +267,7 @@ def _fill_the_dictionaries(self):
             - ``name``: names of the experiment
             - ``ndata``: number of experimental points
         """
-        for exp_dict in self.exp_info:
+        for index, exp_dict in enumerate(self.exp_info[0]):
```
scarlehoff, line 270: Why do you need the index and the enumerate here?

n3fit/src/n3fit/model_trainer.py
```diff
@@ -524,11 +524,25 @@ def _generate_observables(
...
             if not self.mode_hyperopt:
                 log.info("Generating layers for experiment %s", exp_dict["name"])
 
-            exp_layer = model_gen.observable_generator(exp_dict)
+            # Stacked tr-vl mask array for all replicas for this dataset
+            replica_masks = np.stack([e[index]["trmask"] for e in self.exp_info])
+            training_data = np.stack([e[index]["expdata"].flatten() for e in self.exp_info])
+            validation_data = np.stack([e[index]["expdata_vl"].flatten() for e in self.exp_info])
+            invcovmat = np.stack([e[index]["invcovmat"] for e in self.exp_info])
+            invcovmat_vl = np.stack([e[index]["invcovmat_vl"] for e in self.exp_info])
```
scarlehoff, line 536: Please avoid these list comprehensions. Since they are all loops over self.exp_info which is also the outside loop. It's really not clear what is happening here.

n3fit/src/n3fit/model_trainer.py
```diff
@@ -549,8 +563,18 @@ def _generate_observables(
             pos_initial, pos_multiplier = _LM_initial_and_multiplier(
                 all_pos_initial, all_pos_multiplier, max_lambda, positivity_steps
             )
+            replica_masks = np.stack([pos_dict["trmask"] for i in range(len(self.exp_info))])
+            training_data = np.stack(
+                [pos_dict["expdata"].flatten() for i in range(len(self.exp_info))]
+            )
```
scarlehoff, line 569: Please avoid list comprehensions.

n3fit/src/n3fit/stopping.py
```diff
@@ -384,7 +384,7 @@ def e_best_chi2(self):
     @property
     def stop_epoch(self):
         """Epoch in which the fit is stopped"""
-        return self._history.final_epoch + 1
+        return 0 if self._history.final_epoch is None else self._history.final_epoch + 1
```
scarlehoff, line 387: Why the 0?
Wouldn't  it be more correct the final + 1 regardless?

validphys2/src/validphys/config.py
```diff
@@ -1286,7 +1290,7 @@ def produce_rules(
 
         if added_filter_rules:
             for i, rule in enumerate(added_filter_rules):
-                if not isinstance(rule, dict):
+                if not (isinstance(rule, dict) or isinstance(rule, frozendict)):
```
scarlehoff, line 1293:
```suggestion
                if not isinstance(rule, (dict, frozendict)):
```

validphys2/src/validphys/n3fit_data.py
```diff
@@ -98,8 +101,10 @@ def tr_masks(data, replica_trvlseed):
         trmax = int(ndata * frac)
         if trmax == 0:
             # If that number is 0, then get 1 point with probability frac
-            trmax = int(rng.random() < frac)
-        mask = np.concatenate([np.ones(trmax, dtype=bool), np.zeros(ndata - trmax, dtype=bool)])
+            trmax = int(rng.random() < frac) if not parallel_models else 1
```
scarlehoff, line 104: This needs to either be fix or maybe add a warning for when running in parallel.

validphys2/src/validphys/n3fit_data_utils.py
```diff
@@ -8,6 +8,7 @@
 loading their fktables (and applying any necessary cuts).
 """
 import dataclasses
+import functools
```
scarlehoff, line 11:
```suggestion
```

validphys2/src/validphys/utils.py
```diff
@@ -5,14 +5,55 @@
...
 from validobj import ValidationError, parse_input
 
 
+# Since typing.Hashable doesn't check recursively you actually
+# have to try hashing it.
+def is_hashable(obj):
+    try:
+        hash(obj)
+        return True
+    except:
```
scarlehoff, line 25: Please, be explicit with the exceptions that you are expecting to catch.

validphys2/src/validphys/utils.py
```diff
@@ -5,14 +5,55 @@
...
+    # So that we don't infinitely recurse since frozenset and tuples
+    # are Sequences.
+    if is_hashable(obj):
+        return obj
+    elif isinstance(obj, Mapping):
+        return frozendict(obj)
+    elif isinstance(obj, Sequence):
+        return tuple([immute(i) for i in obj])
+    else:
+        raise ValueError("Object is not hashable")
```
scarlehoff, line 39:
```suggestion
    raise ValueError("Object is not hashable")
```

scarlehoff (commented):
Added a few comments.

Honestly, all the changes to validphys (all the freezing) I would rather remove from this PR. Could we do without it?

I'm afraid of possible secondary effects and would prefer to test and think about that part by itself.

n3fit/src/n3fit/layers/losses.py
```diff
@@ -37,12 +38,11 @@ class LossInvcovmat(MetaLayer):
     True
     """
 
-    def __init__(self, invcovmat, y_true, mask=None, covmat=None, **kwargs):
+    def __init__(self, invcovmat, y_true, mask=None, covmat=None, diag=False, **kwargs):
```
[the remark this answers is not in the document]
goord: Ah ok I thought the general idea was to have the covmat stored as a vector for optimization purposes. If we decide to have it as a diagonal matrix (I have no personal preference), the thing needs to be refactored.

n3fit/src/n3fit/model_trainer.py
goord: Because essentially we need to transpose the whole data structure of exp_info to stack the replica experimental masked data. I agree it feels ad-hoc inside the loop. I can create a new data structure and fill it outside the loop.

validphys2/src/validphys/n3fit_data.py
goord: You mean add a warning whenever we are running parallel replicas and have single-point datasets?

goord:
> Added a few comments.
> 
> Honestly, all the changes to validphys (all the freezing) I would rather remove from this PR. Could we do without it?
> 
> I'm afraid of possible secondary effects and would prefer to test and think about that part by itself.

I have squashed all validphys lru cache commits in a single commit that we can move to a separate branch.

n3fit/src/n3fit/layers/losses.py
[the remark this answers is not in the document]
scarlehoff: No, the diagonalization is to move the data to a basis in which there are no correlations between the datapoints. Nothing to do with performance.

n3fit/src/n3fit/model_trainer.py
scarlehoff: Maybe I've skipped something, but it is not being used, is it?

validphys2/src/validphys/n3fit_data.py
scarlehoff: Yes. Something like "putting {dstaset name} in the training set for all replicas" for the first replica.

But not a problem right now. You can leave these warnings for the end.

n3fit/src/n3fit/layers/losses.py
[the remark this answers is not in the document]
goord: https://github.com/NNPDF/nnpdf/blob/200e8f7f744dbb53f2e3aa87c293ed61a4775207/validphys2/src/validphys/n3fit_data.py#L259C1-L264C39

This piece of code stores the covariance matrices as 1-D arrays in the diagonal case. Should we make those regular diagonal matrices?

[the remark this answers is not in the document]
scarlehoff: If you need it to make the code more general later on, yes.
In practice it gets converted into a diagonal 2D matrix before being used for the first time.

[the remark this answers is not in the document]
goord: The diagonal case has been simplified in 680cd8bd89e4a3a9b1d81ff00253cb6f7981125e. The zero-padding was done in the model_gen.py to keep the loss class clean.

n3fit/src/n3fit/model_trainer.py
goord: The exp_info object is a list over experiments, where each entry is a list over replicas, where each entry is a dictionary with strings as keys ("expdata, "expdata_vl", ...) and numpy arrays as values. For the multi-replica observables, we need to collect the replica arrays and stack them.

I will create a separate loop to do this.

scarlehoff: No, I guess the loop cannot be avoided. I meant that it would be much clearer to have an outside loop filling in these lists instead of a list comprehension so that it is clear that they are all the same information (and that it is a double loop on self.exp_info)

At this point the performance is not important, better to do a few extra steps so everything is clearer. The performance is relevant later on the fit.

goord: I fixed it in  de3b55c

n3fit/src/n3fit/stopping.py
goord: I don't exactly recall the issue I ran into here. But to me it seems 0 allows the user to spot that the final epoch was not even created. Maybe we shouldn't fix this at all, it shouldn't happen

n3fit/src/n3fit/layers/mask.py
```diff
@@ -19,29 +21,39 @@ class Mask(MetaLayer):
         c: float
             constant multiplier for every output
         axis: int
-            axis in which to apply the mask
+            axis in which to apply the mask. Currently,
+            only the last axis gives the correct output shape
```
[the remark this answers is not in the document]
goord: Fixed in c07f301cce1ce994fbb2cfcc5d927dc34fb94b75

n3fit/src/n3fit/layers/mask.py
```diff
@@ -19,29 +21,39 @@ class Mask(MetaLayer):
...
     def call(self, ret):
         if self.mask is not None:
-            ret = op.boolean_mask(ret, self.mask, axis=self.axis)
+            flat_res = op.boolean_mask(ret, self.mask, axis=self.axis)
+            output_shape = [-1 if d is None else d for d in ret.get_shape()]
+            # Make sure reshape will succeed: set the last dimension to the unmasked data length and before-last to
+            # the number of replicas
+            output_shape[-1] = self.last_dim
+            output_shape[-2] = self.mask.shape[-2]
+            ret = op.reshape(flat_res, shape=output_shape)
```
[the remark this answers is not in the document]
goord: Fixedin c07f301cce1ce994fbb2cfcc5d927dc34fb94b75

n3fit/src/n3fit/stopping.py
goord: Done in 323be1f5702b6533ec249376664aba5243ac3741

validphys2/src/validphys/n3fit_data.py
goord: Done in 522184dbe6bc12d7d8215eaefbc68cffdfe520ca. Also, when parallel models is set true but the no. replicas is one, the single-point datasets are again randomly distributed among training and validation sets.

goord:
The lru cache changes are now in a separate PR #1945

n3fit/src/n3fit/model_trainer.py
APJansen: Done in [df26706](https://github.com/NNPDF/nnpdf/pull/1788/commits/df267066a0742a1d89ff708dc3b6e4b0685b3fe6)

n3fit/src/n3fit/model_trainer.py
APJansen: Done in [febbfd4](https://github.com/NNPDF/nnpdf/pull/1788/commits/febbfd44a7b514469e0b14acc71381e01704e02f)

APJansen:
I'll take over to fix the last issues. I've fixed a few more comments, rebased on master (trivially), and removed the commit for the hashing.

I think all open comments have been resolved, but since I didn't do them, please check if you agree @scarlehoff. (I just resolved the ones dealing with the hashing as that's its own PR now)

APJansen:
Only slightly nontrivial change I made, in [6bd0fb6](https://github.com/NNPDF/nnpdf/pull/1788/commits/6bd0fb6c1765d0f1d730c345db42585dacf27d13):

`Mask` allows for different shapes of the mask, but it seems this is always the same, `(replicas, features)`, so I removed this flexibility. Also, the test was broken, because it was being tested with an input shape `(features,)`, while in reality as far as I can see it is only ever called with an input shape `(1, replicas, features)`.

APJansen:
Is this ready for the redo-regressions?

scarlehoff:
No, we asked for a lot of stuff to be changed. I'd need to redo the review.

n3fit/src/n3fit/layers/losses.py
```diff
@@ -85,10 +81,17 @@ def call(self, y_pred, **kwargs):
...
+                res = op.tensor_product(tmp[0, :, :], right_dot, axes=1)
+            else:
+                right_dot = op.tensor_product(self.kernel, tmp[0, 0, :], axes=1)
+                res = op.tensor_product(tmp[0, :, :], right_dot, axes=1)
         else:
-            res = op.einsum("bri, ij, brj -> r", tmp, self.kernel, tmp)
+            if len(self.kernel.shape) == 3:
+                res = op.einsum("bri, rij, brj -> r", tmp, self.kernel, tmp)
+            else:
+                res = op.einsum("bri, ij, brj -> r", tmp, self.kernel, tmp)
```
scarlehoff, line 94:
```suggestion
            if len(self.kernel.shape) == 3:
                einstr = "bri, rij, brj -> r"
            elif len(self.kernel.shape) == 2:
                einstr = "bri, ij, brj -> r"
            res = op.einsum(einstr, tmp, self.kernel, tmp)
```

n3fit/src/n3fit/model_gen.py
```diff
@@ -95,10 +111,15 @@ def _generate_experimental_layer(self, pdf):
         else:
             output_layers = [obs(pdf) for obs in self.observables]
 
-        # Finally concatenate all observables (so that experiments are one single entitiy)
-        ret = op.concatenate(output_layers, axis=2)
+        # Finally concatenate all observables (so that experiments are one single entity)
+        ret = op.concatenate(output_layers)
```
scarlehoff, line 115:
```suggestion
        ret = op.concatenate(output_layers, axis=-1)
```

n3fit/src/n3fit/model_trainer.py
```diff
@@ -143,15 +142,13 @@ def __init__(
                 list with the replicas ids to be fitted
         """
         # Save all input information
-        self.exp_info = exp_info
-        if pos_info is None:
-            pos_info = []
+        self.exp_info = list(exp_info)
```
scarlehoff, line 145: Why is the `pos_info is None` conditional removed from here? It should be as possible as `integ_info` to get it as `None`

(actually maybe it should be, the code below,

```python
self.all_info = self.exp_info[0]
if pos_info is not None:
    self.all_info += pos_info
if integ_info is not None:
    self.all_info += integ_info

```

n3fit/src/n3fit/model_trainer.py
```diff
@@ -523,12 +522,43 @@ def _generate_observables(
...
+
+        # Loop over datasets
+        for i in range(len(self.exp_info[0])):
+            # Loop over data fields
+            for key in experiment_data.keys():
+                replica_data = []
+                # Loop over replicas
+                for replica in self.exp_info:
+                    if key in ["expdata", "expdata_vl"]:
+                        replica_data.append(replica[i][key].flatten())
```
scarlehoff, line 542: why do we need to flatten only expdata and exdata_vl ?

n3fit/src/n3fit/model_trainer.py
```diff
@@ -523,12 +522,43 @@ def _generate_observables(
...
+            "expdata": [],
+            "expdata_vl": [],
+            "invcovmat": [],
+            "invcovmat_vl": [],
+        }
+
+        # Loop over datasets
+        for i in range(len(self.exp_info[0])):
+            # Loop over data fields
+            for key in experiment_data.keys():
```
scarlehoff, line 537: Since you are using below `experiment_data.keys()` maybe it is better

```python
# Fill all experiment information for all datasets for all replicas.
for key, info in experiment_data.items()
```

n3fit/src/n3fit/model_trainer.py
```diff
@@ -523,12 +522,43 @@ def _generate_observables(
...
+                    if key in ["expdata", "expdata_vl"]:
+                        replica_data.append(replica[i][key].flatten())
+                    else:
+                        replica_data.append(replica[i][key])
+                # Stack
+                experiment_data[key].append(np.stack(replica_data))
+
         # Now we need to loop over all dictionaries (First exp_info, then pos_info and integ_info)
-        for exp_dict in self.exp_info:
+        for i in range(len(self.exp_info[0])):
```
scarlehoff, line 549:
```suggestion
        for i, exp_dict in enumerate(self.exp_info[0]):
```

Since you are using explicitly `self.exp_info[0][i]` below.

n3fit/src/n3fit/performfit.py
```diff
@@ -163,29 +163,21 @@ def performfit(
...
-        replicas_info = [(replicas, all_experiments, nnseeds)]
+        replicas_info = [(replicas, replica_experiments, nnseeds)]
     else:
+        # Cases 1 and 2 above are a special case of 3 where the replica idx and the seed should
+        # be a list of just one element
         replicas_info = replicas_nnseed_fitting_data_dict
+        for i, info_tuple in enumerate(replicas_info):
+            replica_idxs = info_tuple[0]
+            nnseeds = info_tuple[2]
+            replicas_info[i] = (tuple([replica_idxs]), [info_tuple[1]], tuple([nnseeds]))
```
scarlehoff, line 177: I'm very confused about this. Why does element 1 not get its own name and doesn't become a tuple?

So these are all lists of 1 elements, but you have one list per replica I guess? Why is not coming everything as a list? Usually it should be like that if the other two are lists of many elements.

validphys2/src/validphys/n3fit_data.py
```diff
@@ -74,14 +74,17 @@ def __iter__(self):
...
-def tr_masks(data, replica_trvlseed):
+def tr_masks(data, replica_trvlseed, parallel_models=False, replica=1, replicas=(1,)):
     """Generate the boolean masks used to split data into training and
     validation points. Returns a list of 1-D boolean arrays, one for each
     dataset. Each array has length equal to N_data, the datapoints which
     will be included in the training are ``True`` such that
 
         tr_data = data[tr_mask]
 
+    The single_datapoints_toss flag signals whether single-point datasets
```
scarlehoff, line 85: I think this was never implemented?

validphys2/src/validphys/n3fit_data.py
```diff
@@ -97,9 +100,17 @@ def tr_masks(data, replica_trvlseed):
         # We do this so that a given dataset will always have the same number of points masked
         trmax = int(ndata * frac)
         if trmax == 0:
-            # If that number is 0, then get 1 point with probability frac
-            trmax = int(rng.random() < frac)
-        mask = np.concatenate([np.ones(trmax, dtype=bool), np.zeros(ndata - trmax, dtype=bool)])
+            if parallel_models and len(replicas) > 1:
```
scarlehoff, line 103:
```suggestion
            if parallel_models:
```

I'd do it like this for parallel models regardless of the number of parallel models, otherwise reproducibility becomes (even more) complicated.

n3fit/src/n3fit/model_trainer.py
APJansen: Fixed in [2bbf864](https://github.com/NNPDF/nnpdf/pull/1788/commits/2bbf86457057227ecba60e6d8da9473f2f1b3279)

n3fit/src/n3fit/layers/losses.py
```diff
@@ -85,10 +81,17 @@ def call(self, y_pred, **kwargs):
...
         if tmp.shape[1] == 1:
             # einsum is not well suited for CPU, so use tensordot if not multimodel
-            right_dot = op.tensor_product(self.kernel, tmp[0, 0, :], axes=1)
-            res = op.tensor_product(tmp[0, :, :], right_dot, axes=1)
+            if len(self.kernel.shape) == 3:
+                right_dot = op.tensor_product(self.kernel[0, ...], tmp[0, 0, :], axes=1)
+                res = op.tensor_product(tmp[0, :, :], right_dot, axes=1)
+            else:
+                right_dot = op.tensor_product(self.kernel, tmp[0, 0, :], axes=1)
+                res = op.tensor_product(tmp[0, :, :], right_dot, axes=1)
```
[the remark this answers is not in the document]
APJansen: squeezing seems dangerous, it's not possible to squeeze a specified axis without having it error if the size of that axis is not 1, so you'd have to squeeze the whole thing. This could go wrong if there is only one data point.
I simplified it though in [1d91484](https://github.com/NNPDF/nnpdf/pull/1788/commits/1d9148403d3314d86af9a7f01386687ef52b357b)

validphys2/src/validphys/n3fit_data.py
APJansen: Can confirm that this doesn't appear anywhere else, but I'm not sure what this is about, if the comment can just be removed or if something more needs to be implemented.

n3fit/src/n3fit/model_trainer.py
APJansen: Not sure what you mean here? At this point the values of `experiment_data` are just empty lists, and they are not used?

APJansen:
I've fixed what I could, but I also find anything related to the preprocessing of data very confusing.

n3fit/src/n3fit/layers/losses.py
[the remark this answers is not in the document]
scarlehoff: When you get there you are sure that it is the 1 replica case. Actually, this is not the dense but the loss, shouldn't it always have the replica axis? (thanks to what you did in the multidense)?

n3fit/src/n3fit/model_trainer.py
scarlehoff: That you can do `for i, e in experiment_info.itemd()` and then below do `e.append`

APJansen: done in [a4026e1](https://github.com/NNPDF/nnpdf/pull/1788/commits/a4026e1cd9cd310907de07b2d4f54ec86f2a668e)

n3fit/src/n3fit/layers/losses.py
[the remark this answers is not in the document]
APJansen: Just checked this, it is only the experimental losses that do not have a replica axis (and indeed they shouldn't).

Just adding a length 1 replica axis doesn't work, since it won't match the PDF replica axis, i.e. the einsum `bri, rij, brj -> r` still doesn't work. And copying seems silly and wasteful.

How about just a comment to explain this instead? And perhaps in #1855 we can branch not on the shape of the kernel but on the type of loss (experimental vs train/val)?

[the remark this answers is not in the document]
APJansen: done in [49904a3](https://github.com/NNPDF/nnpdf/pull/1788/commits/49904a35885a48b395ab35fdf95a9cee786e4123)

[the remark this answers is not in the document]
scarlehoff: But couldn't we just do `[invcovmat]` for the experimental loss?

That way `right_dot = op.tensor_product(self.kernel[0, ...], tmp[0, 0, :], axes=1)` would work just the same (and similarly for the einsum)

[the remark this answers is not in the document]
APJansen: For the `tensor_product` it would work because in that case `1 == 1`, but for the einsum case it wouldn't because `r != 1`, all instances of `r` in the einexpr  `bri, rij, brj -> r` must be the same.

n3fit/src/n3fit/model_trainer.py
goord: This must have been an incorrect refactoring or rebase, we should indeed check for pos_info to be not None, and I think the cleanest way is to assign empty lists to the integ_info and pos_info class members whenever these arguments are None, and remove all the checks for None in the code below

goord: Done in rev. db24f63

n3fit/src/n3fit/model_trainer.py
goord: Because for reasons beyond my understanding the expdata and expdata_vl are actually 2D arrays with a single row. Things would be simpler if the trvl mask were to be applied to the observation data too in the loss function, but that can be saved for a later iteration imo.

APJansen:
I noticed I introduced a bug in [49904a3](https://github.com/NNPDF/nnpdf/pull/1788/commits/49904a35885a48b395ab35fdf95a9cee786e4123), got the number of replicas from the batch axis, just pushed a fix.

n3fit/src/n3fit/performfit.py
goord: I agree it's not very readable code @scarlehoff, I made the filling of the replica_info data structure  much more straightforward in commit 66f10689589342177a9a33e621f1318710ed11d1

APJansen: I thought it was still a bit hard to read, I simplified it further in [bc76939](https://github.com/NNPDF/nnpdf/pull/1788/commits/bc76939f2ec25891ac07e3a35fa5d662c35be7d4).

APJansen:
Is this ready now for the redoing of the regressions (once tests pass)? Would be great if we can get this merged, and perhaps even #1936, so that we can turn on some faster hyperopt runs over the weekend to get some more data.

n3fit/src/n3fit/model_trainer.py
```diff
@@ -523,12 +518,43 @@ def _generate_observables(
...
+            # Loop over data fields
+            for key, value in experiment_data.items():
+                replica_data = []
+                # Loop over replicas
+                for replica in self.exp_info:
+                    if key in ["expdata", "expdata_vl"]:
+                        replica_data.append(replica[i][key].flatten())
+                    else:
+                        replica_data.append(replica[i][key])
```
scarlehoff, line 540:
```suggestion
                    replica_data.append(replica[i][key].flatten())
```

Better to flatten the 1-d array as well

scarlehoff (approved):
Just a final comment from me :P

n3fit/src/n3fit/model_trainer.py
APJansen: I just checked, this doesn't work (shape mismatch of 5776(=76*76) vs 76 in DEUTERON), so it's not a 1d array.

scarlehoff: Uh, why is the training data ndata x ndata ? That doesn't look right.

scarlehoff: I just checked and all the quantities are (1, ndata)

scarlehoff: ah, all the quantities that are data ofc, let me change the suggestion to a comment then

APJansen:
@scarlehoff About the flattening, indeed I guess it was the covmat? You wrote that you would change your suggestion to a comment, but then resolved it, should I add a comment there or not?

n3fit/src/n3fit/model_trainer.py
```diff
@@ -523,12 +518,43 @@ def _generate_observables(
...
+
+        # Loop over datasets
+        for i in range(len(self.exp_info[0])):
+            # Loop over data fields
+            for key, value in experiment_data.items():
+                replica_data = []
+                # Loop over replicas
+                for replica in self.exp_info:
+                    if key in ["expdata", "expdata_vl"]:
+                        replica_data.append(replica[i][key].flatten())
```
scarlehoff, line 538:
```suggestion
                        # Save the data with shape (ndata) instead of (1, ndata)
                        replica_data.append(replica[i][key][0])
```

sorry, was in meeting!

PR #1790, opened 2023-08-11
Raise an exception if after many tries the replicas are still negative

### LINKED ISSUE #1783: crash or warn after many iterations generating negative pseudodata
We have a known issue in the pseudodata generation, if the data generated is (always) negative the pseudodata generation gets stuck in an infinite loop.
In general this means there's a bug either in the commondata or the covmat but I think it would be a good idea to warn the user about a possible cause for why the code looks like its frozen while it's taking 100% of a CPU core.

It might also be a good idea to make the code crash after some number of tries (that can be absurdly large). I think after 10^6 we can safely crash?

This is somewhat related to #1556

scarlehoff:
I've empirically chosen the very conservative `1e6` as the number of tries because in my laptop the worst group takes about 5 minutes. I think it can safely be lowered but I preferred to start high.

Fixes #1783 

This is how an error would look like:
```bash
[INFO]: Loading positivity dataset POSXSQ                                                                                                                                                                                                                                                                                   
[INFO]: Loading positivity dataset POSXSB                                                                                                                                                                                                                                                                                   
[INFO]: Loading positivity dataset POSXGL                                                                                                                                                                                                                                                                                   
[ERROR]: Error generating replicas for CDFZRAP_NEW                                                                                                                                                                                                                                                                          
[CRITICAL]: Bug in n3fit ocurred. Please report it.                                                                                                                                                                                                                                                                         
Traceback (most recent call last):                                                                                                                                                                                                                                                                                          
 <etc>                                                                                                                                                                                                                                                                                                 
validphys.pseudodata.ReplicaGenerationError: No valid replica found after 1000000 attempts
```

### DIFF
--- a/validphys2/src/validphys/pseudodata.py
+++ b/validphys2/src/validphys/pseudodata.py
@@ -149,7 +149,8 @@ def make_replica(
 
     max_tries: int
         The stochastic nature of replica generation means one can obtain (unphysical) negative predictions.
-        If after max_tries (default=1e6) no physical configuration is found, the code will raise an Exception
+        If after max_tries (default=1e6) no physical configuration is found,
+        it will raise a :py:class:`ReplicaGenerationError`
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


validphys2/src/validphys/pseudodata.py
```diff
@@ -142,6 +147,10 @@ def make_replica(
     genrep: bool
         Specifies whether computing replicas or not
 
+    max_tries: int
+        The stochastic nature of replica generation means one can obtain (unphysical) negative predictions.
+        If after max_tries (default=1e6) no physical configuration is found, the code will raise an Exception
```
Zaharid, line 152:
```suggestion
        If after max_tries (default=1e6) no physical configuration is found, the code will raise a 
        :py:class:`ReplicaGenerationError`
```

Zaharid (commented):
LGTM.

Could we transform the `for..else` thing into an early return? It saves some indetnation and avoids a relatively obscure construct.

scarlehoff:
If you're fine with the last update I'll merge it.

Zaharid (approved):

PR #1792, opened 2023-08-15
Refactor stopping

APJansen:
This simplifies the `stopping.py` module, reducing the number of classes and the coupling to the outside. It also rewrites the `WriterWrapper`. 

I started looking at this because  #1782 requires changes here, but the changes in this PR are purely a refactor that should do exactly the same. 

## Tests
I test this by running on this branch, renaming the output folder, running on master, and comparing the two. We only expect changes in the timing and versions.
I've tested with:
- [x] `n3fit Basic_runcard_modified.yml 1 -r 5`
- [x] `n3fit Basic_runcard_modified.yml 4 -r 5`
- [x] `n3fit Basic_runcard_modified.yml 3`

In all cases the only differences are the versions and the timings, the rest is identical.
Note I've modified the basic runcard to speed it up, increasing the learning rate so it stops quickly. I think it is still general enough for these tests, in the case with 5 replicas, the best epochs are 77, 72, 77, 72, 75, so not a degenerate case where coincidentally they're all equal or something.

## Comments
- `WriterWrapper` is a class with one method, that gets initialized and immediately after the method gets called, and that's all that happens with it. So it could just as well have been a function. I left it as a class though because it was easier and I'm not sure if everyone will agree.
- Similarly in the stopping module, there were 4 classes and it was quite hard to figure out which calls which etc. I removed one, `ReplicaState`, and moved a lot of the functionality from `FitHistory` to the main `Stopping` class. Now `FitHistory` is essentially nothing more than a list of the 4th class `FitState`, so perhaps I should remove that entirely, keeping only `Stopping` and `FitState`.
- Some of the timings are stored for each replica, but if multiple replicas are run these will be the overal timings. However when you run them one at a time it will be different of course. Also there is nothing better to report on a per-replica basis if run in parallel, so I don't think there is anything better to do and I just added a comment on it.
- In `WriterWrapper` I needed both the `pdf_model` itself and the `pdf_instance` which is the model wrapped by `N3PDF`. I got the former from the latter by accessing its attribute, but that is supposed to be private so need to clean that up.


APJansen:
Hi @scarlehoff, great, thanks for the extra info! The chi2's are saved as well though in `chi2exps.log`, so the history needs to stay in some form, but I don't think it's a problem. 
I think in principle it could be written entirely in terms of [Keras's early stopping callback](https://keras.io/api/callbacks/early_stopping/), was there a reason you didn't use this?
For now though I'll just review what I've done and maybe simplify it some more if possible, I think it's good enough and want to get on with the actual parallel replica stuff.

About the runcards to test, can I just take the basic runcard as the normal fit, and modify it in those 3 ways separately for the others? The first modification being `frac: 1.0` in the `datasets_inputs`, the second `stopping_patience: 1.0`, and the last I'm not sure, just removing the two entries under `posdatasets`?

RoyStegeman:
> I think in principle it could be written entirely in terms of [Keras's early stopping callback](https://keras.io/api/callbacks/early_stopping/), was there a reason you didn't use this?

That's a good question. Would this also work with the boolean positivity requirement?

> The first modification being `frac: 1.0` in the `datasets_inputs`, the second `stopping_patience: 1.0`

Indeed

> and the last I'm not sure, just removing the two entries under posdatasets

If I'm not mistaken the `posdatasets` key cannot be empty since it's [collect](https://docs.nnpdf.science/vp/collect.html)`ed over, so instead of removing it, you can probably put an empty list.

Since I'm somewhat out of the loop, what is the status of the different PRs? 
- This one (#1792) is in draft but from the comments it seems ready for review?
- #1661 can be closed in favour of #1788?
- #1788 is WIP?
- #1782 is also WIP?
- Doing hyperopt (#1726 ?) is the final target, but is waiting on the above ?

RoyStegeman:
Thanks for creating the issue, it's very clear. 

Sure, you can do everything at your own pace, it's your project :). I'll just wait till it's ready to review.

APJansen:
I think it's ready for review, except that the CI is failing, not sure what's going on there...
I did my own checks as discussed and described in detail [here](https://github.com/LHC-NLeSC/nnpdf-scripts/tree/main/tests/stopping), which all pass.

RoyStegeman:
Thanks! I don't think the failing  of the CI is related to this PR.

RoyStegeman:
@APJansen could you try rebasing this on master?

APJansen:
@RoyStegeman Thanks, that fixed it!

scarlehoff:
> I think in principle it could be written entirely in terms of [Keras's early stopping callback](https://keras.io/api/callbacks/early_stopping/), was there a reason you didn't use this?
That's a good question. Would this also work with the boolean positivity requirement?

There's also the secondary problem of using frameworks other than tensorflow for the PDFs (for instance, the quantum PDFs using qibo).

I'd avoid adding the tf dependency where is not explicitly needed. Unless there's a clear advantage in terms of performance (at the end of the day, 99% of the fits and papers are done with tensorflow), in that case it's probably fine to add some inconveniences on the side projects.

APJansen:
Sorry I forgot to add reviewers, is this ok to merge?

Radonirinaunimi:
> Sorry I forgot to add reviewers, is this ok to merge?

I have started having a look at this but still needs a couple of hours/a day.

scarlehoff:
I had a quick look and seems ok. I'll have to look in detail next week.

Did you test some edge cases like no patience / no positivity / infinite patience ?

Also, since you shuffled a lot the stopping module, please remove any "dangling methods" or properties (maybe you did already ofc)

n3fit/src/n3fit/io/writer.py
```diff
@@ -213,98 +218,156 @@
...
+        self.preprocessing = [
+            pdf_object.get_preprocessing_factors() for pdf_object in self.pdf_objects
+        ]
+        self.arc_lengths = [
+            vpinterface.compute_arclength(pdf_object).tolist() for pdf_object in self.pdf_objects
+        ]
+        self.integrability_numbers = [
+            vpinterface.integrability_numbers(pdf_object).tolist()
+            for pdf_object in self.pdf_objects
+        ]
```
Radonirinaunimi, line 276: I think these should be better handled under **one single loop**; it's basically a difference between $\mathcal{O}(n)$ and $\mathcal{O}(3n)$ in time complexity.

n3fit/src/n3fit/io/writer.py
```diff
@@ -213,98 +218,156 @@
...
+        json_dict = jsonfit(
+            best_epoch=self.stopping_object.best_epochs[i],
+            positivity_status=self.stopping_object.positivity_statusses[i],
+            preprocessing=self.preprocessing[i],
+            arc_lengths=self.arc_lengths[i],
+            integrability_numbers=self.integrability_numbers[i],
+            tr_chi2=self.tr_chi2[i],
+            vl_chi2=self.vl_chi2[i],
+            true_chi2=self.true_chi2[i],
+            # Note: last 2 same for all replicas, unless run separately
```
Radonirinaunimi, line 304: What does this comment exactly mean? These two does not necessarily have to be the same even if the replicas were fitted in parallel (if that's what you were referring to).

n3fit/src/n3fit/io/writer.py
```diff
@@ -213,98 +218,156 @@
...
-        json_dict = jsonfit(
-            replica_status, self.pdf_object, tr_chi2, vl_chi2, true_chi2, stop_epoch, self.timings
-        )
-        with open(output_file, "w", encoding="utf-8") as fs:
-            json.dump(json_dict, fs, indent=2, cls = SuperEncoder)
+    def _write_weights(self, i, out_path):
+        log.info(" > Saving the weights for future in %s", out_path)
+        # Extract model out of N3PDF
+        model = self.pdf_objects[i]._models[0]
+        # Need to use "str" here because TF 2.2 has a bug for paths objects (fixed in 2.3)
```
Radonirinaunimi, line 332: I think using a `PurePath` (pathlib) could also fix this.

n3fit/src/n3fit/stopping.py
```diff
@@ -438,49 +336,69 @@ def __init__(
...
+        self._counts = np.zeros(self._n_replicas, dtype=int)
+
+        self._dont_stop = dont_stop
+        self._stop_now = False
+        self._stopping_patience = stopping_patience
+        self._total_epochs = total_epochs
+
+        self._stop_epochs = [total_epochs - 1] * self._n_replicas
+        self._best_epochs = [None] * self._n_replicas
+        self.positivity_statusses = [POS_BAD] * self._n_replicas
```
Radonirinaunimi, line 364:
```suggestion
        self.positivity_statuses = [POS_BAD] * self._n_replicas
```
Also in the subsequent occurrences.

Radonirinaunimi (changes requested):
I did check the corner cases and in all instances the fits ran fine. I have requested very minor modifications which I think are worth addressing.

n3fit/src/n3fit/io/writer.py
APJansen: The "last 2" refers to the two lines below the comment, `timing` and `stop_epoch`. If running multiple replicas in parallel, these are the overall values over all replicas. It's not an issue, just thought it'd be useful to document here.
Shall I just add the word "below" to make it clear?

n3fit/src/n3fit/io/writer.py
APJansen: It shouldn't matter for performance, but it's maybe a bit cleaner, so I changed it.

APJansen:
@Radonirinaunimi I've addressed all your comments, and reran the comparison. @scarlehoff the tests I did are as we discussed, I've documented them [here](https://github.com/LHC-NLeSC/nnpdf-scripts/tree/main/tests/stopping). They all pass also after the last changes.

n3fit/src/n3fit/io/writer.py
Radonirinaunimi: Yes, adding "below" will definitely make it clearer.

Radonirinaunimi (approved):

scarlehoff (commented):
Thanks for this. I've left a few comments but they don't change anything fundamental.

n3fit/src/n3fit/io/writer.py
```diff
@@ -213,98 +219,155 @@
...
+        json_dict = jsonfit(
+            best_epoch=self.stopping_object.best_epochs[i],
+            positivity_status=self.stopping_object.positivity_statuses[i],
+            preprocessing=self.preprocessing[i],
+            arc_lengths=self.arc_lengths[i],
+            integrability_numbers=self.integrability_numbers[i],
+            tr_chi2=self.tr_chi2[i],
+            vl_chi2=self.vl_chi2[i],
+            true_chi2=self.true_chi2[i],
+            # Note: last 2 arguments below are the same for all replicas, unless run separately
```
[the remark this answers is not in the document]
APJansen: It won't be possible to stop individual replicas from training once they are merged into a single model, but of course storing the weights at the best epoch should be implemented at the individual replica level. When I implement that I'll also save the best epoch per replica.

n3fit/src/n3fit/stopping.py
```diff
@@ -438,49 +336,69 @@ def __init__(
...
     def stop_epoch(self):
-        """ Epoch in which the fit is stopped """
+        """Epoch in which the fit is stopped"""
         return self._history.final_epoch + 1
 
+    @property
+    def best_epochs(self):
+        """Epochs in which the best chi2 was found, or the last one if no best was found"""
+        best_or_last_epochs = self._best_epochs
+        for i_replica in np.where([be == None for be in self._best_epochs])[0]:
```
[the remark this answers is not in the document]
APJansen: Agreed, I removed it, thanks

n3fit/src/n3fit/io/writer.py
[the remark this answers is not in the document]
scarlehoff: The `stop_epoch` of each individual replica in that case is the replica at which the training effectively stop (it doesn't matter if internally it didn't).

(however, I'd say the training should be _somehow_ stopped, either masking the chi2 for that replica or swapping the actual values with a constant, so it doesn't bias the rest)

APJansen:
Thanks @scarlehoff, I implemented your comments, and checked that results are still identical. Is it ok to merge now?

PR #1793, opened 2023-08-17
Mamba ci

APJansen:
Because of the issues we have been having with the CI the last month or so, I had a look at using the mamba resolver within conda. In this branch I added this option to the matrix, so the original 4 tests are still there, along with 4 new copies that only differ in that they use the mamba resolver.

It reduces the conda setup step from ~15 minutes (or I've even seen 3 hours on a job from a few weeks ago.. ) to ~1 minute.
And it seems to fix the issue, all of them pass whereas with conda the python 3.9 ubuntu test still fails.

So I would propose to make that the only option

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
         python-version: ["3.9", "3.10"]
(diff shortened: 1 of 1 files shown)


Zaharid (approved):
Thanks for this. Seems to work.

PR #1795, opened 2023-08-18
don't print msr layer summary if msr not enforced

RoyStegeman:


APJansen:
Looks good to me, I think you need to merge master into this to make the tests pass.

APJansen (approved):

PR #1797, opened 2023-08-22
Try building with conda forge only

Zaharid:
Remove the defalts channel, so we can potentially get easier resolutions to the environment and avoid potential licence issues.


RoyStegeman:
You're free now!

PR #1801, opened 2023-08-25
Add `--lite` option to vp-comparefits for quick comparisons

### LINKED ISSUE #1553: Add a `--light` option to comparefits
Do we want this?

I have the feeling we already discussed that at some point?

scarlehoff:
The default remains the normal comparefit but this adds an option to have a quick comparison without the heavy stuff (i.e., without anything that involves doing the fktable convolution with all replicas).

This is quite useful when iterating overfit when you just want to see whether the cuts are the same, how the PDF/distances look etc.

Closes #1553

### DIFF
--- a/validphys2/src/validphys/comparefittemplates/__init__.py
+++ b/validphys2/src/validphys/comparefittemplates/__init__.py
@@ -1,3 +1,4 @@
 import pathlib
 
 template_path = pathlib.Path(__file__).with_name('comparecard.yaml')
+template_lite_path = pathlib.Path(__file__).with_name('comparecard_lite.yaml')
--- a/validphys2/src/validphys/comparefittemplates/comparecard_lite.yaml
+++ b/validphys2/src/validphys/comparefittemplates/comparecard_lite.yaml
@@ -60,27 +60,13 @@ PDFscalespecs:
   - xscale: linear
     Xscaletitle: Linear
 
-Energies:
-  - sqrts: 13000
-    Energytitle: "13 TeV"
-
 Distspecs:
   - ymin: 0
     ymax: 20
 
-pos_use_kin: True
-
-dataset_report:
-  meta: Null
-  template: data.md
-
 pdf_report:
   meta: Null
   template: pdf.md
-
-exponents_report:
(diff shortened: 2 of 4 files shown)


RoyStegeman:
How light is this? I think `summarise_fits` is quite expensive as it also involves calculating the chi2, not just reading info from stored data.

scarlehoff:
Yes. It took 7 minutes in my laptop. I thought about removing it as well but I think that one is needed.

RoyStegeman:
Did you also check memory? Since Mac apparently does this thing of efficiently using swap.

RoyStegeman:
I agree that the chi2 information would be useful to have, but I'm wondering if we want to add the option of only having the average chi2 over replicas? Though I'm not sure since doing that in vp is probably not going to be very clean/minimal. 

I don't think this is a feature I would end up using a lot anyway, so I don't have a very strong opinion about it, but I think the idea is mainly that it provides a quick way of getting a subset of the most relevant information of a vp-report while still being able to produce it on a laptop while getting a coffee? In that case it works well for Mac, but for most peoples Linux configurations it will probably still be too heavy on memory, 

Getting the memory down would be difficult as I think that would require making changes to reportengine, since at the moment I don't believe it has any garbage collector for objects inside the `ChainMap`.

RoyStegeman (approved):
If you want to merge it like this it's fine for me.

Radonirinaunimi:
Tested it on my PC and it finished in 8 minutes, this is definitely much better than the current one! The memory actually never exceeded **5GB** (this was on Arch).

Radonirinaunimi (approved):

scarlehoff:
> Getting the memory down would be difficult as I think that would require making changes to reportengine,

There's an intrinsic memory cost from having all fktables in memory. But we could be more mindful and actually drop the parts that we won't really use anywhere (for instance when performing cuts) but it's not super simple :(

scarlehoff:
Example report: https://vp.nnpdf.science/jv2WTEVCSYG_9AcGI656rA==

As soon as the tests are done I'll merge.

PR #1805, opened 2023-09-06
Test with less tests

scarlehoff:


RoyStegeman:
The tests fail because a different version of `rhash` is being installed. I think simply removing these tests is better than pinning a package for a test that we don't really need anyway.

scarlehoff:
I'll take that as an approval :P

PR #1806, opened 2023-09-18
Exa couplings2

niclaurenti:
This PR implements the alphaem exact running in the Alpha class in the validphys.photon module.
In this way it is consistent with the running used in `EKO` for QED.
Such running also considers the mixed terms between the RGEs of alphas and alphaem (therefore they are evolved together).
Edit: Since the solution must enter in the `fiatlux` integral, and the exact solution is slower than the truncated one, the strategy is to compute $\alpha$ for some reference values in the integration range and then interpolate with `scipy` (the function is continuous since the matching is applied to $\alpha_s$ only but probably not smooth in the matching points)
Edit n. 2: interpolating with `scipy` was still too slow so I interpolated with `numpy`, that is faster
Edit n. 3 (Theoretical issue): the Lux formula uses $\alpha(Q^2)$ to very low values of $Q^2$, below the Landau pole of $\alpha_s$. For this reason I had to remove the mixed terms in the evolution of $\alpha(Q^2)$, i.e. the terms that couple the evolutions of the two couplings. In this way $\alpha$ running is decoupled from $\alpha_s$ running.
It means that the evolution will be slightly inconsistent with the one used in `EKO` (but in the perturbative regions the difference is sub-permil).

I've performed a fit with this branch and it seems to be good:
- report:
https://vp.nnpdf.science/U27tsiK-TS6xejU0tw5trQ==
- Photon comparison with NNPDF40_ecc_qed:
https://vp.nnpdf.science/9gdZCX7wQdyScO6LrrfZbQ==/#PDFscalespecs0_Basespecs1_PDFnormalize1_plot_pdfs_photon
- Photon comparison with NNPDF31_ecc_qed:
https://vp.nnpdf.science/L2bZ5BABQsq4eKhftwwufg==/#PDFscalespecs0_Basespecs1_PDFnormalize1_plot_pdfs_photon


validphys2/src/validphys/photon/compute.py
```diff
@@ -276,13 +321,78 @@ def alpha_em_fixed_flavor(self, q, alpha_ref, qref, nf):
...
+        Returns
+        -------
+        alpha_em at NLO : float
+            target value of a
+        """
+        u = 2 * np.log(q / qref)
+
+        # integration kernel
+        def rge(_t, alpha, beta_qcd_vec, beta_qcd_mix, beta_qed_vec, beta_qed_mix):
+            "RGEs for the couplings. See Eqs. (5-6) of arXiv:hep-ph/9803211"
```
scarlehoff, line 357:
```suggestion
            """RGEs for the couplings. See Eqs. (5-6) of arXiv:hep-ph/9803211"""
```

validphys2/src/validphys/photon/compute.py
```diff
@@ -276,13 +321,78 @@ def alpha_em_fixed_flavor(self, q, alpha_ref, qref, nf):
...
+                -(alpha[1] ** 2)
+                * beta_qed_vec[0]
+                * (
+                    1
+                    + np.sum([alpha[1] ** (k + 1) * b for k, b in enumerate(beta_qed_vec[1:])])
+                    + alpha[0] * beta_qed_mix
+                )
+            )
+            res = np.array([rge_qcd, rge_qed])
+            return res
```
scarlehoff, line 377: Can this integration kernel be moved outside from the class since it doesn't use anything from `self.`?

validphys2/src/validphys/photon/compute.py
```diff
@@ -313,34 +412,67 @@ def set_thresholds_alpha_em(self):
...
+                vec_qed.append(beta.b_qed((0, ord + 2), nf) / (4 * np.pi) ** ord)
+            betas_qcd[nf] = vec_qcd
+            betas_qed[nf] = vec_qed
+            beta_mix_qcd[nf] = beta.b_qcd((2, 1), nf) / (4 * np.pi)
+            beta_mix_qed[nf] = beta.b_qed((1, 2), nf) / (4 * np.pi)
+        return betas_qcd, betas_qed, beta_mix_qcd, beta_mix_qed
+
+
+def apply_match(alphas, ord, L, match):
+    "applying matching conditions to alphas"
```
scarlehoff, line 473:
```suggestion
    """Apply the matching conditions to alpha_s"""
```

Please, add more information to the docstr on what the input is.

scarlehoff (commented):
Thanks for opening the new branch!

I've left a few comments. The general trend is that some more explanation on the docstr (and some comments in the parts where the code might get a bit confusing, specially when many things are happening in the background e.g. due to eko doing stuff) would be good.

validphys2/src/validphys/tests/photon/test_compute.py
```diff
@@ -58,29 +62,139 @@ def test_set_thresholds_alpha_em():
...
+
+def test_couplings_trn():
+    "benchmark the exact running of alpha with EKO"
+    test_theory = API.theoryid(theoryid=THEORY_QED)
+    theory = test_theory.get_description()
+    theory["ModEv"] = 'TRN'
+    alpha = Alpha(theory, 1e8)
+    coupl_ref = [theory["alphas"], theory["alphaqed"]]
+
+    for q in [80, 10, 5]:
```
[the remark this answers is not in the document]
niclaurenti: I would skip this since my fear is that some values of q would make the test fail. For example in `test_exa_interpolation` I had to manually increase the error until I found the one for which the tests were passing, but  maybe there are some values for which the error is larger.
Anyway, with the last commit this PR is done

niclaurenti:
Hi @scarlehoff and @Zaharid , for the reason that I explained in https://github.com/NNPDF/nnpdf/blob/d2ecc69f17f10239fd51e70d29a9adf4f057a5c7/validphys2/src/validphys/photon/compute.py#L500
I had to decouple the $\alpha$ evolution from the $\alpha_s$ evolution. Therefore, all the $\alpha_s$ part in the photon module is now unused. This will cause a little inconsistence with the `EKO` evolution but the differences are very small in the perturbative region of QCD.
So what I can do is the following:
- I can leave it implemented, with some parts that are commented out, so that if some day we will need it or we will find a solution we don't have to reimplement it (and moreover it is not slowing down the code).
- I can remove it
What do you think?

scarlehoff:
I'm in favor of the removal since if/when someone wants to deal for the problem either this would be the smallest part of the job (and they will have to recheck that everything is correct anyway) or (if it is you or someone already involved in the project) would be able to find the commit in which it was removed looking at this PR

(and past experiences tells me that dead code starts smelling after a while and ends up being removed anyway)

That said, if other people want to keep it I won't make a fuss about it.

niclaurenti:
> I'm in favor of the removal since if/when someone wants to deal with the problem either this would be the smallest part of the job (and they will have to recheck that everything is correct anyway) or (if it is you or someone already involved in the project) would be able to find the commit in which it was removed looking at this PR
> 
> (and past experiences tell me that dead code starts smelling after a while and ends up being removed anyway)
> 
> That said, if other people want to keep it I won't make a fuss about it.
> 
> edit: unless there's a plan already of dealing with this in the near future, but from your msg I understand this is something that won't be deal with any time soon

ok then I'll remove it

validphys2/src/validphys/tests/photon/test_compute.py
```diff
@@ -25,62 +29,182 @@ def generate_fiatlux_runcard():
...
-    np.testing.assert_equal(photon.additional_errors.name, "LUXqed17_plus_PDF4LHC15_nnlo_100")
-    np.testing.assert_equal(photon.luxseed, fiatlux_runcard["luxseed"])
-    np.testing.assert_equal(photon.path_to_eko_photon, test_theory.path / "eko_photon.tar")
-    np.testing.assert_equal(photon.q_in, 100.0)
+#     np.testing.assert_equal(photon.replicas, [1, 2, 3])
+#     np.testing.assert_equal(photon.luxpdfset._name, fiatlux_runcard["luxset"].name)
+#     np.testing.assert_equal(photon.additional_errors.name, "LUXqed17_plus_PDF4LHC15_nnlo_100")
+#     np.testing.assert_equal(photon.luxseed, fiatlux_runcard["luxseed"])
+#     np.testing.assert_equal(photon.path_to_eko_photon, test_theory.path / "eko_photon.tar")
+#     np.testing.assert_equal(photon.q_in, 100.0)
```
scarlehoff, line 47: Please remove all these commented lines.

validphys2/src/validphys/tests/photon/test_compute.py
```diff
@@ -94,77 +218,78 @@ def test_betas():
...
+#             # Apply EKO to PDFs
+#             for _, elem in eko.items():
+#                 pdfs_final = np.einsum("ajbk,bk", elem.operator, pdfs_init)
+
+#         photon_Q0 = pdfs_final[ph_id]
+#         photon_fiatlux = XGRID * photon_Q0
+
+#         photon_validphys = photon(XGRID[np.newaxis, :, np.newaxis])[0][0, :, 0]
+
+#         np.testing.assert_allclose(photon_fiatlux, photon_validphys, rtol=1e-7)
```
scarlehoff, line 295: Same as above.

scarlehoff (approved):
Are you running the qed fits already with this branch? (if so I guess everything works so feel free to merge)

validphys2/src/validphys/tests/photon/test_compute.py
niclaurenti: All the `#` were accidentally pushed :)
Fixed it

PR #1809, opened 2023-10-03
Miscelaneous bugfixes (matplotlib & futuretests)

### LINKED ISSUE #1808: Example Future Test does not work
The example [card](https://github.com/NNPDF/nnpdf/blob/master/validphys2/examples/future_test_example.yaml) for **future test** does not work because of a potential bug.
 
Here is the [log](https://data.nnpdf.science/logs/validphys-crash-la8wxjrv.txt).

scarlehoff:
This PR fixes #1808 

and on doing this I realized there were some problems with the latest versions of matplotlib.

1. We were using `ax._get_lines.prop_cycler` which is undocumented and removed in 3.7. I've changed it to the a bit-less-undocumented `color = ax._get_lines.get_next_color()`
2. The transformation to avoid having points on top of each other doesn't work in matplotlib 3.8, hence the pinning of the version in the conda recipe https://github.com/NNPDF/nnpdf/blob/f553482181a9d785a2386ae29c44efc765b74d7f/validphys2/src/validphys/plotutils.py#L289


RoyStegeman:
> The transformation to avoid having points on top of each other doesn't work in matplotlib 3.8, hence the pinning of the version in the conda recipe

Could you point me to where this transformation is? Is this something that will "solve itself" in later versions or does this mean we'll be stuck with matplotlib 3.8 unless we do something ourselves?

validphys2/src/validphys/deltachi2.py
```diff
@@ -219,9 +219,7 @@ def draw(self, pdf, grid, flstate):
 
         errorstd = stats.std_error()
         # color cycle iterable
-        pcycler = ax._get_lines.prop_cycler
-        next_prop = next(pcycler)
-        color = next_prop["color"]
+        color = ax._get_lines.get_next_color()
```
Radonirinaunimi, line 222: `get_next_color` is indeed the best way to iterate on the colors. 

Maybe for absolute consistency (and this is very minor and perhaps not necessary) the only leftover place in which this should be reflected is in the example: https://github.com/NNPDF/nnpdf/blob/f553482181a9d785a2386ae29c44efc765b74d7f/validphys2/examples/API_example.ipynb#L326-L327.

Radonirinaunimi (commented):
Thanks a lot, this indeed fixes the issue! I just have very minor points below. Apart from the `<3.8` part which I might need to take a closer look, this LGTM.

validphys2/src/validphys/checks.py
```diff
@@ -44,8 +44,7 @@ def check_pdf_is_montecarlo(ns, **kwargs):
 
 
 @make_argcheck
-def check_pdf_is_montecarlo_or_symmhessian(ns, **kwargs):
-    pdf = ns['pdf']
+def check_pdf_is_montecarlo_or_symmhessian(pdf, **kwargs):
     etype = pdf.error_type
```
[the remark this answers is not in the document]
scarlehoff: These lines might _need_ the fix as well actually. I only looked at the one that was failing.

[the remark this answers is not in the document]
scarlehoff: I think they don't because of the use of `make_check` instead of `make_argcheck`. The other check, which is newer, uses `make_argcheck` as recommended by reportengine.

So I'll leave them be (with `ns` and `make_check` for the time being...)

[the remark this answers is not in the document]
Radonirinaunimi: That is very likely the case indeed! I am happy with that.

conda-recipe/meta.yaml
```diff
@@ -38,7 +38,7 @@ requirements:
         - pkg-config
         - python # requirements for validphys
         - reportengine >=0.30.25 # see https://github.com/NNPDF/reportengine
-        - matplotlib >=3.3.0
+        - matplotlib >=3.3.0,<3.8
```
RoyStegeman, line 41: Please add an in-line comment pointing to this PR, or even open an issue

RoyStegeman (approved):

validphys2/src/validphys/checks.py
comane, line 48:
```suggestion
@make_check
def check_pdf_is_montecarlo_or_symmhessian(ns, **kwargs):
    pdf = ns['pdf']
    etype = pdf.error_type
```

validphys2/src/validphys/checks.py
```diff
@@ -44,8 +44,7 @@ def check_pdf_is_montecarlo(ns, **kwargs):
 
 
 @make_argcheck
```
comane, line 46:
```suggestion

```

conda-recipe/meta.yaml
scarlehoff:
```suggestion
        - matplotlib >=3.3.0,<3.8  # see https://github.com/NNPDF/nnpdf/pull/1809
```

validphys2/src/validphys/checks.py
```diff
@@ -44,8 +44,7 @@ def check_pdf_is_montecarlo(ns, **kwargs):
...
-def check_pdf_is_montecarlo_or_symmhessian(ns, **kwargs):
-    pdf = ns['pdf']
+def check_pdf_is_montecarlo_or_symmhessian(pdf, **kwargs):
     etype = pdf.error_type
     check(
         etype in {'replicas', 'symhessian'},
```
comane, line 50: Hi @scarlehoff, spotted another bug (my bad)
```suggestion
        etype in {'replicas', 'symmhessian'},
```

RoyStegeman:
image
 
Wait, this doesn't look quite right

validphys2/src/validphys/checks.py
scarlehoff: In reportengine it seems `make_argcheck` is preferred https://github.com/NNPDF/reportengine/blob/79eec2e33c5aab3998fd58e9c3cf84bd2c2696e7/src/reportengine/checks.py#L10

why should we use `make_check` and `ns` instead? It also a bit clearer with `arcgcheck` and `pdf` since that check is only active for PDFs in any case.

comane: Sure, I am ok with that.

Radonirinaunimi:
>     2. The transformation to avoid having points on top of each other doesn't work in matplotlib 3.8, hence the pinning of the version in the conda recipe 

As for this, the issue is more subtle given that [transform.ScaledTranslation](https://github.com/NNPDF/nnpdf/blob/f553482181a9d785a2386ae29c44efc765b74d7f/validphys2/src/validphys/plotutils.py#L298) is expected to have the same behavior as in previous versions, and indeed if one runs the [example](http://matplotlib.org/users/transforms_tutorial.html) in v3.8.0 one gets the expected results. So I am fine with excluding >=3.8.0 for the time being.

validphys2/src/validphys/checks.py
scarlehoff: But what's the reason for using one over the other?

scarlehoff:
I tested a bit and the problem is a bit complicated. Basically, while the transformation works ok in some cases, in others something fails internally in matplotlib and the axis are not updated, so the plot looks like this:

image

by "some cases" I mean some subsets of data.

The same plot, with 3.7 (instead of 3.8) looks like this:

image

(which is correct) I attempted to fix it but couldn't find what was the reason for the failure so for the time being I rather leave it at <3.8. Indeed, looking at the changelogs of matplotlib I could not find anything that would be broken.


> Wait, this doesn't look quite right

I'm using theory 399 in the notebook, which is a low precision theory. If you change it to 400 for instance or 200 it'd be better. But at least now the example runs. We can put something that runs and also looks ok, of course.

Radonirinaunimi (approved):

validphys2/src/validphys/checks.py
comane: Not sure tbh, I was suggesting it because the other functions like `check_pdf_is_montecarlo` make use of it and hence it would be more uniform, but I changed my mind.

I think you are right though that `make_check` should not be used anymore since it is weird that one can access the namespace `ns` from validphys.

I am thinking that we could change the others to be `make_argcheck` as well at this point.

PR #1698, opened 2023-03-21
N3LO fits with IHOU

giacomomagni:
This PR is to introduce IHOU + MHOU theory covmat and allow for N3LO theories.

PR #1810, opened 2023-10-04
nFONLL theory indices up to and including 721

andreab1997:
This is to add the new indeces for the nFONLL theories that are being produced now and also to propose the following booking for them:

- `700` -> NNLO central theory
- `701` -> NLO central theory
- `702` -> NNLO QED central theory
- `703` -> NLO QED central theory
- `704` to `712` -> NNLO sv theories (the only ones that have been added so far)
- `713` to `721` -> NLO sv theories

These will be the theories for the QED / MHOU papers. These are NNLO grids for DIS and NLO grids with NNLO kfactors burned in the grid.

For N3LO: 
- `722` to `???` -> N3LO theories (@giacomomagni add here the final number if you agree)

@RoyStegeman @niclaurenti @AleCandido @felixhekhorn  

EDIT: I realized that this PR can also be used to add the lists of theory ids to be used for the thcovmat construction (in `scalevariationtheoryids`). I did for NNLO for the moment, @giacomomagni add the N3LO ones when you can.


giacomomagni:
So for being annoying, can I continue using theories from 439 to 510?

andreab1997:
> So for being annoying, can I continue using theories from 439 to 510? If this will cause further inconsistencies I'm okay with 722 to 800

I believe we should keep the old theories in the server as they have been used for proceedings and presentations and we should make the results reproducible. Is there a particular reason why you would rather just update the old theories?

giacomomagni:
> Is there a particular reason why you would rather just update the old theories?

My "old" theories never get to the server. Every time they have been computed, there was already a good reason why they were old or deprecated.

I hope that all the material shown so far at any conference had a PRELIMINARY label.

andreab1997:
> > Is there a particular reason why you would rather just update the old theories?
> 
> My "old" theories never get to the server. Every time they have been computed, there was already a good reason why they were old or deprecated.
> 
> I hope that all the material shown so far at any conference had a PRELIMINARY label.

Ok so I would add your old theories to the server and then use these new theoryids for the nFONLL theories but I guess that in your case it won't make much difference if you do it or not. Just let us know what you decide.

AleCandido:
It would be worth to maintain somewhere in the docs a summary, describing what the various ranges are used for.

With IDs up to 7** (even though there are not really 800 theories) it is becoming difficult to find out something (especially if the only way is searching in the *Comment* field).

andreab1997:
> It would be worth to maintain somewhere in the docs a summary, describing what the various ranges are used for.
> 
> With IDs up to 7** (even though there are not really 800 theories) it is becoming difficult to find out something (especially if the only way is searching in the _Comment_ field).

In principle it is already in the NNPDF docs (https://docs.nnpdf.science/theory/theoryindex.html) and for example my old theories are listed there. Sorry for having forgotten to update it for the new theories. I am doing it now

AleCandido:
I said a summary, not the whole table.

Like:
- `-199`: pre-NNPDF4.0
- `200-2xy`: NNPDF4.0
- `400-4xy`: pineline NNPDF4.0 reproduction (w/ aFONLL implementation)
- ...

giacomomagni:
> Ok so I would add your old theories to the server 

To be clear this implies storing 80 theories which are deprecated or not up-to-date anymore. 
Isn't this going to be more confusing ?

andreab1997:
> > Ok so I would add your old theories to the server
> 
> To be clear this implies storing 80 theories which are deprecated or not up-to-date anymore. 

The point is that some of the results we have in the wiki for example are produced with these 80 theories.

>And these theories will not be usable directly in `vp-setupfit`.

Why not?

scarlehoff:
We "sort of agreed" that anything beyond 400 was a experimental theory and so what subject to change.

@andreab1997 is right that we have used these in proceedings/talks but as @giacomomagni said they were preliminary so no worries. We haven't published anything with this. If we want to have a "backup", let's copy all of them to a folder "backup_building_new_pipeline" and that's it.

andreab1997:
Ok then, I do not really agree and I would rather keep reproducible (for next PhDs for example) the results that we have in the wiki. However, I accept your opinion, should I close this PR then? :)

RoyStegeman:
>  I accept your opinion, should I close this PR then? :)

Wait nothing has been agreed upon right?

I'm not sure about Juan's proposal since clearing everything up and storing it elsewhere means that everything that's done in theories 400+ should be nFONLL from this moment on. However, it's not even signed off and both Peter and Tanishq are producing theories in aFONLL that I'm not sure they'd even want to be in nFONLL.

To prevent a mess I'd be happy to create such a backup but it will also mean halting essentially all projects until we really have a final, working and tested nFONLL implementation.

andreab1997:
> > I accept your opinion, should I close this PR then? :)
> 
> Wait nothing has been agreed upon right?
> 
> I'm not sure about Juan's proposal since clearing everything up and storing it elsewhere means that everything that's done in theories 400+ should be nFONLL from this moment on. However, it's not even signed off and both Peter and Tanishq are producing theories in aFONLL that I'm not sure they'd even want to be in nFONLL.
> 
> To prevent a mess I'd be happy to create such a backup but it will also mean halting essentially all projects until we really have a final, working and tested nFONLL implementation.

Yes, this is also my point: the fact that now we use nFONLL as standard for the N3LO, QED and MHOU projects does not mean that people need to stop using aFONLL (eventually with the threshold of the b shifted).

AleCandido:
The only reason to keep using aFONLL is to do polarized/FFs for the time being. Eventually (hopefully very soon), aFONLL will be completely erased from the pineline, since we do not want to maintain two parallel implementations (it is only confusing and generating over-head, moreover, the displacing the b-threshold is a nice exercise, but if it's your only way to compute things is intrinsically flawed).

I agree with @scarlehoff that we should consider everything beyond 400 as temporary, and eventually it will be removed as well. That's why we started using theories from 400, instead of using sequential numbers after 4.0 theories.

If there is anything you want to store properly and keep, start using sequential numbers. In principle, this should happen for all and only papers.
(proceeding with WIP results do not have to be perfectly reproducible, it's understood they are WIP)

AleCandido:
> The only reason to keep using aFONLL is to do polarized/FFs for the time being.

(only because of yadism, everything else should be compatible with nFONLL out-of-the-box, EKO in the first place)

giacomomagni:
> EDIT: I realized that this PR can also be used to add the lists of theory ids to be used for the thcovmat construction (in `scalevariationtheoryids`). I did for NNLO for the moment, @giacomomagni add the N3LO ones when you can.

For the N3LO theories, what I suggest id that we keep using ids from 439 to 510 until the final version is reached. 
(see also #1698 )
Once this will happen (and the paper will be about to be published) we can move them to 722 and so on.

andreab1997:
@RoyStegeman Can this be merged? Do you need to add something else?

RoyStegeman (approved):

Issue #1800, opened 2023-08-25
Hyperopt checkpointing/restarting


goord:
Does the hyperoptimization of the fit currently support checkpointing and restarting the bayesian optimization process?

scarlehoff:
I don't think so.

Btw, we used the `hyperopt` library but since it hasn't been maintained in a while (I see it started having new commits in the last few months, but it has been stale since ~2021) maybe it makes sense to implement some other library...

Back when `hyperopt` was implemented I also implemented [optuna](https://github.com/optuna/optuna) and it worked fine but chose to stay with `hyperopt` because I have a great ability to select the winner technology :)
(for what we were doing hyperopt was much simpler to use and required less tweaking, in any case, what I'm trying to say is feel free to change hyperopt to some other library, the hyperoptimization library is not coupled at all with the rest of the code so it should be easy)

Here's a patch with my old optuna implementation
[patch_optuna.zip](https://github.com/NNPDF/nnpdf/files/12440233/patch_optuna.zip)

the changes needed were really minimal (I'm guessing the patch itself won't work since it is 4 years old) so using other hyperoptimization libraries should also be possible!

Radonirinaunimi:
> Does the hyperoptimization of the fit currently support checkpointing and restarting the bayesian optimization process?

As I mentioned in the slack, this feature is not available. But if picking as a starting point a given `n`-th trial is needed, then this can be easily implemented by loading the `json` file that are already written into disk and use it as the starting input to the `trials` entry.

scarlehoff:
Closed by #1824

PR #1812, opened 2023-10-09
Restart hyperopt

Cmurilochem:
This PR addresses the issue of restarting an hyperoptimization with the Hyperopt library as discussed in #1800.

### DIFF
--- a/n3fit/src/n3fit/stopping.py
+++ b/n3fit/src/n3fit/stopping.py
@@ -356,8 +356,8 @@ class Stopping:
 
         self._dont_stop = dont_stop
         self._stop_now = False
-        self._stopping_patience = stopping_patience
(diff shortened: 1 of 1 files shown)


RoyStegeman:
Hi @Cmurilochem the diffs are a bit difficult to understand, it seems that you have a lot of commits that are already included in master while you're trying to work/merge on top of a branch that does not have these commits?

Perhaps this should be pointed to master and @Radonirinaunimi change should be undone or mabye the `hyperopt_loss` branch should be rebased (though if this was branched off from master you may need to rebase this on top of `hyperopt_loss` if that's what you want)? Either way, before continuing your progress you may want to make sure the history makes sense and you will be able to safely merge by the time that's needed.

PR #1816, opened 2023-10-12
Polarised DIS Commondata

Radonirinaunimi:
The following PR implements the Polarised DIS measurements in the new commondata format.

|  Experiment  | Observable  |  Status  |   Comment  |
|---|---|---|---|
| E155  | $g_1/F_1$  |   | |


Radonirinaunimi:
@toonhasenack, could you please use [black](https://github.com/psf/black) to format the `*.py` file(s)?

buildmaster/E155_NC_9GEV_EP_G1F1RATIO/metadata.yaml
```diff
@@ -0,0 +1,64 @@
...
+    line_by:
+      - Q2
+    figure_by:
+      - k2bins6
+    normalize:
+  y_scale: linear
+      line_by: []
+    figure_by:
+            - k2bins6
+            - Q2
```
Radonirinaunimi, line 48: As per [this guideline](https://github.com/NNPDF/nnpdf/blob/2e39e52258a4b31169a879fedb40513e18697313/doc/sphinx/source/data/new-commondata.rst):
```suggestion
    # Plotting information
    plotting:
        dataset_label: "E155 (g1/F1)"
        plot_x: x
        y_label: '$g_1(x, Q^2) / F_1(x, Q^2)$'
        figure_by:
            - Q2
    kinematic_coverage: [x, y, Q2]
```

Radonirinaunimi:
@toonhasenack, could you also please do the same for the neutron target? Basically the exact same as above but for everything inside `header: {name: G1(Q=N)/F1(Q=N)}` (where `N` denotes neutron). The dataset name should follow the same convention, see edited table above (for [further documentation](https://github.com/NNPDF/nnpdf/blob/2e39e52258a4b31169a879fedb40513e18697313/doc/sphinx/source/data/dataset-naming-convention.rst)).

buildmaster/E155_NC_9GEV_EP_G1F1RATIO/filter.py
```diff
@@ -0,0 +1,94 @@
...
+import glob
+
+
+def read_data(fnames):
+    df = pd.DataFrame()
+    for fname in fnames:
+        with open(fname, "r") as file:
+            data = yaml.safe_load(file)
+
+        x = float(data["dependent_variables"][1]["qualifiers"][4]["value"])
```
Radonirinaunimi, line 13: Being a bit picky here! I know it is exactly the same but for absolute consistency. The neutron one should be with `1` instead.
```suggestion
        x = float(data["dependent_variables"][0]["qualifiers"][4]["value"])
```

.gitignore
```diff
@@ -1,4 +1,4 @@
-# extra files created by cmake
+z# extra files created by cmake
```
giacomomagni, line 1:
```suggestion
# extra files created by cmake
```

Radonirinaunimi: Resolved in [68c65bf](https://github.com/NNPDF/nnpdf/pull/1816/commits/68c65bf8507faaf10ab8c33687125d8ef201fd67).

validphys2/src/validphys/datafiles/new_commondata/SMCSX_NC_17GEV_MUP/metadata.yaml
```diff
@@ -17,21 +17,21 @@ nnpdf_metadata:
   experiment: "SMCSX"
 
 implemented_observables:
-  - observable_name: "G1"
+  - observable_name: "G1F1RATIO"
     observable:
-      description: "SMC small x DIS scattering experiment polarized structure function"
-      label: "$g_1$"
+      description: "SMC small x DIS scattering experiment polarized/unpolarized structure function"
+      label: "$\frac{g_1}{F1}$"
```
Radonirinaunimi, line 23:
```suggestion
      label: "$A_1^p$"
```

validphys2/src/validphys/datafiles/new_commondata/SMCSX_NC_17GEV_MUP/metadata.yaml
```diff
@@ -17,21 +17,21 @@ nnpdf_metadata:
   experiment: "SMCSX"
```
Radonirinaunimi, line 17:
```suggestion
  experiment: "SMC"
```

validphys2/src/validphys/datafiles/new_commondata/SMCSX_NC_17GEV_MUP/metadata.yaml
```diff
@@ -17,21 +17,21 @@ nnpdf_metadata:
...
     tables: [1]
-    npoints: [7] # number of datapoints in each table
+    npoints: [16] # number of datapoints in each table
 
     # Plotting information
     plotting:
       kinematics_override: dis_sqrt_scale
-      dataset_label: "SMCSX (g1)"
-      y_label: "$g_{1,N}(x, Q^2)$"
+      dataset_label: "SMCSX (g1/F1)"
```
Radonirinaunimi, line 33:
```suggestion
      dataset_label: "SMCSX (A1)"
```

validphys2/src/validphys/datafiles/new_commondata/SMCSX_NC_17GEV_MUP/metadata.yaml
```diff
@@ -17,21 +17,21 @@ nnpdf_metadata:
...
+      y_label: "$g_{1,N}(x, Q^2) / F_{1,N}(x, Q^2)$"
```
Radonirinaunimi, line 34:
```suggestion
      y_label: "$A_{1}^{P}(x, Q^2)$"
```

validphys2/src/validphys/datafiles/new_commondata/SMCSX_NC_24GEV_MUD/metadata.yaml
```diff
@@ -17,21 +17,21 @@ nnpdf_metadata:
   experiment: "SMCSX"
 
 implemented_observables:
-  - observable_name: "G1"
+  - observable_name: "G1F1RATIO"
     observable:
-      description: "SMC small x DIS scattering experiment polarized structure function"
-      label: "$g_1$"
+      description: "SMC small x DIS scattering experiment polarized/unpolarized structure function"
+      label: "$\frac{g_1}{F_1}$"
```
Radonirinaunimi, line 23:
```suggestion
      label: "$A_1$"
```

validphys2/src/validphys/datafiles/new_commondata/SMCSX_NC_24GEV_MUD/metadata.yaml
```diff
@@ -17,21 +17,21 @@ nnpdf_metadata:
...
+    ndata: 16
+    tables: [1]
+    npoints: [16] # number of datapoints in each table
 
     # Plotting information
     plotting:
       kinematics_override: dis_sqrt_scale
-      dataset_label: "SMCSX (g1)"
-      y_label: "$g_{1,N}(x, Q^2)$"
+      dataset_label: "SMCSX (g1/F1)"
```
Radonirinaunimi, line 33:
```suggestion
      dataset_label: "SMCSX (A1)"
```

validphys2/src/validphys/datafiles/new_commondata/SMCSX_NC_24GEV_MUD/metadata.yaml
```diff
@@ -17,21 +17,21 @@ nnpdf_metadata:
   experiment: "SMCSX"
```
Radonirinaunimi, line 17:
```suggestion
  experiment: "SMC"
```

validphys2/src/validphys/datafiles/new_commondata/SMCSX_NC_24GEV_MUD/metadata.yaml
```diff
@@ -17,21 +17,21 @@ nnpdf_metadata:
...
+    tables: [1]
+    npoints: [16] # number of datapoints in each table
 
     # Plotting information
     plotting:
       kinematics_override: dis_sqrt_scale
-      dataset_label: "SMCSX (g1)"
-      y_label: "$g_{1,N}(x, Q^2)$"
+      dataset_label: "SMCSX (g1/F1)"
+      y_label: "$g_{1,N}(x, Q^2) / F_{1,N}(x, Q^2)$"
```
Radonirinaunimi, line 34:
```suggestion
      y_label: "$A_{1}^{P}(x, Q^2)$"
```

Radonirinaunimi (commented):
Thanks @toonhasenack for having implemented the missing datasets. Regarding the SMCSX, note that there is an Errata (https://journals.aps.org/prd/abstract/10.1103/PhysRevD.62.079902) so some of the entries in the table need to be changed/fixed.

Given that the actually measured observable is $A_1$ we should keep it as the label in the plots. The `observable_name` can be `G1F1RATIO` since that's what we compute.

enocera:
Thanks @Radonirinaunimi @giacomomagni @toonhasenack. Is my understanding correct that I can now look at all of the implemented data sets, check that I'm happy and if so give you green light for the baseline fits?

Radonirinaunimi:
> Thanks @Radonirinaunimi @giacomomagni @toonhasenack. Is my understanding correct that I can now look at all of the implemented data sets, check that I'm happy and if so give you green light for the baseline fits?

@enocera, yes exactly! Thank you very much!

enocera:
I have checked all the data sets, except the pseudodata for the EIC//EicC and for positivity. Here are my remarks.
1) Answers to the questions in the table at the top of this page
- `COMPASS15_NC_17GEV_MUP_G1` and `COMPASS15_NC_25GEV_MUD_G1`: the selection of tables is good as is, as you consistently select always the table with bins averaged over Q2.
- `E142_NC_58GEV_EN_G1`: I recommend to use Table 3 instead of Table 5. The reason being that, in Table 5, the results of Table 3 have been "evolved" to a common scale by the experimentalists (and we don't really trust them).
- `E143_NC_4GEV_EP_G1` and `E143_NC_4GEV_ED_G1`: I recommend to switch to `CORR`. The reason being that these are two different runs of the same experiment in which the target has been replaced. In this respect, the uncertainty on the beam energy is (mostly) independent in the two runs.
- `E154_NC_9GEV_EN_G1`: I recommend to use Table 1, as done in the current implementation. The reason is the same as for E142: Table 2 has been 'evolved' by the experimentalists to a common Q scale for all bins.
2) Comments
- I noticed that the arXiv reference is missing in several metadata files, even when there is an arXiv number. Is there a reason for that?
- The inelasticity `y` is always set to zero. However, this could be determined from knowledge of `x`, `Q` and the c.m. energy, which are known for each bin of each data set, as y=Q^2/(x(s-Mp^2)), Wouldn't it be good to compute `y` instead of setting it to zero?
- I don't understand why the c.m. energy (as reflected in the name of the data sets) is 4 GeV in `E143_NC_4GEV_EP` and 8 GeV in `E143_NC_8GEV_ED`. Shouldn't it be 4 GeV in both data sets?
- A consideration similar to the above, I guess, also holds for `COMPASS15_NC_17GEV_MUP_G1` and `COMPASS15_NC_25GEV_MUD_G1`, for which the c.m. energy should be 17 GeV in both data sets (therefore the name of the deuteron data set must be changed). Please check if I'm not saying something wrong.
- I would remove outdated or redundant data sets, namely: the old COMPASS implementation (`COMPASS_NC_17GEV_MUP_G1` and `COMPASS_NC_17GEV_MUD_G1`); and the not independent HERMES implementation (`HERMES_NC_7GEV_EN_G1`). In my opinion, this removes sources of confusion for the future, though I leave the decision to you.

enocera:
And of course @Radonirinaunimi @giacomomagni and @toonhasenack thanks for your care in implementing the data. Much appreciated.

Radonirinaunimi:
> I have checked all the data sets, except the pseudodata for the EIC//EicC and for positivity. Here are my remarks.

Thanks a lot @enocera for having carefully checked these implementations. I believe that the positivity and EIC/EIcC datasets do not require reviews as their implementations are rather straightforward.
> 
>     1. Answers to the questions in the table at the top of this page
> 
> 
>     * `COMPASS15_NC_17GEV_MUP_G1` and `COMPASS15_NC_25GEV_MUD_G1`: the selection of tables is good as is, as you consistently select always the table with bins averaged over Q2.
> 
>     * `E142_NC_58GEV_EN_G1`: I recommend to use Table 3 instead of Table 5. The reason being that, in Table 5, the results of Table 3 have been "evolved" to a common scale by the experimentalists (and we don't really trust them).
> 
>     * `E143_NC_4GEV_EP_G1` and `E143_NC_4GEV_ED_G1`: I recommend to switch to `CORR`. The reason being that these are two different runs of the same experiment in which the target has been replaced. In this respect, the uncertainty on the beam energy is (mostly) independent in the two runs.
> 
>     * `E154_NC_9GEV_EN_G1`: I recommend to use Table 1, as done in the current implementation. The reason is the same as for E142: Table 2 has been 'evolved' by the experimentalists to a common Q scale for all bins.
> 
Agreed on everything! We will address these (in addition to some below) and once done we could finally merge this PR into master (thanks @giacomomagni).
> 
>     2. Comments
> 
> 
>     * I noticed that the arXiv reference is missing in several metadata files, even when there is an arXiv number. Is there a reason for that?
> 
This must be an oversight. We'll add the missing the references.

>     * The inelasticity `y` is always set to zero. However, this could be determined from knowledge of `x`, `Q` and the c.m. energy, which are known for each bin of each data set, as y=Q^2/(x(s-Mp^2)), Wouldn't it be good to compute `y` instead of setting it to zero?
> 
Although in all these datasets (except for ATHENA) $y$ is basically immaterial, I agree that we should dump the correct values ($\longleftarrow$ to be added to the list of modifications).

>     * I don't understand why the c.m. energy (as reflected in the name of the data sets) is 4 GeV in `E143_NC_4GEV_EP` and 8 GeV in `E143_NC_8GEV_ED`. Shouldn't it be 4 GeV in both data sets?
> 
>     * A consideration similar to the above, I guess, also holds for `COMPASS15_NC_17GEV_MUP_G1` and `COMPASS15_NC_25GEV_MUD_G1`, for which the c.m. energy should be 17 GeV in both data sets (therefore the name of the deuteron data set must be changed). Please check if I'm not saying something wrong.
> 
Let us actually double check these. I just recall that for some of them, as the values were not explicitly mentioned in the paper, we had to compute the c.o.m.

>     * I would remove outdated or redundant data sets, namely: the old COMPASS implementation (`COMPASS_NC_17GEV_MUP_G1` and `COMPASS_NC_17GEV_MUD_G1`); and the not independent HERMES implementation (`HERMES_NC_7GEV_EN_G1`). In my opinion, this removes sources of confusion for the future, though I leave the decision to you.

Ok, agreed.

giacomomagni:
>  * I don't understand why the c.m. energy (as reflected in the name of the data sets) is 4 GeV in `E143_NC_4GEV_EP` and 8 GeV in `E143_NC_8GEV_ED`. Shouldn't it be 4 GeV in both data sets?
> 
>  * A consideration similar to the above, I guess, also holds for `COMPASS15_NC_17GEV_MUP_G1` and `COMPASS15_NC_25GEV_MUD_G1`, for which the c.m. energy should be 17 GeV in both data sets (therefore the name of the deuteron data set must be changed). Please check if I'm not saying something wrong.

@enocera I think the center of mass energy has been computed from the beam energy with $s = M^2 + 2 M E, 
so this is why deuteron datasets have a larger value. Do you agree on that?

giacomomagni:
To wrap-up the discussion here we have decided to: 

* Remove the inelasticity completely from the kinematics.
* In case multiple beam energies are included in one dataset we will use the `NOTFIXED` labelling in the name instead of `s`.
* In case `s` is not given and there is a single beam energy is provided, we will use the formula above $s = M^2 + 2 M E$ taking into account the mass of the target.

cc @enocera, @Radonirinaunimi

enocera (approved):

PR #1819, opened 2023-10-17
Integrated nuclear fits

Radonirinaunimi:
A re-implementation of the integrated nuclear-PDF fits based on the refactors.

### Remaining to do:
- [ ] Improve and re-organize the `writer` and `vpinterface` modules
- [ ] Accommodate the computation of integrability and arc-lengths to properly account for $A \neq 1$
- [ ] Properly write the predictions onto disk and modify the way evolution is applied accordingly
- [ ] Fix broken tests
- [ ] Extend the report to include nuclear predictions

### Checks:
- [x] Standard fits in which only proton-like data are included ([report](https://vp.nnpdf.science/5xbfImhDT2q7uiD-ATVzEg==/))
- [ ] ...


Radonirinaunimi:
Need to re-started from scratch!

PR #1821, opened 2023-10-20
New jet data

t7phy:


scarlehoff:
Hi @t7phy thanks for this.

Two questions,
1. These four datasets are completely new right?
2. I understand you have fktables already for this (since there is a `theory` block)? They are in theory 600 right?

t7phy:
> Hi @t7phy thanks for this.
> 
> Two questions,
> 1. These four datasets are completely new right?

Yes

> 2. I understand you have fktables already for this (since there is a `theory` block)? They are in theory 600 right?

NLO with SV: 401-409
NNLO: 600
NNLO SV: 601-604 and 606-609

buildmaster/ATLAS_1JET_13TEV_DIF/metadata.yaml
```diff
@@ -0,0 +1,81 @@
...
+        - - ATLAS_1JET_13TEV_DIF_PT-Y_BIN1
+          - ATLAS_1JET_13TEV_DIF_PT-Y_BIN2
+          - ATLAS_1JET_13TEV_DIF_PT-Y_BIN3
+          - ATLAS_1JET_13TEV_DIF_PT-Y_BIN4
+          - ATLAS_1JET_13TEV_DIF_PT-Y_BIN5
+          - ATLAS_1JET_13TEV_DIF_PT-Y_BIN6
+      operation: 'null'
+
+  - observable: {description: "double differential cross section w.r.t. pT and y - altcorr1", label: '$\frac{d^2\sigma}{dp_T dy}$', units: "pb/GeV"}
+    observable_name: PT-Y-ALTCORR1
```
scarlehoff, line 54: This is a question also for @enocera:  PT-Y and PT-Y-ALTCORR1, should the data be different if the difference is only the uncertainties? Shouldn't this be a variant (i.e., shouldn't the data be the same, I see that the numerical values are different...)

For @t7phy, even if they are separate observables (if the data changes then they are I'd say)I  can see the kinematics are exactly the same, you can use just one of them and remove the other.

Btw, you should be able to do that for the uncertainties as well. _if I did everything correctly_ then uncertainties get concatenated so if one of them is shared you can have them in a separate file an add that file in both :)

scarlehoff (commented):
Thank you very much for this. These datasets (with theory 600) mostly work out of the box. I'll put here a list of comments. At the bottom there's a few pictures for the theory-data comparisons.

1. The "process_type" key is missing, when there are no reference (like in this case) you can have a look at the file https://github.com/NNPDF/nnpdf/blob/e005a42cfbb2bc049c5e49556d425952c39e2326/validphys2/src/validphys/plotoptions/kintransforms.py#L35 and use the one that looks the closest (in this case you have JET and DIJET)].
2. You can use `x_scale: log` for a nicer picture. I've added it manually for some of the plots below. This is aesthetics of course.
3. Please use the latex markers `$$` e.g. `$GeV^2$` in the units as well.
4. Please remove the `implemented by` part, git keeps track of all authorship and much more reliably than just the name at the beginning of the file (I know there are some files that have it scattered around the code, mostly from people using sypder I think, we should probably do a pass at some point to remove that).

Everything else seems fine to me. I've tested theory 600. From the plots as you say there seems to be some mismatch between the data and the predictions but I think I agree with you that given the trend it looks more like a grid-problem than a data-problem.

<details><summary> Data-Theory comparisons plots</summary>

Unknown-5
Unknown-4
Unknown-3
Unknown-2
Unknown-1
Unknown
</details>

buildmaster/ATLAS_1JET_13TEV_DIF/metadata.yaml
t7phy: I can answer this, they are indeed the same datasets with different uncertainties and ideally should have variants for the uncertainties. There was a caveat, in that the uncertainties were not symmetrical. When you symmetrize them, the data point shifts a bit, and this resulted in having to treat them separately.

scarlehoff: ahh, okok

Then perfect. The comment remains about the kinematics/uncertainties. I haven't look at them to check whether they are numerically equal, but if they are you can remove one file (like the `x_scale` this as an aesthetical point of course, the commondata works the same).

t7phy: No, the uncertainties are indeed different, thats why their symmetrization changes the the data point. The kinematics on the other hand is the same, but I thought it would be good to have all the files separately as they are listed separately in the metadata

scarlehoff:
Ok, I think this can be merged. I don't see any problems with it and want to have a "cannonical example" in #1813 that people can use for reference. I've even tried to use it in a fit and it works fine.
If it needs to be updated later so be it.

There's a few comments however that might be relevant:
1. We should use always the same notation, and maybe have a collection of variables somewhere. Something like [this](https://github.com/NNPDF/nnpdf/blob/e005a42cfbb2bc049c5e49556d425952c39e2326/validphys2/src/validphys/filters.py#L19). In principle it doesn't matter since we are no longer relying on "k1, k2, k3" but rather defining their names in the metadata. However it would be nice if everybody uses the same "p_T2" or "pt_sqr" or whatever. "y" or "eta" (which might be different but we are using equivalently in some cases) etc.

2. I need to update how the cuts are applied because they are still relying in the k1/k2/k3 order, so if you try to test the cuts with the new reader it will only work if they happen to be in the expected order for the process_type. I'll do this in the next few days.

3. Points 1 and 2 might be a problem for people implementing old commondata in the new format. But until I don't have some examples I won't be able to see the ways in which this can introduce problems so we'll see :P

PR #1823, opened 2023-10-20
add scale variations IDs for theory 600

t7phy:


andreab1997 (commented):
This is correct but surely you also need to update the `theory.db` file. If possible, do it in this PR.

andreab1997 (approved):
Ok looks good to me!

PR #1824, opened 2023-10-24
Restart hyperopt

Cmurilochem:
This PR addresses the issue of restarting an hyperoptimization with the Hyperopt library as discussed in https://github.com/NNPDF/nnpdf/issues/1800.

### 1. `hyper_optimization/filetrials.py`
- To the `FileTrials` class I have added the `from_pkl` and `to_pkl` methods. The last one is a `@classmethod` that is useful to create instances of the class when `tries.pkl` file is available from a previous run.
The `to_pkl` method saves the current state of `FileTrials` to a pickle file, although this is currently being indeed done for every trial in `hyperopt.fmin` directly via the `trials_save_file` argument. 
- In this regard, I also added an attribute `self.pkl_file` which will be responsible for generating a `tries.pkl` file in the same directory as the `tries.json`
- An additional attribute `self._rstate` is also added that will store the last `numpy.random.Generator` of the hyperopt algorithm and will be passed as `rstate` in the `hyperopt.fmin` function so that we warrant that 
by restarting we do so with the same history as if we were doing a direct calculation. The initial fixed seed in `trials.rstate = np.random.default_rng(42)` [here](https://github.com/NNPDF/nnpdf/blob/restart_hyperopt/n3fit/src/n3fit/hyper_optimization/hyper_scan.py#L116C5-L116C46) can still be relaxed and provided as input later.  

### 2. `hyper_optimization/hyper_scan.py`
- In case of restarts, an extra boolean attribute is added to `HyperScanner`, named `self.restart_hyperopt` which is set `true` in case of the `--continue` option in `n3fit` command line (details to be discussed below).
- I have adapted `hyper_scan_wrapper` to allow it to check if `hyperscanner.restart_hyperopt` is `true`. If so, it will generate an initial `FileTrial` instance (`trials`) from `tries.pkl`, which contains by built-in the history of the previous hyperopt and also the `trials.rstate` attribute with the previous numpy random generator.

### 3. `scripts/n3fit_exec.py`
This is perhaps the most fragile of the changes and where I would need help to adapt it properly.
- To the `N3FitApp` I added a new parser `--continue` that will be the keyword to hyperopt restarts. 
- To its `run` method I add a new `self.environment.restart = self.args["continue"]` attribute.
- The way I found to pass this keyword to `HyperScanner` later is to use it in connection with `produce_hyperscanner`. If this is `true`, I then update `hyperscan_config` with `hyperscan_config.update({'restart': 'true'})` and this will later be part of the `HyperScanner`'s
`sampling_dict` argument. 

## Questions and requested feedback

- It looks to me that the adaptations made in the `scripts/n3fit_exec.py` file to allow for `--continue` are not optimal. Maybe a more experienced developer could suggest a more convenient way to do so. 
- Despite all our efforts to make sure that the hyperopt restarts have the same history as if we were just making direct experiments, it seems that, despite having the same hyperparameter guesses, restart calculations will afterall show 
differences in the obtained final losses for different `k` folds. This might be due to the fact that the seeds for the initial `weights` for each k-fold in difference runs are inherently different (see below). 
For example, I have done a test in which I make a simple hyperoptimization with 2 trials, and then restart it to make another 2 trials (4 in total). Then I run another experiment and calculate (with the same runcard) 4 trials directly and compare the results. 
```
Restart 0
 {'validation_losses': '[[ 8.973183]\n [12.285054]]', 'experimental_losses': '[[52.548443]\n [42.66381 ]]', 'hyper_losses': '[[52.548443]\n [42.66381 ]]'}
 {'Adam_clipnorm': [], 'Adam_learning_rate': [], 'Nadam_clipnorm': [2.0117276880696024e-05], 'Nadam_learning_rate': [0.002011336799276246], 'nl2:-0/2': [41.0], 'nl2:-1/2': [32.0], 'nodes_per_layer': ['0'], 'optimizer': ['0']}
Direct 0
 {'validation_losses': '[[ 8.973183]\n [12.285054]]', 'experimental_losses': '[[52.548443]\n [42.66381 ]]', 'hyper_losses': '[[52.548443]\n [42.66381 ]]'}
 {'Adam_clipnorm': [], 'Adam_learning_rate': [], 'Nadam_clipnorm': [2.0117276880696024e-05], 'Nadam_learning_rate': [0.002011336799276246], 'nl2:-0/2': [41.0], 'nl2:-1/2': [32.0], 'nodes_per_layer': ['0'], 'optimizer': ['0']}

Restart 1
 {'validation_losses': '[[14.459031]\n [29.935106]]', 'experimental_losses': '[[64.00286 ]\n [74.955086]]', 'hyper_losses': '[[64.00286 ]\n [74.955086]]'}
 {'Adam_clipnorm': [4.9234053337502195e-06], 'Adam_learning_rate': [0.00013178694123594783], 'Nadam_clipnorm': [], 'Nadam_learning_rate': [], 'nl2:-0/2': [40.0], 'nl2:-1/2': [31.0], 'nodes_per_layer': ['0'], 'optimizer': ['1']}
Direct 1
 {'validation_losses': '[[14.459031]\n [29.935106]]', 'experimental_losses': '[[64.00286 ]\n [74.955086]]', 'hyper_losses': '[[64.00286 ]\n [74.955086]]'}
 {'Adam_clipnorm': [4.9234053337502195e-06], 'Adam_learning_rate': [0.00013178694123594783], 'Nadam_clipnorm': [], 'Nadam_learning_rate': [], 'nl2:-0/2': [40.0], 'nl2:-1/2': [31.0], 'nodes_per_layer': ['0'], 'optimizer': ['1']}

-- restarting and performing more two trials

Restart 2
 {'validation_losses': '[[19.959248]\n [39.484573]]', 'experimental_losses': '[[55.306988]\n [92.67653 ]]', 'hyper_losses': '[[55.306988]\n [92.67653 ]]'}
 {'Adam_clipnorm': [], 'Adam_learning_rate': [], 'Nadam_clipnorm': [5.226179920719625e-06], 'Nadam_learning_rate': [0.00020623358967934892], 'nl2:-0/2': [32.0], 'nl2:-1/2': [13.0], 'nodes_per_layer': ['0'], 'optimizer': ['0']}
Direct 2
 {'validation_losses': '[[19.959248]\n [43.446335]]', 'experimental_losses': '[[ 55.306988]\n [104.45502 ]]', 'hyper_losses': '[[ 55.306988]\n [104.45502 ]]'}
 {'Adam_clipnorm': [], 'Adam_learning_rate': [], 'Nadam_clipnorm': [5.226179920719625e-06], 'Nadam_learning_rate': [0.00020623358967934892], 'nl2:-0/2': [32.0], 'nl2:-1/2': [13.0], 'nodes_per_layer': ['0'], 'optimizer': ['0']}

Restart 3
 {'validation_losses': '[[23.391615]\n [65.55123 ]]', 'experimental_losses': '[[ 69.19588]\n [137.06268]]', 'hyper_losses': '[[ 69.19588]\n [137.06268]]'}
 {'Adam_clipnorm': [1.6662863474168997e-07], 'Adam_learning_rate': [0.0025640118767782183], 'Nadam_clipnorm': [], 'Nadam_learning_rate': [], 'nl2:-0/2': [31.0], 'nl2:-1/2': [17.0], 'nodes_per_layer': ['0'], 'optimizer': ['1']}
Direct 3
 {'validation_losses': '[[23.391615]\n [44.021515]]', 'experimental_losses': '[[69.19588 ]\n [93.464554]]', 'hyper_losses': '[[69.19588 ]\n [93.464554]]'}
 {'Adam_clipnorm': [1.6662863474168997e-07], 'Adam_learning_rate': [0.0025640118767782183], 'Nadam_clipnorm': [], 'Nadam_learning_rate': [], 'nl2:-0/2': [31.0], 'nl2:-1/2': [17.0], 'nodes_per_layer': ['0'], 'optimizer': ['1']}
``` 

By looking at the above results, we can see that Restart 2/3 have the same hyperparameters as Direct 2/3, with the 2 folds having different losses however. Maybe the 1st fold can still keep up with the losses but not the second fold.
With the help of @goord and @APJansen, I investigated this issue and have printed the generated random integers passed as `seeds` to generate the PDF models for each fold in `MoldelTrainer.hyperparametrizable()`; see [here](https://github.com/NNPDF/nnpdf/blob/restart_hyperopt/n3fit/src/n3fit/model_trainer.py#L863). They are shown in the Table below:

|                         |  Trial 0              | |   Trial 1 | |   Trial 2 | |  Trial 3 ||
|--------------|--------|---------|--------|---------|--------|---------|--------|---------|
|                        |   Fold 1 |   Fold 2 |   Fold 1 |   Fold 2  |   Fold 1 |   Fold 2  |  Fold 1 |   Fold 2 |
|  Restart  Job  -random integers|       1181867710       |     461027504          |      1181867710        |       1020231754        |      1181867710        |      **461027504**           |      1181867710      |       **1020231754**        |
|  Direct  Job  - random integers  |        1181867710      |     461027504          |       1181867710       |      1020231754         |      1181867710        |    **1543757328**            |        1181867710     |      **1392765670**         |

As foreseen, it is clear from the table that the `seed`s are different for the second fold every time we run a new calculation, despite the fact that the runs start with the same hyperparameters. This clearly reflects in the different losses shown above. I suspect that if we want to make hyperopt 
runs completely reproducible we could think of alternatives to
```
for k, partition in enumerate(self.kpartitions):
            # Each partition of the kfolding needs to have its own separate model
            # and the seed needs to be updated accordingly
            seeds = self._nn_seeds
            if k > 0:
                seeds = [np.random.randint(0, pow(2, 31)) for _ in seeds]
```
to initialise the seeds.

### 4. `model_trainer.py`
To ensure that these `seeds` are generated in reproducible way, @RoyStegeman helped me to devise a new form that changes the way they are generated by [defining](https://github.com/NNPDF/nnpdf/blob/ba84878353e3ed950426ae6a12674abce2dfb903/n3fit/src/n3fit/model_trainer.py#L848C1-L857C86):
```
        for k, partition in enumerate(self.kpartitions):
            # Each partition of the kfolding needs to have its own separate model
            # and the seed needs to be updated accordingly
            seeds = self._nn_seeds
            if k > 0:
                # generate random integers for each k-fold from the input `nnseeds`
                # we generate new seeds to avoid the integer overflow that may
                # occur when doing k*nnseeds
                rngs = [np.random.default_rng(seed=seed) for seed in seeds]
                seeds = [generator.integers(1, pow(2, 30)) * k for generator in rngs]
```
With all the above modifications, I have repeated my previous 4 trial experiment. The results are shown below for both restart and direct runs:
```
Restart 0
 {'validation_losses': ['2.2993183', '4.4195056'], 'experimental_losses': [10.660690008425245, 13.892794249487705], 'hyper_losses': [19.669736106403093, 21.73920023647384]}
 {'Adadelta_clipnorm': [], 'Adadelta_learning_rate': [], 'RMSprop_learning_rate': [0.015380823956886622], 'activation_per_layer': ['0'], 'dropout': [0.15], 'epochs': [35.0], 'initializer': ['0'], 'multiplier': [1.074400261320179], 'nl2:-0/2': [], 'nl2:-1/2': [], 'nl3:-0/3': [], 'nl3:-1/3': [], 'nl3:-2/3': [], 'nl4:-0/4': [15.0], 'nl4:-1/4': [41.0], 'nl4:-2/4': [36.0], 'nl4:-3/4': [45.0], 'nl5:-0/5': [], 'nl5:-1/5': [], 'nl5:-2/5': [], 'nl5:-3/5': [], 'nl5:-4/5': [], 'nodes_per_layer': ['2'], 'optimizer': ['1'], 'stopping_patience': [0.3600000000000001]}
Direct 0
 {'validation_losses': ['2.2993183', '4.4195056'], 'experimental_losses': [10.660690008425245, 13.892794249487705], 'hyper_losses': [19.669736106403093, 21.73920023647384]}
 {'Adadelta_clipnorm': [], 'Adadelta_learning_rate': [], 'RMSprop_learning_rate': [0.015380823956886622], 'activation_per_layer': ['0'], 'dropout': [0.15], 'epochs': [35.0], 'initializer': ['0'], 'multiplier': [1.074400261320179], 'nl2:-0/2': [], 'nl2:-1/2': [], 'nl3:-0/3': [], 'nl3:-1/3': [], 'nl3:-2/3': [], 'nl4:-0/4': [15.0], 'nl4:-1/4': [41.0], 'nl4:-2/4': [36.0], 'nl4:-3/4': [45.0], 'nl5:-0/5': [], 'nl5:-1/5': [], 'nl5:-2/5': [], 'nl5:-3/5': [], 'nl5:-4/5': [], 'nodes_per_layer': ['2'], 'optimizer': ['1'], 'stopping_patience': [0.3600000000000001]}

Restart 1
 {'validation_losses': ['10.667141', '18.144234'], 'experimental_losses': [14.569936714920344, 25.68137247054303], 'hyper_losses': [46.88904701966194, 52.881341569995435]}
 {'Adadelta_clipnorm': [1.7558937825962389], 'Adadelta_learning_rate': [0.02971486397602543], 'RMSprop_learning_rate': [], 'activation_per_layer': ['0'], 'dropout': [0.03], 'epochs': [30.0], 'initializer': ['0'], 'multiplier': [1.0896393776712885], 'nl2:-0/2': [], 'nl2:-1/2': [], 'nl3:-0/3': [], 'nl3:-1/3': [], 'nl3:-2/3': [], 'nl4:-0/4': [13.0], 'nl4:-1/4': [33.0], 'nl4:-2/4': [12.0], 'nl4:-3/4': [44.0], 'nl5:-0/5': [], 'nl5:-1/5': [], 'nl5:-2/5': [], 'nl5:-3/5': [], 'nl5:-4/5': [], 'nodes_per_layer': ['2'], 'optimizer': ['0'], 'stopping_patience': [0.18000000000000005]}
Direct 1
 {'validation_losses': ['10.667141', '18.144234'], 'experimental_losses': [14.569936714920344, 25.68137247054303], 'hyper_losses': [46.88904701966194, 52.881341569995435]}
 {'Adadelta_clipnorm': [1.7558937825962389], 'Adadelta_learning_rate': [0.02971486397602543], 'RMSprop_learning_rate': [], 'activation_per_layer': ['0'], 'dropout': [0.03], 'epochs': [30.0], 'initializer': ['0'], 'multiplier': [1.0896393776712885], 'nl2:-0/2': [], 'nl2:-1/2': [], 'nl3:-0/3': [], 'nl3:-1/3': [], 'nl3:-2/3': [], 'nl4:-0/4': [13.0], 'nl4:-1/4': [33.0], 'nl4:-2/4': [12.0], 'nl4:-3/4': [44.0], 'nl5:-0/5': [], 'nl5:-1/5': [], 'nl5:-2/5': [], 'nl5:-3/5': [], 'nl5:-4/5': [], 'nodes_per_layer': ['2'], 'optimizer': ['0'], 'stopping_patience': [0.18000000000000005]}

-- restarting and performing more two trials

Restart 2
 {'validation_losses': ['18.18834', '52.55721'], 'experimental_losses': [21.345310585171568, 49.5125512295082], 'hyper_losses': [144.60983298921894, 105.20777437819639]}
 {'Adadelta_clipnorm': [0.8411342478713798], 'Adadelta_learning_rate': [0.04928810632634438], 'RMSprop_learning_rate': [], 'activation_per_layer': ['1'], 'dropout': [0.09], 'epochs': [47.0], 'initializer': ['1'], 'multiplier': [1.0615455307107098], 'nl2:-0/2': [16.0], 'nl2:-1/2': [35.0], 'nl3:-0/3': [], 'nl3:-1/3': [], 'nl3:-2/3': [], 'nl4:-0/4': [], 'nl4:-1/4': [], 'nl4:-2/4': [], 'nl4:-3/4': [], 'nl5:-0/5': [], 'nl5:-1/5': [], 'nl5:-2/5': [], 'nl5:-3/5': [], 'nl5:-4/5': [], 'nodes_per_layer': ['0'], 'optimizer': ['0'], 'stopping_patience': [0.12000000000000002]}
Direct 2
 {'validation_losses': ['18.18834', '52.55721'], 'experimental_losses': [21.345310585171568, 49.5125512295082], 'hyper_losses': [144.60983298921894, 105.20777437819639]}
 {'Adadelta_clipnorm': [0.8411342478713798], 'Adadelta_learning_rate': [0.04928810632634438], 'RMSprop_learning_rate': [], 'activation_per_layer': ['1'], 'dropout': [0.09], 'epochs': [47.0], 'initializer': ['1'], 'multiplier': [1.0615455307107098], 'nl2:-0/2': [16.0], 'nl2:-1/2': [35.0], 'nl3:-0/3': [], 'nl3:-1/3': [], 'nl3:-2/3': [], 'nl4:-0/4': [], 'nl4:-1/4': [], 'nl4:-2/4': [], 'nl4:-3/4': [], 'nl5:-0/5': [], 'nl5:-1/5': [], 'nl5:-2/5': [], 'nl5:-3/5': [], 'nl5:-4/5': [], 'nodes_per_layer': ['0'], 'optimizer': ['0'], 'stopping_patience': [0.12000000000000002]}

Restart 3
 {'validation_losses': ['26.753922', '24.388603'], 'experimental_losses': [52.71014284620098, 35.8982934170082], 'hyper_losses': [82.31994112766945, 3697.219938467043]}
 {'Adadelta_clipnorm': [0.44633727461389994], 'Adadelta_learning_rate': [0.023650226340698025], 'RMSprop_learning_rate': [], 'activation_per_layer': ['1'], 'dropout': [0.09], 'epochs': [26.0], 'initializer': ['1'], 'multiplier': [1.0166524890792967], 'nl2:-0/2': [38.0], 'nl2:-1/2': [34.0], 'nl3:-0/3': [], 'nl3:-1/3': [], 'nl3:-2/3': [], 'nl4:-0/4': [], 'nl4:-1/4': [], 'nl4:-2/4': [], 'nl4:-3/4': [], 'nl5:-0/5': [], 'nl5:-1/5': [], 'nl5:-2/5': [], 'nl5:-3/5': [], 'nl5:-4/5': [], 'nodes_per_layer': ['0'], 'optimizer': ['0'], 'stopping_patience': [0.24000000000000005]}
Direct 3
 {'validation_losses': ['26.753922', '24.388603'], 'experimental_losses': [52.71014284620098, 35.8982934170082], 'hyper_losses': [82.31994112766945, 3697.219938467043]}
 {'Adadelta_clipnorm': [0.44633727461389994], 'Adadelta_learning_rate': [0.023650226340698025], 'RMSprop_learning_rate': [], 'activation_per_layer': ['1'], 'dropout': [0.09], 'epochs': [26.0], 'initializer': ['1'], 'multiplier': [1.0166524890792967], 'nl2:-0/2': [38.0], 'nl2:-1/2': [34.0], 'nl3:-0/3': [], 'nl3:-1/3': [], 'nl3:-2/3': [], 'nl4:-0/4': [], 'nl4:-1/4': [], 'nl4:-2/4': [], 'nl4:-3/4': [], 'nl5:-0/5': [], 'nl5:-1/5': [], 'nl5:-2/5': [], 'nl5:-3/5': [], 'nl5:-4/5': [], 'nodes_per_layer': ['0'], 'optimizer': ['0'], 'stopping_patience': [0.24000000000000005]}
```
|                         |  Trial 0              | |   Trial 1 | |   Trial 2 | |  Trial 3 ||
|--------------|--------|---------|--------|---------|--------|---------|--------|---------|
|                        |   Fold 1 |   Fold 2 |   Fold 1 |   Fold 2  |   Fold 1 |   Fold 2  |  Fold 1 |   Fold 2 |
|  Restart  Job  -random integers|       1872583848       |     203138455          |      1872583848        |      203138455        |      1872583848        |      **203138455**           |      1872583848      |       **203138455**        |
|  Direct  Job  - random integers  |      1872583848       |     203138455          |      1872583848        |      203138455        |      1872583848        |      **203138455**           |      1872583848      |       **203138455**        |

As seen, we are now able to ensure that both the hyperparameter space and the initial `weights` for each k-fold are reproducible when restarting. 

## Note
**As can be seen from the above (last) table, because the seeds to generate the random integers for each k-fold are now derived from the fixed value `self._nn_seeds` [here](https://github.com/NNPDF/nnpdf/blob/ba84878353e3ed950426ae6a12674abce2dfb903/n3fit/src/n3fit/model_trainer.py#L857C22-L857C22), the generated random integers will always be the same in every trial; see https://github.com/NNPDF/nnpdf/pull/1824#discussion_r1379037604. This is an important aspect to keep in mind.**


RoyStegeman:
Thanks @Cmurilochem , do you understand why the test is failing in ubuntu?

Cmurilochem:
> Thanks @Cmurilochem , do you understand why the test is failing in ubuntu?

Hi @RoyStegeman, I added a new test that compares the results of one restart and a direct hyperopt run. This then checks for files and depend on paths and from where you run `pytest`. I will soon try to correct for this.

RoyStegeman:
> I do not know why this is not being run in macos. Do you have any idea?

That's actually a good question. Since you added the `pytest.mark.skip` decorator I would have expected it to be skipped for both ubuntu and macos. Either way, since you're still investigating an issue related to this specific test, let's not worry too much about it now.

Radonirinaunimi:
Thanks a lot @Cmurilochem for these works! Regarding the issue you are facing now, are you sure that the other seeds (tr/vl, MC replicas) aren't also different? In any case, I think that there should be some (simple) ways to trick the [random generators](https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/n3fit_data.py) that it is starting from a $n$-th trial (as you can see from the table, the seeds for the continued hyperopt exactly restarted from the 2nd Trial).

n3fit/src/n3fit/scripts/n3fit_exec.py
```diff
@@ -283,6 +286,7 @@ def run(self):
                 replicas = [replica]
             self.environment.replicas = NSList(replicas, nskey="replica")
             self.environment.hyperopt = self.args["hyperopt"]
+            self.environment.restart = self.args["continue"]
```
RoyStegeman, line 289: Is there a reason you go from "continue" to "restart"? Using different names for the same thing can make code more difficult to read.

n3fit/src/n3fit/model_trainer.py
```diff
@@ -922,7 +926,11 @@ def hyperparametrizable(self, params):
             for model in models.values():
                 model.compile(**params["optimizer"])
 
-            passed = self._train_and_fit(models["training"], stopping_object, epochs=epochs,)
+            passed = self._train_and_fit(
+                models["training"],
+                stopping_object,
+                epochs=epochs,
+            )
```
RoyStegeman, line 933: I don't think this change is necessary, with the configuration as defined in `pyproject.toml` since it was already less than 100 columns. 

I did notice `black` automatically 'unfolds' inputs like this if there is a trailing comma, so please just remove the last comma.

RoyStegeman (commented):
Some comments but not a complete review

n3fit/src/n3fit/model_trainer.py
RoyStegeman: See https://github.com/NNPDF/nnpdf/pull/1828

n3fit/src/n3fit/hyper_optimization/hyper_scan.py
```diff
@@ -88,12 +89,13 @@ def hyper_scan_wrapper(replica_path_set, model_trainer, hyperscanner, max_evals=
...
     A ``tries.json`` file will be saved in the ``replica_path_set`` folder with the information
-    of all trials.
+    of all trials. An additional ``tries.pkl`` file will also be generated in the same folder
+    that stores the previous states of `FileTrials`. This file can be used for restarting purposes.
 
     Parameters
     -----------
         replica_path_set: path
-            folder where to create the json ``tries.json`` file
+            folder where to create the json ``tries.json`` and pickle ``tries.pkl`` files
```
RoyStegeman, line 98:
```suggestion
            folder where to create the ``tries.json`` and ``tries.pkl`` files
```

Cmurilochem:
> Thanks a lot @Cmurilochem for these works! Regarding the issue you are facing now, are you sure that the other seeds (tr/vl, MC replicas) aren't also different? In any case, I think that there should be some (simple) ways to trick the [random generators](https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/n3fit_data.py) that it is starting from a n-th trial (as you can see from the table, the seeds for the continued hyperopt exactly restarted from the 2nd Trial).



> Thanks a lot @Cmurilochem for these works! Regarding the issue you are facing now, are you sure that the other seeds (tr/vl, MC replicas) aren't also different? In any case, I think that there should be some (simple) ways to trick the [random generators](https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/n3fit_data.py) that it is starting from a n-th trial (as you can see from the table, the seeds for the continued hyperopt exactly restarted from the 2nd Trial).

Thanks @Radonirinaunimi. This is something that I will need to check. Thanks for pointing this out. At least, the provisory change to `seeds = [seed * k for seed in seeds]` was enough to solve the problem. My test is passing now ate least locally. I am now struggling a bit to make it work in the CI/CD. I hope to solve it soon.

n3fit/src/n3fit/hyper_optimization/hyper_scan.py
```diff
@@ -109,7 +111,15 @@ def hyper_scan_wrapper(replica_path_set, model_trainer, hyperscanner, max_evals=
     # Tell the trainer we are doing hpyeropt
     model_trainer.set_hyperopt(True, keys=hyperscanner.hyper_keys, status_ok=hyperopt.STATUS_OK)
     # Generate the trials object
-    trials = filetrials.FileTrials(replica_path_set, parameters=hyperscanner.as_dict())
+    trials = FileTrials(replica_path_set, parameters=hyperscanner.as_dict())
+    # Initialize seed for hyperopt
+    trials.rstate = np.random.default_rng(42)
```
RoyStegeman, line 116: Should this always be the same instead of setting it from the runcard and/or depending on the replica number? I think this way we try the same setups for each replica (though randomization in the NN may still result in different performances and thus change the setups `hypeorpt` will try). 

If you want it to be fixed, please put any user defined constants at the top of the module. E.g. `HYPEROPT_SEED = 42` and 
```suggestion
    trials.rstate = np.random.default_rng(HYPEROPT_SEED)
```

RoyStegeman: I suppose you were maybe going to change this anyway since you pointed out some troubles with restarting, but the whole `rstate` implementation that passes it through `FileTrials` seems very convoluted for something that is harcoded and only used in a single place. As it is now everything could be replaced by doing `trials=np.random.default_rng(42)` inside the `fmin` function.

n3fit/src/n3fit/model_trainer.py
```diff
@@ -860,7 +862,9 @@ def hyperparametrizable(self, params):
             # and the seed needs to be updated accordingly
             seeds = self._nn_seeds
             if k > 0:
-                seeds = [np.random.randint(0, pow(2, 31)) for _ in seeds]
+                # seeds = [np.random.randint(0, pow(2, 31)) for _ in seeds]
+                # generate seeds for each k-fold from the input `nnseeds`
+                seeds = [seed * k for seed in seeds]
```
RoyStegeman, line 867: Note that the seed can be `0`, so if you'd like the results to be reproducible you may just want to provide a seed for the numpy generator instead.

RoyStegeman (commented):
And these are my comments for now. Thanks for the work so far!

> I suspect that this might be due to the fact that the seeds for the initial weights for each k-fold in difference runs are inherently different (see below).

I suspect so as well - I noticed you froze the seed for the folds but probably not the tensorflow/numpy/python seeds. If the disagreement is due to the effect of random seeds it's of course not a problem for a real run of the hyperoptimization, though for your test to run I guess you'd have to freeze them.

n3fit/src/n3fit/hyper_optimization/filetrials.py
```diff
@@ -61,9 +63,34 @@ class FileTrials(Trials):
     def __init__(self, replica_path, parameters=None, **kwargs):
         self._store_trial = False
         self._json_file = "{0}/tries.json".format(replica_path)
+        self.pkl_file = "{0}/tries.pkl".format(replica_path)
```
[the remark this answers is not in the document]
Cmurilochem: Done! Thanks!

n3fit/src/n3fit/tests/test_hyperopt.py
```diff
@@ -1,13 +1,90 @@
...
+
+    for i in range(n_trials_total):
+        assert restart_json[i]['misc'] == direct_json[i]['misc']
+        assert restart_json[i]['state'] == direct_json[i]['state']
+        assert restart_json[i]['tid'] == direct_json[i]['tid']
+        assert restart_json[i]['result'] == direct_json[i]['result']
+
+
+if __name__ == '__main__':
+    pytest.main()
```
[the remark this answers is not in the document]
Cmurilochem: Done.

n3fit/src/n3fit/tests/test_hyperopt.py
```diff
@@ -1,13 +1,90 @@
...
     Test hyperoptimization features
 """
+import pathlib
+import shutil
+import subprocess as sp
+import json
+import pytest
 
 from numpy.testing import assert_approx_equal
 from n3fit.hyper_optimization import rewards
```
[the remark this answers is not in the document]
Cmurilochem: Done.

n3fit/src/n3fit/tests/hyperopt/hyper-quickcard.yml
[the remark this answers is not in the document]
Cmurilochem: That is right. Thanks! I noticed that the `tests/test_fit.test_hyperopt` is currently skipped with `@pytest.mark.skip(reason="Still not implemented in parallel mode")`. If I just uncomment it, and make conduct this test with the ` `regressions/hyper-quickcard.yml`` runcard I have the error: `total_time = time() - self.starting_time
TypeError: unsupported operand type(s) for -: 'float' and 'NoneType`. 

As expected, if I try to do my hyperopt restart test with the same `regressions/hyper-quickcard.yml` file I have the same error. I am investigating it right now, and it seems that this has to do with any definition in the `hyperscan_config`.

n3fit/src/n3fit/scripts/n3fit_exec.py
Cmurilochem: Yes, I agree that it makes more sense indeed to just use `--restart` instead of `--continue`. If this is OK, I will make the changes.

RoyStegeman: Sure, please go ahead.

n3fit/src/n3fit/model_trainer.py
Cmurilochem: Done.

RoyStegeman: Thanks! Instead of saying done, in principle whenever you make an edit to a line that has been commented on the conversation can simply be marked as resolved - in many cases this will even be done automatically. That way the conversation is folded and the thread is cleaner.

RoyStegeman: Ah you just did ;)

n3fit/src/n3fit/tests/hyperopt/hyper-quickcard.yml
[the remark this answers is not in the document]
RoyStegeman: The only reason to set it to anything other than 1 is that it'll look nicer when presented, so please just set it to 1.

n3fit/src/n3fit/hyper_optimization/hyper_scan.py
Cmurilochem: > Should this always be the same instead of setting it from the runcard and/or depending on the replica number? I think this way we try the same setups for each replica (though randomization in the NN may still result in different performances and thus change the setups `hypeorpt` will try).
> 
> If you want it to be fixed, please put any user defined constants at the top of the module. E.g. `HYPEROPT_SEED = 42` and

Thanks @RoyStegeman. Done. Indeed, this seed could be later input via runcard and we are making it constant for the moment (unless you say otherwise). The idea here is indeed warrant that the hyperopt keep tracks and show same hyperparameter history in direct and restart runs.

Cmurilochem: > I suppose you were maybe going to change this anyway since you pointed out some troubles with restarting, but the whole `rstate` implementation that passes it through `FileTrials` seems very convoluted for something that is harcoded and only used in a single place. As it is now everything could be replaced by doing `trials=np.random.default_rng(42)` inside the `fmin` function.

`trials.rstate=np.random.default_rng(42)` is set when I am just starting a fresh hyperoptimizarion. This just sets a common starting point for all runs done afresh. Once your run is done or was interrupted, `hyperopt` will save the state of the `FileTrials` for every completed trial in the `tries.pkl` pickle file, including the last used `rstate` as attribute.

Now when I want to restart:
```
    # For restarts, reset the state of `FileTrials` saved in the pickle file
    if hyperscanner.restart_hyperopt:
        pickle_file_to_load = f"{replica_path_set}/tries.pkl"
        log.info("Restarting hyperopt run using the pickle file %s", pickle_file_to_load)
        trials = FileTrials.from_pkl(pickle_file_to_load)
```
`trials` now will contain the history of the previous hyperopt as well as implicitly (via its attribute `trial.rstate`) the `rstate` of the last completed trial. **Note** that now (when restarting) the initial `trial.rstate` is not equal to `np.random.default_rng(42)`.  I suspect that this was your doubt; in summary the whole process then do not just resume in just setting  `trials.rstate=np.random.default_rng(42)`.

n3fit/src/n3fit/model_trainer.py
Cmurilochem: That is true @RoyStegeman; my apologies. I could offer another possibility that seems to work along the lines of generating the same seeds every time.
```
            seeds = self._nn_seeds
            if k > 0:
                seeds_gerator = [np.random.default_rng(seed=seed) for seed in seeds]
                seeds = [generator.random() for generator in seeds_gerator]
```
However, if this condition is not necessary (as you suggested above), I will keep the code as it was previously. Thanks for your very detailed and valuable suggestions

n3fit/src/n3fit/hyper_optimization/hyper_scan.py
RoyStegeman: You're right, apologies, I wasn't paying enough attention.

n3fit/src/n3fit/model_trainer.py
RoyStegeman: I don't think this would work and instead you'd want to do something like `generator.integers(0, pow(2, 31))` as seeds should be integers. Because the seeds for each fold are based on the fixed value `self._nn_seeds` they will still be the same for each fold, is this intentional? Previously you prevented this by multiplying with `k`. 

Either way, feel free to return it as you did originally, I had failed to realize that the `seeds` in this line is already generated using a random number generated and is thus extremely unlikely to be 0 (one way to make it absolutely impossible is by changing the lower bound of that rng to 1 instead of 0).

Cmurilochem: You are completely right @RoyStegeman, my apologies. The correct code would be something like
```
if k > 0:
   seeds_gerator = [np.random.default_rng(seed=seed) for seed in seeds]
   seeds = [generator.integers(1, pow(2, 31)) for generator in seeds_gerator]
```
they do not need to be the same but originated from the same seed.

RoyStegeman: Please have a look at my last commit. Now you'll have a different seed per fold, but same between replicas if the fit is run simultaneously in series. While being different between folds may be helpful in case some seeds end up being particularly good for a given setup and thus doing too well on or hyperopt metric, I don't think they'd need to be different between replicas since the hyperparameters are the things we want to vary in a controlled way. If you'd like the seed to depend on the replica as well, you could probably use `self.replicas` to achieve that. 

Once we start fitting a full distribution we of course do want the different replicas within that distribution to have different seeds, so that is something to keep in mind.

Cmurilochem: Thanks for your help @RoyStegeman. 

I tested here and works perfectly! One thing that was actually raised today during our meeting discussion is whether the seeds (although different in every fold) must also be different for every trial, which is not the case here nor in any of my previous suggestions. So, in summary: should the `seeds = [generator.integers(1, pow(2, 30)) * k for generator in rngs]` also depend (somehow) on the trial number as well ?

In relation to the replica dependence. In the multireplica case I suspect that `self._nn_seeds` (= `nnseeds` argument in `ModelTrainer`) is already a list of seeds for each replica. I am not quite certain but, please, correct if I am wrong. 

Thanks once again for your kind help.

RoyStegeman: > In relation to the replica dependence. In the multireplica case I suspect that self._nn_seeds (= nnseeds argument in ModelTrainer) is already a list of seeds for each replica. I am not quite certain but, please, correct if I am wrong.

This I don't know and personally I'd wait for #1661 before worrying about it, though it's good to keep in mind that we fixed the seed here. 

Between trials of different hyperparameter setups I don't think it's a problem that we re-use the same seed, as long as it's different for fits within that trial. In principle we assume the performance of a fit does not depend on the choice of the seed, in practice there may be certain seeds that are particularly favourable or unfavourable for a given setup but that's something we have to keep in mind even if we have a different seed for each trial.

RoyStegeman:
@Cmurilochem Other than moving the `HYPEROPT_SEED` to the runcard, is there anything else you want to do in this PR before merging?

If not I'll have a quick last look but then I'd say this is done?

n3fit/src/n3fit/hyper_optimization/filetrials.py
```diff
@@ -60,10 +64,35 @@ class FileTrials(Trials):
...
         self._store_trial = False
-        self._json_file = "{0}/tries.json".format(replica_path)
+        self._json_file = replica_path / "tries.json"
+        self.pkl_file = replica_path / "tries.pkl"
         self._parameters = parameters
+        self._rstate = None
         super().__init__(**kwargs)
 
+    @property
+    def rstate(self) -> Generator:
```
Radonirinaunimi, line 74: Right now, the linting complains on the types as by default `self._rstate=None`. You can bypass this by importing:
```python
from typing import Union
``` 

```suggestion
    def rstate(self) -> Union[Generator, None]:
```
But personally, I'd omit any static typing here unless it is done consistently throughout the file, and it reduces the import of modules.

n3fit/src/n3fit/hyper_optimization/filetrials.py
```diff
@@ -95,3 +122,21 @@ def new_trial_ids(self, n):
...
+        """Load and return an instance of `FileTrials` from a pickle file.
+
+        If a pickle file from previous run is present this method can be used
+            to instantiate an initial `FileTrials` object to restart.
+        """
+        try:
+            with open(pickle_filepath, "rb") as file:
+                return pickle.load(file)
+        except FileNotFoundError as err:
+            log.error("Failed to open pickle file: %s", err)
```
Radonirinaunimi, line 142: I am slightly confused by this. What exactly should be the intended behavior if the file is missing? Should the program halt?

n3fit/src/n3fit/hyper_optimization/hyper_scan.py
```diff
@@ -174,6 +190,13 @@ def __init__(self, parameters, sampling_dict, steps=5):
         self.parameters = copy.deepcopy(parameters)
         self.steps = steps
 
+        # adding extra options for restarting
+        restart_config = sampling_dict.get("restart")
+        if restart_config:
+            self.restart_hyperopt = True
+        else:
+            self.restart_hyperopt = False
```
Radonirinaunimi, line 198:
```suggestion
        self.restart_hyperopt = True if restart_config else False
```

n3fit/src/n3fit/tests/test_hyperopt.py
```diff
@@ -1,13 +1,87 @@
...
+
+    # read up generated json files
+    restart_json_path = f"{output_restart}/nnfit/replica_{REPLICA}/tries.json"
+    restart_json = load_data(restart_json_path)
+    direct_json_path = f"{output_direct}/nnfit/replica_{REPLICA}/tries.json"
+    direct_json = load_data(direct_json_path)
+
+    # Remove the 'output_restart' and 'output_direct' directories
+    shutil.rmtree(output_restart, ignore_errors=True)
+    shutil.rmtree(output_direct, ignore_errors=True)
```
Radonirinaunimi, line 77: Why not create `output_restart` and `output_direct` as sub-directories under `tmp_path` and hence avoiding removing them manually?

Radonirinaunimi (commented):
Hi @Cmurilochem, here is a quick review. These are mainly formatting/styling and asking clarifications for various points.

> In relation to the replica dependence. In the multireplica case I suspect that self._nn_seeds (= nnseeds argument in ModelTrainer) is already a list of seeds for each replica. I am not quite certain but, please, correct if I am wrong.

Yes, that is correct! The status of master right now is that it can have different seeds per replica for MCseed and NNseed during a multireplica fit, the only seed that is always the same is the TrVlseed $-$ which #1788 will fix.

n3fit/src/n3fit/model_trainer.py
Cmurilochem: Thanks @RoyStegeman for your clarification. From your thoughts, I suspect that we are OK with the current state of the seeds generation and if so I will proceed with the next part of the PR review by @Radonirinaunimi. Thanks once again for you help and valuable suggestions.

Cmurilochem:
> @Cmurilochem Other than moving the `HYPEROPT_SEED` to the runcard, is there anything else you want to do in this PR before merging?
> 
> If not I'll have a quick last look but then I'd say this is done?

Hi @Radonirinaunimi. I have corrected for all @RoyStegeman's suggestions and now will proceed to your. Thanks for you time in reviewing and for your excellent suggestions.

n3fit/src/n3fit/tests/test_hyperopt.py
Cmurilochem: Done. Thanks!

n3fit/src/n3fit/hyper_optimization/filetrials.py
Cmurilochem: If you are doing a restart and `n3fit` cannot not find the file to restart, the program should halt. Thanks for pointing this out! this is not exactly what the `except` part is actually doing in the snippet you highlight. I have corrected the code to now `raise` this exception properly.

n3fit/src/n3fit/hyper_optimization/filetrials.py
Cmurilochem: I followed your suggestion and removed type hints in order to ensure consistency throughout. I added instead a more descriptive docstring. Thanks!

RoyStegeman (approved):
Thanks, looks good! 

`HYPEROPT_SEED` is going to remain fixed for now?

n3fit/src/n3fit/hyper_optimization/filetrials.py
Radonirinaunimi: Great! Thanks.

Cmurilochem:
> Thanks, looks good!
> 
> `HYPEROPT_SEED` is going to remain fixed for now?

Hi @RoyStegeman that is what we initially thought and implemented so far. But If you think it would be better, I could try to add a new entry into the runcard so that (like other seeds) the user could have control over it. Please, just let me know what do you think ?

RoyStegeman:
Don't worry I'm fine either way, was just making sure it wasn't something you had forgotten about as I had understood you planned to take it from a runcard seed.

Cmurilochem:
> Don't worry I'm fine either way, was just making sure it wasn't something you had forgotten about as I had understood you planned to take it from a runcard seed.

Great! Maybe we could add this feature in the near future as long as we feel the need to do so. I will keep this in mind.

Radonirinaunimi:
Thanks a lot @Cmurilochem! I guess the only minor thing missing in order to merge this PR is a small note in the documentation (at the end of https://docs.nnpdf.science/n3fit/hyperopt.html?highlight=hyperopt) describing how one can restart hyperopt.

Cmurilochem:
> Thanks a lot @Cmurilochem! I guess the only minor thing missing in order to merge this PR is a small note in the documentation (at the end of https://docs.nnpdf.science/n3fit/hyperopt.html?highlight=hyperopt) describing how one can restart hyperopt.

Hi @Radonirinaunimi. I could add a note after [Changing the hyperoptimization target](https://github.com/NNPDF/nnpdf/blob/ba84878353e3ed950426ae6a12674abce2dfb903/doc/sphinx/source/n3fit/hyperopt.rst?plain=1#L319) and let you know after the commit has been done.

RoyStegeman:
Ah good point! It needs to be documented of course. Completely forgot about that :sweat_smile:

doc/sphinx/source/n3fit/hyperopt.rst
```diff
@@ -370,3 +370,19 @@ The figure of merit will be the difference between the :math:`\chi2` of the seco
...
    L_{\rm hyperopt} = \chi^{2}_{(1) \rm pdferr} - \chi^{2}_{(2)}
+
+
+Restarting hyperoptimization runs
+---------------------------------
+
+In addition to the ``tries.json`` files, hyperparameter scans also produce ``tries.pkl`` `pickle <https://docs.python.org/3/library/pickle.html>`_ files,
+which are located in the same directory as the corresponding ``tries.json`` file.
+The generated ``tries.pkl`` file stores the complete history of a previous hyperoptimization run, making it possible to resume the process using the ``hyperopt`` framework.
+To achieve this, you can utilize the ``--restart`` option within the ``n3fit`` command, e.g.,:
```
RoyStegeman, line 381:
```suggestion
To achieve this, you can use the ``--restart`` option within the ``n3fit`` command, e.g.,:
```

Radonirinaunimi (approved):

Cmurilochem:
Hi @RoyStegeman, @Radonirinaunimi and @scarlehoff. Please, let me know whether I could merge this PR after the approval of Roy and Tanjona. Thanks you all for your very valuable suggestions.

RoyStegeman:
Yes, please merge this

PR #1825, opened 2023-10-25
dis+jets implementation

t7phy:


buildmaster/H1_1JET_319GEV_290PB-1_DIF/metadata.yaml
```diff
@@ -0,0 +1,120 @@
...
+    kinematics:
+      variables:
+        pT: {description: "mean transverse momentum", label: '$pT$', units: "$GeV$"}
+        q_sqr: {description: "momentum transferred squared", label: '$q^2$', units: "$GeV^2$"}
+        sqrt_s: {description: "center of mass energy", label: '$\sqrt{s}$', units: "$GeV$"}
+      file: kinematics.yaml
+    data_central: data.yaml
+    data_uncertainties:
+      - uncertainties.yaml
+    kinematic_coverage: [pT, q_sqr, sqrt_s]
```
scarlehoff, line 34: Instead of q_sqr, use Q2 since this is more or less the standard notation

buildmaster/H1_1JET_319GEV_290PB-1_DIF/metadata.yaml
```diff
@@ -0,0 +1,120 @@
...
+  version: 1
+
+version: 1
+version_comment: "NA"
+
+implemented_observables:
+
+  - observable: {description: "double differential cross section w.r.t. q^2 and pT", label: '$\frac{d^2\sigma}{dq^{2}dpT}$', units: "$pb/GeV^3$"}
+    observable_name: PTQ2
+    process_type: JET
```
scarlehoff, line 22: I think it is better to use DIS for the time being for these. They are going to share more logic (cuts and so on) with DIS than with JET (which is usually p p -> single jet)

@enocera or maybe we should create a new process type for DIS + Jet? These are the first datasets we implement for this, right?

buildmaster/H1_2JET_319GEV_351PB-1_DIF/metadata.yaml
```diff
@@ -0,0 +1,73 @@
...
+  version: 1
+
+version: 1
+version_comment: 'NA'
+
+implemented_observables:
+
+  - observable: {description: "double differential cross section w.r.t. q^2 and pT", label: '$\frac{d^2\sigma}{dq^{2}dpT}$', units: "$pb/GeV^3$"}
+    observable_name: PTQ2
+    process_type: DIJET
```
scarlehoff, line 22: Same comment as for JET.

validphys2/src/validphys/cuts/filters.yaml
```diff
@@ -388,3 +388,8 @@
   FNS: FONLL-C
   IC: True
   rule: "Q2 > 8"
+
+- dataset: H1_1JET_319GEV_290PB-1_DIF_PTQ2
+  reason: |
+    There are 8 Q^2 bins, however the grid for the 1st bin is unavailable.
+  rule: "q_sqr >= 8"
```
scarlehoff, line 395: Remember to change this to Q2 as well when you search-and-replace

scarlehoff (commented):
Haven't tested them yet, only had a look through the metadata, some general comments.

buildmaster/H1_1JET_319GEV_290PB-1_DIF/metadata.yaml
enocera: @scarlehoff , @t7phy : my recommendation would be to use two new labels, e.g. DIS_JET and DIS_DIJET. The reason being that the perturbative structure of the partonic cross section shares features both with DIS and with jet/dijet production. So this is relevant when correlating MHOU. Incidentally, note that these DIS+jet data will share with DIS the usual cuts (which are not that relevant, if my recollection is correct), BUT they will also have additional cuts, like pp jets (which are far more relevant). I would include these rules in filters.yaml. Would you like me to do so?

scarlehoff: Yes, that would be great. We have some logic in validphys to apply DIS cuts even if the process type has extra qualifiers so DIS_JET should work fine.
I'm worried about the kinematic variables however, here Q2 is shared but we might need more or a way of mapping between the old (since the other DIS would be old cd) and new so if you add the new cuts we can start testing where (if) that fails...

buildmaster/H1_1JET_319GEV_290PB-1_DIF/metadata.yaml
t7phy: done

validphys2/src/validphys/cuts/filters.yaml
t7phy: done

buildmaster/H1_1JET_319GEV_290PB-1_DIF/metadata.yaml
t7phy: renamed to DIS_JET

buildmaster/H1_2JET_319GEV_351PB-1_DIF/metadata.yaml
t7phy: renamed to DIS_DIJET

scarlehoff:
@enocera one question about the generic DIS cuts that we have in the code (I have a weak recollection of having talk about this or something similar), should they also apply by default to these datasets?

If yes, maybe it is good to add the `x` to the kinematics of DIS+j  since it is used by some of the cuts?

enocera:
The DIS+jet data sets are NOT subject to the generic cuts for DIS we have in the code (certainly not to the x cut; for Q, I'd say that all the data points are above the nominal cut, see Table 2.2 in the NNPDF4.0 paper. So, even if the Q2 cut is there and even if it works, it should have no effect). However, we may want to have a cut in pT, which, I hope, can be implemented in filters.yaml. I need to check the HERA paper to retrieve the optimal value for the cut in pT.

enocera:
The kinematic cut to be implemented is (pT^2+Q^2)>100 GeV2 (that is, one should retain points with pT^2+Q^2 larger than 100 GeV2).

enocera:
Using the aforementioned cut, you should see the reduction of data points summarised in Table 1 of  https://arxiv.org/pdf/2112.01120.pdf.

scarlehoff:
> The kinematic cut to be implemented is (pT^2+Q^2)>100 GeV2

Ok, this should work fine.

> The DIS+jet data sets are NOT subject to the generic cuts for DIS we have in the code

Then we need to find a new name (because I'd like to minimize the changes to other parts of the code). Sadly there's a lot of logic that affect anything that has "DIS" in the name. I will play around a bit with what @t7phy has implemented and propose something.

enocera:
> Then we need to find a new name (because I'd like to minimize the changes to other parts of the code). 

May I suggest HERAJET or something similar?

validphys2/src/validphys/cuts/filters.yaml
```diff
@@ -388,3 +388,8 @@
   FNS: FONLL-C
   IC: True
   rule: "Q2 > 8"
+
+- dataset: H1_1JET_319GEV_290PB-1_DIF_PTQ2
+  reason: |
+    There are 8 Q^2 bins, however the grid for the 1st bin is unavailable.
+  rule: "Q2 >= 8"
```
scarlehoff, line 395: Please move this a bit up in the file to where the dataset cuts are.

buildmaster/H1_1JET_319GEV_290PB-1_DIF/metadata.yaml
```diff
@@ -0,0 +1,120 @@
+setname: 'H1_1JET_319GEV_290PB-1_DIF'
+
+nnpdf_metadata:
+  nnpdf31_process: "1JET"
```
scarlehoff, line 4: As per https://github.com/NNPDF/nnpdf/pull/1825#issuecomment-1786778365 please change all 1JET / DIS_JET to HERAJET
and 2JET / DIS_DIJET to HERADIJET.

(this is also nice in that we will have the nnpdf31_process agree with the process_type for a change)

buildmaster/H1_1JET_319GEV_290PB-1_DIF/metadata.yaml
```diff
@@ -0,0 +1,120 @@
...
+      - uncertainties.yaml
+    kinematic_coverage: [pT, Q2, sqrt_s]
+    plotting:
+      kinematics_override: identity
+      dataset_label: '$H1 Jet 290 pb^{-1} distribution: \frac{d^2\sigma}{dq^{2}dpT}$'
+      plot_x: pT
+      x_scale: log
+      figure_by: 
+        - Q2
+    theory:
```
scarlehoff, line 42:
```suggestion
    theory:
        shifts:
            H1_1JET_319GEV_290PB-1_DIF_PTQ2_BIN1: 6
```

This tells validphys to read the fktables such that the first point in the fktables is actually point 6.

This flag used to be called `apfelcomb::shifts` because it was not needed for any of the pineappl tables but since it clearly will be (particularly if we are not the ones generating the grids) I'll change the name in the reader to simply ``shifts``.

I'm going to update the reader to deal with this.

buildmaster/H1_2JET_319GEV_290PB-1_DIF/metadata.yaml
```diff
@@ -0,0 +1,75 @@
...
+    observable_name: PTQ2
+    process_type: DIS_DIJET
+    ndata: 48
+    tables: [9, 10, 11, 12, 13, 14, 15, 16]
+    kinematics:
+      variables:
+        pT: {description: "mean transverse momentum", label: '$\langle pT \rangle$', units: "$GeV$"}
+        Q2: {description: "momentum transferred squared", label: '$q^2$', units: "$GeV^2$"}
+        sqrt_s: {description: "center of mass energy", label: '$\sqrt{s}$', units: "$GeV$"}
+      file: kinematics_dijet.yaml
```
scarlehoff, line 30:
```suggestion
      file: kinematics.yaml
```

there's no kinematics_dijet

scarlehoff (commented):
Thank you very much for this.

Here's a report with the `plot_fancy` for all these datasets. Some problems at low-pt that hopefully are "fixed" by either the cut that @enocera suggested or by scale variations.

https://vp.nnpdf.science/aVePALDTQ6uNzNRYwr5RCA==

I would remove "distribution: " from the `dataset_label` and maybe don't cover the name with the latex tags.

So instead of 

```
'$H1 Jet 290 pb^{-1} distribution: \frac{1}{\sigma}\frac{d^2\sigma}{dq^{2}dpT}$'
```
I'd do

```
'H1 Jet 290 pb$^{-1} \frac{1}{\sigma}\frac{d^2\sigma}{dq^{2}dpT}$'
```

or if you want to keep all, add spaces in some way
```
'$H1\ Jet\ 290 pb^{-1}\ distribution:\ \frac{1}{\sigma}\frac{d^2\sigma}{dq^{2}dpT}$'
```

scarlehoff:
> The kinematic cut to be implemented is (pT^2+Q^2)>100 GeV2 (that is, one should retain points with pT^2+Q^2 larger than 100 GeV2).

Do we want (pT^2+Q^2)>100 also for Zeus or only for H1?
If we want it also for Zeus I would suggest to sue `pT` instead of `ET` as the kinematic variable for Zeus datasets.

enocera:
The cut should apply to both H1 and ZEUS. @t7phy can you please take care of changing the name of the variable for ZEUS data? Thanks.

validphys2/src/validphys/cuts/filters.yaml
t7phy: which line exactly?

buildmaster/H1_1JET_319GEV_290PB-1_DIF/metadata.yaml
t7phy: done

buildmaster/H1_1JET_319GEV_290PB-1_DIF/metadata.yaml
t7phy: done

validphys2/src/validphys/cuts/filters.yaml
scarlehoff: The exact line doesn't matter, but right now this is together with the `process_type'` while all datasets are at the top.

t7phy:
@scarlehoff @enocera but is it a safe choice to take the mass of a jet/dijet as 0 (by naming ET as pT)? Wouldn't it be better if we implemented cuts on pT and ET separately?

scarlehoff:
What does `pT` (and `ET`) refer to? If this is the pT/eT of the leading jet it should be fine.

t7phy:
> What does `pT` (and `ET`) refer to? If this is the pT/eT of the leading jet it should be fine.

Ok, if they are the same, is it enough to rename the variables or should I also relabel it?

scarlehoff:
_if_ they are the same yes, but they might not be. Opening the paper I see the E has a bar on top most of the time which suggest to me that is some average of both jets. I hope it is defined somewhere in the paper (maybe in the reference for the grids?)

If it is not well defined I agree it is better to leave them separated (but then you will need to implement the pT**2+Q2 >= 100 cut separately for every dataset for which it does apply.
Or create two separate processes type HERAJETPT and HERAJETET so that you can write them as process-level cuts.

t7phy:
There are two Zeus jet datasets which have transverse energy of the jet. The mean transverse energy is only for one of the sets: Zeus dijet. Anyway, for all ZEUS datasets, the minimum value of Q2 is always above 100, so can we just forget it, and leave it as it is?

scarlehoff:
> so can we just forget it, and leave it as it is?

If the difference is for dijet vs jet yes (because they are different processes).
Otherwise the cut will crash when it tries to apply a cut on pT to a dataset without pT in the kinematics.

scarlehoff:
A separated question... wouldn't it be a good idea to use merge the JET and DIJET "observables" and have the "container process" as `HERAJETS` and the observables be `JET` and `DIJET`? (since the data is coming from the same hepdata entry).

Sorry if this is just adding noise, I'm preparing the talk for PDF4LHC and the issue came to mind.

t7phy:
> A separated question... wouldn't it be a good idea to use merge the JET and DIJET "observables" and have the "container process" as `HERAJETS` and the observables be `JET` and `DIJET`? (since the data is coming from the same hepdata entry).
> 
> Sorry if this is just adding noise, I'm preparing the talk for PDF4LHC and the issue came to mind.

Well we had categorized jets and dijets as separate processes so it wouldn't be right to have them together. We had agreed that different processes will be separate even when coming from the same source. Besides, it is worth noting that some of the datasets actually have more observables (besides pT, q2) and we just dont use them for the time being, but who knows if we may need them later on.

scarlehoff:
> Well we had categorized jets and dijets as separate processes 

I guess this is precisely the point that I'm not sure I agree on... one could have JET_pt, DIJET_m2, etc. 
The limitations of having the same process name don't really apply to the new commondata format where the plotting info is defined in the metadata of the observable.

t7phy:
> > Well we had categorized jets and dijets as separate processes
> 
> I guess this is precisely the point that I'm not sure I agree on... one could have JET_pt, DIJET_m2, etc. The limitations of having the same process name don't really apply to the new commondata format where the plotting info is defined in the metadata of the observable.

The idea of having different processes separated was not to do with the format of commondata but rather from an organizational perspective. A way of thinking about this is that we would like to restrict a single dataset to a single set of events.

scarlehoff:
Thanks! Since there's already a branch with the data available, I will try using those old commondata files for PDF4LHC (and, if they don't work because things have changed too much, I can just use the central data from there at least).

enocera:
@scarlehoff OK. Also note that the numbers published in the NNPDF4.0 paper may be incorrect for the H1 low-q 1jet data set (which is called `H1_1JET_319GEV_290PB-1_DIF` in this PR). I get a chi2 (no PDF uncs, no scale vars) which is higher than the number in the paper (3.26 instead of 1.86). I am able to reproduce all the other numbers.

scarlehoff:
Strange, for that one I'm actually getting a (too good) chi2 of 1.01 when I use @t7phy's fktables and yours NP corrections.

scarlehoff:
No, I was using the data from your branch in the `nnpdfcpp` folder.
Could you also add there the systypes files (since the ones available in the branch don't seem to be compatible with these)

scarlehoff:
image

These are the numbers I find with this datasets. I'm going to update the chi2 plots in the presentation and tomorrow I'll try to update also the data-theory comparison plots.

I'm not applying any cuts. While in principle it would be possible I would need to check that the kinematics are the same / order / etc (because the cuts I've implemented for the new ones are looking at the name of the variables).

enocera:
YES!! These are also the numbers that I get with the code I wrote for NNPDF40! Question: do you apply the NP corrections? They must be applied. The reason why the numbers published in NNDPF4.0 are different is possibly because of typos. IT remains to understand why the current PR does not produce the same numbers, but it should be possible by comparing it with the old implementation. And of course the old implementation may be wrong.

scarlehoff:
Yes. I don't understand what's happening with the first one though, but at this point is one among many.

For the Atlas/CMS you think it is ok to use the regularized covmats?

enocera:
@scarlehoff I'm lookingn into the ATLAS/CMS data now and will be done this evening. However I expect that things are OK there - I'm not surprised that some amount of decorrelation is needed.

t7phy:
@enocera @scarlehoff I have changed the H1*351pb-1* datasets to now use the tables given by kt (instead of anti-kt). I think with this, the numbers should match with the old implementation.

t7phy:
This PR is now fully ready with all the suggestions from above implemented.

scarlehoff:
Thanks for this.

I guess you already tested this with the reader and checked that it worked after your changes?

t7phy:
> Thanks for this.
> 
> I guess you already tested this with the reader and checked that it worked after your changes?

Thanks for reminding me of this, I was counting on from pdf4lhc times. I just tried to do a DTC and got this error: `KeyError: 'Label HERAJET not found in PROCESS_DESCRIPTION_LABEL'` , so ofc we need HERAJET and HERADIJET added to possible process names in the reader.

scarlehoff:
Completely forgot about them! 

This is a great candidate to start adding only the entries that are required (such as the q-x mapping)

@enocera @felixhekhorn  how would you compute `x` for the DIS +1/2j? There's no simple x the moment you have a recoil at LO...

felixhekhorn:
> Completely forgot about them!
> 
> This is a great candidate to start adding only the entries that are required (such as the q-x mapping)
> 
> @enocera @felixhekhorn how would you compute `x` for the DIS +1/2j? There's no simple x the moment you have a recoil at LO...

if possible I'd like to change the behaviour on the kinPlot to a more physical one (in my opinion) ... @enocera, I guess there are reasons on why it was chosen in the old way it was chosen, so maybe I'm missing something ...

however, speaking as a theorist I would argue the following way:
1. for fully inclusive DIS, fully inclusive DY and SIA there is a 1:1 mapping at LO between measured variables and the variables in the coll. distribution, because they all rely basically in the same diagram (suitable rotated around)
2. for anything else there is no clear mapping between measured stuff and PDF(/FF) variables, but there is a convolution involved (or in more physical speak: there is a recoiling something), i.e. an integral with running PDF variables
3. what I would suggest for the `x`* mapping is to always map to the *lower* bound of the integral (*a name is just a name, but exactly for this mess we avoid in yadism this name)
   - this would leave 1. unaltered at higher order, since the lower integration bound is always the measured Bjorken-x
   - we would thus need to read the kinPlot not as a point, but rather as a range: the dataset is sensitive between the marker and x=1
   - in the specific case of DIS + 1/2j, I would need to look up the specific formulae ....
   - grafik specifically I don't like this in the old format: there is no PDF at x=1 ...
   - (another option would be to mark the spot where the contribution to the observable is the biggest, but a) this is complicated to determine and b) is a PDF dependent statement)
4. for the `Q2` mapping, I suggest to do the same thing: always use the *lowest* (central) scale
   - same as above: this would leave 1. unaltered, since there is only a static scale choice
   - same as above: we would need to read the kinPlot as a range: the dataset is sensitive between the marker and Q2=oo (which does not harm due to asymptotic freedom)
   - the choice of the *lowest* scale can e.g. be relevant in the case at hand as jet observables can and do have a dynamic scale

scarlehoff:
I'm not sure the x coverage is 100% correct, maybe some of the approximations were too approximated (or I made a mistake coding the formula).

https://vp.nnpdf.science/vxS8mWGySsu-cehRFnhSOA==

Both the process label and the formula can be modified in this file: https://github.com/NNPDF/nnpdf/blob/7e6599a821b6f1d0fb30c705339d3c047da46c99/validphys2/src/validphys/process_options.py#L233

t7phy:
The formula looks correct (from what ERN told us that day). Why do you say the x plot seems incorrect?

scarlehoff:
I've merged and rebase this branch onto master: https://github.com/NNPDF/nnpdf/commit/625372d97d23ff30afcc78c666944de0f794e59a

scarlehoff:
> The formula looks correct (from what ERN told us that day). Why do you say the x plot seems incorrect?

Almost 10^-7 looks too small for me. Maybe there was a square missing.

felixhekhorn:
> > The formula looks correct (from what ERN told us that day). Why do you say the x plot seems incorrect?
> 
> Almost 10^-7 looks too small for me. Maybe there was a square missing.

can you check against the grids?

PR #1826, opened 2023-10-25
Re-implementation of the LHCb Collider DY Datasets

Radonirinaunimi:
The following PR implements the LHCb Collider DY datasets in the new commondata format. The table summarizes the status of the implementation and provides general comments on the comparisons w.r.t. the old implementation. The dataset names in the new tables are the new ones and the mapping to the old ones is tracked in [dataset_names.yml](https://github.com/NNPDF/nnpdf/blob/collider_dy_ncd/buildmaster/dataset_names.yml).

| Dataset Name  | Comparison vs. Old |  General Comments | Status  |
|---|---|---|---|
| LHCB_Z0_7TEV_DIELECTRON_Y  |☑️   |   |  ☑️  |
| LHCB_Z0_8TEV_DIELECTRON_Y  |  ☑️  |   |  ☑️  |
| LHCB_Z0_13TEV_Y-DIELECTRON  |  ☑️  |   |  ☑️  |
|  LHCB_Z0_13TEV_Y-DIMUON |  ☑️  |   |  ☑️  |
| LHCB_Z0_7TEV_MUON_Y |☑️   |   | ☑️   |
| LHCB_DY_7TEV_MUON_Y |  ☑️  |   |  ☑️  |
| LHCB_Z0_8TEV_MUON_Y |☑️   |   | ☑️    |
| LHCB_DY_8TEV_MUON_Y |  ☑️  |   |  ☑️  |
| LHCB_WPWM_7TEV_MUON_Y | ☑️  |   |  ☑️  |
| LHCB_WPWM_8TEV_MUON_Y |  ☑️ |   |  ☑️  |

Final Report: https://vp.nnpdf.science/4ub8cAYxTa2xPP6oM6X4_w==


buildmaster/LHCB_DY_13TEV_DIELECTRON/metadata.yaml
```diff
@@ -0,0 +1,53 @@
...
+  nnpdf31_process: "DY NC"
+  experiment: "LHCb"
+
+implemented_observables:
+- observable_name: "y"
+  observable:
+    description: "Differential cross-section of Z-->ee as a function of Z-rapidity"
+    label: r"$d\isgma / d|y|$"
+    units: "[fb]"
+  process_type: "DY"
```
scarlehoff, line 27: Since this is a reimplementation, please use the same `process_type` that it had before.
For [this dataset/observable](https://github.com/NNPDF/nnpdf/blob/master/nnpdfcpp/data/commondata/DATA_LHCB_Z_13TEV_DIELECTRON.dat) is `EWK_RAP`.

(the process types are [here](https://github.com/NNPDF/nnpdf/blob/94917a4bdddc5e99f1835d61e7a0540f8ef07052/validphys2/src/validphys/filters.py#L19))

buildmaster/LHCB_DY_13TEV_DIELECTRON/metadata.yaml
```diff
@@ -0,0 +1,53 @@
...
+version_comment: "First implementation"
+
+# References
+arXiv:
+  url: "https://arxiv.org/abs/1607.06495"
+iNSPIRE:
+  url: "https://inspirehep.net/literature/1477581"
+hepdata:
+  url: "https://www.hepdata.net/record/ins1477581"
+  version: 1
```
scarlehoff, line 15: Since this dataset and `DIMUON` are coming from the same hepdata, I think it would be better to have them both as

`LHCB_DY_13TEV_Y` for instance and have the observables be `DIMUON` `DIELECTRON`

buildmaster/LHCB_DY_13TEV_DIELECTRON/metadata.yaml
```diff
@@ -0,0 +1,53 @@
+# Previously called `LHCB_Z_13TEV_DIELECTRON`
+# author: "Tanjona R. Rabemananjara"
```
scarlehoff, line 2: Please don't add authorship, git keeps track of that (and more accurately since the data might change with time hence the versioning).

The "previously called" might make more sense as a "version comment" (but in any case not mandatory as it will be in `dataset_names` as well)

buildmaster/LHCB_DY_13TEV_DIELECTRON/metadata.yaml
```diff
@@ -0,0 +1,53 @@
...
+  experiment: "LHCb"
+
+implemented_observables:
+- observable_name: "y"
+  observable:
+    description: "Differential cross-section of Z-->ee as a function of Z-rapidity"
+    label: r"$d\isgma / d|y|$"
+    units: "[fb]"
+  process_type: "DY"
+  ndata: 17
```
scarlehoff, line 28: `ndata` is no longer necessary, we read it directly from the data.

buildmaster/LHCB_DY_13TEV_DIELECTRON/metadata.yaml
```diff
@@ -0,0 +1,53 @@
...
+  tables: [5]
+  npoints: [17]
+
+  # Plotting information
+  plotting:
+      dataset_label: "LHCb $Z\\to ee$"
+      plot_x: y
+      y_label: '$d\sigma_{Z}/dy$ (fb)'
```
scarlehoff, line 36: This is a tricky one and related to the comment above, after finding the `process_type` you will need to look at the `PLOTTINGTYPE_<process_type>` file as well. In this case:

https://github.com/NNPDF/nnpdf/blob/master/nnpdfcpp/data/commondata/PLOTTINGTYPE_EWK_RAP.yaml

So the plotting is missing the kinematic override: `ewk_rap_sqrt_scale`

The kinematic override is then used to prepare the plots.

Btw, not mandatory, but if you write the list in `kinematic_coverage` in the same order as in https://github.com/NNPDF/nnpdf/blob/94917a4bdddc5e99f1835d61e7a0540f8ef07052/validphys2/src/validphys/filters.py#L19 things are more likely to work out of the box.
Of course, the idea is that (eventually) the order is not considered anymore, just the names, but for that we first need to have many datasets implemented so that we can enforce some common naming for the variables.

buildmaster/LHCB_DY_13TEV_DIELECTRON/metadata.yaml
```diff
@@ -0,0 +1,53 @@
...
+  kinematic_coverage: [y, M2, sqrt_s]
+
+  kinematics:
+      variables:
+          var_1: {description: "Z boson rapidity", label: "$y$", units: ""}
+          var_2: {description: "Z boson Mass", label: "$M^2$", units: "$GeV^2$"}
+          var_3: {description: "Center of Mass Energy", label: '$\sqrt{s}$', units: "$GeV$"}
```
scarlehoff, line 43: Please use the variable names as the key.

scarlehoff (commented):
I know this is just a draft, but since I was already reviewing some other dataset PR I thought it was better to add some quick comments so that they are automatically applied to the other datasets that are going to be implemented!

buildmaster/LHCB_DY_13TEV_DIELECTRON/metadata.yaml
cschwan: `\isgma` should be `\sigma`.

Radonirinaunimi:
Thanks for the quick feedback @scarlehoff. I was indeed about to address some of the points you raised above after reviewing everything and tested with https://github.com/NNPDF/nnpdf/pull/1678. I haven't done so yet but I already addressed most of your comments in [cc85bb9](https://github.com/NNPDF/nnpdf/pull/1826/commits/cc85bb9a2ad2f84f6f76062a0aaf90ad82c1959d).

Maybe as a general comment, it would be useful to have https://github.com/NNPDF/nnpdf/pull/1708 up to date, to include `kinematic_override` for example? (This was the reason why I forgot about it although we discussed it)

buildmaster/LHCB_DY_13TEV_DIELECTRON/metadata.yaml
Radonirinaunimi: Yes, they indeed come the same hepdata. However, in the di-electron production the last bin is removed, so I think these really need to be separate as per https://github.com/NNPDF/nnpdf/pull/1813#issuecomment-1755702965.

buildmaster/LHCB_DY_13TEV_DIELECTRON/metadata.yaml
Radonirinaunimi: They are ordered in the same way but the 1st and 3rd variable are named differently, should we change them from `y` to `etay` and `sqrts` to `sqrt_s` to conform with `"EWK_RAP": ("etay", "M2", "sqrts")`?

buildmaster/LHCB_DY_13TEV_DIELECTRON/metadata.yaml
scarlehoff: No, these are different observables so they can have different number of data points (and of course the fktables should be different)

buildmaster/LHCB_DY_13TEV_DIELECTRON/metadata.yaml
scarlehoff: No, just the order is fine (and not even necessary, just making the success a bit more likely, of course, same name even greater chance). The standardization of the naming I would leave for the end.

scarlehoff:
`kinematic_override` might not be necessary (at the moment for the ones that don't use it I asked @t7phy to just put identity for the time being until I decide what to do with the xq plot) but when doing a re-implementation of the old commondata the whole plotting file need to be ported over.
Another option is of course modifying the kinematics so the override is not necessary but I think for the re-implementation it is better to leave them as close to the original as possible.

buildmaster/LHCB_DY_13TEV_DIELECTRON/metadata.yaml
Radonirinaunimi: So, we agree to leave these as they are currently?

scarlehoff: No, I mean that you can have more than one observable in a single metadata.yaml
One of them can be `DIMUON` (and have n datapoints) and the other `DIELECTRON` and have a different number `m` of datapoints.

Radonirinaunimi: Done!

cschwan:
Commit eb4540418cf5d3b53671b3dac6392ed5d9f99a9f adds the dataset that I've been working on, which still has lots of unfilled fields (marked with TODO). If someone could give this a review I'd be grateful!

buildmaster/ATLAS_DY_7TEV_LOMASS_EXT/metadata.yaml
```diff
@@ -0,0 +1,49 @@
...
+      units: "TODO"
+    # data is in table 6, breakdown of systematic uncertainties is in table 7
+    tables: [6, 7]
+    npoints: [6, 0]
+    process_type: "EWK_MLL"
+    plotting:
+      dataset_label: "TODO"
+      plot_x: TODO
+      figure_by:
+        - TODO
```
Radonirinaunimi, line 32: The plotting information you can always refer to the old implementation [PLOTTING_ATLASLOMASSDY11EXT.yaml](https://github.com/NNPDF/nnpdf/blob/master/nnpdfcpp/data/commondata/PLOTTING_ATLASLOMASSDY11EXT.yaml). In your case, `k2` is `mll2`.

buildmaster/ATLAS_DY_7TEV_LOMASS_EXT/metadata.yaml
```diff
@@ -0,0 +1,49 @@
...
+      kinematic_coverage:
+        - TODO
+        - TODO
+        - TODO
```
Radonirinaunimi, line 36: The kinematic coverage is always just the list of kinematic variables.
```suggestion
      kinematic_coverage: [mll, mll2, sqrts]
```

buildmaster/ATLAS_DY_7TEV_LOMASS_EXT/metadata.yaml
```diff
@@ -0,0 +1,49 @@
...
+      kinematics:
+        variables:
+          var_1: {description: "TODO", label: "TODO", units: "TODO"}
+          var_2: {description: "TODO", label: "TODO", units: "TODO"}
+          var_3: {description: "TODO", label: "TODO", units: "TODO"}
```
Radonirinaunimi, line 41:
```suggestion
          mll: {description: "invariant mass of the lepton pair", label: "$m_{\ell \ell}$", units: "$GeV$"}
          mll2: {description: "invariant mass square of the lepton pair", label: "$m_{\ell \ell}$", units: "$GeV^2$"}
          sqrts: {description: "center of mass energy", label: "$\sqrt{s}$", units: "$GeV$"}
```

buildmaster/ATLAS_DY_7TEV_LOMASS_EXT/metadata.yaml
```diff
@@ -0,0 +1,49 @@
...
+arXiv:
+   url: "https://arxiv.org/abs/1404.1212"
+iNSPIRE:
+   url: "https://inspirehep.net/literature/1288706"
+hepdata:
+   url: "https://www.hepdata.net/record/ins1288706"
+   version: 1
+
+nnpdf_metadata:
+   nnpdf31_process: "EWK_MLL"
```
scarlehoff, line 15: The `nnpdf31_process` you can find in the `plotting` file of the "old" commondata that correspond to this dataset. This is (as you correctly filled below) the `process_type`.

buildmaster/ATLAS_DY_7TEV_LOMASS_EXT/metadata.yaml
```diff
@@ -0,0 +1,49 @@
...
+   experiment: "ATLAS"
+
+implemented_observables:
+  - observable_name: "TODO"
+    observable:
+      description: "TODO"
+      label: "TODO"
+      units: "TODO"
```
Radonirinaunimi, line 23:
```suggestion
    observable:
      description: "differential cross-section for Z/gamma* -> ll production as function of the dilepton invariant mass"
      label: "$d\sigma_{Z/\gamma^{*}}/dM_{ll}$"
      units: "[pb / GeV]"
```

buildmaster/ATLAS_DY_7TEV_LOMASS_EXT/metadata.yaml
scarlehoff, line 41: Rather than `var_1`, remember that these need to be the name of the variables 

(I think I'm going write a comment in the documentation because it's already two of you with the same error so probably the documentation was not clear :__ )

buildmaster/ATLAS_DY_7TEV_LOMASS_EXT/metadata.yaml
```diff
@@ -0,0 +1,49 @@
...
+    # data is in table 6, breakdown of systematic uncertainties is in table 7
+    tables: [6, 7]
+    npoints: [6, 0]
+    process_type: "EWK_MLL"
+    plotting:
```
scarlehoff, line 28: Remember to look into the `PLOTTINGTYPE_EWK_MLL` (or something like that) file to fill this part as it can have extra options.

Radonirinaunimi (commented):
Thanks @cschwan! I've tried to address the TODO's below. A minor comment, you should also add the mapping to the old names in [dataset_names.yml](https://github.com/NNPDF/nnpdf/blob/collider_dy_ncd/buildmaster/dataset_names.yml).

Radonirinaunimi:
@scarlehoff, the LHCb data could now be parsed properly (modulo some caveats below). I've produced a data vs theory comparison for the [old](https://vp.nnpdf.science/YukphGh8RiuOn5eXMMj0bQ==/) and [new](https://vp.nnpdf.science/lZVz3sDsThWoT387IOBJug==) implementations.

There are two (three) main issues that are left pending:
- In the old implementation, it is possible to add custom `extra_labels` to organize the plotting (see part below). This doesn't work out of the box here, am I missing something? For the time being, I left `figure_by` empty which renders the plots messy.
https://github.com/NNPDF/nnpdf/blob/efec8052ca70d13af613ed83455f178769c6c0bb/buildmaster/LHCB_WPWM_7TEV_MUON/metadata.yaml#L35-L55
- You'll notice that the `theory_central` for the new implementations are of by a factor of $10^3$. This is exactly the value of the `conversion_factor` in the yamldb. So I am wondering if this is applied correctly?
- (Slightly unrelated), you'll also notice that the report with the new commondata is missing two datasets (w.r.t. old). This is because if I add a few more datasets, then `pandoc` throws a **stacksize error**.

scarlehoff:
Hi @Radonirinaunimi, thanks for this, is this ready for review?

Regarding the questions:

> In the old implementation, it is possible to add custom extra_labels to organize the plotting [...]

I'm trying to have the plotting to be read exactly in the same way. But maybe I was too eager to update the labels with the information from the dictionary. Please leave it as you think it should work because if it doesn't then that's a bug I have to fix!

> You'll notice that the theory_central for the new implementations are of by a factor of . This is exactly the value of the conversion_factor in the yamldb. So I am wondering if this is applied correctly?

The new commondata deprecates the yamldb, which is not read. Conversion factors needs to be put under the theory part of the new commondata.

Radonirinaunimi:
Thanks for the feedback @scarlehoff!

> Hi @Radonirinaunimi, thanks for this, is this ready for review?

Mines (which are currently the ones listed in the summary table above) are ready but there should also be the other experiments (implemented by different people) included here. These are basically the Collider DY experiments for ATLAS and CMS.

I uncommented the `extra_labels` in [f871d7c](https://github.com/NNPDF/nnpdf/pull/1826/commits/f871d7cd2eb9e85b74d62ccf72a10d485b2886c4) (as they should be) keys so currently they will fail at loading.

scarlehoff:
I'd rather have you separated from the rest so that I have less to review in one go (and once they are merged they can serve as an example for the others).

Please add as well the conversion factors and other stuff coming from the yamldb.

Radonirinaunimi:
> Please add as well the conversion factors and other stuff coming from the yamldb.

I have added a new key, `conversion_factor` under `implemented_obserables.theory`. I guess this should be added to the new commondata documentation.

> I'd rather have yours separated from the rest so that I have less to review in one go (and once they are merged they can serve as an example for the others).

What should then be done about the ATLAS low-mass 7TeV and LHCb W enu which have already been added here?

> Edit: one of the reasons is that it would be helpful to have some examples of datasets which have an older companion since datasets_names is empty for now.

Yes, indeed. `datasets_names.yaml` is always up to date.

scarlehoff:
> What should then be done about the ATLAS low-mass 7TeV and LHCb W enu which have already been added here?

I mean, if there are some that are already 100% finished I'd rather have them separated from the rest so that I can review and merge them.

The main reason is that this set of datasets adds things that have never been tested before (such as `datasets_names` or some keys for the plotting) so if I can have already a few finished ones that I can check by themselves it would be good.

Otherwise we can wait until everything is ready, but there's the risk that there's stuff to be changed and then we need to wait until all of them are changed, etc. Better to follow a more _agile_ approach and do minimal PRs.

Radonirinaunimi:
> > What should then be done about the ATLAS low-mass 7TeV and LHCb W enu which have already been added here?
> 
> I mean, if there are some that are already 100% finished I'd rather have them separated from the rest so that I can review and merge them.
> 
> The main reason is that this set of datasets adds things that have never been tested before (such as `datasets_names` or some keys for the plotting) so if I can have already a few finished ones that I can check by themselves it would be good.
> 
> Otherwise we can wait until everything is ready, but there's the risk that there's stuff to be changed and then we need to wait until all of them are changed, etc. Better to follow a more _agile_ approach and do minimal PRs.

This I completely agree 100%. It was my bad for initially thinking that all the collider DY could be collected here.

@cschwan, @niclaurenti, would you mind having the other datasets in different PRs, still pointing to [new_commondata_collected](https://github.com/NNPDF/nnpdf/tree/new_commondata_collected). After this is done, I will clean this one a bit.

cschwan:
I agree that separating them is probably better, then you can proceed implementing/reviewing them quicker!

Radonirinaunimi:
> I agree that separating them is probably better, then you can proceed implementing/reviewing them quicker!

Perfect, thanks! @niclaurenti, could you also do the same? I am going to clean this up but still keep the history.

scarlehoff (commented):
Hi @Radonirinaunimi, many thanks for this.

I've had a first go at this set of datasets, I've add the possibility of using `extra_labels` with label_based `figure_by` to #1678 so now the datasets in this branch would work.

The tests I'm doing for every datasets are three:

1. Check that the central data is exactly the same.
2. Check that computing the chi2 agrees exactly (meaning that the covmat is read and created equally)
3. And finally doing a report.

The test 1 is passed by every dataset.
Test 2 is passed for the following pairs:

```
"LHCBWZMU7TEV", "LHCB_DY_7TEV_MUON_Y"
"LHCBWZMU8TEV", "LHCB_DY_8TEV_MUON_Y"
"LHCBZEE2FB_40", "LHCB_Z0_8TEV_DIELECTRON_Y"
"LHCB_Z_13TEV_DIELECTRON", "LHCB_Z0_13TEV_Y-DIELECTRON"
"LHCB_Z_13TEV_DIMUON", "LHCB_Z0_13TEV_Y-DIMUON"
```

while for these
```
"LHCBZMU7TEV", "LHCB_Z0_7TEV_MUON_Y",
"LHCBZMU8TEV", "LHCB_Z0_8TEV_MUON_Y",
"LHCBWMU7TEV", "LHCB_WPWM_7TEV_MUON_Y",
"LHCBWMU8TEV", "LHCB_WPWM_8TEV_MUON_Y",
"LHCBZ940PB", "LHCB_Z0_7TEV_DIELECTRON_Y"
```

the value of the chi2 is created differently. Please have a look at them!
I've had a look at the covmat and even the diagonal is different.
In particular, for `LHCB_Z0_7TEV_DIELECTRON_Y` there's something going on because the new covmat has values like `10^6`. There might be an order-of-magnitude problem in some uncertainties?

3. I've done the report for the 5 dataset that fully agree so that you can have a look and check that everything is correct. I cannot see any obvious problems but of course you've been implementing them so might know better!
The report is ordered by pairs (so first/second, third/four, etc)

https://vp.nnpdf.science/vY3OvQMqTxybygGVncMkvQ==

Edit: If there is no obvious fix to the 5 that produce a different covmat I would be happy with separating the 5 that are 100% ok and merging them by themselves.

buildmaster/LHCB_Z0_13TEV/metadata.yaml
```diff
@@ -0,0 +1,80 @@
+setname: "LHCB_Z0_13TEV_DIELECTRON"
```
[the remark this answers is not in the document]
Radonirinaunimi: Hi @scarlehoff, thanks a lot for having had a look at these! I will have a detailed look at these failing datasets this evening. The differences must be indeed in the construction/dumping of the uncertainties, although when I did a comparison by eye between the old and new for the first couple of datapoints they seemed to agree (with a sign difference).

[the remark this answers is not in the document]
scarlehoff: Just to say, this one is one of the 100% fine, this change is just that you forgot to remove `_DIELECTRON` in the setname.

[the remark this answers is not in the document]
Radonirinaunimi: Got it! Thanks!

Radonirinaunimi:
@scarlehoff, if you run the checks for the remaining datasets everything should now be Ok. I wasn't able to produce the report which includes all the datasets myself due to the pandoc stacksize error (which I need to find a solution to).

There were basically two main problems. For `LHCB_Z0_7TEV_DIELECTRON_Y`, the issue was that the normalization uncertainty was off by $10^3$. For the rest, since these are separation of the NC and CC components from the same measurement, I previously only dumped the uncertainties for the process-bins. Now, all the correlated uncertainties are dumped and that fixed the issue.

scarlehoff:
Thank you very much! I find agreement for all of them and the plots generally look the same as before afaics

Here's a report with all datasets. Please check that everything looks as you expected. By that I mean, plotting options, axis scale and all that.
(I've used the QCD cfactor for the prediction, and `NRM` for the ones that needed it as well). 

https://vp.nnpdf.science/4ub8cAYxTa2xPP6oM6X4_w==

Once you confirm this is ok, this is ready to merge. However, after looking at #1785 I've realized the old-new mapping syntax might need to change a bit. If it does, I'll add the appropriate change it here (and update the docs) so nothing to do from your side, but just to mention it!

(pinging @enocera in case he wants to have a second look or want to add anything to any of it before merging)

enocera:
@Radonirinaunimi Thanks for your work. @scarlehoff I'd like to have a second look at the implementation before merging, if you don't mind.

Radonirinaunimi:
As far as I am concerned, this is basically ready. So once @enocera approves (thanks for also having a look) we can merge this.

scarlehoff (approved):
I've repeated the checks with the final version fo this branch:

1. Central value are the same
2. The covariance matrix are the same
3. The computed chi2 is the same
4. The t0 chi2 is the same (so multiplicative and additive are equivalent in the new and old implementation)

The report in this comment https://github.com/NNPDF/nnpdf/pull/1826#issuecomment-1810012510 is unchanged since @Radonirinaunimi has not updated any metadata.yaml files (only the uncertainties) in his last commits.

So this one is perfectly ready to go from my side.

Radonirinaunimi:
Thanks a lot for these additional checks @scarlehoff! This is completely done from my side as well.

PR #1830, opened 2023-10-30
Add a new `plot_fancy_dataspecs` and chi2 distributions  with scale variations 

scarlehoff:
With this one can get reports such as https://vp.nnpdf.science/8W7S2JStS3ySBXmea0PEJg==

This is not ready for review (but if anyone wants to have a look feel free). 

A brief explanation of the changes:

In the theorycovariance module (/cc @andreab1997), I've broken down the function `covs_pt_prescrip` so that I can access the internal logic. I've done this in order to manually compute the theory covariance for single datasets without worrying about the processes. This might not be actually necessary but I wanted to ensure that I was not getting unwanted contributions.

I've added a few plotting functions to `dataplots.py`.

- `plot_chi2dist_sv`: this one might not be necessary, it should be possible with the proper change to `config.py` to get a covariance matrix computed on the fly as part of the chi2. I know how to do it only for covmats that are loaded from a fit, but it should be possible. For this I needed to add `results_with_theory_covmat` to `results.py` which would also be unnecessary once I fix the previous point. But before playing with the code I wanted to have something functional since PDF4LHC is coming close.

- `plot_fancy_sv_dataspecs`: theory-data comparison where the errors in the theory are actually the PDF uncertainties and the scale variations added in quadrature, where the scale variations are the diagonal of the theory covariance matrix. This required adding to `results.py` a new class `ThUncertaintiesResult` and a `results_with_scale_variations`. As far as I can see this part will be necessary because within validphys the theory covariance matrix is always added to the experimental covmat while in this case we want to use both contributions separately.

(all other changes in `dataplots.py` are due to `black`)

btw, there are a few functions scattered around the code that claim to do some of these things, however either they no longer work or they use functions that have not been touched in a long time (and that nobody doing MHOU is using anymore) so I've preferred to ignore them. They will be removed in due time.

### DIFF
--- a/validphys2/examples/data_theory_comparison_w_sv.yaml
+++ b/validphys2/examples/data_theory_comparison_w_sv.yaml
@@ -5,7 +5,7 @@ meta:
 
 use_cuts: "internal"
 
-theoryid: 405 # define the central theory
+theoryid: 717 # define the central theory
 point_prescription: "3 point"
 theoryids:
   from_: scale_variation_theories
--- a/validphys2/examples/data_theory_comparison_w_sv.yaml
+++ b/validphys2/examples/data_theory_comparison_w_sv.yaml
@@ -1,0 +1,90 @@
+meta:
+    title: "Example of comparison of data and theory including scale variations"
+    keywords: [data theory comparison, scale variations, example runcard]
+    author: "Validphys examples"
+
+use_cuts: "internal"
+
+theoryid: 405 # define the central theory
+point_prescription: "3 point"
+theoryids:
+  from_: scale_variation_theories
+use_theorycovmat: true
+
+use_pdferr: true
+
+disable_pdferr:
+  use_pdferr: false
+
+dataset_inputs:
+  - { dataset: ATLASWZRAP36PB }
+  - { dataset: BCDMSP }
+
+dataspecs:
+  - pdf: NNPDF40_nnlo_lowprecision
+    speclabel: "pdf1"
+
+  - pdf: Basic_runcard_3replicas_lowprec_221130
+    speclabel: "pdf2"
+
+# for the chi2 comparison
+pdfs:
+  - Basic_runcard_3replicas_lowprec_221130
+  - NNPDF40_nnlo_lowprecision
+
+datanorm:
+  normalize_to: data
+
+template_text: | 
+  Plots with scale variations
+  ===========================
+
+  {@with matched_datasets_from_dataspecs@}
+
+  {@dataset_name@}
+  ----------------
+  {@ with disable_pdferr @}
+  {@ plot_fancy_sv_dataspecs @}
+  {@ datanorm plot_fancy_sv_dataspecs @}
+  {@ endwith @}
+
+  ### chi2 distributions per PDF
+  {@ with dataspecs @}
+  {@ plot_chi2dist_sv @}
+  {@ endwith @}
+
+  {@ endwith @}
+
+  Plots without scale variations
+  ==============================
+
+  {@with matched_datasets_from_dataspecs@}
+
+  {@dataset_name@}
+  ----------------
+  {@ with disable_pdferr @}
+  {@ plot_fancy_dataspecs @}
+  {@ datanorm plot_fancy_dataspecs @}
+  {@ endwith @}
+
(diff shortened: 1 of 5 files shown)


scarlehoff:
This is ready for review. Given #1844, trying to do the theory covmat in the same way as the pdferr risks creating a lot of secondary problems.

And, unlike pdf erros, the scale variations are actively being used. Therefore I think for the time being I prefer to leave the `plot_fancy` and `abs_chi2` plot functions untouched and just have some `_scale variations` extra.

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -322,18 +335,62 @@ def covmat_9pt(name1, name2, deltas1, deltas2):
...
+    deltas2=None,
+    fivetheories=None,
+    seventheories=None,
+):
+    """Utility to compute the covariance matrix by prescription given the
+    shifts with respect to the central value for a given process as well as its name
+    If name2 (and deltas2) are not provided, it is understood that is the same process/dataset
+    """
+    if (name2 is None and deltas2 is not None) or (name2 is not None and deltas2 is None):
+        raise ValueError("WRONG")
```
RoyStegeman, line 361: This message is not very helpful

RoyStegeman (commented):
I still want to have a bit of a closer look later, but it all seems reasonable enough

validphys2/src/validphys/theorycovariance/construction.py
scarlehoff: Ups, forgot that there. Thank you.

validphys2/examples/data_theory_comparison_w_sv.yaml
```diff
@@ -0,0 +1,90 @@
+meta:
+    title: "Example of comparison of data and theory including scale variations"
+    keywords: [data theory comparison, scale variations, example runcard]
+    author: "Validphys examples"
+
+use_cuts: "internal"
+
+theoryid: 405 # define the central theory
```
andreab1997, line 8: I would change this to the new NLO theory, meaning 717

validphys2/src/validphys/results.py
```diff
@@ -469,12 +492,55 @@ def results(dataset: (DataSetSpec), pdf: PDF, covariance_matrix, sqrt_covmat):
...
 
+def results_with_theory_covmat(dataset, results, theory_covmat_dataset):
+    """Returns results with a modfy ``DataResult`` such that the covariance matrix includes
+    also the theory covmat.
+    This can be used to make use of results that consider scale variations without including
+    the theory covmat as part of the covariance matrix used by other validphys function.
+    Most notably, this can be used to compute the chi2 including theory errors while plotting
+    data theory covariance in which the experimental uncertainties are not stained by the thcovmat
+    """
+    # TODO: in principle this function could be removed, and `results` could automagically
```
andreab1997, line 511: I guess you meant `automatically` but I like `automagically` a lot. It expresses very well how I fell when things work with `validphys` :)

scarlehoff: Nono, it is [100% intentional](https://www.oxfordlearnersdictionaries.com/definition/english/automagically)

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -322,18 +335,92 @@ def covmat_9pt(name1, name2, deltas1, deltas2):
...
+    name2=None,
+    deltas2=None,
+    fivetheories=None,
+    seventheories=None,
+):
+    """Utility to compute the covariance matrix by prescription given the
+    shifts with respect to the central value for a pair of processes.
+
+    The processes are defined by the variables ``name1`` and ``name2`` with
+    ``deltas1`` and ``deltas2`` the associated shifts.
```
RoyStegeman, line 360:
```suggestion
    ``deltas1`` and ``deltas2`` the associated shifts wrt the central prediction.
```

Something to make it a bit more clear how delta is defined

RoyStegeman (approved):

scarlehoff:
Since the code has been tagged now, can we merge this?

RoyStegeman:
Yes fine for me

PR #1831, opened 2023-10-30
Generic hessian PDF covmat

comane:
This PR adds to `covmats.pdferr_plus_covmat` the option for an asymmetric hessian set.
The way in which the PDF covmat is computed follows `core.HessianStats`.

TODO:

- Are there some references on how to compute the hessian pdf covmat?
- understand why the ABMP1_6_5_nnlo set (symmhessian) has such a large chi2  (see report below)

https://vp.nnpdf.science/9rRTyXihSsqCDE3h4qmZ5A==

### DIFF
--- a/validphys2/src/validphys/covmats.py
+++ b/validphys2/src/validphys/covmats.py
@@ -685,6 +685,8 @@ def pdferr_plus_covmat(dataset, pdf, covmat_t0_considered):
       the replica theory predictions
     - If the PDF error_type is 'symmhessian', a covariance matrix is estimated using
       formulas from (mc2hessian) https://arxiv.org/pdf/1505.06736.pdf
+    - If the PDF error_type is 'hessian' a covariance matrix is estimated using
+      the hessian formula from core.HessianStats
 
 
     Parameters
@@ -737,7 +739,6 @@ def pdferr_plus_covmat(dataset, pdf, covmat_t0_considered):
         pdf_cov = X @ X.T
 
     elif pdf.error_type == 'hessian':
-        
         rescale_fac = pdf._rescale_factor()
         hessian_eigenvectors = th.error_members
         
--- a/validphys2/src/validphys/covmats.py
+++ b/validphys2/src/validphys/covmats.py
@@ -15,7 +15,7 @@ from validphys.checks import (
     check_data_cuts_match_theorycovmat,
     check_dataset_cuts_match_theorycovmat,
     check_norm_threshold,
-    check_pdf_is_montecarlo_or_symmhessian,
+    check_pdf_is_montecarlo_or_hessian,
     check_speclabels_different,
 )
 from validphys.commondata import loaded_commondata_with_cuts
@@ -677,7 +677,7 @@ def groups_corrmat(groups_covmat):
     return mat
 
 
-@check_pdf_is_montecarlo_or_symmhessian
+@check_pdf_is_montecarlo_or_hessian
 def pdferr_plus_covmat(dataset, pdf, covmat_t0_considered):
     """For a given `dataset`, returns the sum of the covariance matrix given by
     `covmat_t0_considered` and the PDF error:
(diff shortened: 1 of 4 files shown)


scarlehoff:
Thanks!

> understand why the ABMP1_6_5_nnlo set (symmhessian) has such a large chi2

That might be a feature not a bug :P

validphys2/src/validphys/dataplots.py
```diff
@@ -621,6 +621,35 @@ def plot_datasets_chi2(groups_data, groups_chi2):
...
+    collabels = []
+
+    for ds, val in pdf_dict.items():
+        vals.append([chi2.central_result / chi2.ndata for chi2 in val])
+        collabels.append(ds)
+    
+    fig, ax = plotutils.barplot(np.array(vals).T, collabels, datalabels=[f'$\chi^2$, {str(pdf)}' for pdf in pdfs])
+    ax.set_title(r"$\chi^2$ distribution for datasets")
+    ax.legend()
+    return fig
```
scarlehoff, line 650: This function is actually perfect for the reports that I'm doing. Thanks!
I'm going to start using it.
Let me know when you are happy/confident with the hessian covmat and I can review and merge this to master.

comane:
Here is a report in which CT hessian and CT MC are compared.
I would say that in general the difference in the two is compatible with the difference between NNPDF and NNPDF hessian.

https://vp.nnpdf.science/brO4mfBuSnulOJOYxbWXxA==/

comane:
Also ABMP1_6_3_nnlo's chi2 look fine as compared to the ones of ABMP1_6_5_nnlo.
So as you were saying @scarlehoff , that might have been a feature of ABMP1_6_5_nnlo 😁

https://vp.nnpdf.science/yrcojfSKR2iKecuc0hW6Sg==/

validphys2/src/validphys/covmats.py
```diff
@@ -736,6 +738,16 @@ def pdferr_plus_covmat(dataset, pdf, covmat_t0_considered):
         X = X / rescale_fac
         pdf_cov = X @ X.T
 
+    elif pdf.error_type == 'hessian':
+        rescale_fac = pdf._rescale_factor()
+        hessian_eigenvectors = th.error_members
+        
+        # see core.HessianStats
+        X = hessian_eigenvectors[:,0::2] - hessian_eigenvectors[:,1::2]
```
scarlehoff, line 746:
```suggestion
        X = (hessian_eigenvectors[:,0::2] - hessian_eigenvectors[:,1::2])/2.0
```

I think there's a factor of 2 missing here looking at `core.HessianStats`. When I add this I get results which are much closer between the hessian and MC PDF sets.

https://vp.nnpdf.science/kXxyXIE4SDCuIfv0jMcNHA==/

Anyway, with this I think this can be merge. If you accept the suggestion (if it works) and there's nothing else you want to change this would be done.

scarlehoff: I'm actually worried that there might be something else wrong because the change is not that big, but given the one that had the biggest difference in your report now looks like this:

image

(i.e., the difference between the two is even less than for nnpdf40) I guess it would be fine.

comane: Hi @scarlehoff, I think you are right. Having a look at eq. 5 of https://arxiv.org/pdf/1401.0013.pdf, it seems that a factor of two should be there.

> I'm actually worried that there might be something else wrong because the change is not that big,
I am not so worried about this since that might be just because the exp. uncertainty is the largest one.

I guess to check this better we would need a dataset for which the PDF uncertainty is larger/comparable to the experimental one.

scarlehoff: The problem is that I don't even know whether we truly have to expect that they are the same.
For now the check I've done is to ensure that the diagonal is exactly (1 std dev)**2... that happens... so the rest is probably also ok? :__

scarlehoff (approved):

scarlehoff:
I'm going to merge this, since I'll be using it in the presentation on Friday. Afaics it is correct. If you find any problems don't hesitate to open an issue / a new PR of course!

To test this really well we would need to create a (many replicas) MC pdf set from a hessian one such that we have good control. I've checked it with CT14 but they have two version of the MC pdf set which points to some differences being expected anyway and I honestly think we can afford differences in the [third digit](https://vp.nnpdf.science/kXxyXIE4SDCuIfv0jMcNHA==/)

PR #1835, opened 2023-11-06
Remove NNLO specific cuts

andreab1997:
We want to have exactly the same cuts for NLO,NNLO and N3LO fits in the context of the MHOU and N3LO papers. In order to do so, we have to:

- [x] Remove the NNLO specific cuts of LHCb (thus cutting these points also at NLO and N3LO)
- [x] Remove the NNLO specific cuts of D0WMASY and D0WEASY (thus cutting these points also at NLO and N3LO)
- [x] Remove the NNLO specific cuts of CMSDY2D11 and CMSDY2D12 (thus cutting these points also at NLO and N3LO)
- [x]  Remove the NNLO specific cuts of `CMSDY2D11, CMSDY2D12, ATLASLOMASSDY11EXT` (thus cutting these points also at NLO and N3LO)

What do you think? @enocera 

CC: @giacomomagni 

EDIT: From what I understood from this sentence "NNLO- only execute the following rule if the pto is strictly less than 2" from the docstring of the `PerturbativeOrder` class, `NNLO-` actually means `LO` and `NLO`. So doing what I suggested before ( "Increase from `NNLO-` to `N3LO-` the cuts of `CMSDY2D11, CMSDY2D12, ATLASLOMASSDY11EXT`") would not be correct because it would exclude N3LO. Then, I just removed the `PTO` entry also for these three datasets

### DIFF
--- a/n3fit/runcards/reproduce_nnpdf40/NNPDF40_lo_as_01180.yml
+++ b/n3fit/runcards/reproduce_nnpdf40/NNPDF40_lo_as_01180.yml
@@ -199,27 +199,6 @@ cuts_intersection_spec:
   pdf: NNPDF31_nlo_as_0118
   dataset_inputs: *NNLODatasets
 
-added_filter_rules:
-      - dataset: CMSDY2D11
-        reason: "Reproduce 4.0 cuts"
-        PTO: NNLO- 
-        local_variables:
-          M: sqrt(M2)
-          max_M: 200.0
-        rule: "M <= max_M"
-      - dataset: CMSDY2D11
-        reason: "Reproduce 4.0 cuts"
-        PTO: NNLO-
-        local_variables:
-          M: sqrt(M2)
-          min_M: 30.0
-          max_rapidity: 2.2
-        rule: "M >= min_M and etay <= max_rapidity"
-      - dataset: ATLASLOMASSDY11EXT
-        reason: "Reproduce 4.0 cuts"
-        PTO: NNLO-
-        rule: "idat >= 2"
-
 
 ############################################################
 datacuts:
--- a/n3fit/runcards/reproduce_nnpdf40/NNPDF40_lo_as_01180.yml
+++ b/n3fit/runcards/reproduce_nnpdf40/NNPDF40_lo_as_01180.yml
@@ -199,6 +199,26 @@ cuts_intersection_spec:
   pdf: NNPDF31_nlo_as_0118
   dataset_inputs: *NNLODatasets
 
+added_filter_rules:
+      - dataset: CMSDY2D11
+        reason: "Reproduce 4.0 cuts"
+        PTO: NNLO- 
+        local_variables:
+          M: sqrt(M2)
+          max_M: 200.0
+        rule: "M <= max_M"
+      - dataset: CMSDY2D11
+        reason: "Reproduce 4.0 cuts"
+        PTO: NNLO-
+        local_variables:
+          M: sqrt(M2)
+          min_M: 30.0
+          max_rapidity: 2.2
+        rule: "M >= min_M and etay <= max_rapidity"
+      - dataset: ATLASLOMASSDY11EXT
+        reason: "Reproduce 4.0 cuts"
+        PTO: NNLO-
+        rule: "idat >= 2"
 
 
 ############################################################
--- a/n3fit/runcards/reproduce_nnpdf40/NNPDF40_lo_as_01180.yml
+++ b/n3fit/runcards/reproduce_nnpdf40/NNPDF40_lo_as_01180.yml
@@ -200,6 +200,7 @@ cuts_intersection_spec:
   dataset_inputs: *NNLODatasets
 
 
+
 ############################################################
 datacuts:
   t0pdfset: 210506-01-rs-nnpdf40-lo-flav # PDF set to generate t0 covmat
--- a/n3fit/runcards/reproduce_nnpdf40/NNPDF40_lo_pch_as_01180.yml
+++ b/n3fit/runcards/reproduce_nnpdf40/NNPDF40_lo_pch_as_01180.yml
@@ -200,6 +200,7 @@ cuts_intersection_spec:
   dataset_inputs: *NNLODatasets
 
 
+
 ############################################################
 datacuts:
   t0pdfset: 210503-01-rs-nnpdf40-lo-pch # PDF set to generate t0 covmat
--- a/n3fit/runcards/reproduce_nnpdf40/NNPDF40_lo_pch_as_01180.yml
+++ b/n3fit/runcards/reproduce_nnpdf40/NNPDF40_lo_pch_as_01180.yml
@@ -199,26 +199,6 @@ cuts_intersection_spec:
   pdf: NNPDF31_nlo_as_0118
   dataset_inputs: *NNLODatasets
 
-added_filter_rules:
-      - dataset: CMSDY2D11
-        reason: "Reproduce 4.0 cuts"
-        PTO: NNLO- 
-        local_variables:
-          M: sqrt(M2)
-          max_M: 200.0
-        rule: "M <= max_M"
-      - dataset: CMSDY2D11
-        reason: "Reproduce 4.0 cuts"
-        PTO: NNLO-
-        local_variables:
-          M: sqrt(M2)
-          min_M: 30.0
-          max_rapidity: 2.2
-        rule: "M >= min_M and etay <= max_rapidity"
-      - dataset: ATLASLOMASSDY11EXT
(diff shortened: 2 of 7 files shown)


enocera:
Thanks @andreab1997. I'll review this PR later today.

enocera:
@andreab1997 Thanks for your work. It seems to me that cuts are now the same at NLO and NNLO for all LHC data, as desired. I understand that the more conservative cuts are always preferred, which is good. I have a concern, however. Suppose that we merge this PR. How would we be able to use the published NNPDF4.0 cuts? Is there a possibility to use different filter rules in different fits?

andreab1997:
> @andreab1997 Thanks for your work. It seems to me that cuts are now the same at NLO and NNLO for all LHC data, as desired. I understand that the more conservative cuts are always preferred, which is good. I have a concern, however. Suppose that we merge this PR. How would we be able to use the published NNPDF4.0 cuts? Is there a possibility to use different filter rules in different fits?

Yes this is indeed an issue. However, my first goal was to have something with which @giacomomagni and I can quickly run the fits (since it seems we are a bit in a rush...). Of course at the moment we cannot merge this so, if the goal is to have the final MHOU and N3LO fits done with tagged code, we first have to find a permanent solution for this.

I will try to propose one as soon as possible.

scarlehoff:
It is already possible to add extra filter rules directly in the runcard.

Right now I don’t remember wheter they would get appended or substituted (and I cannot check) but the technology is already there.

scarlehoff:
On that note. I would not merge this PR but rather prepare the set of cuts that we want to have added and manually specify them in the runcards.

In this way the default fit would be 4.0, which I believe we want for the time being.

Once the 3 in-preparation papers are done we can discuss promoting them to default cuts.

enocera:
II agree with @scarlehoff 's plan.

andreab1997:
> On that note. I would not merge this PR but rather prepare the set of cuts that we want to have added and manually specify them in the runcards.
> 
> In this way the default fit would be 4.0, which I believe we want for the time being.
> 
> Once the 3 in-preparation papers are done we can discuss promoting them to default cuts.

Yes, I also agree. I opened this PR exactly to have clear which are the cuts that we need (and I guess @enocera confirmed that what is now in `filters.yaml` is what we need for the MHOU and N3LO paper). My plan is now to understand how to specify these cuts in the runcard and then restore `filter.yaml` to the NNPDF4.0 status.

andreab1997:
So this is my new proposal. 
The idea is to use the `added_filter_rules` key introduced in #1745 that allows to manually add cuts to the ones defined in `filters.yaml`. However it only allows to add cuts, i.e. not to remove or replace cuts. So, in order to keep being able to reproduce both the NLO and the NNLO fits of NNPDF4.0, the proposed solution is to keep only the less conservative cuts in `filters.yaml` (which in practice means to change only three datasets, one of which is not even used by default in the fits) and then add the rest in the runcard. 
Of course, to reproduce the NNLO fits nothing changes, i.e. the current runcards would work, while, to reproduce the NLO fits, it is needed to add ~10 lines to the runcard (as shown in the following example runcard, in particular under the speclabel "Default NLO 4.0").

This is the example runcard: 
[test_cuts.txt](https://github.com/NNPDF/nnpdf/files/13283130/test_cuts.txt)

and this is the report obtained with it: https://vp.nnpdf.science/m3BC4hOmSGuU2iUOs2yVYw== 

As you can see @enocera and @scarlehoff , the number of points in the NNLO fit is the usual one we have in NNPDF4.0 (first column) and to obtain it you can just use the old runcards, same for the NLO fit (second column) but you need to add some lines to the runcard, while in the third column we have a NLO fits but with the same number of points of the NNLO (as we want for the MHOU and N3LO fit).

Let me know what you think (and keep in mind that if this is ok for you I will also change the runcard examples).

CC: @giacomomagni

scarlehoff:
I'm ok with this proposal. Just one question, did you try this in a fit (more specifically, with the 4.0 NLO fit runcard adding these extra rules).

I want to make sure there's nothing else happening during the NLO that might be missed when just creating the table. The fit/setupfit action(s) override some option and would like to know for sure that the specific runcards for 4.0 behave in the same way (once the cutsnare added).

andreab1997:
> I'm ok with this proposal. Just one question, did you try this in a fit (more specifically, with the 4.0 NLO fit runcard adding these extra rules).
> 
> I want to make sure there's nothing else happening during the NLO that might be missed when just creating the table. The fit/setupfit action(s) override some option and would like to know for sure that the specific runcards for 4.0 behave in the same way (once the cutsnare added).
> 
> (also please update the example runcards for the reproduction of 4.0 accordingly where needed)

I just tried with vp-setupfit and it seems to work: I obtain 3530 points for the NNPDF4.0 NLO fit and 3541 points for the NLO fits with NNLO cuts (as you can see here https://vp.nnpdf.science/hNleDurmQvCN6wxMkZ0Qgg==/ these are the correct numbers for a MHOU-like fit). The two runcards I used (I will put them in the examples folder once this PR is approved) are these: 
[231010-ab-baseline-nlo-global-nFONLL_nlo40cuts.txt](https://github.com/NNPDF/nnpdf/files/13284314/231010-ab-baseline-nlo-global-nFONLL_nlo40cuts.txt)
[231010-ab-baseline-nlo-global-nFONLL_newnlocut.txt](https://github.com/NNPDF/nnpdf/files/13284318/231010-ab-baseline-nlo-global-nFONLL_newnlocut.txt)

giacomomagni:
if we are all happy, can we merge ?

andreab1997:
> if we are all happy, can we merge ?

I still need to update the examples runcards

andreab1997:
I would say that now we should have everything. Let me just point out a few details just to remember:

- In order to reproduce 4.0 NLO and LO it is now needed to add a few lines to the runcard, as I did now in the examples runcards.  I checked that using them I get the same number of points of the 4.0 NLO 
- Just for the record, I actually get two points less (4424) than the 4.0 NLO (that you can see for example [here](https://vp.nnpdf.science/7EVzNuveQzykex2LVQFLEQ==/) has 4426 points). The reason is that about one year ago @enocera introduced a cut on the last point of `CMS_TTBAR_2D_DIFF_MTT_TRAP_NORM` and `ATLAS_TOPDIFF_DILEPT_8TEV_TTRAPNORM`. So it is perfectly reasonable to get two point less now but the cuts introduced in this PR are working
- Do you want me to also add this runcard? [test_cuts.txt](https://github.com/NNPDF/nnpdf/files/13283130/test_cuts.txt) 

If you are happy, I would say this can be merged. @enocera @scarlehoff @giacomomagni

RoyStegeman:
@enocera @scarlehoff could you please have a look at this when you have time? Since it's blocking the creation of an environment to do the fits.

enocera (approved):
That looks good to me.

scarlehoff:
@enocera how sure are you that these new cuts are going to be always there? Are we sure it won't be a case of "oh, we don't actually like these new fits, go back to the previous situation"?

andreab1997:
So, speaking with @scarlehoff, we concluded that a more elegant way of achieving this (i.e. having at NLO the cuts of NNLO) is to use the `use_cuts: fromfit` key. I tried now and it seems to work, so I am putting now in  `filters.yaml` only the cuts that are relevant for N3LO (and restoring the rest). @enocera @giacomomagni

andreab1997:
Please @giacomomagni check that all the N3LO cuts are ok after [d54898b](https://github.com/NNPDF/nnpdf/pull/1835/commits/d54898b21d328608e7ff8aee63ec18d9b72f83d8)

@RoyStegeman I guess this can be merged today, nothing controversial is here anymore

scarlehoff:
Hi @andreab1997 I think you left the runcards and remove the N3LO filters?

andreab1997:
> Hi @andreab1997 I think you left the runcards and remove the N3LO filters?

I am sorry, my bad. I did now

enocera:
> @enocera how sure are you that these new cuts are going to be always there? Are we sure it won't be a case of "oh, we don't actually like these new fits, go back to the previous situation"?

Of course I'm not sure. And I think it's likely that at some point we (you) will be asked to come back to the NNPDF4.0 default cuts. But that was exactly my initial concern. And I understood that there is the machinery in place to handle the possibility of having NNPDF4.0 cuts and more restrictive cuts for the MHOU and N3LO papers. Is your point on deciding what the default should be?

scarlehoff:
The problem is having less restrictive cuts.

In any case, I think the solution that @andreab1997 posted (using exactly the NNLO fit cuts, since that's the real goal) is the best.

giacomomagni (approved):

PR #1837, opened 2023-11-07
NCD TTB ATLAS

t7phy:


scarlehoff (commented):
Thanks! Here the report of differences for this one. As with the total, it is important to know whether the differences are bug or features. In particular for `ATLAS_TTBAR_8TEV_LJ_DIF` the data itself is different (not by much)
while for `ATLAS_TTBAR_8TEV_2L_DIF` the problem of some observables might be just the treatment of some uncertainties.

```
 > Testing old: ATLAS_TOPDIFF_DILEPT_8TEV_TTM vs new: ATLAS_TTBAR_8TEV_2L_DIF_MTTBAR
# Differences in the computation of chi2 545.25 vs 757.23
    The covmats are different even the diagonal

 > Testing old: ATLAS_TOPDIFF_DILEPT_8TEV_TTMNORM vs new: ATLAS_TTBAR_8TEV_2L_DIF_MTTBAR-NORM
 > Everything ok!
 > Testing old: ATLAS_TOPDIFF_DILEPT_8TEV_TTRAP vs new: ATLAS_TTBAR_8TEV_2L_DIF_YTTBAR
# Differences in the computation of chi2 341.03 vs 403.29
    The covmats are different even the diagonal
> Testing old: ATLAS_TOPDIFF_DILEPT_8TEV_TTRAPNORM vs new: ATLAS_TTBAR_8TEV_2L_DIF_YTTBAR-NORM
> Everything ok!
```

The data is different for `ATLAS_TTBAR_8TEV_LJ_DIF` it is important to know whether it is a feature or a bug. The errors range from a few per-mille and a percent for every dataset other than `ATLAS_TTBAR_8TEV_LJ_DIF_MTTBAR` for which it goes to a few percent:
```                          ratio
1      1.138890  1.118623  1.018117
2      1.115040  1.106053  1.008125
3      0.678057  0.681068  0.995580
4      0.349925  0.352835  0.991754
5      0.146934  0.148915  0.986694
6      0.036742  0.037503  0.979696
7      0.004360  0.004613  0.945229
```

t7phy:
@scarlehoff  all the distributions in ATLAS 13 TEV LJ are new, everything else is old

scarlehoff:
Report for those https://vp.nnpdf.science/xUkXJO0vQiWlx1s3MEQW0g==

t7phy:
> For ATLAS_TTBAR_8TEV_LJ_DIF the only two that appear in NNPDF40 are
> 
> ATLAS_TTB_DIFF_8TEV_LJ_TRAPNORM and ATLAS_TTB_DIFF_8TEV_LJ_TTRAPNORM
> 
> And the differences there are sub-% so I think it is ok. The covmats are quite different so we need a legacy variant.

This seems to be the cases where the dataset differences and the covmat differences come from the treatment of assym uncer.

PR #1841, opened 2023-11-14
remove qed input from `ekobox.apply.apply_pdf`

RoyStegeman:
This input was removed from the function in eko

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -51,7 +51,7 @@ requirements:
         - sphinxcontrib-bibtex
         - curio >=1.0
         - pineappl >=0.5.8
-        - eko >=0.13.5,<0.14
+        - eko >=0.14.0
         - banana-hep >=0.6.8
         - fiatlux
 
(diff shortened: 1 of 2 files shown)


niclaurenti:
Now this line

https://github.com/NNPDF/nnpdf/blob/d2b942be2e35ee909095d8832ee8264fc96b4355/n3fit/src/evolven3fit_new/evolve.py#L101

can be removed.
And also `qed` from 
https://github.com/NNPDF/nnpdf/blob/d2b942be2e35ee909095d8832ee8264fc96b4355/n3fit/src/evolven3fit_new/evolve.py#L115
and
https://github.com/NNPDF/nnpdf/blob/d2b942be2e35ee909095d8832ee8264fc96b4355/n3fit/src/evolven3fit_new/evolve.py#L148

niclaurenti (approved):

RoyStegeman:
I think in eko `0.13.5` the function had a default input, so I'm not sure any test would fail. Either way, you're right that now the tag is there we want to update the nnpdf dependency

PR #1842, opened 2023-11-14
update pineappl dependency to 0.6.2

RoyStegeman:
There were some bugfixes in pineappl since 0.5.8. Not sure if they affect anything we do in this repo but we may as well update it.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -50,7 +50,7 @@ requirements:
         - sphinx_rtd_theme >0.5
         - sphinxcontrib-bibtex
         - curio >=1.0
-        - pineappl >=0.5.8
+        - pineappl >=0.6.2
         - eko >=0.13.5,<0.14
         - banana-hep >=0.6.8
         - fiatlux


scarlehoff:
If tests pass I'm happy with this.

However, just to make sure... this last version of pineappl is compatible with eko 0.14 ? (cc @AleCandido @cschwan @felixhekhorn) It should not matter for the nnpdf code, but just so that we make sure that people have a conda environment in which the different tools can talk with each other.

AleCandido:
> However, just to make sure... this last version of pineappl is compatible with eko 0.14 ? (cc @AleCandido @cschwan @felixhekhorn) It should not matter for the nnpdf code, but just so that we make sure that people have a conda environment in which the different tools can talk with each other.

The PineAPPL interface we use has to be actually compatible with Pineko, that is making use of it.
There is also a direct interface between PineAPPL and EKO, that allows one to use the CLI, but this is not used in the Pineline (and it might be slightly broken because of the fix of the introduction of `matching_order` in the theory card, but I guess that it would just be ignored).

RoyStegeman:
This PR doesn't change much other than that it pins a lower bound. In practice 0.6.2 is already being downloaded in most cases: https://github.com/NNPDF/nnpdf/actions/runs/6862782721/job/18661228431#step:6:969

scarlehoff (approved):
Perfect then.

Issue #1843, opened 2023-11-14
When using `pdferr` in `plot_fancy` the pdf errors are used twice.


scarlehoff:
If you use `plot_fancy` with `use_pdferr: True` in the runcard, the PDF errors will be added to the covmat.
However, since the experimental errors are computed _from_ the covmat and this is modified, the resulting error bands include the pdf errors as well. This is a problem since `plot_fancy` includes the PDF error already in the prediction.

PR #1844, opened 2023-11-14
Raise an exception when use_pdferr is used together with plot_fancy

scarlehoff:
Addresses #1843 

I'm just adding a check to `plot_fancy`. I've taken the liberty to also run `black` since apparently it was missing from `dataplots.py`?

The only change that is not from black is in line 397   (sorry @RoyStegeman, I must've somehow missed `dataplots.py` when I passed black to every file...)

### DIFF
--- a/validphys2/src/validphys/checks.py
+++ b/validphys2/src/validphys/checks.py
@@ -54,7 +54,10 @@ def check_pdf_is_montecarlo_or_hessian(pdf, **kwargs):
 
 @make_argcheck
 def check_not_using_pdferr(use_pdferr=False, **kwargs):
-    check(not use_pdferr, "The flag 'use_pdferr' must be `False` to use this function. This is to avoid including the PDF error in the uncertainty bars of the experimental datapoints.")
+    check(
+        not use_pdferr,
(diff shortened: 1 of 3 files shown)


RoyStegeman:
> sorry @RoyStegeman, I must've somehow missed `dataplots.py` when I passed black to every file...

I think these diffs are due to my own change: https://github.com/NNPDF/nnpdf/commit/ccf3c62486759407c2fb79eda475756ff1efa5fc

validphys2/src/validphys/checks.py
```diff
@@ -52,6 +52,11 @@ def check_pdf_is_montecarlo_or_hessian(pdf, **kwargs):
     )
 
 
+@make_argcheck
+def check_not_using_pdferr(use_pdferr, **kwargs):
+    check(not use_pdferr, "The flag 'use_pdferr' must be off to use this function")
```
RoyStegeman, line 57:
```suggestion
    check(not use_pdferr, "The flag 'use_pdferr' must be `False` to use this function. This is to avoid including the PDF error in the uncertainty bars of the experimental datapoints.")
```
Or something like that, anyway I think a bit more explanation would be helpful

RoyStegeman (approved):

PR #1850, opened 2023-11-16
use mamba in docker construction

RoyStegeman:
Use mamba to build the docker to avoid passing the 6hour limit on the github CI

Since I force pushed it's not longer visible, but using mamba reduce the time of the "Build docker image and export environment" step to 3 minutes: https://github.com/NNPDF/nnpdf/actions/runs/6889900597/job/18741833824

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -8,7 +8,7 @@ on:
 
 # some general variables
 env:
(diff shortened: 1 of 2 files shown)


andreab1997 (approved):
This looks good to me but I am not an expert so probably better wait for @scarlehoff

PR #1852, opened 2023-11-16
create reproduce_nnpdf40_updated environment

RoyStegeman:
This adds the environment and runcards for the `_updated` nnpdf4.0 and the three new variants (QED, MHOU, N3LO. 

The environment can be installed with  `conda env create -f nnpdf40_updated.yml`


n3fit/runcards/reproduce_nnpdf40_updated/nnpdf40_updated.yml
```diff
@@ -0,0 +1,264 @@
+name: nnpdf40_updated
+channels:
+  - https://packages.nnpdf.science/public
+  - conda-forge
+  - defaults
```
scarlehoff, line 5: Shouldn't this be `defaults` then `conda-forge`? as in https://github.com/NNPDF/binary-bootstrap

RoyStegeman: Is there a good reason `defaults` is first in binary bootstrap? I think it doesn't matter much

scarlehoff: That it is what god intended.

More seriously, don't know. But I guess the `reproduce_xxx` and the binary bootstrap should align. The fact that you are running with this and it works (also the qed fits are running with this environment I guess?) means that probably updating the bootstrap repository is quicker.

RoyStegeman: I guess this is just the way it was ordered in my `.condarc` when I produced this file. In the env for nnpdf4.0 it's actually the other way around. 

I guess (it's really not much more then a guess) that when creating the env this overrides the `.condarc` (or at least the `channels`). There will always remain some ambiguity e.g. the defaults for strict channel priority I believe can depend on the specific version of conda that's being used. 

That being said, I assume the build numbers differ between the different channels and thus there is a unique solution to this configuration of packages that does not depend on the order of the channels. Maybe that's not the case?

scarlehoff:
I think we should close this since it is no longer final.

RoyStegeman:
I was planning just to update this branch rather than open a new one

scarlehoff:
oh, so no new tag, just an update of the environment?

RoyStegeman:
Yes there will be a new tag, but what would that have to do with this branch? I can replace the file in this PR with one that has the new tag

scarlehoff:
cleanliness, but ok

What's your plan for this? https://github.com/NNPDF/nnpdf/issues/1853

scarlehoff:
I think this can be merged, right?

RoyStegeman:
I was planning to wait until the NLO and NNLO fits are approved, and in that case I'd like to also upload their runcards.

PR #1857, opened 2023-11-22
Add qcd exa to theory.db as theory 798

niclaurenti:
Adding QCD EXA theory:
- `798` full QCD `EXA` theory


niclaurenti:
@andreab1997 @giacomomagni @RoyStegeman you are keeping track of the numbering of the theories more than me. Are 800 and 801 free?

RoyStegeman:
I think you should be able to use 798 and 799, I'm not aware of anyone else already having reserved theories

giacomomagni:
> I think you should be able to use 798 and 799, I'm not aware of anyone else already having reserved theories

it should be okay, n3lo last at 797 and we are currently using 800 for polarized stuff.

RoyStegeman:
Did you want to merge this?

niclaurenti:
@RoyStegeman I delete the ``mixed'' theory and then yes

PR #1861, opened 2023-11-24
Implement a python-only installation

scarlehoff:
This pull requests implements a python-only installation defined in the [pyproject.toml](https://python-poetry.org/docs/pyproject/) file.

While it might look like the PR is changing a lot of files, this is mostly a reordering, so let me describe all changes here:

## Build system
I'm using [poetry](https://python-poetry.org/docs/) as the build system as we need some more advance features than what setuptools offers. For the user however nothing changes, you can use `poetry` for development but you can also utilize `pip install .` or `pip install -e .` just the same.

### cmake
This PR doesn't remove cmake, but it has been modified so that it doesn't install the python stuff.

### conda
`conda` will not be removed since it is needed for `lhapdf` and `pandoc`.
We also needed it for now to get some actual package for `reportengine` which at the moment is broken in pip (or the repo).

### How to use
If you want to already start using this installation method, you can just go to the root of the repository and do:

```
pip install .
```

If you want to install all extras you can also do:
```
pip install .[docs,tests,qed]
```

This will install both the `n3fit` and `validphys` packages. The datafiles are no longer in `${CONDA_PREFIX}/share/NNPDF` but rather they are part of the `validphys` package.

If you are under a `conda` installation with `lhapdf` and `pandoc` you are done. If you are not, you need to install `lhapdf` and `pandoc` by yourself. In another PR (#1799) I will also add a possibility to work without `lhapdf`, first as an experimental thing, later as something more useful.

## Data and Theory locations
Up to now, the cmake installation would put everything (data, theories, etc) into a share/NNPDF folder in the conda environment. Since a pure python installation cannot go around copying data to arbitrary folders in the system, I've changed a bit this configuration:

### Datafiles and theory.db
I believe that datafiles _need_ to be synchronised with the status of the code. Therefore, they are now installed as data coming from the python package.
The same is true for `theory.db`.
This also brings another advantage, now you don't need to reinstall (or copy) when you are working on implementing new data, you can just do `pip install -e .` and also the data and theory will be editable, so development becomes much easier.
It also allows you to share the very heave share folder between NNPDF installations.
This will be all in `validphys2/src/validphys/datafiles`

### Theories and results
Theories and results are downloaded from the server. They should be shared among every possible installation that you have in your system (no need to have 6 copies of `theory_200`). 

## Other considerations
If I've done everything correctly, nobody should see any difference upon using the code implemented in this PR. Nor developers or users. Everything will work the same as it always has. Even custom configurations like using a non-default NNPDF_PROFILE_PATH will work out of the box (as this takes precedence over the new location).

The only difference will be a practical one (you will notice that theories that you thought you already downloaded are being downloaded again).

### Possible breaking changes
I cannot think of any situation in which this PR introduces a breaking change. Of course, if people is directly adding or removing stuff from the `share/NNPDF` folder, without using a custom nnprofile, now they have to move it to the new location... but this has always been discouraged.

### Testing method
Beyond the tests, I'm running all runcards in the examples folder of validphys. Mainly to make sure that we are not using any of the files in the C++ folders that I'm willingly ignoring.

This is the script that I'm using for testing in a clean environment should anyone want to do the same:

With cmake (by-the-<s>book</s>docs installation):
```bash
ENVIR=condapdf_clean
conda create -n ${ENVIR} python==3.10 -y
conda activate ${ENVIR}
conda install nnpdf --only-deps -y
conda install pytest hypothesis pytest-mpl -y
cmake -S . -B build_folder  -DCMAKE_INSTALL_PREFIX=${CONDA_PREFIX}
cmake --build build_folder
cmake --install build_folder
pytest --pyargs --mpl validphys n3fit
```

Note that as per [the docs](https://docs.nnpdf.science/get-started/installation.html#installation-from-source) you might need some extra stuff if you are using mac os or if you are missing some compilers in your system. And if you go that route you might also need to follow other instructions there.

I'll remove the python installation from cmake (with the appropriate changes to the documentation and with a big warning from cmake) in another PR.


AleCandido (commented):
All the changes that are not related to imports or data are due to Black, correct?

AleCandido (commented):
Please, run `poetry lock` and commit the lock file as well

scarlehoff:
I've rearranged things here and there (e.g., things like theory.db is under data, but theories are now in a different place, so I needed to change theoryid anyway so I modified a bit that class). There are not that many changes due to black

scarlehoff:
This PR is ready for review. Once I tested everything works (might need more includes of data files) next batch of changes will be another PR pointing to this one to remove a lot of libnnpdf/nnpdfcpp that is not used by buildmaster or evolven3fit.

n3fit/src/evolven3fit_new/__init__.py
```diff
@@ -0,0 +1 @@
+from validphys._version import __version__, __version_tuple__
```
AleCandido, line 1: Why do we need a `__version_tuple__`?

CMakeLists.txt
```diff
@@ -148,54 +145,8 @@ add_subdirectory(nnpdfcpp)
...
-    install(CODE "execute_process(COMMAND ${PYTHON_EXECUTABLE} -m pip install -e ${PROJECT_SOURCE_DIR}/n3fit)")
-else(N3_DEV)
-    install(CODE "execute_process(COMMAND ${PYTHON_EXECUTABLE} -m pip install --no-deps --ignore-installed ${PROJECT_SOURCE_DIR}/n3fit)")
-    if(BURN_TAG)
-        install(CODE "set(LIBRARY \"n3fit\")")
-        install(SCRIPT ${PROJECT_SOURCE_DIR}/version.cmake)
-    endif(BURN_TAG)
-endif(N3_DEV)
+if(NNPDF_DEV)
+    install(CODE "execute_process(COMMAND ${PYTHON_EXECUTABLE} -m pip install -e ${PROJECT_SOURCE_DIR})")
```
AleCandido, line 149: Is dev mode with CMake required?

scarlehoff: Yes, the idea is that when people use manually `cmake` they get the development version as it has been done until now (I forgot the default though, my bad, fixing it now)!

Instead when they get the conda package is not the dev version of course.

n3fit/src/evolven3fit_new/__init__.py
scarlehoff: I saw that it was for free with the dynamic version package so I decided to add it as well.

CMakeLists.txt
AleCandido: > Instead when they get the conda package is not the dev version of course.

This for sure, no way out.

> Yes, the idea is that when people use manually `cmake` they get the development version as it has been done until now (I forgot the default though, my bad, fixing it now)!

So don't you want to encourage `pip` installation instead? Under the hood, it will happen anyhow, so keeping a single frontend would simplify the interactions. Isn't it?

pyproject.toml
```diff
@@ -1,3 +1,128 @@
...
+# Generic dependencies (i.e., validphys)
+python = "^3.9"
+matplotlib = ">=3.3.0,<3.8"
+pineappl = "^0.6.2"
+pandas = "<2"
+numpy = "*"
+validobj = "*"
+prompt_toolkit = "*"
+# Reportengine (and its dependencies) need to be installed in a bit more manual way
+reportengine = { git = "https://github.com/NNPDF/reportengine", rev = "3bb2b1d"}
```
RoyStegeman, line 73:
```suggestion
reportengine = { git = "https://github.com/NNPDF/reportengine", branch = "master"}
```

validphys2/src/validphys/nnprofile.yaml
```diff
@@ -0,0 +1,42 @@
+# The local resource location default to the XDG Base Directories unless explicitly defined here
+
+# results_path: ${XDG_DATA_CONFIG}/NNPDF/results
+# theories_path: ${XDG_DATA_CONFIG}/NNPDF/theories
+# hyperscan_path: ${XDG_DATA_CONFIG}/NNPDF/hyperscans
+# validphys_cache_path: ${XDG_DATA_CONFIG}/NNPDF/vp-cache
```
RoyStegeman, line 6:
```suggestion
# results_path: ${XDG_DATA_HOME}/NNPDF/results
# theories_path: ${XDG_DATA_HOME}/NNPDF/theories
# hyperscan_path: ${XDG_DATA_HOME}/NNPDF/hyperscans
# validphys_cache_path: ${XDG_DATA_HOME}/NNPDF/vp-cache
```

validphys2/src/validphys/core.py
```diff
@@ -1,12 +1,8 @@
 # -*- coding: utf-8 -*-
 """
 Core datastructures used in the validphys data model.
-Created on Wed Mar  9 15:19:52 2016
-
-@author: Zahari Kassabov
```
RoyStegeman, line 6: :cry:

RoyStegeman:
Why did you put the datafiles folder under validphys and not just in the nnpdf root dir?

validphys2/src/validphys/nnprofile.yaml
scarlehoff: ups, thanks

RoyStegeman:
That's what I assumed but you should still be able to put it outside vp no? https://docs.python.org/3.3/distutils/setupscript.html#installing-additional-files

Maybe that only makes the structure more confusing though

scarlehoff:
I'm not sure, in the poetry documentation (which is what I'm using to build/install) there's no mention about supporting moving things from one location to another.

AleCandido:
Poetry intentionally avoids implementing some features (like pre/post-install hooks) in order to avoid some variability (and keep everything simpler).

https://python-poetry.org/docs/pyproject/#include-and-exclude

True that the feature you're claiming is quite declarative, so they could. But I guess the idea is that it would be surprising to install a Python package (from PyPI or anywhere else) and it then modifies other paths outside the environment.
E.g. you might think that an environment is isolated, but then you and up modifying thins user-wide (not system-wide, just because you will be prevented by missing `sudo`).

If that's required, then you should have vp copying/moving the files from its folder to somewhere else at runtime.
But honestly, I'd avoid this (if possible).

AleCandido:
The other option is `vp-get data`, that I would like :P

scarlehoff:
We can add later a link to the folder in the root of the repository to make things easier. But right now it's in `nnpdfcpp/data` which is not necessarily more intuitive so I'll leave it like it is.

pyproject.toml
```diff
@@ -1,3 +1,128 @@
...
+]
+classifiers = [
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Topic :: Scientific/Engineering",
+    "Topic :: Scientific/Engineering :: Physics",
+    "License :: OSI Approved :: GNU General Public License v3 (GPLv3)",
+]
+license = "GPL-3.0-or-later"
+requires-python = ">=3.8"
```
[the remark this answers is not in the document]
scarlehoff: (the reason is I missed this comment)

scarlehoff:
I might inadvertently left something intended for #1864 here. I wanted to leave libnnpdf alone in this PR and moved all cmake removal to the other PR.
Let me fix it!

libnnpdf/CMakeLists.txt
```diff
@@ -58,6 +58,9 @@ target_link_libraries(nnpdf ${LHAPDF_LIBRARIES} ${GSL_LDFLAGS} ${SQLITE3_LDFLAGS
 
 install(FILES ${PROJECT_SOURCE_DIR}/libnnpdf/scripts/nnpdf.pc DESTINATION lib/pkgconfig)
 install(DIRECTORY src/NNPDF DESTINATION include)
+
+file(WRITE ${PROJECT_SOURCE_DIR}/libnnpdf/REAMDE.md "The share folder of NNPDF has been moved, see: https://github.com/NNPDF/nnpdf/pull/1861\n")
+install(FILES ${PROJECT_SOURCE_DIR}/libnnpdf/REAMDE.md DESTINATION share/NNPDF)
```
RoyStegeman, line 63:
```suggestion
file(WRITE ${PROJECT_SOURCE_DIR}/libnnpdf/README.md "The share folder of NNPDF has been moved, see: https://github.com/NNPDF/nnpdf/pull/1861\n")
install(FILES ${PROJECT_SOURCE_DIR}/libnnpdf/README.md DESTINATION share/NNPDF)
```

RoyStegeman (approved):
To me this looks fine. I didn't check each file individually but the changes as you describe in the main comment seem perfectly fine and the moving around of the folders didn't seem accidentally to break anything.

Before merging though, I do think we should make it very clear to everyone using the code that the data storage will be changed.

scarlehoff:
> I do think we should make it very clear to everyone using the code that the data storage will be changed.

Beyond saying it in the code meeting not sure what to do.
However, I'm happy waiting for the final-final-final environment so that possible confusion don't delay the releases.

RoyStegeman:
Yes we can mention it at the CM and maybe in the telegram channel, more than that we don't really need to do I think. 

Please do indeed wait for the final environment, thanks.

scarlehoff:
Ok, so here's my proposal for `nnprofile.yaml`:

The first thing the code will do is to look at the environment variable `NNPDF_PROFILE_PATH`. This hasn't changed. If you were setting it with that, you can continue doing so (however, note that there's a new key `theory_path` that you should set as well and you should absolutely remove `data_path`!)

Then the code will try to look at the standard location `${XDG_CONFIG_HOME}/NNPDF/nnprofile.yaml` which defaults to `~/.config/NNPDF/nnprofile.yaml`, so if you want to set a global profile that will work for all your NNPDF installation you should do it there. A normal use case would be to set there
```yaml
nnpdf_share: ~/.local/share/NNPDF
```

Then you can have several installations of NNPDF and they all will share the downloaded resources so you don't have to download them again.
In a cluster you might want to have something like
```yaml
nnpdf_share: /mnt/separated_disk/NNPDF
```

You can also set _only_ the theory path
```yaml
nnpdf_share: /mnt/separated_disk/NNPDF_theories
```

Differently from before, the validphys installation now includes a `nnprofile_default.yaml` with the default values which are used in case the custom profiles don't define them, so you only need to include in your custom profile the ones that you actually want to change (most of the time the location of the share directory, `nnpdf_share`.

### The default issue
Now, the issue with the default.... after the discussion yesterday and thinking about this morning, there's an use-case that we need to keep supporting which is having separate conda environments with separated theories... and the only way of doing that is autodiscovery. So the default is the special key (which you can also use in your `~/.config/NNPDF/nnprofile.yaml`

```yaml
nnpdf_share: RELATIVE_TO_PYTHON
```

This will be the default and should default to `${CONDA_PREFIX}` or to `${VIRTUAL_ENV}`. If you have a folder `RELATIVE_TO_PYTHON` in your system that you were expecting to use, I'm sorry but I don't pity you.

### Something that _will_ change
Up to now, in `share/NNPDF/data` the theory database, the theories downloaded and the commondata were together. Now theories go to their own folder `share/NNPDF/theories` (unless you say otherwise in the profile).
If you don't want to redownload your data just modify the name of the `data` directory to `theories` and you're done.

Please 
@Radonirinaunimi @RoyStegeman 
tell me whether you are ok with this solution and let me know any problems or issues real or imagined that you can think about!
I'll now test this in a few different systems (and modify things accordingly if I find any problems).

validphys2/src/validphys/nnprofile_default.yaml
```diff
@@ -0,0 +1,68 @@
+#
+# Default `nnprofile.yaml` for NNPDF.
+# The location of a custom profile can be given with the `NNPDF_PROFILE_PATH` enviroment variable
+# otherwise by default ${XDG_DATA_HOME}/.config/NNPDF/nnprofile.yaml will be read
```
RoyStegeman, line 4: /```suggestion
# otherwise by default ${XDG_CONFIG_HOME}/.config/NNPDF/nnprofile.yaml will be read
```

File: validphys2/src/validphys/nnprofile_default.yaml [hunk 9]
```diff
@@ -0,0 +1,68 @@
...
+
+# Remote resource locations
+fit_urls:
+    - 'https://data.nnpdf.science/fits/'
+    - 'https://nnpdf.web.cern.ch/nnpdf/fits/'
+
+fit_index: 'fitdata.json'
+
+hyperscan_urls:
+  - 'https://data.nnpdf.science/hyperscans/'
```
RoyStegeman, line 44:
```suggestion
    - 'https://data.nnpdf.science/hyperscans/'
```

Because my linter doesn't like inconsistencies :)

RoyStegeman (approved):
Fine for me. Thanks!

Radonirinaunimi (approved):
Thanks a lot for this! Finally was able to test this carefully and everything worked fine (tested it both on my local computers and our Amsterdam cluster).

Issue #1862, opened 2023-11-24
Speeding up `vp-setupfit`


giacomomagni:
It looks to me that `vp-setupfit` is really slowed down by the fact that predictions are computed for all the replicas: 

https://github.com/NNPDF/nnpdf/blob/ce6c05cacec95dee36c577503c9a1109daa2908a/validphys2/src/validphys/convolution.py#L132-L180

even though only the central prediction is then needed: 

 https://github.com/NNPDF/nnpdf/blob/ce6c05cacec95dee36c577503c9a1109daa2908a/validphys2/src/validphys/theorycovariance/construction.py#L179
 
Since by default set we are using `210715-n3fit-1000-001` which contains 1000 replicas, this could be a major wast of time. 
 
Similarly as done for the `t0pdfset` here there is no need to load the entire pdf set: 
 https://github.com/NNPDF/nnpdf/blob/ce6c05cacec95dee36c577503c9a1109daa2908a/validphys2/src/validphys/lhapdfset.py#L44-L52
 
 Thanks to @andreab1997 and @RoyStegeman for the assistance

scarlehoff:
The PDF object in validphys can be loaded as a "t0 pdf" at any point with `load_t0`:

https://github.com/NNPDF/nnpdf/blob/ce6c05cacec95dee36c577503c9a1109daa2908a/validphys2/src/validphys/core.py#L202

I guess the thcovmat should be using that instead. Alternatively, `central_predictions` can be used, instead of just `predictions`.

I'd try to help but I honestly don't know which functions are used within the constructions and which are not... but... assuming that the predictions used are taking the predictions from `procs_results_theory` it would be enough swapping that variable by`theoryids_procs_central_values`

RoyStegeman:
I don't think it's as straightforward as that since e.g. `theoryids_procs_central_values` still calculates the predictions for all replicas, but then only returns the central prediction.

Since I believe for pdfs inside `theorycovmatconfig` we are never interested in anything beyond replica0 I think the cleanest thing to do would be something similar to what is now done for t0pdfs.

scarlehoff:
If that's the case you can have just a function similar to `theoryids_procs_central_values` that uses the `central_predictions` instead of `predictions`. Or that loads the PDF as a t0pdf.

[...]

PR #1863, opened 2023-11-24
calculate only predicitons for rep0 in thcovmat construction

RoyStegeman:
addresses #1862


andreab1997 (approved):
Looks good to me!

scarlehoff (approved):
lgtm, thanks!

I get a factor of ~x2 when creating the theory covmat in my laptop.

RoyStegeman:
> I get a factor of ~x2 when creating the theory covmat in my laptop

Good, the amount of time saved depends on the number of replicas of course.

PR #1864, opened 2023-11-24
Minimize the effect of cmake/C++ across the code

scarlehoff:
Depends on #1861

This PR:
- Removes C++ code that is no longer used anywhere (i.e., `nnpdfcpp`)
- Isolates the pieces that we want to keep for the time being (i.e., `evolven3fit`)
- Removes conda C++ dependencies that are no longer needed


scarlehoff:
I think this is the most I can clean the repository from cmake and C++ code without breaking `evolven3fit` and `buildmaster`.

PR #1865, opened 2023-11-24
Theory covmat for higher twist 

RoyStegeman:
We want to add a covmat to account for higher twist effects. Therefore the covmat should be constructed similar to a 3pt prescirption where $\Delta^\pm = \pm \sigma_\mathrm{obs}\frac{c}{Q^2(1-x)}$ for come coefficient $c$.

impact on the datapoints is as follows: https://vp.nnpdf.science/xee4YTRKSAmRmN3iUZfkIg==


RoyStegeman:
Just a quick comment (also just as a reminder to myself) - we should allow for both theory covmat through scalevar and HT covmat to be added simultaneously.

scarlehoff:
A comment so I don't forget, I think before doing this we should remove all the thcovmat stuff thst is not being used. That would facilitate adding new components (because one would know where to look...).

RoyStegeman:
Since I don't expect this PR to be one of the ones that's open for an infinite amount of time I think that would imply removing that functionality now-ish. I don't mind, but I had understood you wanted to keep it as it currently is to make sure nobody is using it (especially `make_scale_var_covmat` and down from there)

scarlehoff:
No, I put the exception to make sure nobody was using it but I'd rather remove it as soon as possible.

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -182,10 +183,10 @@ def dataset_names(data_input):
     return [el.name for el in data_input]
 
 
-ProcessInfo = namedtuple("ProcessInfo", ("theory", "namelist", "sizes"))
+ProcessInfo = namedtuple("ProcessInfo", ("theory", "namelist", "sizes", "data"))
 
 
-def combine_by_type(each_dataset_results_bytheory, dataset_names):
+def combine_by_type(each_dataset_results_bytheory, dataset_names, groups_dataset_inputs_loaded_cd_with_cuts):
```
RoyStegeman, line 189: I don't think this would work since `each_dataset_results_bytheory` collects over `theorids`. If you'd like to use this `combine_by_type` function you could consider replacing `each_dataset_results_bytheory` by a `produce_` rule. You could do this in such a way where if the thcovmat being calculated is the usual MHOU it loops over different theoryids, but if the theorycovmat is a HT comvat the predictions corresponding to the up and down shift are calculated from a single predicion but provided in the same format as the `each_dataset_results_bytheory`.

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -199,6 +225,234 @@ def covmat_9pt(name1, name2, deltas1, deltas2):
...
+
+    # ABMP parametrisation
+    x = [0.0, 0.1, 0.3, 0.5, 0.7, 0.9, 1]
+    y_2 = [0.023, -0.032, -0.005, 0.025, 0.051, 0.003, 0.0]
+    y_T = [-0.319, -0.134, -0.052, 0.071, 0.030, 0.003, 0.0]
+    H_2 = scint.CubicSpline(x, y_2)
+    H_2 = np.vectorize(H_2)
+    H_T = scint.CubicSpline(x, y_T)
+    def H_L(x):
+        return np.power(x, 0.05)*(H_2(x) - H_T(x))
```
RoyStegeman, line 367:
```suggestion
        return (H_2(x) - np.power(x, 1.05)*H_T(x))
```
Looking at the paper this prefactor is there only for F_T

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -199,6 +225,234 @@ def covmat_9pt(name1, name2, deltas1, deltas2):
...
+            N_L = - N_2 * yL / yp
+
+    if "CHORUS_CC" in experiment_name:
+        x = kwargs.get('xB')
+        y = kwargs.get('y')
+        Q2 = kwargs.get('Q2')
+        yL = np.power(y, 2)
+        Gf = 1
+        Mh = 1
+        MW2 = 1
```
RoyStegeman, line 448: - Gf is Fermi's constant, the value can be copied from the theorycard 
- Mh is the proton mass, also in the theory card
- MW2 is the W boson mass squard, also in the theory card

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -199,6 +225,234 @@ def covmat_9pt(name1, name2, deltas1, deltas2):
...
+                for j, exp2 in enumerate(process_info.namelist[proc2]):
+                    if ht_pt_prescription == 5:
+                        s = np.outer(deltas["(+,0)"][i], deltas["(+,0)"][j]) + \
+                            np.outer(deltas["(0,+)"][i], deltas["(0,+)"][j])
+                    start_locs = (start_proc_by_exp[exp1], start_proc_by_exp[exp2])
+                    covmats[start_locs] = s
+    return covmats
+
+
+def compute_normalisation_by_experiment(experiment_name, **kwargs):
```
RoyStegeman, line 424: Why are you collecting them as kwargs if they are being used in the funtion?

P.S. python is not cpp, to keep things more readable it helps to be consistent and e.g. use `**` for power and `dictname['keyname']` everywhere

Issue #1858, opened 2023-11-23
Sort dataset diffs


felixhekhorn:
Can we please sort [Dataset differences and cuts](https://vp.nnpdf.science/WyHREKMwQiyD2lXbmWFV4Q==/#dataset-differences-and-cuts)? with whatever unique sort you like - e.g. alphabetic ... else I can only guess by the length that this is the same as in [this report](https://vp.nnpdf.science/XiaPwCL9QjiQW0TOvlVpCw==/#dataset-differences-and-cuts)

I guess it should be somewhere here:
https://github.com/NNPDF/nnpdf/blob/ce6c05cacec95dee36c577503c9a1109daa2908a/validphys2/src/validphys/fitdata.py#L296

scarlehoff:
> else I can only guess by the length that this is the same

I'm a bit confused. When they are not the same you get a list of things that are different, e.g.: https://vp.nnpdf.science/BMvnWVZaRSGjBF9mAzFKiQ==/#dataset-differences-and-cuts

felixhekhorn:
report 1 is A vs. B and report 2 is C vs. D - if possible I'd like to catch in a glance that they don't differ by dataset but by theory setting

felixhekhorn:
Adding more (relevant) information is a good I idea I think ... but I ask for a small (i.e. quick) improvement first :innocent:

PR #1867, opened 2023-11-27
Order dataset differences by name of the commondata

scarlehoff:
This addresses #1858 in the following ways:

1. Adds a mention to the theories being used at the end of the comparison
2. Adds the name of the commondata (the one we actually use in the runcard) so that it is easier to read for those of us that know the dataset by its runcard name
3. Orders by the setname, this has (or will have) the benefit of automatically implying an order in 1. experiment, 2. process, 3. energy

Example report: https://vp.nnpdf.science/QHB504AaT0WbicohYul0cA==

Let me know if this is what you had in mind @felixhekhorn

### DIFF
--- a/validphys2/src/validphys/fitdata.py
+++ b/validphys2/src/validphys/fitdata.py
@@ -23,12 +23,7 @@ from validphys.plotoptions import get_info
 # TODO: Add more stuff here as needed for postfit
 LITERAL_FILES = ['chi2exps.log']
 REPLICA_FILES = ['.dat', '.json']
-FIT_SUMRULES = [
-    "momentum",
-    "uvalence",
-    "dvalence",
-    "svalence",
-]
+FIT_SUMRULES = ["momentum", "uvalence", "dvalence", "svalence"]
 
 # t = blessings.Terminal()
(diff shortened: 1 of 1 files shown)


felixhekhorn:
A side note: I guess for the new CD you should avoid using `\rm` in the `dataset_label` since the renderer (I couldn't identify it - is it native?) doesn't seem to like that:
grafik

scarlehoff:
It's reportengine magic I guess. Don't really know what the problem with `rm` is, but it renders ok in the plots (e.g. https://vp.nnpdf.science/xJZivqlbQhSYY-tPbWMvUQ==/matched_datasets_from_dataspecs47_dataset_report_report.html).

felixhekhorn:
> It's reportengine magic I guess. Don't really know what the problem with `rm` is, but it renders ok in the plots (e.g. https://vp.nnpdf.science/xJZivqlbQhSYY-tPbWMvUQ==/matched_datasets_from_dataspecs47_dataset_report_report.html).

yeah, it depends on the renderer (mpl might use a different then vp) ... however, are the `rm` really required?

Issue #1860, opened 2023-11-24
Error in comparing two fits in NNPDF determination


TheWaleedHussain:
I am trying to compare two fits using command "vp-comparefits -i"  but everytime I'm getting same error. I have tried to run this command by using different python versions but it is still not working. Tell me what to do now. Here is the error that I'm getting:

![Error](https://github.com/NNPDF/nnpdf/assets/146311717/e6d6b4e3-4a66-4bc3-aeb0-75fdd90f9420)

scarlehoff:
Could you tell us the version of the code you are using?

If you could upload somewhere the fit(s) that are  producing this errors so that we can reproduce them that would be really helpful. Thanks.

TheWaleedHussain:
I am using the following version of code:

`python -c "import validphys as vp ; print(vp.__version__)"
4.0.7.4+gce6c05cac`

I am trying to compare these two fits:

`230217_tutorialfit` : (https://data.nnpdf.science/fits/230217_tutorialfit.tar.gzl)
`230217_tutorialfit_2` : (https://data.nnpdf.science/fits/230217_tutorialfit_2.tar.gz)

scarlehoff:
Many thanks for the report. This is fixed in PR #1868 

For the time being I 've updated the `230217_tutorialfit` so that it is a bit more smooth and it should not produce the error any longer.

PR #1868, opened 2023-11-27
Add the central value as extreme when the percentiles are outside

scarlehoff:
Fixes issue #1860

This (and other reports that we have received) about the negative error bars are due a combination of factors:
 1. The fact that the central value is the average of the replicas is only true for the PDF and linear quantities (so, not for hadronic observable such as some of the positivity obs)
 2. While the PDF is gaussian that's not a problem, but in fringe regions of the phase space (which are probed by the positivity observables) the PDF can have non gaussianities, which are even more pronounced if you use a fit with not many replicas.

I've added the central value as the extreme value to bypass this problem.

I could be convinced of raising an exception in this situations, but we might actually want to plot a non-gaussian PDF...

### DIFF
--- a/validphys2/src/validphys/core.py
+++ b/validphys2/src/validphys/core.py
@@ -750,6 +750,15 @@ class ThCovMatSpec:
 
 # TODO: Decide if we want methods or properties
 class Stats:
+    """Class holding statistical information about the object used in validphys
+    By the PDF convention, the central value corresponds to member 0.
+    This is equal to the mean of the error members only for the PDF
+    itself and for linear quantities such as DIS-type observables.
+
+    If you want to obtain the average of the central value you can do:
+    ``np.mean(stats_instance.error_members, axis=0)``
+    """
+
     def __init__(self, data):
         """`data `should be N_pdf*N_bins"""
         self.data = np.atleast_2d(data)
@@ -793,10 +802,14 @@ class MCStats(Stats):
         return np.mean(np.power(self.error_members() - self.central_value(), order), axis=0)
 
     def errorbar68(self):
-        # Use nanpercentile here because we can have e.g. 0/0==nan normalization
-        # somewhere.
+        # Use nanpercentile here because we can have e.g. 0/0==nan normalization somewhere
         down = np.nanpercentile(self.error_members(), 15.87, axis=0)
         up = np.nanpercentile(self.error_members(), 84.13, axis=0)
+        # The central value is not necessarily the average of the error members
+        # and small non-gaussianities on the PDF can lead to errors here, so block with the cv
+        cv = self.central_value()
+        down = np.maximum(down, cv)
+        up = np.maximum(up, cv)
         return down, up
 
     def sample_values(self, size):


validphys2/src/validphys/core.py
```diff
@@ -793,10 +802,14 @@ def moment(self, order):
         return np.mean(np.power(self.error_members() - self.central_value(), order), axis=0)
 
     def errorbar68(self):
-        # Use nanpercentile here because we can have e.g. 0/0==nan normalization
-        # somewhere.
+        # Use nanpercentile here because we can have e.g. 0/0==nan normalization somewhere
         down = np.nanpercentile(self.error_members(), 15.87, axis=0)
         up = np.nanpercentile(self.error_members(), 84.13, axis=0)
+        # The central value is not necessarily the average of the error members
```
RoyStegeman, line 808:
```suggestion
        # The central value is not necessarily the median of the error members
```
It is instead _precisely_ the average of the error members.

For a gaussian distribution where we take enough samples (i.e. replicas)  the mean and median will be very close (distance decreasing as 1/sqrt(Nrep)) but if gaussianity breaks down and we have a small amount of samples the distance between mean and median can be large.

validphys2/src/validphys/core.py
```diff
@@ -750,6 +750,15 @@ def __str__(self):
 
 # TODO: Decide if we want methods or properties
 class Stats:
+    """Class holding statistical information about the object used in validphys
+    By the PDF convention, the central value corresponds to member 0.
+    This is equal to the mean of the error members only for the PDF
+    itself and for linear quantities such as DIS-type observables.
+
+    If you want to obtain the average of the central value you can do:
+    ``np.mean(stats_instance.error_members, axis=0)``
```
RoyStegeman, line 759: I don't really understand this sentence?

RoyStegeman (commented):
Thanks, I was indeed going to ask to keep the error and just make a clearer error message.

validphys2/src/validphys/core.py
scarlehoff:
```suggestion
    If you want to obtain the average of the error members you can do:
    ``np.mean(stats_instance.error_members, axis=0)``
```

validphys2/src/validphys/core.py
```diff
@@ -750,6 +750,15 @@ def __str__(self):
 
 # TODO: Decide if we want methods or properties
 class Stats:
+    """Class holding statistical information about the object used in validphys
+    By the PDF convention, the central value corresponds to member 0.
+    This is equal to the mean of the error members only for the PDF
+    itself and for linear quantities such as DIS-type observables.
+
+    If you want to obtain the average of the error members you can do:
+    ``np.mean(stats_instance.error_members, axis=0)``
```
RoyStegeman, line 759: I still don't really understand why you mention this? Is it just to explain the above in a different way? Since I think the above explanation of "rep0 is defined the average over the error members" is just fine

RoyStegeman (approved):

validphys2/src/validphys/core.py
scarlehoff: Because np.mean(error_members) is not the same as data[0] which is the central value, for instance in a simple hadronic case we have <f^2> vs <f>^2

RoyStegeman: Okay I see, I was indeed just thinking of PDFs. Maybe it's worth stressing indeed that these "vp objects" can be any function of a PDF set. Even though in principle this explanation makes sense and I agree the misunderstanding is my fault, the "By the PDF convention, the central value corresponds to member 0" made me think of PDFs - probably because I don't really talk about member 0 when it's anything other than PDFs.

RoyStegeman:
Thanks, I like the change. Sorry for being a pain ;)

PR #1869, opened 2023-11-28
Re-implementation of the CMS Collider DY Datasets

Radonirinaunimi:
The following PR implements the CMS collider DY in the new commondata format. The status is summarized in the table below.

☑️  in **Comparison vs. Old** means that the results are fully identical while 🔴 means that comparisons are available but noticeable differences are perceived.

| Dataset Name  | Comparison vs. Old |  General Comments | Status  |
|---|---|---|---|
| CMS_WP_7TEV_MUON_ASY  | ☑️  | Slight differences in the input rawdata vs hepdata entries. In addition, the order in which the correlated uncertainties are slightly different (see last dataset below). |   ☑️  |
| CMS_WP_7TEV_ELECTRON_ASY  | ☑️   |   | ☑️   |
| CMS_Z0_7TEV_DIMUON_2D  | ☑️   |   | ☑️   |
| CMS_WPWM_8TEV_MUON_Y  |  ☑️    | The entries of the correlated (systematics) are exactly the same but they are ordered differently. Somehow, in the new implementation, the treatment ADD (hence the correlated statistics) always go first. There might a mistake in the implementation since the correlation matrix from the systematic uncertainties are both used to generate artificial systematic and statistical uncertainties.  |  ☑️   |


Radonirinaunimi:
This is now ready for review as it completes the CMS collider DY implementation.

PR #1870, opened 2023-11-28
Fix typo in validphys

niclaurenti:
Fix typo in validphys


RoyStegeman (approved):
Did you check for any other instances where this is wrong?

niclaurenti:
> Did you check for any other instances where this is wrong?

No, but I can search if it is wrong in other places. As soon as I have time I'll do it

niclaurenti:
@RoyStegeman I couldn't find any other places where it is wrong

PR #1872, opened 2023-11-29
Set `reportengine version ==0.30.28`

scarlehoff:
This is the version used in the environment in #1852. Newer versions would change quite a few things and are currently untested together with this repository. Since there are effectively no maintainers for `reportengine` I think we should stop updates until someone can commit the time to actually making sure that everything works ok.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -37,7 +37,7 @@ requirements:
         - apfel
         - pkg-config
         - python # requirements for validphys
-        - reportengine >=0.30.25 # see https://github.com/NNPDF/reportengine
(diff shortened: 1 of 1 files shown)


conda-recipe/meta.yaml
```diff
@@ -37,7 +37,7 @@ requirements:
         - apfel
         - pkg-config
         - python # requirements for validphys
-        - reportengine >=0.30.25 # see https://github.com/NNPDF/reportengine
+        - reportengine >=0.30.25,<=0.30.28 # see https://github.com/NNPDF/reportengine
```
scarlehoff, line 40:
```suggestion
        - reportengine ==0.30.28  # see https://github.com/NNPDF/reportengine
```

RoyStegeman (approved):
Thanks

PR #1875, opened 2023-12-01
Allow photon PDF to reports

niclaurenti:
Allow photon PDF in reports through the argument `--photon` in `vp-comparefits`

### DIFF
--- a/validphys2/src/validphys/scripts/vp_comparefits.py
+++ b/validphys2/src/validphys/scripts/vp_comparefits.py
@@ -79,7 +79,7 @@ class CompareFitApp(App):
         parser.add_argument(
             '-p',
             '--photon',
-            help="Add photon PDF to the report",
+            help="Use LUX basis (which include the photon) for the report",
             action='store_true')
 
         parser.set_defaults()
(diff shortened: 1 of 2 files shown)


niclaurenti:
This is the output of the report:
https://vp.nnpdf.science/7DxRLas3S_a3zqixFEzADg==

validphys2/src/validphys/scripts/vp_comparefits.py
```diff
@@ -76,6 +76,11 @@ def add_positional_arguments(self, parser):
             '--lite',
             help="Smaller version of the usual comparefit fit",
             action='store_true')
+        parser.add_argument(
+            '-p',
+            '--photon',
+            help="Add photon PDF to the report",
```
scarlehoff, line 82:
```suggestion
            help="Use LUX basis (which include the photon) for the report",
```

to be a bit more specific on what the flag is doing.

scarlehoff (approved):
I'm fine with this soluton

RoyStegeman:
I think you can just add the photon to `FLAVOUR` without changing `default_elements` and do `copy.deepcopy` for the `LUX_FLAVOUR` where the photon is in the `default_elements`. Same as we did for evolution

RoyStegeman (approved):
Thanks, LGTM!

PR #1876, opened 2023-12-01
Add two extra colors

scarlehoff:
Right now we are using this list of 8 https://scottplot.net/cookbook/4.1/colors/#dark-pastel
But of course, we added cbar and photon so now we need 10

See e.g.:

https://vp.nnpdf.science/7DxRLas3S_a3zqixFEzADg==/figures/Scales0_Normalize_Basespecs0_PDFscalespecs0_Distspecs0_plot_pdfdistances.pdf

So I've added two extra colors at the end. If you don't like them suggestions are welcome. One example here:

image


scarlehoff:
It's also fine by me. Feel free to update the two files.

Issue #1856, opened 2023-11-21
Add 3FNS  in default grids.


giacomomagni:
Would it be possible to switch to a default grid in the evolution that starts from 1 GeV ?
we should just adopt this grid as default: 

https://github.com/NNPDF/nnpdf/blob/ce6c05cacec95dee36c577503c9a1109daa2908a/n3fit/src/evolven3fit_new/utils.py#L67

and change these lines:

https://github.com/NNPDF/nnpdf/blob/ce6c05cacec95dee36c577503c9a1109daa2908a/n3fit/src/evolven3fit_new/utils.py#L188-L194

This way all the uploaded fits, will automatically include IC in 3FNS and there is no need to do further work. 

Thanks. 

@RoyStegeman , @andreab1997 @scarlehoff

scarlehoff:
What do you mean exactly by this?
In principle the grid is defined by the eko in the theory. So you mean that, when creating the evolution eko for the theory, you would like to have another option available, right?

Theories below < 400 should all have ekos that start at Q=1.65 GeV

For theories above, the ekos can be updated to use Q=1 GeV as the starting evolution and that won't be a problem.

Of course I would be in favour of having a `--qgrid_option grid_option_N` with several defaults to choose from.

RoyStegeman:
In what case would those grids even be used? Looking at where `generate_q2grid` is called it will be used if there is a theory without an eko.tar in the folder. One perhaps strange thing is that it looks like the `Q0` from the theory will be ignored if `Qfin` and `Q_points` are not passed. 

Either way, I don't really understand why those specific grids were defined the way they are  (maybe @scarlehoff does) so for me it's fine if both the nf=4 and nf=3 are the same default grids starting from 1 GeV.

scarlehoff:
The grids were defined to exactly reproduce the grids used in the fits in NNPDF4.0 (and those need to be kept of course since we often want to reproduce NNPDF4.0). 

>  if there is a theory without an eko.tar in the folder

Given how slow generating ekos is, there should be none (and maybe we should remove that option altogether).

> for me it's fine if both the nf=4 and nf=3 are the same default grids starting from 1 GeV.

I would not change any defaults because it already took some tuning to be able to reproduce 4.0 and I would like to really substitute `evolven3fit` by `evolven3fit_new` and remove apfel.

giacomomagni:
My point is to have all the `lhapdf` sets that we produce including already the FFNS `nf=3`. 
[...]

PR #1879, opened 2023-12-04
Add --legacy40 flag for evolven3fit_new to use the grid for nf=3 instead of nf=4

scarlehoff:
So that the `eko` that are being used to produce many of the results can easily be reproduced with `evolven3fit_new`.

This deals with #1856.


n3fit/src/evolven3fit_new/utils.py
```diff
@@ -178,15 +178,17 @@ def get_theoryID_from_runcard(usr_path):
...
+def generate_q2grid(Q0, Qfin, Q_points, match_dict, nf0=None, ic=False):
     """Generate the q2grid used in the final evolved pdfs or use the default grid if Qfin or Q_points is
     not provided.
 
     match_dict contains the couples (mass : factor) where factor is the number to be multiplied to mass
     in order to obtain the relative matching scale.
     """
     if Qfin is None and Q_points is None:
-        if nf0 == 4:
+        if nf0 == 4 and ic:
```
giacomomagni, line 189:
```suggestion
        if nf0 == 3 or ic:
```
this way you can remove the last branch.

n3fit/src/n3fit/scripts/evolven3fit_new.py
```diff
@@ -34,6 +34,9 @@ def construct_eko_parser(subparsers):
     parser.add_argument(
         "-p", "--x-grid-points", default=None, type=int, help="Number of points of the x-grid"
     )
+    parser.add_argument(
+        "--ic", action="store_true", help="Use the nf=3 grid even if the fit was run with nf=4"
```
giacomomagni, line 38:
```suggestion
        "--ic", action="store_true", help="Use the nf0=3 grid even if the fit was run with nf0=4"
```

n3fit/src/evolven3fit_new/utils.py
```diff
@@ -178,15 +178,17 @@ def get_theoryID_from_runcard(usr_path):
...
+            return Q2GRID_Nf03
+        elif nf0 == 4:
             return Q2GRID_Nf04
         elif nf0 == 3:
             return Q2GRID_Nf03
```
scarlehoff, line 194:
```suggestion
```

giacomomagni (approved):

scarlehoff:
Ok, given the discussion let me change this to:

1. Use always a "complete down to 1 GeV grid'
2. Add a few extra points before 1.65 GeV that are basically the same that are already in the nf=3

And remove the different conditions. This means the result of `evolven3fit` and `evolven3fit_new` would be different until I update the 4.0 theories but I'll do that little by little in the next few weeks (and theory 200 is already done so the most important one is safe).

giacomomagni (commented):
LGTM. Stupid question: i'd have to redo also my ekos if I've used `Q2GRID_Nf03` before right?

scarlehoff:
>  i'd have to redo also my ekos if I've used Q2GRID_Nf03 before right?

I'd say not urgent (and specially not for N3LO). 
The problem of having used nf=3 is that you would have some differences at the boundaries (because they happen at slightly different values of Q). But this kind of 1-to-1 comparison might be relevant (if it ever is) only for QED which is really supposed to be "4.0 with the photon".

scarlehoff:
With `--legacy40` the resulting LHAPDF grid file agrees with the apfel evolution at 1e-3 with some bigger than that differences in the thresholds and around them (Q ~ 4.92 GeV). This differenced was already there in the past and @AleCandido and @felixhekhorn convinced me that the eko result was correct then so I'm happy to go with it.

The default eko should be finish soon, then I'll check that the results above 1.65 GeV are also consistent and this can be merged.

PR #1881, opened 2023-12-04
Merge prefactors into single layer

APJansen:
Join the preprocessing layers into one.

This also required changing the way weights are extracted and set, as a replica now is only a single slice of a weight tensor. Locally for me the way I implemented it is giving errors that seem specific to Mac, running on the GPU.

## Status: Done, ready for review
- [x] I started changing stuff in vpinterface, but I think what I should do instead is similar to what I did before with the PDF as a whole: use the `replica_axis` flag I made there, to keep everything as is in the single replica models (i.e. not have an extra replica axis of length 1 there) so nothing there has to change.
- [x] It is producing very different results, because the initialization is different. Keeping it the same would be very ugly, but I can do it temporarily just to verify that that's the only source of differences.
- [x] Fix test_fit

Edit: the changes of this PR changed the layout of the weights and thus weights saved for earlier versions of the code need to be restructured so they remain equivalent. The following script (comment: https://github.com/NNPDF/nnpdf/pull/1881#issuecomment-1857581109) achieves exactly that:

<details>

<summary>test script</summary>

```python
import numpy as np
import pickle

from n3fit.layers import Preprocessing


flav_info = [
    {"fl": i, "largex": [0, 1], "smallx": [1, 2]}
    for i in ["u", "ubar", "d", "dbar", "c", "g", "s", "sbar"]
]

np.random.seed(42)
x_in = np.random.rand(1, 10, 1)

branch = "master"

if branch == "parallel-prefactor":
    print("Running parallel-prefactor")
    weights_parallel = {}
    outs_parallel = {}
    for large_x in [True, False]:
       for num_replicas in [1, 3]:
           preproc = Preprocessing(
                   flav_info=flav_info,
                   input_shape=(1,),
                   name="preprocessing_factor",
                   seed=42,
                   large_x=large_x,
                   num_replicas=num_replicas,
               )
           preproc.build((None, 1,))
           weights_parallel[(large_x, num_replicas)] = preproc.get_weights()
           outs_parallel[(large_x, num_replicas)] = preproc(x_in)
    with open("tmp/preproc_weights.pkl", "wb") as f:
       pickle.dump(weights_parallel, f)
    with open("tmp/preproc_outs.pkl", "wb") as f:
       pickle.dump(outs_parallel, f)

if branch == "master":
    print("Running master")
    with open("tmp/preproc_weights.pkl", "rb") as f:
        weights_parallel = pickle.load(f)
    with open("tmp/preproc_outs.pkl", "rb") as f:
        outs_parallel = pickle.load(f)
    for large_x in [True, False]:
        for num_replicas in [1, 3]:
            print(f"large_x: {large_x}, num_replicas: {num_replicas}")
            preprocs = []
            for i_replica in range(num_replicas):
                preproc = Preprocessing(
                    flav_info=flav_info,
                    input_shape=(1,),
                    name=f"preprocessing_factor_{i_replica}",
                    seed=42,
                    large_x=large_x,
                )
                preproc.build((None, 1))

                weights_this_replica = []
                for w in weights_parallel[(large_x, num_replicas)]:
                    weights_this_replica.append(w[i_replica])

                preproc.set_weights(weights_this_replica)
                preprocs.append(preproc)
            outs = np.stack(
                [preprocs[i_replica](x_in) for i_replica in range(num_replicas)], axis=1
            )
            np.testing.assert_equal(outs, outs_parallel[(large_x, num_replicas)])
```


APJansen:
@scarlehoff @RoyStegeman @Radonirinaunimi Just so you know, I rebased this and updated the status above, in case you want to have a look. But it's still WIP so no need to review yet.

n3fit/src/n3fit/model_gen.py
```diff
@@ -669,19 +669,18 @@ def pdfNN_layer_generator(
     else:
         sumrule_layer = lambda x: x
 
+    compute_preprocessing_factor = Preprocessing(
+        flav_info=flav_info,
+        input_shape=(1,),
+        name="preprocessing_factor",
+        seed=42,  # TODO Aron: figure out what to do here
```
scarlehoff, line 676: Here it is absolutely impossible to keep the same random result than before, so just do `seed=replica_seed + number_of_layers` or whatever takes you out from a seed previously used in a replicable manner and that's it.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -379,8 +379,10 @@ def set_replica_weights(self, weights, i_replica=0):
                 the replica number to set, defaulting to 0
         """
         self.get_layer(f"{NN_PREFIX}_{i_replica}").set_weights(weights[NN_PREFIX])
-        self.get_layer(f"{PREPROCESSING_PREFIX}_{i_replica}").set_weights(
-            weights[PREPROCESSING_PREFIX]
+        set_layer_replica_weights(
+            layer=self.get_layer(PREPROCESSING_PREFIX),
+            weights=weights[PREPROCESSING_PREFIX],
+            i_replica=i_replica,
```
scarlehoff, line 385: Instead of doing it in this way, which breaks down the layer into smaller layers, it might be more convenient to do the following:

1. Get all weights, now you have a numpy array
2. Modify the subset that you actually need to modify.
3. Set all weights.

You are getting and resetting the same bunch of numbers for all replicas, it is true, but setting weights by replica will only happen once per fit so the performance hit is more than acceptable. This greatly simply the code and you avoid playing with fire...

scarlehoff (commented):
I've left a suggestion which I think might simplify the code and bypass completely the problem you have in mac (and, probably, bypass future problems with other tensorflow versions because setting all weights will always work, breaking down a layer into smaller variables is more tricky).

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
APJansen: Thanks, good idea, it seems to work!

APJansen:
I have done a manual test, first creating a prefactor layer in this branch and getting the output on some test input. Then in master create the same layers, and set the weights to those in this branch. Comparing outputs, they're identical, both for 1 and for 3 replicas.

This is also the reason the tests are still failing, it's only some regression tests that fail, so they should be updated.
Can you remind me how this was done?

Also, I changed the name of the layer from `preprocessing_{i_replica}` to just `preprocessing`.
Are you ok with that, for both the multiple replica model and the single replica ones, with the latter also keeping a length 1 axis dimension? Contrary to the comment in the initial post, I don't think it's worth differentiating this. For the output of the PDF it was worth it because this saved updating validphys, but I don't think that looks into the internals of the model, so here it's not necessary.

The same will apply of course to the `NN_{i_replica}` layers, and in both cases saved models will need to be updated because of this change in structure.

<details>

<summary>test script</summary>

```python
import numpy as np
import pickle

from n3fit.layers import Preprocessing


flav_info = [
    {"fl": i, "largex": [0, 1], "smallx": [1, 2]}
    for i in ["u", "ubar", "d", "dbar", "c", "g", "s", "sbar"]
]

np.random.seed(42)
x_in = np.random.rand(1, 10, 1)

branch = "master"

if branch == "parallel-prefactor":
    print("Running parallel-prefactor")
    weights_parallel = {}
    outs_parallel = {}
    for large_x in [True, False]:
       for num_replicas in [1, 3]:
           preproc = Preprocessing(
                   flav_info=flav_info,
                   input_shape=(1,),
                   name="preprocessing_factor",
                   seed=42,
                   large_x=large_x,
                   num_replicas=num_replicas,
               )
           preproc.build((None, 1,))
           weights_parallel[(large_x, num_replicas)] = preproc.get_weights()
           outs_parallel[(large_x, num_replicas)] = preproc(x_in)
    with open("tmp/preproc_weights.pkl", "wb") as f:
       pickle.dump(weights_parallel, f)
    with open("tmp/preproc_outs.pkl", "wb") as f:
       pickle.dump(outs_parallel, f)

if branch == "master":
    print("Running master")
    with open("tmp/preproc_weights.pkl", "rb") as f:
        weights_parallel = pickle.load(f)
    with open("tmp/preproc_outs.pkl", "rb") as f:
        outs_parallel = pickle.load(f)
    for large_x in [True, False]:
        for num_replicas in [1, 3]:
            print(f"large_x: {large_x}, num_replicas: {num_replicas}")
            preprocs = []
            for i_replica in range(num_replicas):
                preproc = Preprocessing(
                    flav_info=flav_info,
                    input_shape=(1,),
                    name=f"preprocessing_factor_{i_replica}",
                    seed=42,
                    large_x=large_x,
                )
                preproc.build((None, 1))

                weights_this_replica = []
                for w in weights_parallel[(large_x, num_replicas)]:
                    weights_this_replica.append(w[i_replica])

                preproc.set_weights(weights_this_replica)
                preprocs.append(preproc)
            outs = np.stack(
                [preprocs[i_replica](x_in) for i_replica in range(num_replicas)], axis=1
            )
            np.testing.assert_equal(outs, outs_parallel[(large_x, num_replicas)])
```

scarlehoff:
The tests for 3.11 are pointing to something else though:

```
        TypeError: 'NoneType' object cannot be interpreted as an integer
```

RE the label of the replicas. No I don't think that matters, it's completely internal after all.

For the regression instead, just used the `quickcard_1.json` (and similarly names) to update the ones that failed.
It needs to be done manually because they don't fail often enough so that we've automatized the process... (obv, for you they fail more because you're touching the fitting code directly)

APJansen:
I just repeated the procedure above on Snellius, updating the weights_1.h5 and quickcard_1.json from a run on Snellius and running the corresponding test. There as well it still fails.

RoyStegeman:
> That should fix it right? 

Looking at what exactly the test does this should indeed not work. Namely, it uses the `hdf5` file to initialise the NN, and subsequently still performs the fit. So when doing the steps as you describe, the `json` files you saved correspond to the starting point while the `json` files they're compared to correspond to the end of the fit. One possible solution is of course to run the fit twice: once to produce the `hdf5` file, and then load the weights from that file to produce the `json` files needed for the comparison. 

It's not so clear to me why the the weights are loaded from the file as opposed to just initialised using the rng in this regression test (maybe to ensure they are always the same?). And of course I don't know how the current `hdf5` files were generated (in what kind of fit setup), though in principle for this regression test that doesn't matter. @scarlehoff can probably clarify on these points.

scarlehoff:
Just to be precise, "fixing the regression test" here should mean:

1. Converting the current weights into the new format (and if you can please leave the script here in the PR for future reference, since it might be necessary to use it in the future; it's not a widely used feature so I'm ok with this compromise, but I don't think we should merge something that breaks all previous weights)
2. Seeing how the numbers change due to this changes (hopefully not much when the starting point is the same).
3. Updating the numbers in the `.json` file only once 2. is properly checked.

APJansen:
Ah I see, that makes sense, I'll look into it, thanks!

APJansen:
Fixed! And ready for review.

I don't know if you think it's necessary to do more tests on top of the CI, but if you do I would propose to postpone that and just do it in #1905, which branches off of this.

scarlehoff:
For merging, I'd say the same as with:
https://github.com/NNPDF/nnpdf/pull/1888#issuecomment-1885356414

Btw, one point that I wanted to raise during the meeting this morning (which I could not attend) is that please either squash or rebase (given the amount of commits -and intermediate merges- probably you need both) these PR such that their history is not mixed with master.

APJansen:
> Btw, one point that I wanted to raise during the meeting this morning (which I could not attend) is that please either squash or rebase (given the amount of commits -and intermediate merges- probably you need both) these PR such that their history is not mixed with master.

Yes I tried that yesterday (on nn-layers-refactor), but I got some error I didn't understand, so I merged instead, but I can take another look.

scarlehoff:
Probably at this point squashing first and then rebase is the best thing to do.

RE your comment in the other PR, if it takes more than a few weeks yes. I was honestly hoping it was going to be all done by December x)

APJansen:
Rebasing worked, if you want I can also squash the commits into one per PR.

scarlehoff:
If we need to go the route of having a PR collecting all finished PR it would be a good idea (so that each of the branches becomes one commit in that PR) but I'm still hopeful that the papers will be out by next / the following week and that we will be able to continue normally xD

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -46,7 +46,8 @@
 }
 
 NN_PREFIX = "NN"
-PREPROCESSING_PREFIX = "preprocessing_factor"
+NN_LAYER = "NNs"
+PREPROESSING_LAYER = "preprocessing_factor"
```
scarlehoff, line 50:
```suggestion
PREPROCESSING_LAYER = "preprocessing_factor"
```

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -353,14 +354,12 @@ def get_replica_weights(self, i_replica):
...
-            tf.Variable(w, name=w.name)
-            for w in self.get_layer(f"{PREPROCESSING_PREFIX}_{i_replica}").weights
-        ]
-        weights = {NN_PREFIX: NN_weights, PREPROCESSING_PREFIX: prepro_weights}
+        weights = {}
+        for layer_type in [NN_LAYER, PREPROESSING_LAYER]:
+            weights[layer_type] = [
+                tf.Variable(w, name=w.name)
+                for w in get_layer_replica_weights(self.get_layer(layer_type), i_replica)
+            ]
```
scarlehoff, line 362:
```suggestion
        weights = defaultdict(list)
        for layer_type in [NN_LAYER, PREPROCESSING_LAYER]:
            layer_ref = self.get_layer(layer_type)
            for w in get_layer_replica_weights(layer_ref, i_replica):
                weights[layer_type].append(tf.Variable(w, name=w.name))
```

This doesn't get called many times, right? (For the same number of lines I'd rather have a more explicit approach, otherwise the list comprehension might be useful)

Also, why do you need `tf.Variable` here if you are already doing `tf.Variable` in `get_layer_replica_weights`?
If needed, wouldn't it be better to do it in that function?

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -378,10 +377,10 @@ def set_replica_weights(self, weights, i_replica=0):
...
                 the replica number to set, defaulting to 0
         """
-        self.get_layer(f"{NN_PREFIX}_{i_replica}").set_weights(weights[NN_PREFIX])
-        self.get_layer(f"{PREPROCESSING_PREFIX}_{i_replica}").set_weights(
-            weights[PREPROCESSING_PREFIX]
-        )
+        for layer_type in [NN_LAYER, PREPROESSING_LAYER]:
+            set_layer_replica_weights(
+                layer=self.get_layer(layer_type), weights=weights[layer_type], i_replica=i_replica
+            )
```
scarlehoff, line 383:
```suggestion
        for l_type in [NN_LAYER, PREPROESSING_LAYER]:
            layer = self.get_layer(l_type)
            set_layer_replica_weights(layer=layer, weights=weights[l_type], i_replica=i_replica)
```

Yes, yes, feel free to ignore this one while sighing

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -411,51 +410,78 @@ def load_identical_replicas(self, model_file):
...
-        weights = {}
 
-        with h5py.File(model_file, 'r') as f:
-            # look at layers of the form NN_i and take the lowest i
-            i_replica = 0
-            while f"{NN_PREFIX}_{i_replica}" not in f:
-                i_replica += 1
+def stacked_single_replicas(layer):
+    """
+    Check if the layer consists of stacked single replicas (Only happens for NN layers)
```
scarlehoff, line 423: could you add a bit more to this docstr about why this is needed?

Maybe since it is a check, a better function name would be `is_stacked_single_replicas`?

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -46,7 +46,8 @@
 }
 
 NN_PREFIX = "NN"
-PREPROCESSING_PREFIX = "preprocessing_factor"
+NN_LAYER = "NNs"
```
scarlehoff, line 49: Is the fact that `NN_LAYER` is prefixed with `NN` like `NN_PREFIX` a necessary thing? Otherwise it could be a problem in the future.

If it is not necessary, maybe `NN_LAYER` could be `Many_NN` (or I don't know, any other name, it should only be a name).

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -411,51 +410,78 @@ def load_identical_replicas(self, model_file):
...
+    if not isinstance(layer, MetaModel):
+        return False
+    return f"{NN_PREFIX}_0" in [sublayer.name for sublayer in layer.layers]
+
+
+def get_layer_replica_weights(layer, i_replica: int):
+    """
+    Get the weights for the given single replica `i_replica`,
+    from a `layer` that has weights for all replicas.
+
```
scarlehoff, line 444:
```suggestion

    Note that the layer could be a complete  a complete NN with many separated sub_layers
    each of which containing weights for all replicas together.
    This functions separates the per-replica weights and returns the list of weight as if the
    input ``layer`` were made of _only_ replica ``i_replica``.
```

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -411,51 +410,78 @@ def load_identical_replicas(self, model_file):
...
+    """
+    if not isinstance(layer, MetaModel):
+        return False
+    return f"{NN_PREFIX}_0" in [sublayer.name for sublayer in layer.layers]
+
+
+def get_layer_replica_weights(layer, i_replica: int):
+    """
+    Get the weights for the given single replica `i_replica`,
+    from a `layer` that has weights for all replicas.
```
scarlehoff, line 443:
```suggestion
    Get the weights for the given single replica ``i_replica``,
    from a ``layer`` that has weights for all replicas.
```

I think otherwise sphinx doesn't consider it code but rather italic

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -411,51 +410,78 @@ def load_identical_replicas(self, model_file):
...
+        weights = [tf.Variable(w[i_replica : i_replica + 1], name=w.name) for w in layer.weights]
+
+    return weights
+
+
+def set_layer_replica_weights(layer, weights, i_replica: int):
+    """
+    Set the weights for the given single replica `i_replica`,
+    for a `layer` that has weights for all replicas.
+
```
scarlehoff, line 469:
```suggestion
    Set the weights for the given single replica ``i_replica``,
    When the input ``layer`` contains weights for many replicas, ensures that
    only those corresponding to replica ``i_replica`` are updated.
```

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -411,51 +410,78 @@ def load_identical_replicas(self, model_file):
...
-        # have to put them in the same order
-        weights_ordered = []
-        weights_model_order = [w.name for w in self.get_replica_weights(0)[weights_key]]
-        for w in weights_model_order:
-            for w_h5 in weights:
-                if w_h5.name == w:
-                    weights_ordered.append(w_h5)
-
-        return weights_ordered
+    layer.set_weights(full_weights)
```
scarlehoff, line 487: If you want to play around, I'm guess you could use https://www.tensorflow.org/api_docs/python/tf/tensor_scatter_nd_update (and gather in get) to avoid the numpy part.

On the other hand, probably sticking to numpy and set_weights is much more robust towards possible tensorflow upgrades :P

n3fit/src/n3fit/layers/msr_normalization.py
```diff
@@ -96,8 +98,10 @@ def call(self, pdf_integrated, photon_integral):
         divisors = op.gather(y, self.divisor_indices, axis=0)
 
         # Fill in the rest of the flavours with 1
+        # (Note: using y.shape in the output_shape below gives an error in Python 3.11)
```
scarlehoff, line 101:
```suggestion
```

I think it is even better like this tbh. Out of curiosity, is the error due to python 3.11 or 3.11 getting a newer version of tf?

n3fit/src/n3fit/model_gen.py
```diff
@@ -572,18 +572,14 @@ def pdfNN_layer_generator(
...
-                input_shape=(1,),
-                name=f"preprocessing_factor_{i_replica}",
-                seed=replica_seed + number_of_layers,
-                large_x=not subtract_one,
-            )
-        )
+    compute_preprocessing_factor = Preprocessing(
+        flav_info=flav_info,
+        input_shape=(1,),
+        name="preprocessing_factor",
```
scarlehoff, line 578: Please, import this from `MetaModel` to make sure that it is consistent.

n3fit/src/n3fit/model_gen.py
```diff
@@ -572,18 +572,14 @@ def pdfNN_layer_generator(
...
+        seed=seed[0] + number_of_layers,
```
scarlehoff, line 579: This means that training replicas 1 to 10 and 2 to 10 won't give me the same initial results, right?

However, would 1 to 10 and 1 to 9 produce the exact same result from 1 to 9? I would like to know that.

n3fit/src/n3fit/model_trainer.py
```diff
@@ -454,7 +454,7 @@ def _model_generation(self, xinput, pdf_model, partition, partition_idx):
             training.summary()
             pdf_model = training.get_layer("PDFs")
             pdf_model.summary()
-            nn_model = pdf_model.get_layer("NN_0")
+            nn_model = pdf_model.get_layer("NNs")
```
scarlehoff, line 457: Same as above, please take this directly from the constants set in MetaModel

n3fit/src/n3fit/tests/test_modelgen.py
```diff
@@ -25,7 +25,7 @@
 
 
 def test_generate_dense_network():
-    nn = generate_nn("dense", **COMMON_ARGS)[0]
+    nn = generate_nn("dense", **COMMON_ARGS).get_layer("NN_0")
```
scarlehoff, line 28: as above
the same below

n3fit/src/n3fit/vpinterface.py
```diff
@@ -224,13 +224,7 @@ def get_preprocessing_factors(self, replica=None):
...
             replica = 1
         # Replicas start counting in 1 so:
-        preprocessing_layers = self._models[replica - 1].get_layer_re(r"preprocessing_factor_\d")
-        if len(preprocessing_layers) > 1:
-            # We really don't want to fail at this point, but print a warning at least...
-            log.warning("More than one preprocessing layer found within the model!")
-        elif len(preprocessing_layers) < 1:
-            log.warning("No preprocessing layer found within the model!")
-        preprocessing_layer = preprocessing_layers[0]
+        preprocessing_layer = self._models[replica - 1].get_layer("preprocessing_factor")
```
scarlehoff, line 227: and here

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
APJansen: No that shouldn't be necessary, I renamed it to `NN_LAYER_ALL_REPLICAS`, and same for `PREPROCESSING`, is that ok?

scarlehoff: I think my message was confusing :__

I meant `NN_LAYER = "Many_NN"` so that `NN_LAYER` doesn't use the `NN` prefix! The name of the variable itsel fis ok.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
APJansen: Haha no I agree it's nicer, fixed :)

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
APJansen: I think you're right that `tf.Variable` is not needed here, so it just becomes a oneliner.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
APJansen: In the context of this PR, for the preprocessing layer we extract replica weights by taking the right slice of the multi-replica preprocessing layer, while the NN (dense) layers are not joined yet and need to be extracted in a different way. So here this could be replaced with simply a check if the layer is a preprocessing or a nn layer.
However in the next PR, NNs will also be joined together into one layer, so that won't be possible, and still we need to preserve the old way of extracting weights just for the dense_per_flavour layer, where replicas are not joined. (I guess it's still possible to do there, but I didn't want to waste time on that. I have added `is_` and explained a little, is that ok?

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
APJansen: Ok how about `all_NNs`?

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
scarlehoff: Yes. My comment was mostly about giving it some context.

n3fit/src/n3fit/layers/msr_normalization.py
APJansen: I'm not sure to be honest, the fix works regardless and it didn't look worse, so I didn't investigate further.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
scarlehoff: Whatever name not `NN`-prefixed works, if every other module uses this constant it can be changed for something more beautiful at any point.

n3fit/src/n3fit/model_gen.py
scarlehoff: Well, you can add a `constants.py` or something if you don't like the import from `n3fit.backends`. If they are supposed to be the same they should be the same. But in `model_gen` you need to import from `n3fit.backends` anyway so it doesn't create any circular problems.

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
APJansen: Yes I agree the numpy part is probably more robust. I had something using just tensorflow before which worked on some versions but not on my mac..

n3fit/src/n3fit/model_gen.py
APJansen: You mean `n3fit 1 -r 10` vs `n3fit 2 -r 10`, and `n3fit 1 -r 10` versus `n3fit 1 -r 9`? I'll test, I'm actually not sure how those arguments get fed into `pdfNN_layer_generator`.

scarlehoff: Running in parallel I mean. You normally start at replica 1 but it is not necessarily always like that.

The reason why I'm asking is that I think right now if you instantiate, say, 10 replicas (to fit them in parallel for instance) each of them will get a separate seed for the generation of the pseudo-data and, until now, a separate seed for each of the Neural Networks.
As long as the seeds were separated that meant the initial point could be reproducible.

Instead, now all replicas are being instantiated with the first seed, so obviously if you fit 1 to 10 you will get a result different from 2 to 10 (because the corresponding first seed is different).

But I am wondering whether 1 to 10 would be equivalent to 1 to 9 for the first 9 replicas.

Then, one thing to be done for when parallel fit becomes the standard is recover that reproducibility. For now reproducibility is limited to sequential CPU replicas. Reproducibility cannot be achieved in any easy way in GPU, but at least the initial state can be made to be the same.

APJansen: I checked `n3fit 1 -r 4`, `n3fit 2 -r 4`, and `n3fit 1 -r 5`, and there were no similar results anywhere. For the `MultiDense` layers I did make sure that initialization is identical to master, here I did not.

Perhaps it's easiest to use the solution from there: [MultiInitializer](https://github.com/NNPDF/nnpdf/blob/341e39d60fb22383bd6c63559141288190360e86/n3fit/src/n3fit/backends/keras_backend/multi_dense.py#L148)?

scarlehoff: If it works out of the box for the preprocessing, yes indeed.
But if it doesn't we can leave that for later.

Btw, since you're at it, check that with seed[0] + number_of_layers we are not being silly and using the same seed twice. That was something that worked well when every layer was using seed +=1 but not sure whether that's the case anymore.

APJansen: If it's ok to leave for later, I'll try it in the next PR #1905, to save some hassle of moving the code between branches.

APJansen: The seeds should still be ok, I haven't changed the seeding of the NN layers, neither in this PR nor in the next.

n3fit/src/n3fit/model_gen.py
APJansen: A separate `constants.py` makes sense, but then lots of other stuff should move there too, I'd say let's leave that for later ;P

scarlehoff (approved):
Just to confirm, is this finished now? (can it join the bulk of PRs to be merge to master?)

Does anybody else want to look at it? (not sure whether @Radonirinaunimi or @RoyStegeman already did have look)

Radonirinaunimi:
> Just to confirm, is this finished now? (can it join the bulk of PRs to be merge to master?)
> 
> Does anybody else want to look at it? (not sure whether @Radonirinaunimi or @RoyStegeman already did have look)

Yes, thanks! I'd like to have at look at this and #1788 today (before the CM).

Radonirinaunimi:
> #1788 needs to be first squashed and rebased on top of what-will-be-master before being ready to be merged though

Yes, this I understood. But we can nonetheless start reviewing that, no? Or, do you expect to have many merged changes into master that'll later affect the review of that PR?

scarlehoff:
> do you expect to have many merged changes into master that'll later affect the review of that PR?

I don't know. But wanted to warn you just in case!

APJansen:
> Just to confirm, is this finished now? (can it join the bulk of PRs to be merge to master?)

Yes if you are happy with it this is ready to be merged!

> Or, do you expect to have many merged changes into master that'll later affect the review of that PR?

I don't think so, just did a quick test merging this branch into #1788 (without pushing) and there were no merge conflicts at all.

scarlehoff:
btw @APJansen did you already rebased / merged on top / with of #1914 ?

scarlehoff:
> but a quick test run fails with ModuleNotFoundError: No module named 'validphys._version'

You need to reinstall `pip install -e .` would work but also cmake.

Radonirinaunimi (approved):
Apart from the pending issue regarding the full reproducibility, this also LGTM.

Below are just some tiny fixes regarding docstrings.

APJansen:
The clean install worked straight away! :)

And I've rebased this onto the latest master and squashed it into a single commit.

PR #1884, opened 2023-12-07
Set to 0 not active flavours

scarlehoff:
In the previous evolution (I think apfel?) was setting to 0 the flavours that were not active.

Instead, with eko, a non-evolved flavour just remains unchanged, which means that values of `-1e-7` in the bottom column are repeated for every value of Q until the threshold is reached.

Not a big issue in general, but if people have code that just breaks when it sees a negative value this would create unnecessary problems and we want our grids to be as problem free as possible.

### DIFF
--- a/n3fit/src/n3fit/io/writer.py
+++ b/n3fit/src/n3fit/io/writer.py
@@ -219,7 +219,7 @@ XGRID = np.array(
 
 
 class WriterWrapper:
-    def __init__(self, replica_numbers, pdf_objects, stopping_object, all_chi2s, q2, timings):
+    def __init__(self, replica_numbers, pdf_objects, stopping_object, all_chi2s, theory, timings):
         """
         Initializes the writer for all replicas.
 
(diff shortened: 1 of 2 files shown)


n3fit/src/n3fit/io/writer.py
```diff
@@ -570,14 +565,33 @@ def storefit(
...
     # build exportgrid
     xgrid = XGRID.reshape(-1, 1)
 
+    q0 = theory.get_description().get("Q0")
+    q20 = q0**2
+
+    # Set the active flavours, by default u, d, s
+    active_flavours = 3
+
+    # Now check whether c, b, t are active
```
RoyStegeman, line 580:
```suggestion
    # Now check whether c, b, t are active at Q0
```

n3fit/src/n3fit/io/writer.py
```diff
@@ -519,44 +514,44 @@ def evln2lha(evln):
...
-        - 3 * evln[7]
-        - 2 * evln[8]
-    ) / 120
-
-    lha[11] = (5 * evln[1] - 6 * evln[12] + evln[13] + 5 * evln[3] - 6 * evln[7] + evln[8]) / 60
-
-    lha[1] = (5 * evln[1] - 6 * evln[12] + evln[13] - 5 * evln[3] + 6 * evln[7] - evln[8]) / 60
-
-    lha[12] = (evln[1] - evln[13] + evln[3] - evln[8]) / 12
+    # charm
```
RoyStegeman, line 517:
```suggestion

    # if a heavy quark is not active at Q0 (the scale at which the output of the fit is stored),
    # keep the PDF values at 0.0 to prevent small negative values due to numerical instabilities

    # charm
```

RoyStegeman (approved):
Seems fine to me. Thanks!

PR #1886, opened 2023-12-07
[New Commondata Format] Old Jets datasets

comane:
This PR converts the following datasets (and versions therein) into the new commondata format

- CMS_2JET_7TEV
- CMS_1JET_8TEV
- ATLAS_2JET_7TEV_R06
- ATLAS_1JET_8TEV_R06
- ATLAS_1JET_8TEV_R06_DEC

A report comparing legacy vs New:
https://vp.nnpdf.science/BAIRR312QyuMF3hKa5-9HQ==


buildmaster/dataset_names.yml
```diff
@@ -0,0 +1,11 @@
...
+  dataset: ATLAS_1JET_8TEV_R06_PTY
+  variant: decorrelated
+
+ATLAS_1JET_8TEV_R06: ATLAS_1JET_8TEV_R06_PTY
+
+ATLAS_2JET_7TEV_R06: ATLAS_2JET_7TEV_R06_M12Y
+
+CMS_1JET_8TEV: CMS_1JET_8TEV_PTY
+
+CMS_2JET_7TEV: CMS_2JET_7TEV_M12Y_BIN1
```
scarlehoff, line 11:
```suggestion
CMS_2JET_7TEV: CMS_2JET_7TEV_M12Y
```

buildmaster/ATLAS_1JET_8TEV_R06/metadata.yaml
```diff
@@ -0,0 +1,74 @@
+# Created on Mar 2023, author: Mark N. Costantini
+
+# Generalia
+setname: "ATLAS_1JET_8TEV_R06"
+
+version: 0
```
scarlehoff, line 6:
```suggestion
version: 1
```

This is just a convention ofc

buildmaster/ATLAS_2JET_7TEV_R06/metadata.yaml
```diff
@@ -0,0 +1,81 @@
+# Created on Mar 2023, author: Mark N. Costantini
+
+# Generalia
+setname: "ATLAS_2JET_7TEV_R06"
+
+version: 0
```
scarlehoff, line 6:
```suggestion
version: 1
```

buildmaster/CMS_1JET_8TEV/metadata.yaml
```diff
@@ -0,0 +1,70 @@
+# Created on Apr 2023, author: Mark N. Costantini
+
+setname: "CMS_1JET_8TEV"
+
+version: 0
```
scarlehoff, line 5:
```suggestion
version: 1
```

buildmaster/CMS_1JET_8TEV/metadata.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+    ndata: 239
+    
+    tables: [1,2,3,4,5,6]
+    process_type: JET
+
+    kinematics:
+        variables:
+          y: {description: "jet rapidity", label: "$|y|$", units: ""}
+          p_T2: {description: "jet transverse momentum", label: "$k_{T}$", units: ""}
```
scarlehoff, line 41: Should this be pT or pT^2? (why the 2?)

buildmaster/CMS_2JET_7TEV/metadata.yaml
```diff
@@ -0,0 +1,67 @@
...
+      dataset_label: "CMS dijets 7 TeV"
+      kinematics_override: identity
+      x_scale: log
+      plot_x: m12
+      figure_by: 
+          - ydiff
+
+    kinematics:
+        variables:
+          ydiff: {description: "rapidity separation", label: "$y$", units: ""}
```
scarlehoff, line 52:
```suggestion
          ydiff: {description: "rapidity separation", label: r"$\Delta y$", units: ""}
```

buildmaster/CMS_2JET_7TEV/metadata.yaml
```diff
@@ -0,0 +1,67 @@
+# Created on Mar 2023, author: Mark N. Costantini
+
+# Generalia
+
+setname: "CMS_2JET_7TEV"
+
+version: 0
```
scarlehoff, line 7:
```suggestion
version: 1
```

buildmaster/CMS_1JET_8TEV/metadata.yaml
```diff
@@ -0,0 +1,70 @@
...
+        - - CMS_1JET_8TEV
+      operation: 'null'
+
+    # Plotting information
+    plotting:
+        dataset_label: "CMS jets 8 TeV"
+        plot_x: y
+        figure_by:
+            - p_T2
+    kinematic_coverage: [y, p_T2, sqrt_s]
```
scarlehoff, line 68: You are missing plotting information for this one from `PLOTTINGTYPE_JET`
```
kinematics_override: jet_sqrt_scale
```

scarlehoff (commented):
Thank you very much! (and thank you as well for the very detailed report!). I've left some comments with some modifications. Some of them I've added locally to add the report with the plot (like the `kinematics_override` key in `CMS_1JET_8TEV` which might make the plot look differently to what you expected)

Please have a look at the plot and check whether everything looks as you expected:
https://vp.nnpdf.science/yWSiPUzmTfyZwHKQttJY-g==

(please add units for the different m12/pt)

As per the datasets themselves:
- `ATLAS_1JET_8TEV_R06`:
	- no variant: the t0 chi2 is different, which points to some difference in the treatment of MULT / ADD 
	- decorrelated: the covmats are quite different (shouldn't they be the same for `DEC`?) which obviously leads to a very different chi2
- `CMS_1JET_8TEV_PTY`: there is a difference of a factor of two between the computed chi2, the off-diagonal entries of the covmat is different. They are also different with the `bugged` version which I expected would reproduce the old dataset (or maybe I should not expect the old one to be reproducible at all?)

For these two the results are completely equivalent (including t0 and everything)
- `ATLAS_2JET_7TEV_R06_M12Y`: I needed to use the `bugged` variant (so I've added that to `dataset_names.yaml`
- `CMS_2JET_7TEV_M12Y`

comane:
> * `ATLAS_1JET_8TEV_R06`:
>   
>   * no variant: the t0 chi2 is different, which points to some difference in the treatment of MULT / ADD


This should have been fixed now.

>   * decorrelated: the covmats are quite different (shouldn't they be the same for `DEC`?) which obviously leads to a very different chi2

The covariance matrices are supposed to be quite different (see https://vp.nnpdf.science/5DpNBePRSne1TZ1Sfky6Kg==/#dataset-atlas_1jet_8tev_r06_dec-variant-decorrelated) and the chi2 too (old value ~ 0.7, new value = 2.6).

comane:
> * `CMS_1JET_8TEV_PTY`: there is a difference of a factor of two between the computed chi2, the off-diagonal entries of the covmat is different. They are also different with the `bugged` version which I expected would reproduce the old dataset (or maybe I should not expect the old one to be reproducible at all?)

This I find surprising and it is possibly a mistake. I had found a bug for this dataset that was changing the chi2, but not by a factor of 2. 
In the old report https://vp.nnpdf.science/5DpNBePRSne1TZ1Sfky6Kg==/#dataset-cms_1jet_8tev-variant-nominal

Did you compute the chi2 with the cuts applied?

PR #1888, opened 2023-12-08
Nn layers refactor

APJansen:
This PR does two things that both should leave everything identical:

### 1. Pull together the 3 functions that were responsible for generating the neural network layers:
   - `generate_dense_network`
   - `generate_dense_per_flavor_network`
   - `generate_nn`

Only the last one remains, the first two had a lot of overlap.
I have also pulled in the loop over replicas out of `pdf_NN_layer_generator` into `generate_nn`

This is everything up to and including[this commit](https://github.com/NNPDF/nnpdf/pull/1888/commits/cfc3b8d6f315811a5d37abf4e7403fc70157a6bb).
This PR may be easier to follow commit by commit.

### 2. Reverse the order of the loops over replicas and layers

This is the actual point, currently we do: for all replicas, for all layers, create the layer.
To accomodate the upcoming multi-replica layers, where one layer contains all replicas, the order needed to change.

## Status

I think there are 3 relevant choices to test for:
- dropout or not
- dense layers or dense_per_flavour layers
- 1 or multiple replicas

The dense_per_flavour layers aren't compatible with multiple replicas or with dropout (they could be, but in the first case it doesn't pass a check, and the second case just wasn't implemented), but the dropout and replicas choices should be independent, so we have:

default layer:
| replicas \ dropout | yes | no |
| --- | --- | --- |
| 1 | X | X |
| multiple | X | X |

- [x] dense_per_flavour, 1 replica, no dropout

For each of these I have taken a simple runcard, ran it for 100 epochs, and compared the last 3 digits of the chi2 between this branch and master (well, replica_axis_first, which is ready to be merged). (Xs mean they pass this test)

## TODO:
- [x] Fix dense_per_flavour case, getting a size mismatch after training, at single replica models
- [x] Either test and fix or explicitly exclude the combination of dense_per_flavour and dropout (as it was, but no longer is)
- [x] Update tests which still use the old `generate_dense_network` and `generate_dense_per_flavour_network`
- [x] Check if a third step is possible without changing numerics: directly concatenating each layer over replicas. Actually no this isn't possible until the `MultiDense` layer itself is implemented.

### DIFF
--- a/n3fit/src/n3fit/model_gen.py
+++ b/n3fit/src/n3fit/model_gen.py
@@ -299,74 +299,6 @@ def observable_generator(
     return layer_info
 
 
-# Network generation functions
-def generate_dense_network(
-    nodes_in: int,
-    nodes: int,
-    activations: List[str],
-    initializer_name: str = "glorot_normal",
-    seed: int = 0,
-    regularizer: str = None,
-):
-    """
-    Generates a dense network
-    """
-    layer_type = "dense"
-    custom_args = {"kernel_regularizer": regularizer}
-
-    def initializer_generator(initializer_name, seed):
-        return MetaLayer.select_initializer(initializer_name, seed=seed)
-
-    list_of_pdf_layers = []
-    for i, (nodes_out, activation) in enumerate(zip(nodes, activations)):
-        init = initializer_generator(initializer_name, seed + i)
-        layer = base_layer_selector(
-            layer_type,
-            kernel_initializer=init,
-            units=nodes_out,
-            activation=activation,
-            input_shape=(nodes_in,),
-            **custom_args,
-        )
-        list_of_pdf_layers.append(layer)
-        nodes_in = int(nodes_out)
-
-    return list_of_pdf_layers
-
-
-def generate_dense_per_flavour_network(
-    nodes_in, nodes, activations, initializer_name="glorot_normal", seed=0, basis_size=8
-):
-    """
-    For each flavour generates a dense network of the chosen size
-
-    """
-    layer_type = "dense_per_flavour"
-    custom_args = {"basis_size": basis_size}
-
-    def initializer_generator(initializer_name, seed):
-        initializers = [
-            MetaLayer.select_initializer(initializer_name, seed=seed + b) for b in range(basis_size)
-        ]
-        return initializers
-
-    list_of_pdf_layers = []
-    for i, (nodes_out, activation) in enumerate(zip(nodes, activations)):
-        init = initializer_generator(initializer_name, seed + i)
-        layer = base_layer_selector(
-            layer_type,
-            kernel_initializer=init,
-            units=nodes_out,
-            activation=activation,
-            input_shape=(nodes_in,),
-            **custom_args,
-        )
-        list_of_pdf_layers.append(layer)
-        nodes_in = int(nodes_out)
-
-    return list_of_pdf_layers
-
-
 def generate_pdf_model(
     nodes: List[int] = None,
     activations: List[str] = None,
@@ -655,7 +587,7 @@ def pdfNN_layer_generator(
 
     nn_replicas = generate_nn(
         layer_type=layer_type,
-        input_dimensions=nn_input_dimensions,
+        nodes_in=nn_input_dimensions,
         nodes=nodes,
         activations=activations,
         initializer_name=initializer_name,
@@ -748,7 +680,7 @@ def pdfNN_layer_generator(
 
 def generate_nn(
     layer_type: str,
-    input_dimensions: int,
+    nodes_in: int,
     nodes: List[int],
     activations: List[str],
     initializer_name: str,
@@ -762,8 +694,9 @@ def generate_nn(
     Create the part of the model that contains all of the actual neural network
     layers.
     """
-    x = Input(shape=(None, input_dimensions), batch_size=1, name='xgrids_processed')
+    x = Input(shape=(None, nodes_in), batch_size=1, name='xgrids_processed')
 
+    custom_args = {}
     if layer_type == "dense_per_flavour":
         # set the arguments that will define the layer
         # but careful, the last layer must be nodes = 1
@@ -771,24 +704,37 @@ def generate_nn(
         # is set to the number of nodes of the last layer when it should
         # come from the runcard
         nodes[-1] = 1
+        custom_args['basis_size'] = last_layer_nodes
 
-    constant_args = {
-        'nodes_in': input_dimensions,
-        'nodes': nodes,
-        'activations': activations,
-        'initializer_name': initializer_name,
-    }
-    if layer_type == "dense_per_flavour":
-        constant_args['basis_size'] = last_layer_nodes
-        layers_generator = generate_dense_per_flavour_network
-    if layer_type == "dense":
+        def initializer_generator(initializer_name, seed):
+            initializers = [
+                MetaLayer.select_initializer(initializer_name, seed=seed + b)
+                for b in range(basis_size)
+            ]
+            return initializers
+
+    elif layer_type == "dense":
         reg = regularizer_selector(regularizer, **regularizer_args)
-        constant_args['regularizer'] = reg
-        layers_generator = generate_dense_network
+        custom_args['regularizer'] = reg
+
+        def initializer_generator(initializer_name, seed):
+            return MetaLayer.select_initializer(initializer_name, seed=seed)
 
     models = []
     for i_replica, replica_seed in enumerate(replica_seeds):
-        list_of_pdf_layers = layers_generator(**constant_args, seed=replica_seed)
+        list_of_pdf_layers = []
+        for i_layer, (nodes_out, activation) in enumerate(zip(nodes, activations)):
+            init = initializer_generator(initializer_name, replica_seed + i_layer)
+            layer = base_layer_selector(
+                layer_type,
+                kernel_initializer=init,
+                units=nodes_out,
+                activation=activation,
+                input_shape=(nodes_in,),
+                **custom_args,
+            )
+            list_of_pdf_layers.append(layer)
+            nodes_in = int(nodes_out)
 
         # add dropout as second to last layer
         if dropout > 0:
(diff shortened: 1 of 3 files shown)


APJansen:
@scarlehoff @RoyStegeman Does this look good to you, the plan described above I mean?

And just in case it has changed since the last time I asked: do we still want to maintain the dense_per_flavour layer?

APJansen:
Ok no problem, working now!

The issue I had with the per_flavour layer came from this old TODO about the basis_size coming from the last entry of the nodes list, while it should come form the runcard. I was overwriting `nodes[-1]` with 1 in this case, but this changed the original which was reused for the single replica model.
Is it worth fixing this old TODO? There's a `basis` argument in the runcard, which is a list, so we can just take its length, but I'm not sure how to add dependence on runcard arguments.

Also, I don't think it makes sense to add the possibility of combining these layers with dropout, and indeed it wasn't possible before, so I just raised an error in that case.

n3fit/src/n3fit/model_gen.py
```diff
@@ -780,44 +680,123 @@ def compute_unnormalized_pdf(x, postfix=""):
...
-    common_args = {
-        'nodes_in': input_dimensions,
-        'nodes': nodes,
-        'activations': activations,
-        'initializer_name': initializer_name,
-        'seed': replica_seed,
-    }
-    if layer_type == "dense":
+    if dropout > 0 and layer_type == "dense_per_flavour":
+        raise ValueError("Dropout is not supported for dense_per_flavour layers")
```
RoyStegeman, line 726: Please add this check at the level of `n3fit_checks_action`

n3fit/src/n3fit/model_gen.py
```diff
@@ -780,44 +680,120 @@ def compute_unnormalized_pdf(x, postfix=""):
...
-        list_of_pdf_layers = generate_dense_network(
-            **common_args, dropout_rate=dropout, regularizer=reg
-        )
-    elif layer_type == "dense_per_flavour":
-        list_of_pdf_layers = generate_dense_per_flavour_network(
-            **common_args, basis_size=last_layer_nodes
-        )
+        custom_args['regularizer'] = reg
+
+        def initializer_generator(initializer_name, seed, i_layer):
```
scarlehoff, line 751:
```suggestion
        def initializer_generator(seed, i_layer):
```

Since the initializer name is fixed, it doesn't need to enter here.

n3fit/src/n3fit/model_gen.py
```diff
@@ -780,44 +680,120 @@ def compute_unnormalized_pdf(x, postfix=""):
...
+        # set the arguments that will define the layer
+        # but careful, the last layer must be nodes = 1
+        # TODO the mismatch is due to the fact that basis_size
+        # is set to the number of nodes of the last layer when it should
+        # come from the runcard
+        nodes_list[-1] = 1
+        basis_size = last_layer_nodes
+        custom_args['basis_size'] = basis_size
+
+        def initializer_generator(initializer_name, seed, i_layer):
```
scarlehoff, line 739:
```suggestion
        def initializer_generator(seed, i_layer):
```

n3fit/src/n3fit/model_gen.py
```diff
@@ -780,44 +680,120 @@ def compute_unnormalized_pdf(x, postfix=""):
...
+            base_layer_selector(
+                layer_type,
+                kernel_initializer=init,
+                units=nodes_out,
+                activation=activation,
+                input_shape=(nodes_in,),
+                **custom_args,
+            )
+            for init in inits
+        ]
```
scarlehoff, line 773:
```suggestion
        for replica_seed in replica_seeds:
            init = initializer_generator(replica_seed, i_layer)
            layers = base_layer_selector(
                    layer_type,
                    kernel_initializer=init,
                    units=nodes_out,
                    activation=activation,
                    input_shape=(nodes_in,),
                    **custom_args,
                )
```

I think like this it is better for readability (I think you don't need `inits` later right? otherwise of course no)
I've also removed the initializer name.

n3fit/src/n3fit/model_gen.py
```diff
@@ -780,44 +680,120 @@ def compute_unnormalized_pdf(x, postfix=""):
...
+    if dropout > 0:
+        dropout_layer = base_layer_selector("dropout", rate=dropout)
+        list_of_pdf_layers.insert(-2, dropout_layer)
+
+    # In case of per flavour network, concatenate at the last layer
+    if layer_type == "dense_per_flavour":
+        concat = base_layer_selector("concatenate")
+        list_of_pdf_layers[-1] = [lambda x: concat(layer(x)) for layer in list_of_pdf_layers[-1]]
+
+    # ... then apply them to the input to create the models
```
scarlehoff, line 787: The ... makes me think that this applies to the `dense_per_flavour` only 
```suggestion
    # Apply all layers to the input to create the models
```

n3fit/src/n3fit/model_gen.py
```diff
@@ -780,44 +680,120 @@ def compute_unnormalized_pdf(x, postfix=""):
...
+    xs = [layer(x) for layer in list_of_pdf_layers[0]]
+    for layers in list_of_pdf_layers[1:]:
+        if type(layers) is list:
```
scarlehoff, line 790: Please, add a comment here.

n3fit/src/n3fit/model_gen.py
```diff
@@ -780,44 +680,120 @@ def compute_unnormalized_pdf(x, postfix=""):
...
 
-    # Note: using a Sequential model would be more appropriate, but it would require
-    # creating a MetaSequential model.
-    x = Input(shape=(None, input_dimensions), batch_size=1, name='xgrids_processed')
-    pdf = x
-    for layer in list_of_pdf_layers:
-        pdf = layer(pdf)
+    models = [
+        MetaModel({'NN_input': x}, pdf, name=f"NN_{i_replica}") for i_replica, pdf in enumerate(xs)
+    ]
```
scarlehoff, line 797: Please change the name of `xs` here since you are doing a loop in `xs` which contains a call to `x` which is however not the content of `xs`!

I think it makes sense to change it to `pdfs` or `pdf_final_layers` or something like this.

scarlehoff (commented):
lgtm, they are all suggestions for style / loops / naming

n3fit/src/n3fit/model_gen.py
```diff
@@ -780,44 +680,120 @@ def compute_unnormalized_pdf(x, postfix=""):
...
-    elif layer_type == "dense_per_flavour":
-        list_of_pdf_layers = generate_dense_per_flavour_network(
-            **common_args, basis_size=last_layer_nodes
-        )
+        custom_args['regularizer'] = reg
+
+        def initializer_generator(initializer_name, seed, i_layer):
+            seed += i_layer
+            return MetaLayer.select_initializer(initializer_name, seed=seed)
+
```
Radonirinaunimi, line 754: Maybe add a `raise ValueError` here if `layer_type` is not either `dense` or `dense_per_flavour`. This would at least remove the warning that `initializer_generator` below is possible unbound.

Ideally, this would have been checked in [n3fit.checks](initializer_generator) (?).

```suggestion
    else:
        raise ValueError(f"{layer_type} not allowed. Can only be 'dense' or 'dense_per_flavour'")
```

Radonirinaunimi (commented):
Looks very good!

n3fit/src/n3fit/model_gen.py
APJansen: Ah yes I agree that was confusing.

n3fit/src/n3fit/model_gen.py
APJansen: done!

n3fit/src/n3fit/model_gen.py
APJansen: You mean on why the if statement is needed? I added a comment, it's because dropout is shared between layers. I could also remove the if statement and replace `dropout_layer` with `[dropout_layer for _ in range(num_replicas)]` or something.

n3fit/src/n3fit/model_gen.py
APJansen: Agreed, I think I was trying to anticipate how it will look with multi dense layers, but it doesn't matter.

n3fit/src/n3fit/model_gen.py
APJansen: You mean a warning you get in your IDE, from a linter or something? 
I can add a check in checks, but I think that won't remove this?

Radonirinaunimi: That will remove the warning (but this is a very minor part of it). The main reason it to have it crash with informative error message if it is not one of the two.

scarlehoff: I would vote for a `check_layer_accepted_type` or something in `n3fit/checks.py` rather than having the check here, otherwise it will fail very late (after having loaded all fktables!)

APJansen: Ok, done!

APJansen: I called it from `wrapper_check_NN` , which is already in `n3fit_checks_provider.py`, and it seems to work fine for me, is that not enough?

scarlehoff: Oh, missed that line! Yes, that's enough.

I'll have a second look later today if I can and merge it. Thanks!

Radonirinaunimi: > I called it from `wrapper_check_NN` , which is already in `n3fit_checks_provider.py`, and it seems to work fine for me, is that not enough?

Yes, this would be the ideal solution indeed.

APJansen: Ok great, thanks :)

scarlehoff (approved):
lgtm

RoyStegeman (approved):

Radonirinaunimi (approved):

scarlehoff:
Please don't merge this yet!

APJansen:
Thanks, I won't merge yet! When is the next tag expected?

RoyStegeman:
Hopefully once #1901 is merged

scarlehoff:
> Hopefully once https://github.com/NNPDF/nnpdf/pull/1901 is merged

And when the papers are out...

APJansen:
> And when the papers are out...

The ... seems to indicate that that will take a while? ;P

Maybe it's worth creating a general waiting-for-next-tag branch or something, so that we can keep master as is but also not block further development?

PR #1890, opened 2023-12-11
Enable conda package for 3.11

scarlehoff:
In https://github.com/NNPDF/nnpdf/pull/1871 I've been able to test with 3.11 (although I had to change some test thresholds).

This PR tries to prepare the package for python 3.11. If this works well (even lowering the threshold) I think it would be a good idea to merge before the final tag so that we have 3.9 to 3.11 with 4.0.7

Note that for 4.0.8 if everything goes according to plan the conda package for nnpdf will not depend on any version of python since nnpdf will be python-only.

Edit:
Turns out the problem with conda was a pinned libarchive many moons ago (which we pinned because of a problem with conda!)

During the testing I've also removed *eigen* since the tensorflow package in anaconda has not been updated in a while. I've tested in my computer (with linux and python 3.9) and there is no memory leak anymore it seems?

However, I think I'm going to have a slightly different conda-recipe for 3.9/3.10 and 3.11 to avoid changing things too much for the next tag.

### DIFF
--- a/conda-recipe/conda_build_config.yaml
+++ b/conda-recipe/conda_build_config.yaml
@@ -1,7 +1,7 @@
 #For some reason the resolver decides to use an old version of numpy
 #without this
 numpy:
-    - 1.21
+    - 1.24
 
 pin_run_as_build:
     lhapdf: x.x.x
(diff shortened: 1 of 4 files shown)


RoyStegeman (approved):
This is fine to merge, thanks!

conda-recipe/conda_build_config.yaml
```diff
@@ -1,7 +1,7 @@
 #For some reason the resolver decides to use an old version of numpy
 #without this
 numpy:
-    - 1.21
+    - 1.24
```
RoyStegeman, line 4: Should we check if this is still true?

PR #1866, opened 2023-11-26
Re-implementation of some of the ATLAS Collider DY Datasets

Radonirinaunimi:
The following PR implements some of the ATLAS collider DY in the new commondata format. The status is summarized in the table below.

☑️  in **Comparison vs. Old** means that the results are fully identical while 🔴 means that comparisons are available but noticeable differences are perceived.

| Dataset Name  | Comparison vs. Old |  General Comments | Status  |
|---|---|---|---|
| ATLAS_DY_7TEV_EMUON_Y  | ☑️  | The old implementation used a luminosity uncertainty of 3.5% while in HepData it is 3.4%  |    ☑️   |
| ATLAS_DY_7TEV_DILEPTON_Y-CRAP  | ☑️   | The new implementation (from HepData) is missing one source of uncorrelated systematics  |    ☑️   |
| ATLAS_DY_7TEV_DILEPTON_Y-FRAP  | ☑️  | The new implementation (from HepData) is missing one source of uncorrelated systematics  |    ☑️   |
| ATLAS_WPWM_8TEV_MUON_Y |   | FK tables are missing but old commondata exists  |  ☑️  |
| ATLAS_Z0_8TEV_LOWMASS_2D |  ☑️    |   |   ☑️   |
| ATLAS_Z0_8TEV_HIGHMASS_2D | 🔴  |  Slight differences in treatment of asymmetric systematic correlated uncertainties  |   ☑️   |
| ATLAS_Z0_8TEV_3D_CRAP |   | Could not find dataset and FK tables to compare the implementation to  | ☑️  |
| ATLAS_Z0_8TEV_3D_FRAP |   | Could not find dataset and FK tables to compare the implementation to  | ☑️  |
| ATLAS_DY_13TEV_FID |☑️   | Just needs to fix the plotting to be as a function of Gauge bosons  |  ☑️ |
| ATLAS_Z0_8TEV_20FB_PT-INVDIST | |  |   ☑️  |
| ATLAS_Z0_8TEV_20FB_PT-RAPDIST | |  |   ☑️  |

Remains TODO:
- [ ] Define the plotting entries to be exactly the same as before

Issue #1846, opened 2023-11-15
Complete porting old Collider DY into the new format


Radonirinaunimi:
To complete the porting of the old collider DY data into the new commondata format, the following datasets need to be converted:

| Dataset Name  | Paper  | Link to Old filter  |
|---|---|---|
| CDF Z differential    |  [paper](http://arxiv.org/abs/0908.3914)   |  [filter](https://github.com/NNPDF/nnpdf/blob/master/buildmaster/filters/CDF.cc)  |
|  D0 Z differential   |  [paper](http://arxiv.org/abs/hep-ex/0702025)   |  [filter](https://github.com/NNPDF/nnpdf/blob/master/buildmaster/filters/D0.cc)   |
|   D0 W electron asymmetry  |  [paper](http://arxiv.org/abs/1309.2591)   |  [filter](https://github.com/NNPDF/nnpdf/blob/master/buildmaster/filters/D0.cc)   |
|  D0 W muon asymmetry  |  [paper](http://arxiv.org/abs/1412.2862)   |  [filter](https://github.com/NNPDF/nnpdf/blob/master/buildmaster/filters/D0.cc)   |
| ~ATLAS low-mass DY 7 TeV~  | [paper](http://arxiv.org/abs/1404.1212)  | [filter](https://github.com/NNPDF/nnpdf/blob/master/buildmaster/filters/ATLASLOMASSDY11.cc)  |
| ATLAS high-mass DY 7 TeV  | [paper](http://arxiv.org/abs/1305.4192)  | [filter](https://github.com/NNPDF/nnpdf/blob/master/buildmaster/filters/ATLAS.cc)  |
| ATLAS W, Z 7TeV $(\mathcal{L}=35~\rm{pb}^{−1})$  | [paper](http://arxiv.org/abs/1109.5141)  |  [filter](https://github.com/NNPDF/nnpdf/blob/master/buildmaster/filters/ATLAS.cc) |
| ATLAS W, Z 7 TeV $(\mathcal{L}=4.6~\rm{pb}^{−1})$  | [paper](http://arxiv.org/abs/1612.03016)  | [filter](https://github.com/NNPDF/nnpdf/blob/master/buildmaster/filters/ATLASWZRAP11.cc)  |
|  CMS W electron asymmetry 7 TeV | [paper](http://arxiv.org/abs/1206.2598)  |  [filter](https://github.com/NNPDF/nnpdf/blob/master/buildmaster/filters/CMS.cc) |
[...]

PR #1893, opened 2023-12-12
Polarized proton-PDF fits

Radonirinaunimi:
The following PR is meant to implement the required changes to allow fitting polarized proton PDFs.

### DIFF
--- a/n3fit/src/n3fit/layers/msr_normalization.py
+++ b/n3fit/src/n3fit/layers/msr_normalization.py
@@ -21,28 +21,24 @@ IDX = {
     'v15': 6,
     'v24': 7,
     'v35': 8,
+    't3': 9,
+    't8': 10,
+    't15': 11,
+    't24': 12,
+    't35': 13,
 }
 MSR_COMPONENTS = ['g']
 MSR_DENOMINATORS = {'g': 'g'}
 # The VSR normalization factor of component f is given by
 # VSR_CONSTANTS[f] / VSR_DENOMINATORS[f]
 VSR_COMPONENTS = ['v', 'v35', 'v24', 'v3', 'v8', 'v15']
-VSR_CONSTANTS = {
-    'v': 3.0,
-    'v35': 3.0,
-    'v24': 3.0,
-    'v3': 1.0,
-    'v8': 3.0,
-    'v15': 3.0,
-}
-VSR_DENOMINATORS = {
-    'v': 'v',
-    'v35': 'v',
-    'v24': 'v',
-    'v3': 'v3',
-    'v8': 'v8',
-    'v15': 'v15',
-}
+VSR_CONSTANTS = {'v': 3.0, 'v35': 3.0, 'v24': 3.0, 'v3': 1.0, 'v8': 3.0, 'v15': 3.0}
+VSR_DENOMINATORS = {'v': 'v', 'v35': 'v', 'v24': 'v', 'v3': 'v3', 'v8': 'v8', 'v15': 'v15'}
+
+# The following lays out the SR for Polarised PDFs
+TSR_COMPONENTS = ['t3', 't8']
+TSR_DENOMINATORS = {'t3': 't3', 't8': 't8'}
+TSR_CONSTANTS = {'t3': 1.2701, 't8': 0.5850}  # +/- 0.0025  # +/- 0.0250
 
 
 class MSR_Normalization(MetaLayer):
@@ -52,6 +48,7 @@ class MSR_Normalization(MetaLayer):
 
     _msr_enabled = False
     _vsr_enabled = False
+    _tsr_enabled = False
 
     def __init__(self, mode: str = "ALL", replicas: int = 1, **kwargs):
         if mode == True or mode.upper() == "ALL":
@@ -61,6 +58,8 @@ class MSR_Normalization(MetaLayer):
             self._msr_enabled = True
         elif mode.upper() == "VSR":
             self._vsr_enabled = True
+        elif mode.upper() == "TSR":
+            self._tsr_enabled = True
         else:
             raise ValueError(f"Mode {mode} not accepted for sum rules")
 
@@ -75,6 +74,13 @@ class MSR_Normalization(MetaLayer):
             self.vsr_factors = op.numpy_to_tensor(
                 [np.repeat(VSR_CONSTANTS[c], replicas) for c in VSR_COMPONENTS]
             )
+        if self._tsr_enabled:
+            self.divisor_indices += [IDX[TSR_DENOMINATORS[c]] for c in TSR_COMPONENTS]
+            indices += [IDX[c] for c in TSR_COMPONENTS]
+            self.tsr_factors = op.numpy_to_tensor(
+                [np.repeat(TSR_CONSTANTS[c], replicas) for c in TSR_COMPONENTS]
+            )
+
         # Need this extra dimension for the scatter_to_one operation
         self.indices = [[i] for i in indices]
 
@@ -113,6 +119,8 @@ class MSR_Normalization(MetaLayer):
             ]
         if self._vsr_enabled:
             numerators += [self.vsr_factors]
+        if self._tsr_enabled:
+            numerators += [self.tsr_factors]
 
         numerators = op.concatenate(numerators, axis=0)
         divisors = op.gather(y, self.divisor_indices, axis=0)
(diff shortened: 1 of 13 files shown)


giacomomagni:
Just to remind ourself that once #1896 is merged, we need to adjust also `evolven3fit_new`

Radonirinaunimi:
> Just to remind ourself that once #1896 is merged, we need to adjust also `evolven3fit_new`

In this branch, `evolven3fit_new` has already been adjusted. I will now reflect the merged change from #1896 to here.

scarlehoff:
Take into account that the reader for the new commondata will keep being rebased on top of master until the new commondata is finished (and that will include also changes to evolven3fit).

I think it would be a good idea to pause the polarized fit and finish the new commondata first...

n3fit/src/n3fit/layers/msr_normalization.py
```diff
@@ -61,6 +62,8 @@ def __init__(self, mode: str = "ALL", replicas: int = 1, **kwargs):
             self._msr_enabled = True
         elif mode.upper() == "VSR":
             self._vsr_enabled = True
+        elif mode.upper() == "TSR":
+            self._tsr_enabled = True
```
scarlehoff, line 66: I think mode.upper() == "ALL" should also activate _tsr for consistency.

Otherwise adding an ALL_polarized might make sense.

n3fit/src/n3fit/performfit.py
```diff
@@ -9,11 +9,74 @@
...
+            list of xgrid values to compute predictions from
+
+        """
+        # `res` is of shape (n_replicas, n_fl=14, n_x, n_q2=1)
+        res = evolution.grid_values(pdfset, pids, xgrid, [q0value])
+        # TODO: for the time being select the Central replicas,
+        # but modify the following to also returns the STD
+        return np.squeeze(res[0].swapaxes(0, -1))  # Shape=(nx, n_fl=14)
+
+    return compute_asx
```
scarlehoff, line 77: Please, use directly validphys for this. You can get results from validphys in whatever basis you want.

Radonirinaunimi:
> Take into account that the reader for the new commondata will keep being rebased on top of master until the new commondata is finished (and that will include also changes to evolven3fit).
> 
> I think it would be a good idea to pause the polarized fit and finish the new commondata first...

On this side, there isn't really anything I could do unless we resolve this [comment](https://github.com/NNPDF/nnpdf/pull/1866#issuecomment-1846843682) from #1866. Although I have already implemented the CMS datasets listed in #1846, I haven't pushed them yet since most of them also contain the differences/issues above.

n3fit/src/n3fit/layers/msr_normalization.py
Radonirinaunimi: Actually, the other sum rules except for $T_3$ and $T_8$ must not be enforced in the polarised fits as they are no longer applicable. I've removed the other sum rules when running `vp-comparefits --use_polarized` as that was very confusing.

n3fit/src/n3fit/performfit.py
Radonirinaunimi: Yes, that'd be indeed the plan as per the todo. Once we are confident that this is the conceptual approach to enforce positivity I'll do so.

n3fit/src/n3fit/layers/msr_normalization.py
scarlehoff: ah! Ok, then that's what the bug :P

None of them are?

validphys2/src/validphys/pdfbases.py
```diff
@@ -621,6 +623,17 @@ def f_(transform_func):
 LUX_FLAVOUR = copy.deepcopy(FLAVOUR)
 LUX_FLAVOUR.default_elements=('u', 'ubar', 'd', 'dbar', 's', 'sbar', 'c', 'cbar', 'g', 'photon')
 
+EVOL_POL = LinearBasis.from_mapping({
+    r'\Sigma'  : {'u': 1, 'ubar': 1, 'd': 1, 'dbar': 1, 's': 1, 'sbar': 1, 'c': 1, 'cbar': 1 ,'b':1, 'bbar': 1, 't': 1, 'tbar': 1},
```
giacomomagni, line 627: I have a question, for my uderstanding:
I'm not 100% sure I understand this. At the fitting scale singlet should contain only `u,d,s`, is this what is happening here?

validphys2/src/validphys/pdfbases.py
```diff
@@ -803,6 +816,18 @@ def fitbasis_to_NN31IC(flav_info, fitbasis):
...
 
+
+    elif fitbasis == 'EVOL_POL':  # With Perturbative Charm
+        sng = {'sng': 1, 'v': 0, 'v3': 0, 'v8': 0, 't3': 0, 't8': 0, 't15': 0, 'g': 0 }
+        v = {'sng': 0, 'v': 1, 'v3': 0, 'v8': 0, 't3': 0, 't8': 0, 't15': 0, 'g': 0 }
+        v3 = {'sng': 0, 'v': 0, 'v3': 1, 'v8': 0, 't3': 0, 't8': 0, 't15': 0, 'g': 0 }
+        v8 = {'sng': 0, 'v': 0, 'v3': 0, 'v8': 1, 't3': 0, 't8': 0, 't15': 0, 'g': 0 }
+        t3 = {'sng': 0, 'v': 0, 'v3': 0, 'v8': 0, 't3': 1, 't8': 0, 't15': 0, 'g': 0 }
+        t8 = {'sng': 0, 'v': 0, 'v3': 0, 'v8': 0, 't3': 0, 't8': 1, 't15': 0, 'g': 0 }
+        cp = {'sng': 0, 'v': 0, 'v3': 0, 'v8': 0, 't3': 0, 't8': 0, 't15': 0, 'g': 0 }
```
giacomomagni, line 827: Also here I'm confused... why we return all the elements in evolution basis, but `cp` ?

validphys2/src/validphys/pdfbases.py
```diff
@@ -803,6 +816,18 @@ def fitbasis_to_NN31IC(flav_info, fitbasis):
         g = {'u': 0, 'ubar': 0, 'd': 0, 'dbar': 0, 's': 0, 'sbar': 0, 'c': 0, 'cbar': 0, 'g': 1 }
         v15 = {'u': 1, 'ubar': -1, 'd': 1, 'dbar': -1, 's': 1, 'sbar': -1, 'c': -3, 'cbar': 3, 'g': 0 }
 
+
+    elif fitbasis == 'EVOL_POL':  # With Perturbative Charm
+        sng = {'sng': 1, 'v': 0, 'v3': 0, 'v8': 0, 't3': 0, 't8': 0, 't15': 0, 'g': 0 }
+        v = {'sng': 0, 'v': 1, 'v3': 0, 'v8': 0, 't3': 0, 't8': 0, 't15': 0, 'g': 0 }
```
giacomomagni, line 822: Shouldn't we remove `v,v3,v8` at all for the time being?

n3fit/src/n3fit/layers/msr_normalization.py
Radonirinaunimi: Unfortunately, I haven't found "the" bugs yet (if they are present) :/

But currently, I have added a check here:
https://github.com/NNPDF/nnpdf/blob/7ee6897d2aec206a69d86fc6f152abd2b26d1f3c/n3fit/src/n3fit/checks.py#L71-L80
that for polarised fits, the `sum_rules` must always be (and ony) `TSR`.

validphys2/src/validphys/pdfbases.py
Radonirinaunimi: These are needed to rotate the outputs of the NN which are always expressed in `NN31IC`.

validphys2/src/validphys/pdfbases.py
Radonirinaunimi: Same as above, these are needed for the rotation into `NN3IC`. This is perfectly fine since all the non-active flavors will always be zero from the fit anyway.

validphys2/src/validphys/pdfbases.py
Radonirinaunimi: Ah, let me double check the appropriate expressions here.

Radonirinaunimi: This part is just used to rotate a given PDF in the flavor basis into `EVOL_POL`. This is not used in the fit but rather in the report to show the result in the `EVOL_POL` basis, so if $t,b$ are zero from the fits then they will not contribute into $\Sigma$.

PR #1895, opened 2023-12-14
ZPT 13 TeV dataset

comane:
This PR implements tabulated figure 6, uncertainty is taken form tables 5 and 7: 

https://www.hepdata.net/record/ins1768911


buildmaster/ATLAS_Z_13TEV_PT/metadata.yaml
```diff
@@ -0,0 +1,107 @@
+# Created on Oct 2023, author: Mark N. Costantini
+
+ # Generalia
```
scarlehoff, line 3:
```suggestion
```

buildmaster/ATLAS_Z_13TEV_PT/metadata.yaml
```diff
@@ -0,0 +1,107 @@
...
+    data_uncertainties:
+      - uncertainties_E.yaml
+    
+
+    # Theory
+    theory:
+      FK_tables:
+        - - ATLAS_Z0J_13TEV_PTE
+        - - ATLAS_Z0_13TEV_PTE_TOT
+      operation: 'null'
```
scarlehoff, line 65:
```suggestion
      operation: 'ratio'
```

and also below

scarlehoff (commented):
This needs to be moved to the folder within validphys (and probably the same thing with respect to the process_type applies here)

(also the folder name should be Z0J)

validphys2/src/validphys/datafiles/new_commondata/ATLAS_Z0J_13TEV_PT/metadata.yaml
```diff
@@ -0,0 +1,60 @@
...
+    data_uncertainties:
+      - uncertainties.yaml
+    
+
+    # Theory
+    theory:
+      FK_tables:
+        - - ATLAS_Z0J_13TEV_PTLL
+        - - ATLAS_Z0_13TEV_PTLL_TOT
+      operation: 'null'
```
scarlehoff, line 60:
```suggestion
      operation: 'RATIO'
```

scarlehoff (commented):
image

image

I've removed from this comparison the first few points (they are very badly described but that's expected) and the last few points (the experimental uncertainties are huge so it was impossible to see anything in the plot)

Scale variations might have a big impact here, this is only the central value, but wanted to check how the plot looked. I'll revise the paper also to make sure I'm applying the right cuts and kinematic variables since I ran this with the previous implementation of the data, so I was looking at the wrong figures in the paper.

(it should take much less to recompute now since I have already warmup grids and those should not change much regardless of the cuts)

scarlehoff:
I'm looking at this again now, I have some questions and comments:

1. This is the right record and version, right? https://www.hepdata.net/record/92377?version=3
2. I'm going to run ZpT above 30 GeV to avoid the problems at low pT
3. I've noticed there's a factor of ¿bin size? between your data and Table 5 (the born level) but it is not _exactly_ the bin size, why is that?

scarlehoff:
Since I've only computed the predictions from pT > 30 GeV onwards, it is necessary to change the metadata to:

```
    theory:
      shifts:
        ATLAS_Z0J_13TEV_PT_LL.NNLO.pT: 9
      FK_tables:
        - - ATLAS_Z0J_13TEV_PT_LL.NLO.pT
        - - ATLAS_Z0J_13TEV_TOT.NLO.cross
      operation: 'ratio'
```

and add the following filter rules to `cuts/filters.yaml`

```
- dataset: ATLAS_Z0J_13TEV_PT_LL
  reason: Avoid the region where resummation effects become important.
  rule: "pT >= 30"
- dataset: ATLAS_Z0J_13TEV_PT_LL
  reason: Avoid the large pT region where the statistics are not very good.
  rule: "pT <= 600"
```

(the choice of pt < 600 can be modified a bit)

comane:
> Since I've only computed the predictions from pT > 30 GeV onwards, it is necessary to change the metadata to:
> 
> ```yaml
>     theory:
>       shifts:
>         ATLAS_Z0J_13TEV_PT_LL.NNLO.pT: 9
>       FK_tables:
>         - - ATLAS_Z0J_13TEV_PT_LL.NLO.pT
>         - - ATLAS_Z0J_13TEV_TOT.NLO.cross
>       operation: 'ratio'
> ```

Hi @scarlehoff I am a bit confused, why are the FK_tables .NLO here?

comane:
> I'm looking at this again now, I have some questions and comments:
> 
> 1. This is the right record and version, right? https://www.hepdata.net/record/92377?version=3
> 2. I'm going to run ZpT above 30 GeV to avoid the problems at low pT
> 3. I've noticed there's a factor of ¿bin size? between your data and Table 5 (the born level) but it is not _exactly_ the bin size, why is that?
> 4. There's one bin missing, not very important since it is the last one and the statistics are terrible, but is that intentional?

Yes I think that the version is correct. Note that central values are taken from tabulated figure 6 and 5 is used for the uncertainties as was discussed in meetings.
This should also answer 3. and 4.

scarlehoff:
> Hi @scarlehoff I am a bit confused, why are the FK_tables .NLO here?

It's a name I was using for testing, sorry. The actual name is not important, but the `shitfs` are `operation` are.

> Note that central values are taken from tabulated figure 6 

Then the table entry in the metadata should be changed too (it should be 26 if it is this one: https://doi.org/10.17182/hepdata.92377.v3/t26)

Re 3. do you understand this factor? I'm just wondering what is this.

comane:
> Re 3. do you understand this factor? I'm just wondering what is this.

You mean between the central values of Table 5 and the central values of Tabulated 6?

I don't know why tbh, but this is the reason of why we switched to Tabulated 6 instead of Table 5 which was originally implemented

scarlehoff (approved):

comane:
Hi @scarlehoff, thanks for reviewing.
I only added a comment to the metadata to specify that central values and uncertainties are not taken from the same tables.

Will merge once the test finish

PR #1896, opened 2023-12-15
Add th 800 to theory db

giacomomagni:
This PR is to add th 800 for polarized evoution. 

```
kbthr: 4.0
IC: 0
Q0: 1.0
```


RoyStegeman (approved):
Why is kbthr still set to 4? 

Would you mind moving 800 and 390 such that the theories are in order of theoryid?

giacomomagni:
> Why is kbthr still set to 4?
> 
Because we are using aFONLL for the time being. 
But I actually made a mistake, as here `kbthr` should not be changed as the evolution eko will be incorrect, will fix it.

> Would you mind moving 800 and 390 such that the theories are in order of theoryid?

I can do it, do you have a script to do that by chance?

RoyStegeman:
> I can do it, do you have a script to do that by chance?

No I would export it as a csv, change that, and then load it again. 

Ah you were of course just using a GUI to add the new theory but that doesn't allow you to change the order. We should really get rid of this database... though while it's still there I do think we should try to keep it organised.

scarlehoff:
The only reason I haven't gotten rid of this database is that the cpp code (like evolven3fit) needs it.

As soon as the python installation is finished and evolven3fit replaced, the theorydb is going next. It will probably arrive to 2024 alive but I can promise you it doesn't get to Februrary.

giacomomagni:
Okay this is the snippet I used.
Could you please check I've not messed up anything?

```
import sqlite3
import pandas as pd

conn = sqlite3.connect("theory.db")
cur = conn.cursor()
data = cur.execute("SELECT * FROM TheoryIndex")

columns = conn.execute("PRAGMA table_info('TheoryIndex')")

headers = [v for (_, v, _, _, _, _) in columns]

lines = [th for th in data]
df = pd.DataFrame(lines, columns=headers)

# sort
df.sort_values('ID', inplace=True)

conn2 = sqlite3.connect("sorted_theory.db")

#push the dataframe to sql 
df.to_sql("TheoryIndex", conn2, if_exists="replace", index=False)
```

RoyStegeman (approved):
The script looks good. Thanks!

giacomomagni:
Okay thanks, I'll merge it tomorrow.

PR #1898, opened 2023-12-18
Fix photon replicas generation + make vp-nextfitruncard use Q0 instead of Qmin for iterating preprocessing

niclaurenti:
Opening this PR to fix a bug in the photon sector: in the case of single replica fits, after that the `Photon` object was computed in

https://github.com/NNPDF/nnpdf/blob/72008b25884f65848fe9741b2842e2bc927638fc/n3fit/src/n3fit/model_trainer.py#L825

it was recomputed in

https://github.com/NNPDF/nnpdf/blob/72008b25884f65848fe9741b2842e2bc927638fc/n3fit/src/n3fit/model_gen.py#L445

with `replicas=[1]`.
Since  `replicas` is a list containing the replicas IDs of the different replicas, it was computing the replica n. 1 for all the replicas (sorry for repeating replicas 5 times in a single sentence).
Moreover, the `Photon` object already contains just one photon in the single replica fits, so there's no need to recompute it imposing just 1 replica.


niclaurenti:
@APJansen I didn't understand why the photon was recomputed in 

https://github.com/NNPDF/nnpdf/blob/72008b25884f65848fe9741b2842e2bc927638fc/n3fit/src/n3fit/model_gen.py#L445

but probably there was a reason.
Given what I wrote in the first comment of this PR, is it going to work anyway with the changes I did?

scarlehoff:
Probably there was some miscommunication between "the photon is not run in parallel but with single replicas" and "it is run with just replica 1".

With this changes is it working as expected? It looks reasonable.

Regarding the check, it should already be done in `checks.py`

RoyStegeman:
I agree that recomputing is not necessary, but I believe this was just done to create a placeholder model for when the single model that contains multiple replicas is finished training and further downstream we just treat it as a bunch of individual models. For this there need to be placeholder models for the information from the single model to be copied to, if that makes sense. I.e. I believe this photon would be overwritten, is that not correct?

scarlehoff:
But since the photon is just one, you can use the same model which will be a single-replica one. This might fail with more replicas but it will error out before.

niclaurenti:
> Regarding the check, it should already be done in `checks.py`

Ok then I'll remove it

niclaurenti:
@scarlehoff @RoyStegeman I have run a fit after the bugfix: it fixes the problem of the photon band but it doesn't fix the variation of the effective exponents.

Before the bugfix:
https://vp.nnpdf.science/ebvyFn_iRpOQhXa_hOacRA==/exponents_report_report.html#tables

After the bugfix:
https://vp.nnpdf.science/0RutKm1UQkewJ0SQGQrSmg==/exponents_report_report.html#tables

RoyStegeman:
> Ok then I'll remove it

Can add it as a decorator to `performfit` instead?

n3fit/src/n3fit/checks.py
```diff
@@ -408,3 +408,11 @@ def check_fiatlux_pdfs_id(replicas, fiatlux, replica_path):
             raise CheckError(
                 f"Cannot generate a photon replica with id larger than the number of replicas of the PDFs set {luxset.name}:\nreplica id={max_id}, replicas of {luxset.name} = {pdfs_ids}"
             )
+
+@make_argcheck
+def check_multireplica_qed(replicas, fiatlux):
+    if fiatlux is not None:
+        if len(replicas) > 1:
+            raise CheckError(
+                "At the moment, running a multireplica QED fits is deprecated."
```
RoyStegeman, line 417: It's not deprecated, it's not possible at all

n3fit/src/n3fit/checks.py
```diff
@@ -408,3 +408,11 @@ def check_fiatlux_pdfs_id(replicas, fiatlux, replica_path):
...
+            )
```
RoyStegeman, line 418:
```suggestion
            )

```

n3fit/src/n3fit/checks.py
niclaurenti: Well, actually it is possible. The problem is that all the photons will be computed in series so it will take around 30 min * 100 replicas = 50 hours

scarlehoff:
```suggestion
                "At the moment, running a multireplica QED fits is not allowed."
```

This is correct in any case.

RoyStegeman:
https://github.com/NNPDF/nnpdf/blob/72008b25884f65848fe9741b2842e2bc927638fc/n3fit/src/n3fit/backends/keras_backend/MetaModel.py#L403-L407

Apologies for not being very involved here. Did you see this line? In principle I think this can be avoided for the photon since it's not fitted but precomputed so it's known before the fit, and indeed we don't want to compute a photon simply to have it as a placeholder only to discard it later (which you already addressed)

scarlehoff:
Indeed, in that line `"add_photons"` is never in the list `self.layers` and with Niccolo's change it is not necessary.

Of course, the right fix is for the single_replica_generator to take the index of the replica and with that take only the necessary photon.

Let me remove this part of the code and add a comment around the photon. I don't think we want to try to fix it now (with the testing associated to it).

scarlehoff:
I made the commit in this branch since @niclaurenti is using it to run the fits and it should be done later today. If these work we can merge this :D

RoyStegeman (approved):

PR #1902, opened 2023-12-21
Trvl mask layers diag

goord:
Fixed diagonal covmat case with replica mask layers


goord:
100-replica GPU test shows no statistical difference with trvl_mask-layers branch: [](https://vp.nnpdf.science/7aUQ7_9fTzy9_wDHwf8_DQ==)

PR #1904, opened 2024-01-03
modifications to remove the charm valence sum rule

enocera:
This branch contains modifications to remove the charm valence sum rule in ccbar asymm fits. Needed to reply to the the referee. A report comparing fits w/ and w/ the charm valence sum rule is [here](https://vp.nnpdf.science/3xe03GKkQ_uIvPXD7qOW_A==/).
This PR is not intended to be merged, but only to allow other people to check that what I did is sensible.


tgiani:
@enocera the code looks good to me, but looking at the report I don t understand why in the fit without sumrules for V15 all the valence sumrules seems to be less precise

scarlehoff:
As said by email, I think this is correct, but by lifting the (c+cb)=0 condition you also lift (s+sb)=0 which explains the report.

tgiani:
ah right I see, you re right. Thank you!

enocera:
I close this PR, as it will be superseded by https://github.com/NNPDF/nnpdf/tree/no_csr

PR #1905, opened 2024-01-09
Multi dense layer

APJansen:
Still WIP, but it will be very simple once #1888 and #1881 are merged.

Just creating draft PR now to see the CI, as locally I have some issues with seeds in tensorflow again.


APJansen:
After rebasing this is still failing a single test (`test_fit` on the qed runcard), on a single entry, on a single python vesion:
```
 x: array([1.319493, 1.199529, 1.054267, 5.157313, 1.068909])
 y: array([1.31949 , 1.199594, 1.054267, 5.075661, 1.068905])
```

(also the new `test the python installation` tests, what are these for? From their name it seems like they shouldn't be testing the code)

scarlehoff:
It tests the almost pure python (as opposed of conda) installation, which means testing whether the code installed work. Seeing just whether it installs is not very useful. We need to test more, not less!

In any case, try resubmitting the test. In any case, I'm working on updating the tests to make them more robust but also much more restrictive: https://github.com/NNPDF/nnpdf/pull/1924
in case you want to wait for that

APJansen:
I restarted them but I'm pretty sure it's not going to change anything, as this has been the case for several runs already. I'm not sure what to make of it, I thought it's just a numerical difference that happens to be over the threshold in some versions and under in others, and that updating the regression values will fix it. But it's kinda weird that the other components differ by ~10^-5 and this one by 10^-1..

APJansen:
Still failing some regression tests unfortunately, with a huge difference in stop epoch on the normal fit for some reason.. 
About the others, the mismatch is only in a very small percentage, but I also see a lot of zeroes in the outputs, so I'm not sure how significant it actually is.

scarlehoff:
The 3 first and two last columns (photon, anti-top, anti-bottom - bottom, top) should be 0. That also means that approx only 9/14 of the data contain actual information. Still, only 7 numbers changed so it doesn't look worrying. Probably connected to the changes  you see in the fitbot results.
All of these might be due to the change of shape because maybe enough operations have changed order during the optimization.

For the normal fit the change is much bigger. As it should, as this tests checks the initialization and you said that it was supposed to change:

> Note that many initializers' statistics depend on the shape of the input, so just using a single initializer out of the box not only will give different results because it is seeded differently, it will actually be statistically different.

If you are happy with these changes https://github.com/NNPDF/nnpdf/blob/b06492e9f7f927fd8a24c7bf4cfa1d9eaee390be/extra_tests/regression_checks.py#L44 (change regenerate to True and run it once in your computer)

----------

In order to be happy with this changes, please do a complete fit trying to reproduce `231213-02-rs-nnlo-baseline-updated` so that we can be sure that while the initialization has changed the final fits are the same (note that the fitbot doesn't test this since it also starts with fixed weights).

This is a good time to change the initialization random numbers since we are at the start of the 4.0.8 tag, but we need to do it carefully!

APJansen:
> For the normal fit the change is much bigger. As it should, as this tests checks the initialization and you said that it was supposed to change:

No actually initialization should be identical, thats what [MultiInitializer](https://github.com/NNPDF/nnpdf/blob/3873cd15fe55130b09c4193291c8b825d5bc4199/n3fit/src/n3fit/backends/keras_backend/multi_dense.py#L148) is supposed to take care of.

I just updated my comment to say that locally the difference is only 9 epochs, so I don't understand why it's so different in the CI.

scarlehoff:
Did you check that the initialization is the same between master and this branch for that runcard and replica?

scarlehoff:
It could also be that the numerical problems that grow enough to be seen in the fitbot (or the fit without replicas) grow even more in the fit without starting weights. 

But, tbh, if the shape for one single replica doesn't change why should any numerical problems arise?

APJansen:
Yes just checked, weights are identical after initialization (both the NN and the prefactors).

The shape does change, there's an extra replica axis, so I do expect small numerical differences, but not that big..

The good news is that this other discrepancy in the regular tests has disappeared.

About reproducing `231213-02-rs-nnlo-baseline-updated`, how do I do that? Which runcard?

Initialization of the prefactors would change in #1923 if I get that to work btw.

scarlehoff:
> The shape does change, there's an extra replica axis, so I do expect small numerical differences, but not that big..

But only with an extra (1,) everywhere, right?
Why do you need this extra `(1,)` in the bias though?

> About reproducing 231213-02-rs-nnlo-baseline-updated, how do I do that? Which runcard?

The runcard you can get from that fit `vp-get fit 231213-02-rs-nnlo-baseline-updated` , but if the initialization is not supposed to change it should not be necessary (it will be in #1923 I guess)

APJansen:
> but if the initialization is not supposed to change it should not be necessary (it will be in https://github.com/NNPDF/nnpdf/pull/1923 I guess)

Ok great :) If these new regression tests all pass would you consider it sufficiently tested, or is there another comparison I should do?

> But only with an extra (1,) everywhere, right?
> Why do you need this extra (1,) in the bias though?

Yes that's right, the bias shape in general will be (R, 1, F) for R replicas and a layer with F output features, the reason is that the output shape before adding the bias is (R, N, F), so without this extra 1 it wouldn't broadcast correctly.

scarlehoff:
> Ok great :) If these new regression tests all pass would you consider it sufficiently tested, or is there another comparison I should do?

I mean, the more tests you do the better. The important thing is to make sure we understand where the differences are coming from.

btw, it would be good to also have a parallel replicas regression test, if you can add it in #1923 it would be great.

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
```diff
@@ -0,0 +1,179 @@
...
+        applying each replica separately and concatenating along the last axis.
+        If the input already contains multiple replica outputs, it is equivalent
+        to applying each replica to its corresponding input.
+        """
+        if inputs.dtype.base_dtype != self._compute_dtype_object.base_dtype:
+            inputs = tf.cast(inputs, dtype=self._compute_dtype_object)
+
+        input_axes = 'brnf' if self.replica_input else 'bnf'
+        einrule = input_axes + ',rfg->brng'
+        outputs = tf.einsum(einrule, inputs, self.kernel)
```
scarlehoff, line 115: replica_input is fixed upon initialization? Because if so it would be better to move some of this logic to the initialization part.

Also, I'm wondering whether the numerics are coming from einsum, which is different from tensordot (and the order of operations is most likely to be different).

Since the axis that would need to be passed to tensordot are known as soon as `replica_input` is known, it could be a good idea to have some logic outside that will generate the `ein_rule` and the tensordot axes rules so that it is easy to change between one and the other for

1) testing
2) benchmark (in the past `einsum` was much faster in gpu and much slower in cpu... and we don't want that)

APJansen:
I see that at the first epoch, results are identical. At the second the validation chi2's start to differ, at about 10^-8 relatively, and at the second the training chi2 similarly, and at say epoch 20 the relative difference is still around 10^-7.

So all looks good to me locally, I think the only thing is this big difference in the CI for the normal fit.

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
APJansen: No I mean just matching my local master to this branch, didn't get that either.

I'm not sure it's possible to do the general case with a tensordot though, how do you implement `"brnf,rfg->brng"`?

scarlehoff: I'm thinking about the single replica case which is the one that needs to match.

So, you don't get the same values even if you squeeze out the extra (1,)?
We keep having these small differences and it would be good to understand where are they coming from.

Maybe is the sum rules, or maybe we need to move to float64 at least for part of the calculation.

APJansen: Ah so a flag which switches the behavior only in the single replica case? Sure, I can try a bit more to get it exactly the same. I think the differences are again just different tensorflow operations, I don't know exactly what the dense layer uses, can try to match that.

scarlehoff: > Ah so a flag which switches the behavior only in the single replica case?

Yes. Didn't you have that already at some point? (I remember having this conversation about eigen and tensordot already)

> I don't know exactly what the dense layer

I think it uses tensordot. But a way of isolating the issue might be to generate a dense layer and use it in that call function.

APJansen: Hmm actually there is no difference in the outputs of `MultiDense` for one replica compared to `Dense`, as I verified locally in this test, which gives exact zeroes, both after the last commit removing einsum for 1 replica, but also before.

<details>
<summary>Script</summary>

```python
import numpy as np

from tensorflow.keras.layers import Dense
from tensorflow.keras.initializers import RandomNormal
from n3fit.backends.keras_backend.multi_dense import MultiDense

replica_input = False  # check both cases

np.random.seed(1)
if replica_input:
    test_input = 10**3 * np.random.rand(1, 1, 7, 1)
else:
    test_input = 10**3 * np.random.rand(1, 7, 1)

d_ini = RandomNormal(stddev=0.1, seed=42)
d = Dense(11, kernel_initializer=d_ini)

md_ini = RandomNormal(stddev=0.1, seed=0)
md = MultiDense(replica_seeds=[42], units=11, replica_input=replica_input, kernel_initializer=md_ini)

if replica_input:
    print(d(test_input).shape, md(test_input).shape)
    print(d(test_input) - md(test_input))
else:
    print(d(test_input).shape, md(test_input)[:, 0].shape)
    print(d(test_input) - md(test_input)[:, 0])
```
</details>

APJansen: I see now that the last commit did make the difference much smaller in the CI, on the differing stop epoch in the first test, which first was about a factor 2 and now it's only ~20 epochs, about a percent. 

I did another test expanding on the previous, where I turn both layers into models and train them for a bit, and still they give identical predictions.

I don't think it's worth chasing these differences, it can be a huge rabbit hole, and we have been happy to ignore them as long as they are small several times already.

scarlehoff: Yes. But we have also introduced some bugs because we were not rigorous enough. Now the papers are published (and the people involved in the work need to write their respective thesis) so I'd rather be extra careful. 

If we understand that the difference is coming from dense vs multidense that's fine, we regenerate the regression tests and we continue from there... but right now we still don't know what's introducing this difference. 

Also, understanding what kind of operations screw up precision might be useful as well. We are working under the assumption that float32 is enough for the precision we claim in the PDFs, but if a change in the order of operations actually has an effect beyond 1% (which is the relative precision we are testing against) we will need to move to float64.

APJansen: I have done some more testing, printing outputs (min, max and mean) after each layer in the PDF. I see that the results are identical in the first forward pass. Then the first backward pass happens, I don't see inside that. Then, the first validation happens, where outputs start to differ, but only for the third (out of 3) dense layer, not for the first two, and not for the prefactors (but of course also for everything that depends on the third layer).

I've also repeated this with 4 rather than 3 layers, expecting to find either that it was still only the last layer, or it was layer 3 and 4 that had changed outputs. But actually now it's all layers. (all dense layers and everything that depends on them, still not the preprocessing layers ofcourse).

I've also repeated the toy example, see script below, now turning it into a model of 5 layers, all dense or all multidense, and training for 100 epochs, and I still find 0 difference.

So it's probably something in the backward pass, something to do with the multi dense layer, but that only creates a difference in the presence some of the later layers... 

If there's any other way to validate this PR, that's going to be much easier than to trace this back to exactly what's going on, I don't think I'll be able to do that.
I'm running the runcard of `231213-02-rs-nnlo-baseline-updated` now, with 120 replicas.

<details>
<summary>script</summary>

```python
import numpy as np
import tensorflow as tf

from tensorflow.keras.layers import Dense, Input, Lambda
from tensorflow.keras.initializers import RandomNormal
from n3fit.backends.keras_backend.multi_dense import MultiDense
from tensorflow.keras.models import Model

np.random.seed(1)
input_layer = Input(shape=(7, 1))
test_input = 10**3 * np.random.rand(1, 7, 1)
test_output = 10**3 * np.random.rand(1, 7, 11)

# create model with dense layers
ds = []
for i in range(5):
    d_ini = RandomNormal(stddev=0.1, seed=42+i)
    d = Dense(11, kernel_initializer=d_ini)
    ds.append(d)

ds_output = input_layer
for d in ds:
    ds_output = d(ds_output)

ds_output = Lambda(lambda nns: tf.stack(nns, axis=1), name=f"stack_replicas")([ds_output])


model_d = Model(inputs=input_layer, outputs=ds_output)
model_d.compile(optimizer='adam', loss='mse')
model_d.summary()

# create model with multi-dense layers (one replica)
mds = []
for i in range(5):
    md_ini = RandomNormal(stddev=0.1, seed=0)
    md = MultiDense(replica_seeds=[42+i], units=11, replica_input=(i != 0), kernel_initializer=md_ini)
    mds.append(md)

mds_output = input_layer
for md in mds:
    mds_output = md(mds_output)

model_md = Model(inputs=input_layer, outputs=mds_output)
model_md.compile(optimizer='adam', loss='mse')
model_md.summary()

# fit both
model_d.fit(test_input, test_output, epochs=100)
model_md.fit(test_input, test_output, epochs=100)

# compare
print(model_d(test_input) - model_md(test_input))
print(np.mean(model_d(test_input)))
print(np.mean(model_md(test_input)))

```

</summary>

scarlehoff: > If there's any other way to validate this PR

It's not so much validating this particular PR but rather understanding where the problem was coming from (which you just did). I want to be sure that we are not in a situation in 2/3 months from now where one of the results of the papers needs to be reproduced and it doesn't work and we don't understand why.


Could you do the exact same test but using double precision instead of single precision? That would be invaluable in this scenario because if it is a problem with numerical precision that would be a good motive to move to double for at least production runs (and the tests).
We don't need that kind of precision for things like hyperopt, but a change of a % because of an unknown change in the order of operation in the "black box" of the code is a bit worrying. If we can avoid it just by having a flag to run in double it would be great.

Thanks. And apologies for being very strict, but for now I think we need to be extremely careful ^^U

APJansen: Is there an existing flag for this? I found that for Keras one can set `tf.keras.backend.set_floatx('float64')`, which will default the weights to float64. But this creates all sorts of issues with the inputs which are all still float32.

APJansen: I added the flag in #1929, it should be working but there is an unrelated problem with the CI that I tried to fix in #1930, but apparently that wasn't it.

scarlehoff: Did the numerical issue go away when using `double`?

APJansen: I just checked, they became a lot less. A difference in the 10th digit in the min/max, 5th digit in the mean, of the last dense layer. Much less in earlier layers, and only in the 14th digit in the final PDF itself. (In the first validation step)
The exact same test when setting the flag to false (i.e. not using double precision) gives differences in the third digit of the min/max and second in the mean.

scarlehoff: Ok, then we have been too optimistic until know with doing everything in single precision.
Thanks for this. 

Is it fine if I push directly to the branch where you add the double precision flag? I will update all regression tests to use it.
I'm wondering whether we should make it default...

scarlehoff: (or, if you have a machine with linux in which you can re-generate the regression test you can also do it yourself ofc)

APJansen: > Is it fine if I push directly to the branch where you add the double precision flag? I will update all regression tests to use it.

Sure, no problem, I'll leave the branch alone. I don't have access to a linux machine.

> I'm wondering whether we should make it default...

I still think it's not really a problem though. I didn't mention, the difference in the final PDF, without double precision, is only in the 7th digit. No final results should be sensitive to that (in a statistically significant way), the learning process itself is much noisier than that.
Of course for reproducibility in the tests I agree it is nice to use.

scarlehoff: > I didn't mention, the difference in the final PDF, without double precision, is only in the 7th digit.

Ah, ok. That's very good to hear.
Then let's leave it just for tests for the time being.

APJansen: > I didn't mention, the difference in the final PDF, without double precision, is only in the 7th digit.

Just to be clear because I realized it might be interpreted differently, by final I meant after all the layers, so the normalized PDF, not the final epoch or something, still talking about the first validation step

scarlehoff: Ah, I see.

Well, you said you are running 231213-02-rs-nnlo-baseline-updated, that will give us the final PDF.

APJansen: Yes, something went wrong so running again, just to be sure this should be ok right?
```bash
vp-setupfit runcard-repro.yml
n3fit runcard-repro.yml 1 -r 120
evolven3fit runcard-repro 120
postfit 1 runcard-repro --at-least-nrep
```
where I took runcard-repro.yml from the fit you mention (and adapted it to run in parallel, have to set same_trvl_per_replica to true though).

scarlehoff: No, don´t run like this `n3fit runcard-repro.yml 1 -r 120`
Run it one replica at a time (we want to compare with the fit that @RoyStegeman did)

Also, for postfit you would want 100 replicas.

APJansen: Hmm but that wouldnt test the multi replica implementation. Should we perhaps get #1788 merged first and then do the above test with same_trvl_per_replica True?

APJansen: Alright, running them. Should I use `evolven3fit_new`? I thought that had become the regular `evolven3fit` but the command is not recognised. Perhaps I need to reinstall again?

scarlehoff: I left it as `evolven3fit_new` https://github.com/NNPDF/nnpdf/blob/e4a19925e06c93002956bac498338a1123ef2964/pyproject.toml#L43C1-L43C16 since I assumed people would have it already as part of their scripts with that name.

APJansen: Here is the [comparison](https://vp.nnpdf.science/FGwCRitnQhmqBYBjeQWS7Q==/). (Perhaps some parts are missing, I wasn't sure if the comparison really completed or quit early, if so let me know and I'll rerun it)

APJansen: @scarlehoff We just had a look at the report during the meeting, and it looked good to us. If you agree and merge #1929 I can rebase and regenerate the regression tests.

scarlehoff: Yes, please!

If you cannot regenerate the tests I can do so.

scarlehoff: Well, actually, go on without regenerating the tests.
Better to finish the PR as if they were working and we update them only at the end!

n3fit/src/n3fit/checks.py
```diff
@@ -385,8 +385,8 @@ def check_consistent_parallel(parameters, parallel_models, same_trvl_per_replica
             "Replicas cannot be run in parallel with different training/validation "
             " masks, please set `same_trvl_per_replica` to True in the runcard"
         )
-    if parameters.get("layer_type") != "dense":
-        raise CheckError("Parallelization has only been tested with layer_type=='dense'")
+    if parameters.get("layer_type") == "dense_per_flavour":
+        raise CheckError("Parallelization has not been tested with layer_type=='dense_per_flavour'")
```
RoyStegeman, line 389: Why this change? I think the previous one was more correct, though given there is another check that raises an error for anything but `dense` and `dense_per_flavour` it's effectively the same thing.

n3fit/src/n3fit/backends/keras_backend/base_layers.py
```diff
@@ -17,34 +17,44 @@
...
+from tensorflow import expand_dims, math, nn
+from tensorflow.keras.layers import (  # pylint: disable=unused-import
+    Dropout,
+    Input,
+    Lambda,
+    concatenate,
+)
 from tensorflow.keras.layers import Dense as KerasDense
-from tensorflow import expand_dims
+from tensorflow.keras.layers import LSTM, Concatenate  # pylint: disable=unused-import
```
RoyStegeman, line 28: `from tensorflow.keras.layers import` is done twice

n3fit/src/n3fit/tests/test_multidense.py
```diff
@@ -0,0 +1,68 @@
+import numpy as np
+import tensorflow as tf
+from tensorflow.keras import Sequential
+from tensorflow.keras.initializers import GlorotUniform
+from tensorflow.keras.layers import Dense
+
+from n3fit.backends.keras_backend.multi_dense import MultiDense
+from n3fit.model_gen import generate_nn
```
RoyStegeman, line 8:
```suggestion
```

RoyStegeman (commented):
Looks good. I believe @scarlehoff is also happy with the numerical stability?

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
```diff
@@ -0,0 +1,191 @@
...
+            if self.replica_input:
+                self.matmul = lambda inputs: tf.tensordot(inputs, self.kernel[0], [[-1], [0]])
+            else:
+                # Manually add replica dimension
+                self.matmul = lambda inputs: tf.expand_dims(
+                    tf.tensordot(inputs, self.kernel[0], [[-1], [0]]), axis=1
+                )
+        else:
+            einrule = f"b{'r' if self.replica_input else ''}nf,rfg->brng"
+            self.matmul = lambda inputs: tf.einsum(einrule, inputs, self.kernel)
```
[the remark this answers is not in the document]
scarlehoff: Mostly efficiency.

RE the other point, I'd be happy to swap if someone spends the time of reproducing the fits and checking that they are the exactly same. But it needs to be done for all of them.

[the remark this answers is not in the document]
RoyStegeman: > also we can swap by the end of the year when the theses being written with respect to these 3 papers are finished

This was indeed what I was thinking, if we can remove it from about a year from now we can add a note to say that, by then we can presumably be a bit less strict about the numerical precision with which the current results are reproduced. Unless for efficiency we want to keep it like this anyway.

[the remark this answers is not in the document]
scarlehoff: If efficiency is still an issue I'd keep it, but last time I profiled the code was many versions of tensorflow ago.

But indeed, I think in a year we can relax a bit the numerical stability of the results.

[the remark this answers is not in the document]
RoyStegeman: So we just need a comment linking to this conversation to provide context for the exception of the 1 replica case ;)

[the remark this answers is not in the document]
scarlehoff: or... a benchmark

tbh I'm not sure why is it slower, but it is probably also the backpropagation given that's what changes...

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
```diff
@@ -0,0 +1,191 @@
...
+        self.built = True
+
+        if self.replicas == 1:
+            if self.replica_input:
+                self.matmul = lambda inputs: tf.tensordot(inputs, self.kernel[0], [[-1], [0]])
+            else:
+                # Manually add replica dimension
+                self.matmul = lambda inputs: tf.expand_dims(
+                    tf.tensordot(inputs, self.kernel[0], [[-1], [0]]), axis=1
+                )
```
scarlehoff, line 110:
```suggestion
            # Using tensordot here for numerical stability with 4.0 fits
            # TODO: benchmark against the replica-agnostic einsum below and make that default
            if self.replica_input:
                axis = 1
            else:
                axis = 0
            self.matmul = lambda inputs: tf.tensordot(inputs, self.kernel[0], [[-1], [0]], axis=axis)
```
Remind me, why can `self.replica_input` be `True` or `False`?

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
```diff
@@ -0,0 +1,191 @@
...
+        else:
+            einrule = f"b{'r' if self.replica_input else ''}nf,rfg->brng"
```
scarlehoff, line 112:
```suggestion
            einrule = "brnf,rfg->brng"  if self.replica_input else "bnf,rfg->brng"
```

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
RoyStegeman: > why can self.replica_input be True or False? 

First layers shares the same inputs for all replicas, in which case it's `False`. Only the later layers have an additional index for the replicas

scarlehoff: But then in the 1 replica case it doesn't matter.

RoyStegeman: I think it's just for convenience of the output shape

n3fit/src/n3fit/backends/keras_backend/base_layers.py
APJansen: I agree this is a bit of a mess, I was surprised isort didn't pick it up, but in fact it's worse, isort caused this ^^
I think it wants to keep them alphabetical, but also not edit the pylint comment. In fact only `Input` requires it.

Of course it could be fixed by just having that comment on all of them, or removing it and using it somewhere just for that purpose, but I don't think that's an improvement.

Also I was confused by the `concatenate` (lower case `c`) import, I don't think that's used at all. There's `op.concatenate`, `np.concatenate` (using `tf.concat`), and there's the `Concatenate` layer which is accessed through `base_layer_selector` with the string `"concatenate"`, but this import is used nowhere.
So I removed it, here and in `backends/__init__.py`.

Do you agree on both points?

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
```diff
@@ -0,0 +1,191 @@
...
+    >>>     for r in range(replicas)
+    >>>     ]
+    >>> gridsize, features = 100, 2
+    >>> multi_dense_model.build(input_shape=(None, gridsize, features))
+    >>> for single_model in single_models:
+    >>>     single_model.build(input_shape=(None, gridsize, features))
+    >>> test_input = tf.random.uniform(shape=(1, gridsize, features))
+    >>> multi_dense_output = multi_dense_model(test_input)
+    >>> single_dense_output = tf.stack([single_model(test_input) for single_model in single_models], axis=1)
+    >>> tf.reduce_all(tf.equal(multi_dense_output, single_dense_output))
```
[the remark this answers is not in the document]
APJansen: Fair enough, I'll just remove it then.

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
APJansen: > Also, remind me, why can self.replica_input be True or False? (and will that happen in the 1 replica case?)

As Roy said, it's because the input to the first layer is just the x's, which doesn't have a replica dimension yet. (That's independent of the number of replicas being 1 or more, in both cases there is no extra dimension to the first layer's input, but there is to all following layers).

I agree it's not very well named, perhaps I can replace it with just `is_first_layer`?

> Can't you do something like a reshape with the first dimension being (1,) which would work in both cases?

It would be something like this:
```python
            output_shape = input_shape[:-1] + (self.units,)
            if self.replica_input:  # Have to manually add the replica dimension
                output_shape = (1,) + output_shape

            self.matmul = lambda inputs: tf.reshape(
                    tf.tensordot(inputs, self.kernel[0], [[-1], [0]]), output_shape
            )
```
avoiding the if statement, but this doesn't work as the batch dimension is `None` when this is called. It can be set manually of course, but then I think it starts to get less clear than what we started with.

n3fit/src/n3fit/backends/keras_backend/base_layers.py
scarlehoff: It might have been used in the past and forgotten there.

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
scarlehoff: I'm quite confused, why in this case you don't need axis?

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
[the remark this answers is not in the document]
APJansen: I thought it was a little slower, but a quick test on my laptop (replacing `if self.replicas == 1` with `False`), on the production runcard I get 0.44s for the tensordot and 0.42s for the einsum (per epoch, average over the second 100 epochs). Anyway since we have to keep it for now, I'll put in the comment, and it'll have to be benchmarked a bit more when it could be removed.

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
APJansen: I'm quite confused by your question, what are you referring to exactly? ;P

Maybe this answers your question: note the `self.kernel[0]`, that removes the replica axis from the kernel. (In a failed attempt to reproduce the numerics exactly). I'll see what happens if I remove that, but I don't think the axis will end up in the right place.

scarlehoff: The original code was:

```python
           if self.replica_input:
                self.matmul = lambda inputs: tf.tensordot(inputs, self.kernel[0], [[-1], [0]])
            else:
                # Manually add replica dimension
                self.matmul = lambda inputs: tf.expand_dims(
                    tf.tensordot(inputs, self.kernel[0], [[-1], [0]]), axis=1
                )
```

The tensordot call in your new example is the same but now you don't seem to need the axis argument while before you did.

APJansen: No the `axis=` is an argument to `tf.expand_dims`, which in the (broken) code above is replaced by `tf.reshape` with a predetermined shape.

scarlehoff: ah!!! Then we were talking cross-purposes, I missed, the `)` 

That makes it much easier. Let me change my suggestion.

```suggestion
            self.matmul =  lambda inputs: tf.tensordot(inputs, self.kernel[0], [[-1], [0]])
            if not self.replica_input:
                # Manually add replica dimension
                self.matmul = lambda x: tf.expand_dims(self.matmul(x), axis=1))
```
or some flavour thereof.

My main concern was to make clear that it was the same operation happening in both cases, Having an extra `expan_dims` afterwards I'm less worried about. I thought the `tensordot` was different in the two cases, sorry.

APJansen: Yeah the way black formats this is not the clearest.

Ah yes that looks good, just had to modify it slightly to prevent it from going into an infinite recursion.

APJansen: I've rewritten `replica_input` to (`not`) `is_first_layer`, I think it's much clearer like that.

APJansen:
Just so it doesn't get lost in a now resolved long conversation, here is the [comparison](https://vp.nnpdf.science/FGwCRitnQhmqBYBjeQWS7Q==/) I ran.

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
```diff
@@ -0,0 +1,163 @@
...
+from tensorflow.keras.initializers import Initializer
+from tensorflow.keras.layers import Dense
+
+
+class MultiDense(Dense):
+    """
+    Dense layer for multiple replicas at the same time.
+
+    For the first layer in the network, the input shape is (batch_size, gridsize, features),
+    still without a replica axis.
```
scarlehoff, line 13:
```suggestion
    For architectures in which the input is shared across replicas, it is useful to have an input shape
    wihtout a replica axis, e.g., `(batch_size, gridsize, features)`. The `is_first_layer` argument
    can be used so that a replica axis is created upon output.
```

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
```diff
@@ -0,0 +1,163 @@
...
+        self.replicas = len(replica_seeds)
+        self.replica_seeds = replica_seeds
+        self.kernel_initializer = MultiInitializer(
+            single_initializer=kernel_initializer, replica_seeds=replica_seeds
+        )
+        self.bias_initializer = MultiInitializer(
+            single_initializer=self.bias_initializer, replica_seeds=replica_seeds
+        )
+        self.is_first_layer = is_first_layer
+
```
scarlehoff, line 48:
```suggestion

        # Definition of the convolution between the input of the layer and the kernel parameters
        # it is defined during the build stage.
```

n3fit/src/n3fit/model_gen.py
```diff
@@ -750,17 +750,28 @@ def initializer_generator(seed, i_layer):
...
     list_of_pdf_layers = []
     for i_layer, (nodes_out, activation) in enumerate(zip(nodes_list, activations)):
-        layers = [
-            base_layer_selector(
+        if layer_type == "dense":
+            layers = base_layer_selector(
                 layer_type,
-                kernel_initializer=initializer_generator(replica_seed, i_layer),
+                replica_seeds=replica_seeds,
+                kernel_initializer=initializer_generator(0, i_layer),
```
scarlehoff, line 757: One question, since the seeds come in `replica_seeds` and the `initializer_generator` function is built specifically for this branch... wouldn't it be better to remove the extra input?

scarlehoff (commented):
Thanks for this. I like the `is_first_layer` argument much more!

Just a few small comments.
For the rest, please regenerate the regression test using a linux computer (and probably a conda installation I guess) so that they pass.

other than that I think I'm happy with it!

n3fit/src/n3fit/model_gen.py
APJansen: I'm not sure what precisely you propose? Note that we need 3 pieces of information: `replica_seeds`, `i_layer` as this is added to the replica seeds, and the initializer class. Currently the last two are combined by setting `i_layer` as the "single_initializer"'s seed and extracting it to add to the replica seeds in `MultiInitializer`. How did you want to rewrite this?

scarlehoff: Sorry, I meant that I think `initializer_generator(0, i_layer)` could be defined simply as `initializer_generator(i_layer)` given that the function is defined before specifically for this branch of the conditional I believe.

APJansen: Thanks, makes sense. In fact looking at that section of code again it was unnecessarily complicated, I've simplified it a lot (and tested that results are still identical with 1 or 2 replicas, or 1 with the per flavour layer).

(I had to do the comparisons on snellius, as I'm having some bizarre issue after updating my mac, where two subsequent identical runs give different results. I'll open an issue about it.)

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
APJansen: What was your intention for this change? To me it makes it less clear, and downplays the impact of this argument, as it actually changes the operation the layer does (from applying all weights to the input, to applying slice i of the weights along the replica axis to slice i of the inputs along the replica axis), it's only this trivial in the single replica case. I updated the documentation to make this a bit more clear, let me know what you think.

n3fit/src/n3fit/model_gen.py
```diff
@@ -728,50 +727,48 @@ def generate_nn(
...
-            return MetaLayer.select_initializer(initializer_name, seed=seed)
-
-    # First create all the layers...
-    # list_of_pdf_layers[d][r] is the layer at depth d for replica r
-    list_of_pdf_layers = []
-    for i_layer, (nodes_out, activation) in enumerate(zip(nodes_list, activations)):
-        if layer_type == "dense":
-            layers = base_layer_selector(
+        def layer_generator(i_layer, nodes_out, activation):
+            """Generate the ``i_layer``-th layer for all replicas."""
```
scarlehoff, line 756: I guess just to be a pain, but since this now really only works for the MultiDense, please do `elif "dense"` so that it is obvious that this it can only enter here in the multidense case.

And maybe in the docstring of this internal function or as a comment, add something like "Generation of layer with MetaLayer.MultiDense".

This is just to avoid a future in which someone adds a new layer type and thinks this should work out of the box. The fact that the initializer don't get the seeds directly but that they go to the layer is confusing if you don't know what layer is being used here.

scarlehoff (approved):
Just added a comment to further clarify what happens inside the layer selection (I agree that it is much clearer now nonetheless).

RE the problem with mac... anything that has to do with regression is basically broken, that's why at some point we just decided to skip (some? all?) regression tests in mac.

n3fit/src/n3fit/model_gen.py
APJansen: But `if A: ... elif B: ...` will generate complaints from eg pylint right? That's why I just had the `else: # dense`. I can add an else that throws an exception, but I thought the policy was to have these kinds of checks all at the start through reportengine?

scarlehoff: Yes, the checks should be done at the beginning at it is indeed done.

But that doesn't mean you cannot throw an exception if that helps clarify the code. Here in particular someone adding a new layer type might remove the check from `checks.py` and not even read this part of the code thinking that it would work. Adding a `raise ValueError(f"{layer_type=} not recognized during model generation")` is a perfectly valid option.

scarlehoff: adding, just to clarify where I'm coming from, most of this comments on clarifications are because I'm thinking of the situation in which a student arrives in three years fresh from their undergraduate, with no knowledge of n3fit or probably even python and is asked to check what would happen if one uses a fancy new network architecture and check what happens.

This is a situation I obviously didn't have in mind when I first wrote n3fit (I guess I was the one ho was asked to do something and my n3fit was validphys :P) but now I'd rather have redundant explanations, checks, etc when possible rather than following a more pure approach.

APJansen: Ok, done!

APJansen: Yes it makes sense, no problem!

APJansen:
Yes, trying to right now. Having some weird issue on snellius where when I run the regression checks it can't find reportengine, even though it's installed and I can run fits just fine.

I thought it'd be very convenient if this can be fully automated like the fitbot, and made an attempt at doing so in #1944.

APJansen:
> Yes. Also the weights for the fitbot I think need to be updated.

What do you mean here? If you just mean the structure of the `developing_weights.h5`, I changed those already.

scarlehoff:
Maybe it didn't survive the rebase: https://github.com/NNPDF/nnpdf/actions/runs/7928211177/job/21646032244#step:11:448

APJansen:
Ah ok, I'll redo it

APJansen:
Was confused for a while, this always gets me: there's the group names, but theres also the attribute layer_names which has to be changed independently. The `-a` flag in `h5tree -va developing_weights.h5` shows these as well, didn't know about that, very useful :)

RoyStegeman (approved):
Please update the reference fit for the fitbot report with NNBOT-1c97e2a73-2024-02-16

scarlehoff:
tbh, having regression tests for "exactly the same fit" we might want to keep the fitbot fixed between tags (unless big changes happen) as to show the cumulative change...

We should discuss this point in AMS

RoyStegeman:
Putting it that way there are two use-cases for the fitbot:
1. checking the cumulative change between tags
2. checking the impact of a single PR (which in some cases we want to be 0 or otherwise extremely small, not just statistically equivalent)

Given that to assess point one we will most likely end up looking at a global fit anyway while using the latest published version as a reference, I think the fitbot provides more value for point 2. 

We could of course also ask the bot for two reports (assuming that doesn't break the github action time constraints)... 

We can indeed discuss in AMS and leave it for now

scarlehoff:
> We could of course also ask the bot for two reports (assuming that doesn't break the github action time constraints).

tbh, this is key. There was a time when we were hitting the time constrain (which is 6 hours I think?) and the bot is the maximum we could get away under that time. Now it takes about one hour so we can safely add more reports.

Radonirinaunimi (approved):
I don't have further comments in addition to those that have been raised above. It looks great to me!

APJansen:
Great, thanks for the reviews everyone, so we leave the fitbot as is for now and I can merge this?

Btw, about the redo-regressions workflow, it's not working perfectly, in that it creates a new commit but doesn't trigger the other tests again, so if it's the final commit the PR cannot be merged. (Here I put the label on while the tests were just starting, and they do continue but you have to go to the previous commit to see them). Not sure what the best solution is, but the simplest which I did here is just to make another (trivial) commit.

scarlehoff:
I think it is fine to force merge the PR provided the previous checks all passed other than the regression. By the time the regression label is used the PRs should be well tested.

RoyStegeman:
The bot probably doesn't have the right privileges, similar to people who are not members of the NNPDF github organisation. A solution would probably be to create a token with those privileges and allow the github action to use that when pushing.

scarlehoff:
No, I think github actions cannot trigger more actions by design.

RoyStegeman:
Yes because it doesn't have the permissions. I think something like this [fine-grained-personal-access-token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#fine-grained-personal-access-tokens) might solve it.

scarlehoff:
I think we can simply add a `on_workflow_call` option and do it manually the same that the label is added manually...

> Yes because it doesn't have the permissions. 

Maybe it has changed or I'm misremembering, but I think at some point it was simply not possible because it could easily cause an infinite recursion.

RoyStegeman:
I see, perhaps you're right, I didn't look that far into it.

PR #1906, opened 2024-01-09
Fix alphaem running

niclaurenti:
Implement APFEL running of alphaem in photon module: below tau mass alpha runs with nl=2, below muon mass nl=1 ecc.
Moreover below a given (arbitrary) threshold nf is set to zero.


scarlehoff:
You need to wait for conda to pick up the new release of eko. We are still using conda.

niclaurenti:
> You need to wait for conda to pick up the new release of eko. We are still using conda.

Yes I know, I just pushed it and when it will be updated in conda-forge I will rerun the tests

validphys2/src/validphys/photon/alpha.py
```diff
@@ -0,0 +1,241 @@
...
+            nl = 3
+        return nf, nl
+
+
+def rge(_t, alpha, beta_qed_vec):
+    """RGEs for the running of alphaem"""
+    rge_qed = -(alpha**2) * (
+        beta_qed_vec[0] + np.sum([alpha ** (k + 1) * b for k, b in enumerate(beta_qed_vec[1:])])
+    )
+    return rge_qed
```
scarlehoff, line 241: Since this function is only used in this module and only as functional argument to `solve_inp` maybe it would be better to change it to `_rge` and instead of use `alpha` as the input (which could be confused with the class `Alpha` also defined in this module) use `alpha_ref` so that it is clear that it is just a value.

validphys2/src/validphys/photon/alpha.py
```diff
@@ -0,0 +1,241 @@
...
+        alpha_em: float
+            target value of a
+        """
+        u = 2 * np.log(q / qref)
+
+        # solve RGE
+        res = solve_ivp(
+            rge, (0, u), (alphaem_ref,), args=[self.betas_qed[(nf, nl)]], method="Radau", rtol=1e-6
+        )
+        return res.y[0][-1]
```
scarlehoff, line 144: As far as I can see this is the only place where rge is used I think.
Please add a comment before the return on why you select [0][-1] so that it is clear for future generations :P

validphys2/src/validphys/photon/alpha.py
```diff
@@ -0,0 +1,241 @@
...
+        ----------
+        q : float
+            target scale
+        alph_aem_ref : float
+            reference value of alpha_em
+        qref: float
+            reference scale
+        nf: int
+            number of flavors
+
```
scarlehoff, line 102: nl is missing.

validphys2/src/validphys/photon/alpha.py
```diff
@@ -0,0 +1,241 @@
+import numpy as np
```
scarlehoff, line 1: Please add a small docstr to this module (and also below to the class `Alpha`)

validphys2/src/validphys/photon/alpha.py
```diff
@@ -0,0 +1,241 @@
...
+        Parameters
+        ----------
+        q : float
+            target scale
+        alph_aem_ref : float
+            reference value of alpha_em
+        qref: float
+            reference scale
+        nf: int
+            number of flavors
```
scarlehoff, line 131: same as above

validphys2/src/validphys/photon/alpha.py
```diff
@@ -0,0 +1,241 @@
...
+            target value of a
+        """
+        if (nf, nl) == (0, 0):
+            return alphaem_ref
+        lmu = 2 * np.log(q / qref)
+        den = 1.0 + self.betas_qed[(nf, nl)][0] * alphaem_ref * lmu
+        alpha_LO = alphaem_ref / den
+        alpha_NLO = alpha_LO * (
+            1 - self.betas_qed[(nf, nl)][1] / self.betas_qed[(nf, nl)][0] * alpha_LO * np.log(den)
+        )
```
scarlehoff, line 115: I think `self.betas_qed[(nf, nl)][1]` would not exist if QED is at LO. I think there should be some "if theory qed = 1" to return alpha_LO

validphys2/src/validphys/photon/alpha.py
niclaurenti: in this case `alpha` is not $\alpha_{ref}$ but $\alpha(t)$. I will think to a more clear name

validphys2/src/validphys/photon/alpha.py
```diff
@@ -0,0 +1,254 @@
...
+            return self.alphaem_fixed_flavor_trn(q, alphaem_ref, qref, nf, nl)
+        
+        u = 2 * np.log(q / qref)
+
+        # solve RGE
+        res = solve_ivp(
+            _rge, (0, u), (alphaem_ref,), args=[self.betas_qed[(nf, nl)]], method="Radau", rtol=1e-6
+        )
+        # taking fist (and only) element of y since it is a 1-D differential equation.
+        # Then taking the last element since it is the requested value
```
scarlehoff, line 156:
```suggestion
        #  then the last element of the array which corresponds to alpha(q)
```

validphys2/src/validphys/photon/alpha.py
```diff
@@ -0,0 +1,254 @@
...
+            electromagnetic coupling
+        """
+        nf, nl = self.find_region(q)
+        return self.alphaem_fixed_flavor(
+            q, self.alphaem_thresh[(nf, nl)], self.thresh[(nf, nl)], nf, nl
+        )
+
+    def alphaem_fixed_flavor_trn(self, q, alphaem_ref, qref, nf, nl):
+        """
+        Compute the running alphaem for nf fixed at NLO, using truncated method.
```
scarlehoff, line 91:
```suggestion
        Compute the running alphaem for nf fixed at, at least, NLO, using truncated method.
```

scarlehoff (commented):
I guess this branch is what has been used for the latest fits to the ML right?

niclaurenti:
> I guess this branch is what has been used for the latest fits to the ML right?

Yes!

niclaurenti:
Moreover, until this PR is not merged some tests will be broken on master since it is downloading the new eko version

validphys2/src/validphys/photon/alpha.py
```diff
@@ -0,0 +1,254 @@
...
+
+from .constants import ME, MMU, MQL, MTAU
+
+
+class Alpha:
+    """Class that solves the RGE for alpha_qed"""
+    def __init__(self, theory, q2max):
+        self.theory = theory
+        self.alpha_em_ref = theory["alphaqed"]
+        self.qref = self.theory["Qref"]
```
scarlehoff, line 16:
```suggestion
        self.qref = theory["Qref"]
```

For consistency

scarlehoff (approved):
I noticed... the stability of eko releases is yet another thing that should be addressed but anyway...

PR #1907, opened 2024-01-10
move fiatlux import inside Photon class

RoyStegeman:
Some people were having a problem installing or running fiatlux (or simply didn't install it) while not wishing to do a QED fit. It makes sense to only import fiatlux in case the fit is a QED fit.

### DIFF
--- a/validphys2/src/validphys/photon/compute.py
+++ b/validphys2/src/validphys/photon/compute.py
@@ -2,7 +2,6 @@
 import logging
 import tempfile
 
-import fiatlux
 import numpy as np
 from scipy.integrate import solve_ivp, trapezoid
 from scipy.interpolate import interp1d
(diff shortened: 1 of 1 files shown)


scarlehoff:
The python installation PRs already do that in order to be able to install without fiatlux with the goal of compartmentalizing a bit the code.

So I would rather not put this in master because it is yet another conflict to fix... people can just do it manually if they really need to.

In any case

> or simply didn't install it

Then they can install it

> Some people were having a problem installing

Then they should open an issue and we should fix that instead of hiding the problem.

RoyStegeman:
> Then they should open an issue that we should fix instead of hiding the problem.

I was going to reply that I did raise the issue: https://github.com/conda-forge/fiatlux-feedstock/pull/5

But I failed to realise it had already been resolved: https://github.com/conda-forge/fiatlux-feedstock/pull/8

PR #1911, opened 2024-01-17
FHMRUVV N3LO fits

giacomomagni:
This PR is to add the settings needed to reproduce fits with the FHMRVVU splitting functions.

a3917f5a350b07b6b8471158dfa6562a194bc71d


giacomomagni:
@scarlehoff with this PR and the associated theories one should be able to run it.
One can actually think of merging this (if really needed), the only inconvenient is that I chose theory ID from `1000`
because I didn't want to interfere with anything else.

giacomomagni:
Hi think we should now merge also this PR as soon as possible, as it seems these results will be in the n3lo paper

RoyStegeman:
Are the theories already on the server? If you want to merge it would be better for the theories to be changed to 8... (whichever are next in line)

giacomomagni:
yes they should be... I picked 1000 totally randomly because I didn't know what was the scope of this at the beginning.

PR #1913, opened 2024-01-19
Pre-merge branch for batch of changes

scarlehoff:
Dummy branch to merge the latest changes to a single branch and run tests before merging to master.

Incidentally, I just found a bug either on the way github calls the workflow or the way in which we configured it.


scarlehoff:
@APJansen I'm going to run the bot on my batch of branches and then I will merge yours and do the same. I'll try to rebase them (they should be 100% independent and try to fix any small conflicts myself)

I will do:

#1881 
then
#1888 

Is the order correct? Or should I only merge one of them because it contains the other?

scarlehoff:
Ok! Perfect.

I now need to understand how to tell github actions that this is not, in fact, master...

PR #1918, opened 2024-01-22
Include the theory covariance matrix in the chi2 by dataset

andreab1997:


andreab1997:
@scarlehoff this should work, and this is the report for the chi2 by dataset with the thcovmat: https://vp.nnpdf.science/T9CF-K8mRsi8CP3sLWVulw== . Please check if this is sensible to you. 
To me It seems sensible, however the numbers are not 100% the same in every case of the same obtained with your runcard (i.e. this https://vp.nnpdf.science/NTNlolpYRvSScta_JUCZfg== for NLO and this https://vp.nnpdf.science/GprPuOJoTOOaFt-Mg0OkBg== for NNLO). The numbers are usually very very close but not exactly the same, so it does not bother me too much. However, do you have any idea.


PS: The code is still in draft because I am still trying to make `each_dataset_chi2` a function in such a way I can avoid to overwrite functions that are used somewhere else.

andreab1997:
> @scarlehoff this should work, and this is the report for the chi2 by dataset with the thcovmat: https://vp.nnpdf.science/T9CF-K8mRsi8CP3sLWVulw== . Please check if this is sensible to you. To me It seems sensible, however the numbers are not 100% the same in every case of the same obtained with your runcard (i.e. this https://vp.nnpdf.science/NTNlolpYRvSScta_JUCZfg== for NLO and this https://vp.nnpdf.science/GprPuOJoTOOaFt-Mg0OkBg== for NNLO). The numbers are usually very very close but not exactly the same, so it does not bother me too much. However, do you have any idea?
> 
> PS: The code is still in draft because I am still trying to make `each_dataset_chi2` a function in such a way I can avoid to overwrite functions that are used somewhere else.
> 
> EDIT: I solved the problem of making `each_dataset_chi2` a function in another way and now the changes should be minimal. It is now ready for review then

One possibility is that your function to compute the theory covariance matrix for a single dataset use the PDF defined in the runcard to produce the thcovmat, while the thcovmat of the fit is produced by a different PDF. Can you confirm this? @scarlehoff

scarlehoff:
I don't understand the question, the PDF used to compute should be the PDF which is accessible in the relevant namespace.

In this example https://github.com/NNPDF/nnpdf/blob/master/validphys2/examples/data_theory_comparison_w_sv.yaml the PDF is defined in the dataspecs for instance. As you loop over the dataspecs that PDF is visible.

I haven't tried it, but I would expect that if you added a

```yaml
theory_options:
    pdf: my_special_pdf
    theoryids:
        from_: scale_variation_theories
    use_theorycovmat: true
```

and then do `{@with theory_options@} ---- {@endwith@}`, everything in the middle would use `my_special_pdf`.

Could you add an example runcard? (that way we can more easily understand what the problem is as well)

However, 
> while the thcovmat of the fit is produced by a different PDF.

Indeed, this will always be true, if nothing else because in the example above there is no fit to get the th covmat from.

scarlehoff:
Ahh, I see what you mean now, sorry.

Could you add a runcard with an example using this new way as well?

andreab1997:
> Ahh, I see what you mean now, sorry.
> 
> Could you add a runcard with an example using this new way as well?

Yes I can but note that this new method is not new. This is using exactly the same function that is used in the usual comparison reports

scarlehoff:
But it'd be useful to see how it would work in a minimal example to get the numbers for a few datasets so that one doesn't have to either create a full report or reverse-engineer how is it done there.
And, as you saw, it produces different results depending on whether this is done with the PDF that was the product of the fit or the PDF that was used during the fit to create the thcovmat.

andreab1997:
[example_chi2_by_dataset_thcovmat.txt](https://github.com/NNPDF/nnpdf/files/14026494/example_chi2_by_dataset_thcovmat.txt)
[chi2_by_dataset.txt](https://github.com/NNPDF/nnpdf/files/14026512/chi2_by_dataset.txt)
Would these be okay?

scarlehoff:
Yes, something small like that. Thank you.
I guess you can even remove the t0 part.

You can use:

```yaml

template_text: |
  $\chi^2$ by dataset
  -------------------
  ### Plot
  {@plot_fits_datasets_chi2@}
  ### Table
  {@ProcessGroup fits_chi2_table(show_total=true)@}
```

to avoid having to have two files.

andreab1997:
> Yes, something small like that. Thank you. I guess you can even remove the t0 part.
> 
> You can use:
> 
> ```yaml
> template_text: |
>   $\chi^2$ by dataset
>   -------------------
>   ### Plot
>   {@plot_fits_datasets_chi2@}
>   ### Table
>   {@ProcessGroup fits_chi2_table(show_total=true)@}
> ```
> 
> to avoid having to have two files.

Yes sure, let me do that

andreab1997:
@scarlehoff I agreed to all your comments. I guess this can be merged now, right?

scarlehoff (approved):
Thanks!

PR #1921, opened 2024-01-26
Parallel hyperoptimization with MongoDB

Cmurilochem:
# Aim

This PR aims to implement parallel hyperoptimizations using MongoDB datasets and mongo workers. **This will enable us to calculate several trials simultaneously.**

# Strategy

Similarly to [FileTrials](https://github.com/NNPDF/nnpdf/blob/c3f896a7d07e8287a31f0ba28e66d28c4989a9ab/n3fit/src/n3fit/hyper_optimization/filetrials.py#L52), the main idea is to implement a [MongoFileTrials](https://github.com/NNPDF/nnpdf/blob/9c8ef1432d008d765e05c9b9f48e6eeb32fb94e4/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py#L102) class that inherits from [MongoTrials](https://github.com/hyperopt/hyperopt/blob/d8515263a3168ba8f111179d29e74ffb09670610/hyperopt/mongoexp.py#L660). This new MongoFileTrials class will then be the one we will instantiate before calling hyperopt `fmin`

# Tasks

- [x]  Implement `MongoFileTrials`
- [x] Parse MongoDB option to `n3fit` command and `HyperScanner`
- [x] Adapt `hyper_scan_wrapper` to allow for parallel evaluation of `fmin` trials
- [x] Add `MondoDB` and `pymongo` as dependencies
- [x] Add unit/integration test
- [x] Quantify performance improvement
- [x] Run test on snellius
- [x] Add documentation
- [x] Add restarting options

# Usage

### Local Machine (for simple tests only)
First, make sure that you have `MongoDB` installed either via `conda` (not sure if available in the latest `conda` version) or `apt-get/brew`. Also `pymongo` is necessary but this can be easily  installed via `pip` (it has already been added as dependency). 

In the latest version of the code in this PR, `n3fit` is adapted to run automatically (by internal subprocessing) both `mongod` (that generates `MongoDB` databases) and `hyperopt-mongo-worker` (that launches mongo workers). 

To run parallel hyperopts with `n3fit`, do:

```
n3fit hyper-quickcard.yml 1 -r N_replicas -o dir_output_name --hyperopt N_trials --parallel-hyperopt --num-mongo-workers N
```
where `N` defines the number of mongo workers you want to launch in parallel. Indeed, `N` will define the number of trials we are calculating simultaneously. If you want to restart jobs, make sure you have ` dir_output_name ` in your current path and do:
```
n3fit hyper-quickcard.yml 1 -r N_replicas -o dir_output_name --hyperopt N_trials --parallel-hyperopt --num-mongo-workers N --restart
```

### Snellius

Here is a complete slurm script showing how we would run a hyperopt experiment in parallel in snellius (including restarts if needed):
```bash
#!/bin/bash
#SBATCH --nodes=1
#SBATCH --partition gpu
#SBATCH --gpus-per-node=4
#SBATCH --time 24:00:00
#SBATCH --output=logs/parallel_slurm-%j.out


# Print job info
echo "Job started on $(hostname) at $(date)"


# conda env
ENVNAME=py_nnpdf-master-gpu

# calc details
RUNCARD="hyper-quickcard.yml"
REPLICAS=2
TRIALS=30
DIR_OUTPUT_NAME="test_hyperopt"
RESTART=false

# number of mongo workers to lauch
N_MONGOWORKERS=4


# activate conda environment
source ~/.bashrc
anaconda
conda activate $ENVNAME


# set up cudnn to run on the gpu
CUDNN_PATH=$(dirname $(python -c "import nvidia.cudnn;print(nvidia.cudnn.__file__)"))
echo "CUDNN path: $CUDNN_PATH"
export LD_LIBRARY_PATH="$CONDA_PREFIX/lib/:$CUDNN_PATH/lib:$LD_LIBRARY_PATH"
echo "LD_LIBRARY_PATH: $LD_LIBRARY_PATH"

# Verify GPU usage
ngpus=$(python3 -c "import tensorflow as tf; print(len(tf.config.list_physical_devices('GPU')))")
ngpus_list=$(python3 -c "import tensorflow as tf; print(tf.config.list_physical_devices('GPU'))")

echo "List of physical devices '$ngpus_list'"

if [ ${ngpus} -eq 0 ]; then
    echo "GPUs not being used!"
else
    echo "Using GPUs!"
    echo "Num GPUs Available: ${ngpus}"
fi


# Run n3fit

echo "Changing directory to $TMPDIR"
cp "runcards/$RUNCARD" $TMPDIR
if [ ${RESTART} == "true" ]; then
    cp -r $DIR_OUTPUT_NAME $TMPDIR
fi
cd $TMPDIR


echo "Running n3fit..."

if [ ${RESTART} == "true" ]; then

    echo "Restarting job...."
    echo "n3fit '$TMPDIR/$RUNCARD' 1 -r $REPLICAS --hyperopt $TRIALS -o $DIR_OUTPUT_NAME --parallel-hyperopt --num-mongo-workers $N_MONGOWORKERS --restart"

    n3fit "$TMPDIR/$RUNCARD" 1 -r $REPLICAS --hyperopt $TRIALS -o $DIR_OUTPUT_NAME --parallel-hyperopt --num-mongo-workers $N_MONGOWORKERS --restart

else

    echo "n3fit '$TMPDIR/$RUNCARD' 1 -r $REPLICAS --hyperopt $TRIALS -o $DIR_OUTPUT_NAME --parallel-hyperopt --num-mongo-workers $N_MONGOWORKERS"

    n3fit "$TMPDIR/$RUNCARD" 1 -r $REPLICAS --hyperopt $TRIALS -o $DIR_OUTPUT_NAME --parallel-hyperopt --num-mongo-workers $N_MONGOWORKERS

fi


echo "Copying outputs to $SLURM_SUBMIT_DIR ..."
cp -r "$TMPDIR/$DIR_OUTPUT_NAME" $SLURM_SUBMIT_DIR


echo "Returning to $SLURM_SUBMIT_DIR ..."
cd $SLURM_SUBMIT_DIR


echo "Job completed at $(date)"
```
This would be run by doing:
```
sbatch minimal_parallel_hyperopt.slurm --exclusive
```
Here, each mongo worker selected (4) sees and run in one separate GPU:

top
nvidia_smi

as implemented [here](https://github.com/NNPDF/nnpdf/blob/9c8ef1432d008d765e05c9b9f48e6eeb32fb94e4/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py#L222-L234). In this run, we are then calculating 4 trials in parallel.

We could also set up our experiment to run 2 mongo workers in each gpu (8 trials in parallel), e.g., by using `N_MONGOWORKERS=8` in the script above. In this case, we would observe:

8_wks_top

8_wks_smi

# Performance assessment

## Local Machine

I have just made a very quick test in my local pc to assess the possible performance improvement with parallel hyperopts. I used the `hyper-quickcard.yml` card from `n3fit/tests/regression` (with minor modifications) and run it for 10 trials and 2 replicas varying the number of simultaneously launched mongo workers. The results are summarised in the figure below:

parallel_performance

The results look encouraging *a priori*. 

## Snellius

For the snellius tests, I have employed  the slurm script above as model and a more complete [runcard.txt](https://github.com/NNPDF/nnpdf/files/14229434/runcard.txt). I ran 10 trials with 2 replicas with varying numbers of mongo workers. The final results (after several fine tunings in the code) are plotted in the figure below:

snellius_parallel_performance

It shows the variations of the total clock run time of each job as a function of the number of launched mongo workers. The idea here is that each mongo worker is responsible for one trial in hyperopt, so the more mongo workers we launch the more trials we calculate simultaneously.

I also tested the possibility that we launch more than 1 mongo worker per gpu; see right (light grey) part of the figure. This is actually where we observe the best performance and improvement. So, as seen, a job with 8 mongo workers (2 per gpu) is nearly ~8x faster than a serial hyperopt.


scarlehoff:
Hi @Cmurilochem do we absolutely need mongodb for this? And if so, is there no pip package (or, at worst, conda-forge package?)

Using the defaults channel introduces licensing problems. 

(If there's no other solution so be it, but we can't add it to the conda recipe)

Cmurilochem:
> Hi @Cmurilochem do we absolutely need mongodb for this? And if so, is there no pip package (or, at worst, conda-forge package?)
> 
> Using the defaults channel introduces licensing problems.
> 
> (If there's no other solution so be it, but we can't add it to the conda recipe)

Hi @scarlehoff. Thanks for your help. I created a test for parallel hyperopt and wanted to make it run. The only way is to use `mongodb` though. Yes, I see. There is no pip for it unfortunately, only conda. It runs nicely in the `Test python installation` but never in the "Tests". 

If you so suggest (mainly to avoid add one more depency apart from `lhapdf` and `pandoc`), I could skip the test for now and remove `mongodb` from the conda install.

scarlehoff:
The problem of the dependency is separated (if needed, we can add it).

But wouldn't it be possible to use it from`conda-forge`? https://anaconda.org/conda-forge/mongodb
That way it can be added to the conda recipe.

Cmurilochem:
> The problem of the dependency is separated (if needed, we can add it).
> 
> But wouldn't it be possible to use it from`conda-forge`? https://anaconda.org/conda-forge/mongodb That way it can be added to the conda recipe.
> 
> Edit: otherwise we simply don't add it to the conda-recipe and if one wants to run with mongodb they will have to procure that by themselves. It's not a big problem, I just hoped the conda-forge version worked, but I see it is failing...

Yes...my test with `conda install -c conda-forge mongodb --yes` failed indeed. I will try to add it anyway and see what happens. Let's see...

APJansen:
> The tests have shown that there may exist an immense communication overhead, with workers acting in a kind of chaotic way. This is a point that I will investigate in more details.

Why do you say it's due to communication? That should be very minimal. Seems to me it's memory usage, more than one just doesn't fit on one GPU. In the screenshot you posted, you see that the memory usage is close to 100%. Or... is that this tensorflow thing, where it just reserves all the memory it can?
Even if it is the memory though, I expect that to improve with coming PRs, we can discuss that later.

> I ran 10 trials with 2 replicas with varying numbers of mongo workers, using one gpu each. The results are plotted in the figure below:

I'm still confused by this plot. First of all, the parallel/sequential is always referring to parallelization in trials right? Not in replicas?
Is it true that using mongoDB with only one worker is 1.6 times slower per trial than using the old method, for a test job that takes an hour? That seems very strange?
Then the scaling to 2 and 3 GPUs looks good but it seems to plateau there, any idea why? Communication is only once per trial right, so that shouldn't be it. Are they waiting for each other to finish or something?

scarlehoff:
> Or... is that this tensorflow thing, where it just reserves all the memory it can?

Tensorflow allocates all memory for itself in the GPU. You need to do this to control it https://www.tensorflow.org/guide/gpu#limiting_gpu_memory_growth

APJansen:
Thanks Juan, I remember we did this before. Did you manage to change this memory allocation Carlos?
I added

```python
gpus = tf.config.list_physical_devices('GPU')
for gpu in gpus:
    tf.config.experimental.set_memory_growth(gpu, True)
```
at the start of [this function](https://github.com/NNPDF/nnpdf/blob/ec7640473d393df6db0ff3870e2bb9a21b350cd0/n3fit/src/n3fit/backends/keras_backend/internal_state.py#L87). (to measure the change in memory usage in #1936)

That also allowed me to easily compare the effect on performance. With 100 replicas and the production runcard, it made it 1% slower, which may just be random variation.

I don't know if it will solve the problems here, but maybe it makes sense to just always use this @scarlehoff? I can make a small separate PR for it if you agree.

Cmurilochem:
> Thanks Juan, I remember we did this before. Edit: just noticed you already have it Carlos :)
> 
> I checked quickly the effect on performance, with 100 replicas and the production runcard, it made it 1% slower, which may just be random variation.
> 
> I don't know if it will solve the problems here, but maybe it makes sense to just always use this @scarlehoff? I can make a small separate PR for it if you agree.

Hi @APJansen and @scarlehoff. Thanks for your help. @Cmurilochem I had a spelling mistake while setting `TF_FORCE_GPU_ALLOW_GROWTH`. Instead of `my_env["TF_FORCE_GPU_ALLOW_GROWTH"] = "True"`, this should be `my_env["TF_FORCE_GPU_ALLOW_GROWTH"] = "true"`. Because of this the environment variable was not set and I was having memory fragmentation problems. It seems to be fine for me now. I am able to run 2 workers is the same CPU.

APJansen:
Ah great :) Looks promising, and it's actually still roughly twice as fast as one worker per GPU?

I saw when I was testing for #1936 that GPU usage is often around 90% at 100 replicas, but still (with the changes there) I was able to run 500 replicas as well, with still better than linear scaling in the number of replicas. So that's not a limit somehow. So perhaps we'll be able to run with even more than 2 per GPU.
But I would say for this PR, just make sure it's configurable and working with 2, and we'll test the limits later.

Cmurilochem:
> Ah great :) Looks promising, and it's actually still roughly twice as fast as one worker per GPU?
> 
> I saw when I was testing for #1936 that GPU usage is often around 90% at 100 replicas, but still (with the changes there) I was able to run 500 replicas as well, with still better than linear scaling in the number of replicas. So that's not a limit somehow. So perhaps we'll be able to run with even more than 2 per GPU. But I would say for this PR, just make sure it's configurable and working with 2, and we'll test the limits later.

Yes, indeed. I made a test in which I run 8 workers, 2 per gpu, and it is ~1.8 faster than the job with 4 workers, one in each gpu. I will run more performance tests and update the PR and graph soon. I also see that the `TF_FORCE_GPU_ALLOW_GROWTH` option does not alter much the single-gpu jobs as you obtained.

n3fit/src/n3fit/hyper_optimization/hyper_scan.py
```diff
@@ -114,28 +115,64 @@ def hyper_scan_wrapper(replica_path_set, model_trainer, hyperscanner, max_evals=
...
+            fn=model_trainer.hyperparametrizable,
+            space=hyperscanner.as_dict(),
+            algo=hyperopt.tpe.suggest,
+            max_evals=max_evals,
+            show_progressbar=False,
+            trials=trials,
+            rstate=trials.rstate,
+            trials_save_file=trials.pkl_file,
+        )
+
```
APJansen, line 175:
```suggestion
       # Call to hyperopt.fmin
    fmin_args = dict(
        fn=model_trainer.hyperparametrizable,
        space=hyperscanner.as_dict(),
        algo=hyperopt.tpe.suggest,
        max_evals=max_evals,
        trials=trials,
        rstate=trials.rstate,
    )
    if hyperscanner.parallel_hyperopt:
        trials.start_mongo_workers()
        best = hyperopt.fmin(**fmin_args, show_progressbar=True, max_queue_len=trials.num_workers)
        trials.stop_mongo_workers()
    else:
        best = hyperopt.fmin(**fmin_args, show_progressbar=False, trials_save_file=trials.pkl_file)
```

Perhaps something like this instead, to make it clear what the two branches have in common?

Cmurilochem: Looks nice. Thanks for your suggestion. I will accept it.

Cmurilochem:
Hi @APJansen. Thanks for spotting a problem in the test. You were right; this was due to `Popen` in `start_mongo_database`. If you run 2 consecutive tests it might be the case that during the second time `mongod` is still running in the background which makes your calculator to restart unexpectedly. Just on case, I added new a function that terminates this process before the end of the test. I think it is OK now: I could run it several times and it passed. Please, let me know if you are still facing this problem.

doc/sphinx/source/n3fit/hyperopt.rst
```diff
@@ -386,3 +386,44 @@ To achieve this, you can use the ``--restart`` option within the ``n3fit`` comma
...
+
+ 1. **Initiate the MongoDB database:** Start by setting up the database in your current directory.
+ This database is referred to as ``hyperopt`` in the following instructions. You can initiate it with the command:
+
+  .. code-block:: bash
+
+    mongod --dbpath ./hyperopt
+
+  By default, ``mongod`` uses port ``27017``. This is also the default port for the ``n3fit --db-port`` option.
+  If you wish to use a different port, specify it as follows: ``mongod --dbpath ./hyperopt --port YOUR_PORT_NUMBER``.
```
Radonirinaunimi, line 409:
```suggestion
  If you wish to use a different port, specify it as follows: ``mongod --dbpath ./hyperopt --db-port YOUR_PORT_NUMBER``.
```

n3fit/src/n3fit/scripts/n3fit_exec.py
```diff
@@ -261,6 +271,20 @@ def check_positive(value):
 
         parser.add_argument("--hyperopt", help="Enable hyperopt scan", default=None, type=int)
         parser.add_argument("--restart", help="Enable hyperopt restarts", action="store_true")
+        parser.add_argument(
+            "--parallel-hyperopt",
+            help="Enable hyperopt run in parallel with MongoDB",
+            action="store_true",
+        )
```
Radonirinaunimi, line 278: I should have asked this already in the other PR but what should be the intended behavior if some random user sets both `--restart --parallel-hyperopt`?

doc/sphinx/source/n3fit/hyperopt.rst
```diff
@@ -386,3 +386,44 @@ To achieve this, you can use the ``--restart`` option within the ``n3fit`` comma
...
+simultaneous evaluation of multiple trials.
+
+To set up and run a parallelized hyperopt search, follow these steps:
+
+ 1. **Initiate the MongoDB database:** Start by setting up the database in your current directory.
+ This database is referred to as ``hyperopt`` in the following instructions. You can initiate it with the command:
+
+  .. code-block:: bash
+
+    mongod --dbpath ./hyperopt
```
Radonirinaunimi, line 406: Maybe for more clarity? If so, the subsequent occurrences also need to be updated.
```suggestion
 This database is referred to as ``hyperopt-db`` in the following instructions. You can initiate it with the command:

  .. code-block:: bash

    mongod --dbpath ./hyperopt-db
```

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
```diff
@@ -0,0 +1,250 @@
+"""
+    Hyperopt trial object for parallel hyperoptimization with MongoDB.
+    Data are fetched from MongoDB databases and stored in the form of json files within the nnfit folder
+"""
+import json
+import logging
+import os
+import subprocess
+
+from bson import SON, ObjectId
```
Radonirinaunimi, line 10: `bson` should also be included in the dependencies.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
```diff
@@ -0,0 +1,250 @@
...
+    Retrieve a list of all physical GPU devices available in the system.
+
+    Returns
+    -------
+        list: A list of TensorFlow physical devices of type 'GPU'.
+    """
+    return tf.config.list_physical_devices('GPU')
+
+
+def set_tf_visible_device(gpu_id, devices_list):
```
Radonirinaunimi, line 30: Is this now a deprecated function?
```suggestion
```

Radonirinaunimi (commented):
@Cmurilochem Are there still technical blocking issues in this PR? AFAIC, this already looks really good.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
Cmurilochem: Hi @Radonirinaunimi. Thanks! this `bson` package comes with`pymongo`. We already have `pymongo` in the dependencies.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
Cmurilochem: Thanks @Radonirinaunimi! I made it but I never used. I removed it completely. Also, should we move `get_physical_gpus` to the keras_backend like in `operations.py` ? Note that to make it work I am already importing tensorflow from `mongofiletrials.py`

Radonirinaunimi: If this should be moved, actually, it might make more sense to put it in `keras_backend.internal_state` instead (?).

Cmurilochem: Thanks! It seems more logic indeed. Done.

Cmurilochem:
Hi @Radonirinaunimi, @scarlehoff and @goord. I've made some changes to allow for an automatic restarting using mongodb as we discussed yesterday. They are:
- Added a new class [MongodRunner](https://github.com/NNPDF/nnpdf/blob/f294a855f1e1cf522ffc4fbe0bca77acaf13afca/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py#L62) to also automate the generation of the mongodb database. This was important as for restarting we should have already the previous database uncompressed before launching `mongod`. 
- Added two new methods to [MongoFileTrials](https://github.com/NNPDF/nnpdf/blob/f294a855f1e1cf522ffc4fbe0bca77acaf13afca/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py#L108): [compress_mongodb_database](https://github.com/NNPDF/nnpdf/blob/f294a855f1e1cf522ffc4fbe0bca77acaf13afca/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py#L259C9-L259C35) and [extract_mongodb_database](https://github.com/NNPDF/nnpdf/blob/f294a855f1e1cf522ffc4fbe0bca77acaf13afca/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py#L278) that will be responsible to save (after all runs) and extract (in the case of restarts) the database as needed.

I will be soon pushing the updated docs...

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
```diff
@@ -0,0 +1,290 @@
...
+            try:
+                worker.terminate()
+                worker.wait()
+                log.info(f"Stopped mongo worker {self.workers.index(worker)+1}/{self.num_workers}")
+            except Exception as err:
+                log.error(
+                    f"Failed to stop mongo worker {self.workers.index(worker)+1}/{self.num_workers}: {err}"
+                )
+
+    def compress_mongodb_database(self):
```
Radonirinaunimi, line 259: To compress the database, I think it would be best and neater to use [tarfile](https://docs.python.org/3/library/tarfile.html) as this is more pythonic. In this way, you can simply do:
```python
import tarfile

with tarfile.open(dest_filename_tar, "w") as tar:
    tar.add(path_folder_to_compress)
```

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
```diff
@@ -0,0 +1,290 @@
...
+            log.info(f"Compressing MongoDB database into {self.database_tar_file}")
+            subprocess.run(
+                ['tar', '-cvf', f'{self.database_tar_file}', f'{self.db_name}'] + glob.glob('65*'),
+                check=True,
+            )
+        except subprocess.CalledProcessError as err:
+            raise RuntimeError(f"Error compressing the database: {err}")
+
+    @staticmethod
+    def extract_mongodb_database(database_tar_file):
```
Radonirinaunimi, line 278: In the same way, you can use `tarfile` to untar the compressed database.
```python
import tarfile

with tarfile.open(path_to_tar) as tar:
    tar.extractall(dest)
```

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
```diff
@@ -0,0 +1,290 @@
...
+                log.info(f"Stopped mongo worker {self.workers.index(worker)+1}/{self.num_workers}")
+            except Exception as err:
+                log.error(
+                    f"Failed to stop mongo worker {self.workers.index(worker)+1}/{self.num_workers}: {err}"
+                )
+
+    def compress_mongodb_database(self):
+        """Saves MongoDB database as tar file"""
+        # check if the database exist
+        if not os.path.exists(f"{self.db_name}" and not glob.glob('65*')):
```
Radonirinaunimi, line 262: Out of curiosity, why does it have to start with `65`?

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
Radonirinaunimi, line 278: A question though, didn't we say to untar the compressed database into a temporary directory? In this case using [tempfile](https://docs.python.org/3/library/tempfile.html). I don't think the user needs a copy of the untar folders in the working directory.

Radonirinaunimi (commented):
Thanks @Cmurilochem for the detailed summary. Here are some comments.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
goord: Right, important here is that python's tempfile allows the user to set the environment variable TMPDIR to have control over where the db will be extracted. On snellius scratch-local nodes, this is exactly the fast node-local SSD

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
```diff
@@ -0,0 +1,290 @@
...
+                "--mongo",
+                f"{self.db_host}:{self.db_port}/{self.db_name}",
+            ]
+            if workdir:
+                args.extend(["--workdir", workdir])
+            if exp_key:
+                args.extend(["--exp-key", exp_key])
+            args.extend(["--poll-interval", str(poll_interval)])
+            if use_subprocesses:
+                args.append("--no-subprocesses")
```
APJansen, line 223: I'm confused by this, if you want to use subprocesses, don't use subprocesses? And then it doesn't change the behavior below.
I see this is an option for the mongo worker, which is manually started inside a subprocess. I suppose the reason is to be able to set the GPU it sees through the environment variable? And perhaps these lines are because you don't want to both manually use a subprocess and have mongodb do it as well?

But now `use_subprocesses` defaults to `False` and is never changed, so you are doing both. I have no idea if that's desired or not, but at least this is very confusing.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
```diff
@@ -0,0 +1,290 @@
...
+            log.warning("No GPUs found in the system.")
+
+        # launch mongo workers
+        for i in range(self.num_workers):
+            # construct the command to start a hyperopt-mongo-worker
+            args = [
+                "hyperopt-mongo-worker",
+                "--mongo",
+                f"{self.db_host}:{self.db_port}/{self.db_name}",
+            ]
```
APJansen, line 216:
```suggestion
                f"{self.db_host}:{self.db_port}/{self.db_name}",
                 "--max-consecutive-failures", "10",
                 "--reserve-timeout", "600",
            ]
```

This fixes the timeout issue we had, not sure if one of the two and/or lower settings would suffice, but this was enough for the hyperopt runcard, and I think there's no harm in having these higher than necessary.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
Cmurilochem: Hi @APJansen. I made a test with `use_subprocesses=True` and it basically means that every worker must wait the other workers to finish the task before going on. You are essentially making the job in serial. That is way this option is turned off by default.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
Cmurilochem: Done!

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
Cmurilochem: Thanks for pointing this out. After research for more info I noticed that the names of these extra directories might change depending on the system settings. To avoid this problem I added an extra options to `mongod` `--directoryperdb` which puts everything we need inside `hyperopt-db`. So, we do not need these `65` directories anymore to restart.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
Cmurilochem: I am not sure if we actually need a `temp` file after the above modifications. Now, the only thing that will appear after extracting the tar file is the directory `hyperopt-db`. But if you think that it still necessary I could add. As you can see here https://github.com/NNPDF/nnpdf/blob/0c7eb7f789df28b26597486a691f62607a7c79e6/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py#L294
I just included a path argument. If this is not in the same directory where the mongo workers are also launched we might make their life more difficult to find the database, in this case they must be informed the database whereabouts.

Radonirinaunimi: Ok, let me do some experiments with this and maybe add comments (if still required) later.

Cmurilochem: Thanks for your help @Radonirinaunimi. I'd appreciate vey much. Meanwhile, I will work on CI/CD and update the docs.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
Cmurilochem: Thanks @APJansen. I have added these options to the [start_mongo_workers](https://github.com/NNPDF/nnpdf/blob/5628cba72bcac6d3ac226b2b1d2b8bce821cb748/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py#L217) method but opted for now to leave their values on the defaults.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
APJansen: Ok, can you then just change `use_subprocesses` to `no_subprocesses`? Otherwise this is confusing.

Cmurilochem: Thanks! Done.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
Cmurilochem: I made some tests and it appears to have no downsides. I added these values as default.

Radonirinaunimi:
@Cmurilochem, Is the current documentation correct/complete? By following the current documentation I get the following:
```bash
[INFO]: Started mongo worker 1/1
INFO:hyperopt.mongoexp:PROTOCOL mongo
INFO:hyperopt.mongoexp:USERNAME None
INFO:hyperopt.mongoexp:HOSTNAME localhost
INFO:hyperopt.mongoexp:PORT 27017
INFO:hyperopt.mongoexp:PATH /hyperopt-db/jobs
INFO:hyperopt.mongoexp:AUTH DB None
INFO:hyperopt.mongoexp:DB hyperopt-db
INFO:hyperopt.mongoexp:COLLECTION jobs
INFO:hyperopt.mongoexp:no job found, sleeping for 0.3s
INFO:hyperopt.mongoexp:no job found, sleeping for 0.6s
INFO:hyperopt.mongoexp:no job found, sleeping for 0.5s
...
```

Cmurilochem:
> @Cmurilochem, Is the current documentation correct/complete? By following the current documentation I get the following:
> 
> ```shell
> [INFO]: Started mongo worker 1/1
> INFO:hyperopt.mongoexp:PROTOCOL mongo
> INFO:hyperopt.mongoexp:USERNAME None
> INFO:hyperopt.mongoexp:HOSTNAME localhost
> INFO:hyperopt.mongoexp:PORT 27017
> INFO:hyperopt.mongoexp:PATH /hyperopt-db/jobs
> INFO:hyperopt.mongoexp:AUTH DB None
> INFO:hyperopt.mongoexp:DB hyperopt-db
> INFO:hyperopt.mongoexp:COLLECTION jobs
> INFO:hyperopt.mongoexp:no job found, sleeping for 0.3s
> INFO:hyperopt.mongoexp:no job found, sleeping for 0.6s
> INFO:hyperopt.mongoexp:no job found, sleeping for 0.5s
> ...
> ```

Hi @Radonirinaunimi. Yes, I updated the docs on Monday. Are your using:
```bash
n3fit runcard.yml 1 -r N_replicas --hyperopt N_trials --parallel-hyperopt --num-mongo-workers N
```
It looks that you are just using the default number of mongo worker, 1. So, you could use `--num-mongo-workers 2` to set up 2 workers and so on. What is exactly going on ? The worker never gets a job ? For the tests I've done, It is expected that the worker sleep for a few fraction of seconds to find its job. Have you installed `pymongo<4` and `mongodb` ? Note: it seems that is a bug in `pymongo>=4` that still persists; (here)[https://github.com/hyperopt/hyperopt/pull/845].

APJansen:
Note that this waiting for jobs seemed to increase with the number of replicas, we had to set the timeout for that higher, it can be over 15 minutes. No idea why and a bit annoying, but in the grand scheme of things not a big issue.

Radonirinaunimi:
@Cmurilochem, @APJansen  I started with 2 workers first but the waiting for jobs took a while so I explored first with only just one. After letting it run for a couple of minutes, however, the jobs started at last. Although this is not indeed an issue, maybe there is something we could do to improve this. I am still experimenting so will provide feedback/suggestions later.

Radonirinaunimi:
Running on two Tesla V100 with 150 replicas and setting `--num-mongo-workers 2`, things are looking very good in that the gpus are fully exploited.

hyperopt-gpu-2

I am bit surprised though that still significant operations are utilizing the system's RAM (only ~20gb are used by other processes).

Cmurilochem:
> Running on two Tesla V100 with 150 replicas and setting `--num-mongo-workers 2`, things are looking very good in that the gpus are fully exploited.
> 
> hyperopt-gpu-2
> 
> I am bit surprised though that still significant operations are utilizing the system's RAM (only ~20gb are used by other processes).

Hi @Radonirinaunimi. Yes, indeed. It looks like that you have lots of `hyperopt-mongo-worker` processes. From your settings, we should expect only two working within that job, unless you were running more than one experiment at the same time ?

Radonirinaunimi:
> Yes, indeed. It looks like that you have lots of hyperopt-mongo-worker processes.

How can you see this from the screenshots? And how are workers distributed across gpus?

This was the exact command I ran:
```bash
n3fit runcard.yml 1 -r 150 --hyperopt 15 --parallel-hyperopt --num-mongo-workers 2
```

doc/sphinx/source/n3fit/hyperopt.rst
```diff
@@ -386,3 +386,37 @@ To achieve this, you can use the ``--restart`` option within the ``n3fit`` comma
 
 The above command example is effective when the number of saved trials in the ``test_run/nnfit/replica_1/tries.pkl`` is
 less than ``20``. If there are ``20`` or more saved trials, ``n3fit`` will simply terminate, displaying the best results.
+
+
+Running hyperoptimizations in parallel with MongoDB
+---------------------------------------------------
+
+In NNPDF, you can effectively run hyperoptimization experiments in parallel using `MongoDB <https://www.mongodb.com>`_.
```
Radonirinaunimi, line 394:
```suggestion
It is possible to run hyperoptimization scans in parallel using `MongoDB <https://www.mongodb.com>`_.
```

doc/sphinx/source/n3fit/hyperopt.rst
```diff
@@ -386,3 +386,37 @@ To achieve this, you can use the ``--restart`` option within the ``n3fit`` comma
...
+  n3fit hyper-quickcard.yml 1 -r N_replicas --hyperopt N_trials --parallel-hyperopt --num-mongo-workers N
+
+Here, ``N`` represents the number of MongoDB workers you wish to launch in parallel.
+Each mongo worker handles one trial in Hyperopt. So, launching more workers allows for the simultaneous calculation of a greater number of trials.
+Note that there is no need to manually launch MongoDB databases or mongo workers prior to using ``n3fit``,
+as the ``mongod`` and ``hyperopt-mongo-worker`` commands are automatically executed
+by :meth:`~n3fit.hyper_optimization.mongofiletrials.MongodRunner.start` and
+:meth:`~n3fit.hyper_optimization.mongofiletrials.MongoFileTrials.start_mongo_workers` methods, respectivelly.
+By default, the ``host`` and ``port`` arguments are set to ``localhost`` and ``27017``, while the database is named ``hyperopt-db``.
+If necessary, you can modify these settings using the ``n3fit --db-host`` , ``n3fit --db-port`` and ``n3fit --db-name`` options.
```
Radonirinaunimi, line 412:
```suggestion
If necessary, it is possible modify these settings using the ``n3fit --db-host`` , ``n3fit --db-port`` and ``n3fit --db-name`` options.
```

doc/sphinx/source/n3fit/hyperopt.rst
```diff
@@ -386,3 +386,37 @@ To achieve this, you can use the ``--restart`` option within the ``n3fit`` comma
...
+
+To resume a hyperopt experiment, add the ``--restart`` option to the ``n3fit`` command:
+
+.. code-block:: bash
+
+  n3fit hyper-quickcard.yml 1 -r N_replicas --hyperopt N_trials --parallel-hyperopt --num-mongo-workers N --restart
+
+Note that, unlike in serial execution, parallel hyperoptimization runs do not generate ``tries.pkl`` files.
+Instead, MongoDB databases are saved as ``hyperopt-db.tar.gz`` files inside ``replica_path`` directory.
+These are conveniently extracted for reuse in restart runs.
```
Radonirinaunimi, line 422: I am actually wondering if we should either (a) store the history of the database inside the `fit_path` or (b) append to `hyperopt-db` the name of the runcard? The reason is that if one runs different hyperopts with different runcard (with for example different dataset selections) inside a given directory then everything will mix up.

n3fit/src/n3fit/hyper_optimization/hyper_scan.py
```diff
@@ -114,28 +116,62 @@ def hyper_scan_wrapper(replica_path_set, model_trainer, hyperscanner, max_evals=
...
     # Tell the trainer we are doing hpyeropt
     model_trainer.set_hyperopt(True, keys=hyperscanner.hyper_keys, status_ok=hyperopt.STATUS_OK)
+
+    if hyperscanner.restart_hyperopt:
+        # For parallel hyperopt restarts, extract the database tar file
+        if hyperscanner.parallel_hyperopt:
+            tar_file_to_extract = f"{replica_path_set}/{hyperscanner.db_name}.tar.gz"
+            log.info("Restarting hyperopt run using the MongoDB database %s", tar_file_to_extract)
+            path = os.getcwd()
+            MongoFileTrials.extract_mongodb_database(tar_file_to_extract, path)
```
Radonirinaunimi, line 126:
```suggestion
            MongoFileTrials.extract_mongodb_database(tar_file_to_extract, path=os.getcwd())
```

Radonirinaunimi (commented):
This looks good. I only have one minor/major question below.

doc/sphinx/source/n3fit/hyperopt.rst
Cmurilochem: Yes, you are certainly right. we could leave the name of the tar file as general as `hyperopt-db.tar.gz` inside the `fit_path` but compress instead the database name + runcard name ? What do you think ?

Radonirinaunimi: A per the discussion, please do whichever is easier (implementation-wise). Option (a) would be nicer but if it turns out to introduce more (unneeded) complications then we should go for (b).

APJansen: We implemented option b. It was failing the hyperopt tests locally, but for seemingly unrelated reasons (I had to install pdfflow?). Let's see if it passes the CI.

Cmurilochem: The CI/CD will now pass!!

Radonirinaunimi: Perfect! Thanks

Radonirinaunimi (approved):
Tested it again and everything is ok and consistent. This is now good to go. Thanks again!

Cmurilochem:
> Tested it again and everything is ok and consistent. This is now good to go. Thanks again!

Great! Thanks! I will just now rebase on top of the new master...

Issue #1916, opened 2024-01-22
When running replicas in parallel, make it so the initial state of the network does not depend on the number of replicas.


scarlehoff:
At the moment when running replicas in parallel, the initial state of the network depends on the number of replicas (i.e., it is not the same to run 1 to 20 than doing two runs of 1 to 10, 10 to 20).
While GPU reproducibility is difficult to achieve, the initial state should at least (even if it needs to be done artificially) forced to be the same.

See https://github.com/NNPDF/nnpdf/pull/1881#discussion_r1461978430

(note that all released fits are run sequentially)

scarlehoff:
This is fixed by #1923

PR #1923, opened 2024-01-29
Make weight initialization reproducible

APJansen:
This is an attempt to address #1916, branching off of #1905 because it uses the `MultiInitializer` to initialize the preprocessing weights.

I have checked that when doing any of:
- `n3fit runcard.yaml 1 -r 3`
- `n3fit runcard.yaml 2 -r 3`
- `n3fit runcard.yaml 1 -r 2`

the preprocessing weights, NN weights and PDF output right after creation are identical for replica number 2.

Even after rebasing on trvl-mask-layers though, results do not remain identical.

I don't know where the difference is coming from, as the tr/vl masks, and the invcovmats, are also the same for replica 2.


scarlehoff:
Rebase all these branches on current master. There was a bug in #1881 that I've corrected in #1922 and that would make the final fit be different.

APJansen:
The reason this isn't working yet is of course precisely because it needs both #1788 and #1905, for proper seeds of the train/val splits and the weights respectively. Will continue once those are merged.

APJansen:
I've added a test, do you agree that this is what we want to test here? And do you have an idea why it's not finding the runcards? It works locally (if I remove the linux mark)

scarlehoff:
They should be in the regression folder I think. Locally it works because I guess you are running `pytest` in the folder where the tests are.

n3fit/src/n3fit/tests/quickcard-parallel.yml
```diff
@@ -0,0 +1,79 @@
...
+
+############################################################
+# frac: training fraction
+# ewk: apply ewk k-factors
+# sys: systematics treatment (see systypes)
+dataset_inputs:
+- { dataset: NMC, frac: 0.5 }
+- { dataset: SLACP, frac: 0.5}
+- { dataset: CMSZDIFF12, frac: 0.5, cfac: ['QCD'], sys: 10 }
+- { dataset: ATLASTTBARTOT8TEV, frac: 1.0, cfac: ['QCD'] }
```
scarlehoff, line 18:
```suggestion
```

remove all these datasets. Let's just do it as fast as possible. Same for the other runcard.
This affects only the experimental part and it is not what we are testing here.

n3fit/src/n3fit/tests/quickcard-parallel.yml
```diff
@@ -0,0 +1,79 @@
...
+  t0pdfset: NNPDF40_nnlo_as_01180     # PDF set to generate t0 covmat
+  q2min        : 3.49                # Q2 minimum
+  w2min        : 12.5                # W2 minimum
+
+############################################################
+theory:
+  theoryid: 399        # database id
+
+############################################################
+genrep: True    # on = generate MC replicas, False = use real data
```
scarlehoff, line 31:
```suggestion
genrep: false
```

n3fit/src/n3fit/tests/quickcard-parallel.yml
```diff
@@ -0,0 +1,79 @@
...
+
+parameters: # This defines the parameter dictionary that is passed to the Model Trainer
+  nodes_per_layer: [15, 10, 8]
+  activation_per_layer: ['sigmoid', 'tanh', 'linear']
+  initializer: 'glorot_normal'
+  optimizer:
+    optimizer_name: 'RMSprop'
+    learning_rate: 0.00001
+    clipnorm: 1e-4
+  epochs: 300
```
scarlehoff, line 44:
```suggestion
  epochs: 1
```

Same changes for the other runcard.

scarlehoff (commented):
> I have checked that when doing any of:
> ```
> n3fit runcard.yaml 1 -r 3
> n3fit runcard.yaml 2 -r 3
> n3fit runcard.yaml 1 -r 2
> ```
> the preprocessing weights, NN weights and PDF output right after creation are identical for replica number 2.

I'd say the check missing is that `n3fit runcard.yaml 2` is also identical.

However, to avoid the "replica 1 problem" I would do instead:

```
n3fit runcard.yaml 1 -r 3
n3fit runcard.yaml 2 -r 3
n3fit runcard.yaml 3
```

And check that replica 3 is the same.

With the test_fit itself... What about using `save: weights.h5` and checking that the weights for replica 2 are the same? If `epochs = 1` or `epochs = 0` (not sure) they will be the initial ones...

APJansen:
What is the "replica 1 problem"? I suppose it's related to why you started using varying replica numbers in the regression tests, but I never knew the reason for that?

scarlehoff:
> What is the "replica 1 problem"? 

Turns out that we were only testing the first replica so after some of the multireplica stuff was merged it actually only worked for the first replica when running sequentially.

Also when seeding, if we are missing some seed we might not notice for replica one and only realise from replica 2 onwards.

APJansen:
I found the issue: it's the constraints on the preprocessing weights. Removing them makes all the weights identical to like `1e-8`, with constraints the trainable preprocessing coefficients are completely different.
I don't understand why or how to solve it though, this is just supposed to clip the weights to the specified range right?

APJansen:
The constraint was a simple fix, but still not passing.

Checking the weight differences between my 3 ways of running (in a temporary script below for now), I find relative differences below `1e-6` after 1 epoch, with the exception of the biases, which are initialized to 0 and can have relative differences even above `0.1`. I guess this is not surprising given that they start at 0 and there are some numerical differences, though I hoped there wouldn't be any differences.

<details>
<summary>script</summary>

```python
import h5py
import numpy as np

TEMPDIR = f"/private/var/folders/lt/xy7j0k1j4tdf_k8p_87tb6300000gn/T/pytest-of-aronjansen/pytest-{testnr}/test_multireplica_runs_quickca1"

weights = {}
for name in ['a', 'b', 'c']:
    weight_path = f"{TEMPDIR}/{name}/quickcard-parallel/nnfit/replica_2/weights.h5"
    weights[name] = h5py.File(weight_path, 'r')


def extract_all_weights(file):
    weights = {}
    for key in file.keys():
        if isinstance(file[key], h5py.Group):
            weights[key] = extract_all_weights(file[key])
        else:
            weights[key] = file[key][()]
    return weights

# compute diff of nested dicts
def diff(d1, d2):
    d = {}
    for key in d1:
        if isinstance(d1[key], dict):
            d[key] = diff(d1[key], d2[key])
        else:
            reldiff = np.mean(np.abs(
                (d1[key] - d2[key]) / (d1[key] + d2[key])
                 ))
            d[key] = reldiff
            if reldiff > 1e-5:
                print(f"key: {key}, relative diff: {reldiff}, first: {d1[key]}, second: {d2[key]}")
            else:
                print(f"key: {key}, relative diff: {reldiff}")
    return d

# Comparing a and b:
a = extract_all_weights(weights['a'])
b = extract_all_weights(weights['b'])
c = extract_all_weights(weights['c'])

print("Comparing a and b:")
diff(a, b)
print("Comparing a and c:")
diff(a, c)
print("Comparing b and c:")
diff(b, c)
```
</details>

scarlehoff:
What if you set the learning rate to 0.? (but anyway, having the weights, even without biases, being the same at epoch 1 is a good enough test for the initialization being the same, which is the goal here)

APJansen:
I have added a test on the constraint that purposefully violates it for one replica, then runs one training step to have it corrected, checks that it's indeed satisfied afterward, and that the other replicas were not affected. This is in [6b42580](https://github.com/NNPDF/nnpdf/pull/1923/commits/6b42580cc821c65ae1e13d1597006f893d6ba7f4)

In a separate commit [df8781f](https://github.com/NNPDF/nnpdf/pull/1923/commits/df8781fc54178ef0a99e6f7f3e68cb8e8628c1da) I have simplified the constraint, please check that you agree with this:
As it was, `norms` is just identical to `w`, as it sums over an axis of length 1 and keeps dims (the shape of the weights that this is applied to is `(replicas, 1)`, and was `(1,)` before my earlier changes).
Then `self.rate` is I think always 1 (this I'm not sure about?), so `desired = K.clip(w, self.min_value, self.max_value)`. And then the final normalization is totally unnecessary.

scarlehoff:
I would keep it closer to the original (with norm and so on) https://github.com/keras-team/keras/blob/601488fd4c1468ae7872e132e0f1c9843df54182/keras/constraints.py#L236 on the basis that since it is there it might be for a reason

(and also, that way it is closer to what MinMaxNorm does, it is annoying that they don't provide `MinMax` and that's it)

APJansen:
Ok, I'll revert that. I have to reduce the threshold on the weight differences from `1e-15` to `1e-7` though because of this, but I guess that's good enough.

APJansen:
To come back to your earlier question though, about using a learning rate of 0.0, I had tried this before and to my great confusion the results still changed. That is exactly because of this constraint. I would still argue it's better to change this, the division by epsilon plus the norm in the original is clearly because in that case they want to normalize in the first place, and avoid division by 0.

scarlehoff:
>  1e-7 though because of this, but I guess that's good enough.

Yes, it is!

APJansen:
I've changed the cases to those you suggested:

```
n3fit runcard.yaml 1 -r 3
n3fit runcard.yaml 2 -r 3
n3fit runcard.yaml 3
```
checking on replica 3.

It's passing now, except for 2 of the regression checks, and two tests using python 3.12: the original preprocessing test, which is also a regression test, and, curiously, the test introduced here, but only in the sequential case.

On the regressions: I think the changes in weight initialization shouldn't affect the sequential case, which is how those are run. But the modification of the constraint on the preprocessing weights does change things in all cases, and indeed that's the where the reported violation in the regressions is. Not sure about the 3.12 changes.

scarlehoff:
For 3.12, some `seed=0` are now understood by (keras or tensorflow, one of the two) as `seed=None` and will put a random seed in.
I see that you added a `seed=0`, it could be that.

APJansen:
Ah great, that sounds like it's the issue. I hope I fixed it.

APJansen:
Yep that fixed it :)

n3fit/src/n3fit/backends/keras_backend/base_layers.py
```diff
@@ -128,6 +128,7 @@ def apply_dense(xinput):
         {
             "input_shape": (1,),
             "replica_seeds": None,
+            "base_seed": 0,
```
scarlehoff, line 131:
```suggestion
            "base_seed": 1,
```

maybe better start at one? Or would that be a problem somewhere down the line?

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
```diff
@@ -30,23 +30,28 @@ class MultiDense(Dense):
...
+    base_seed: int (default: 0)
+        Base seed for the single replica initializer to which the replica seeds are added.
     """
 
     def __init__(
         self,
         replica_seeds: List[int],
         kernel_initializer: Initializer,
         is_first_layer: bool = False,
+        base_seed: int = 0,
```
scarlehoff, line 42:
```suggestion
        base_seed: int = 1,
```

n3fit/src/n3fit/backends/keras_backend/multi_dense.py
```diff
@@ -30,23 +30,28 @@ class MultiDense(Dense):
     is_first_layer: bool (default: False)
         Whether this is the first MultiDense layer in the network, and so the input shape
         does not contain a replica axis.
+    base_seed: int (default: 0)
```
scarlehoff, line 33:
```suggestion
    base_seed: int (default: 1)
```

n3fit/src/n3fit/layers/preprocessing.py
```diff
@@ -75,20 +72,25 @@ def generate_weight(self, name: str, kind: str, dictionary: dict, set_to_zero: b
...
             trainable = False
         else:
             minval, maxval = dictionary[kind]
             trainable = dictionary.get("trainable", True)
-            # Set the initializer and move the seed one up
-            initializer = MetaLayer.select_initializer(
-                "random_uniform", minval=minval, maxval=maxval, seed=self.seed
+            # Seeds will be set in the wrapper MultiInitializer
+            # Note: keras 3 interprets a seed of 0 as None and replaces it with a random seed,
+            # so we set it to 1 here and subtract it later when the replica seed is set
```
scarlehoff, line 82: where is it set to 1?

n3fit/src/n3fit/layers/preprocessing.py
```diff
@@ -75,20 +72,25 @@ def generate_weight(self, name: str, kind: str, dictionary: dict, set_to_zero: b
...
+            single_replica_initializer = MetaLayer.select_initializer(
+                "random_uniform", minval=minval, maxval=maxval
             )
-            self.seed += 1
             # If we are training, constrain the weights to be within the limits
             if trainable:
                 constraint = constraints.MinMaxWeight(minval, maxval)
 
+        initializer = MultiInitializer(single_replica_initializer, self.replica_seeds, base_seed=0)
```
scarlehoff, line 90:
```suggestion
        initializer = MultiInitializer(single_replica_initializer, self.replica_seeds, base_seed=1)
```

scarlehoff (commented):
Not sure whether it makes sense or is it needed to change those 0 to 1. It might be that they never enter keras alone (or that you wanted 0 so that it the user uses 0 they indeed get the keras behaviour)

Leaving the comment just to make sure that those 0 are intended.

APJansen:
The base seed is always added to the replica seeds, which never results in a 0. I'm also ok with changing it to 1, but it will change all the regressions of course.

n3fit/src/n3fit/layers/preprocessing.py
APJansen: Sorry this was an earlier approach I tried, will remove this.

scarlehoff:
> The base seed is always added to the replica seeds, which never results in a 0.

Then it is fine. I just wanted to make sure it was intended and not an oversight!

APJansen:
Ok, yes, by passing the base seed separately rather than taking it from the single replica initializer, we now just override whatever the random seed was that Keras chose (rather than adding the replica seed to it).

n3fit/src/n3fit/layers/preprocessing.py
```diff
@@ -31,28 +33,23 @@ class Preprocessing(MetaLayer):
...
             Whether large x preprocessing factor should be active
-        seed: int
-            seed for the initializer of the random alpha and beta values
-        num_replicas: int (default 1)
-            The number of replicas
     """
 
     def __init__(
         self,
+        replica_seeds: Optional[List[int]],
```
scarlehoff, line 40:
```suggestion
        replica_seeds: List[int],
```

otherwise add a default

n3fit/src/n3fit/layers/preprocessing.py
```diff
@@ -31,28 +33,23 @@ class Preprocessing(MetaLayer):
...
-        num_replicas: int = 1,
         **kwargs,
     ):
         if flav_info is None:
             raise ValueError(
                 "Trying to instantiate a preprocessing factor with no basis information"
             )
         self.flav_info = flav_info
-        self.seed = seed
+        self.replica_seeds = replica_seeds
```
scarlehoff, line 50:
```suggestion
        # This variable contains the next seed to be used to generate weights
        self._replica_seeds = replica_seeds
```

n3fit/src/n3fit/layers/preprocessing.py
```diff
@@ -75,20 +72,23 @@ def generate_weight(self, name: str, kind: str, dictionary: dict, set_to_zero: b
...
+            # Seeds will be set in the wrapper MultiInitializer
+            single_replica_initializer = MetaLayer.select_initializer(
+                "random_uniform", minval=minval, maxval=maxval
             )
-            self.seed += 1
             # If we are training, constrain the weights to be within the limits
             if trainable:
                 constraint = constraints.MinMaxWeight(minval, maxval)
 
+        initializer = MultiInitializer(single_replica_initializer, self.replica_seeds, base_seed=0)
```
scarlehoff, line 88:
```suggestion
        initializer = MultiInitializer(single_replica_initializer, self._replica_seeds, base_seed=0)
```

n3fit/src/n3fit/layers/preprocessing.py
```diff
@@ -75,20 +72,23 @@ def generate_weight(self, name: str, kind: str, dictionary: dict, set_to_zero: b
...
+        # increment seeds for the next coefficient
+        self.replica_seeds = [seed + 1 for seed in self.replica_seeds]
```
scarlehoff, line 90:
```suggestion
        self._replica_seeds = [seed + 1 for seed in self._replica_seeds]
```

n3fit/src/n3fit/tests/test_fit.py
```diff
@@ -17,6 +17,8 @@
 import shutil
 import subprocess as sp
 
+import h5py
+import numpy as np
```
scarlehoff, line 21:
```suggestion
```

n3fit/src/n3fit/tests/test_fit.py
```diff
@@ -201,3 +205,41 @@ def test_weirdbasis(tmp_path, timing=30):
...
+    for key in file_1.keys():
+        # The bias is initialized to 0 and will have possibly large relative differences
+        if 'bias' in key:
+            continue
+        if isinstance(file_1[key], h5py.Group):
+            compare_weights(option_1, option_2, file_1[key], file_2[key])
+        else:
+            weight_name = file_1[key].name
+            err_msg = f"Difference between runs `n3fit {option_1}` and `n3fit {option_2}` in weights {weight_name}"
+            np.testing.assert_allclose(
```
scarlehoff, line 243:
```suggestion
            assert_allclose(
```

not sure if you were using numpy somewhere else

scarlehoff (approved):
thanks, lgtm

scarlehoff:
Update also the fitbot with the latest one once it finishes running.

APJansen:
I don't see the fitbot results?

scarlehoff:
It is still running but in a previous commit. It seems it cannor un in the commit of another bot.

APJansen:
We're having some silly issue with one hyperopt test only on python 3.11, where the test fails on the phi2 hyperopt loss. In the CI it's `0.0`, locally it's `1e-10` and locally on master it's `1e7`... 
Practically we can just make the test pass by either changing the assert from `> 0` to `>= 0.0` or removing the option to have phi2 as a loss completely.
But I have no idea where this change is coming from. Running a 1 trial hyperopt with 200 epochs locally on master vs this branch, differences in phi and everything else are order 1 (which is fine as the preprocessing initialization changed), not order 17..

APJansen:
No idea.. that could be but it doesn't explain that locally I also get very different results (with the same environment, in python 3.9). Also checked several seeds, the order of magnitude remains +7 for master and -10 here..

scarlehoff:
Is this the value of phi2? Does it start in the same point?
There might be a 1/(almost 0) division somewhere? That in one case it gets removed / set to 0 and in the other it goes through?

I agree that's very weird.

APJansen:
I found the explanation. The seed was being set as an int, which then uses the same for all replicas. Didn't look into details but I assume this causes the phi2 statistic to be vanishingly small. That must happen for any version. Still that's larger than 0 so it's ok. For some reason, maybe as you mention Juan the addition of packages causing some version differences, in python 3.11 it was exactly 0, thus failing the test, but the actual issue was that it was practically zero anywhere. I just changed the seed from and int to a list of 2 different ints, and locally it's now again order +7.

PR #1924, opened 2024-01-29
Add more regression tests

scarlehoff:
This would've caught several errors that found their way into master (and that, for instante, delayed the qed paper!)

Of course, this is going to introduce a lot of false positives, but I think the trade-off will be worth it (and if not, we can always roll it back or make the check a bit more tolerant).

This PR:
1. Adds check on the final `.exportgrid` produced by the regression tests inside the n3fit tests.
2. Adds bunch of extra regression tests for fits as a separate run workflow
3. Separates `integrability` and `arc_lenghts` into their own category so they are checked with a greater tolerance. `integrability` is too close to 1e-8 and asking for too much precision generates problem. `arc_length` on the other hand involves an integration using `scipy` (I believe this is what you are seeing in your other PR @APJansen). Maybe there's some environmental variable that makes it more robust...

Note that this separate workflow at the moment only deals with fits. The plan is to also add validphys runcards in the future.


APJansen:
That looks convenient for sure, maybe add a label like for the fitbot to run this on the CI?

I ran the extra tests locally and failed 3, the first one because the stop epoch was different (actual 1663, desired 1667), and the other two because of some mismatched elements, only 7 / 2744, these are numbers 1 and 7.

Note in the test overview they appear as `runcard_replica1` etc, rather than say `central_16`. Not sure if it's possible but it would be very convenient to see it like that. Or if not, at least change the dictionary to an ordered dict or a list of tuples or something, so we can see which is which.

scarlehoff:
> That looks convenient for sure, maybe add a label like for the fitbot to run this on the CI?

The idea is that it is run for every commit.

> Not sure if it's possible but it would be very convenient to see it like that. Or if not, at least change the dictionary to an ordered dict or a list of tuples or something, so we can see which is which.

I don't know whether this can be changed to show the content, but in any case the order of the dict is guaranteed for all the python versions we use.

extra_tests/regression_checks.py
```diff
@@ -0,0 +1,45 @@
...
+        "no_msr": 92,
+        "no_sumrules": 18,
+        "no_vsr": 54,
+        "trainable_prepro": 61,
+        "no_lagrange": 27
+        }
+
+@pytest.mark.parametrize("runcard_replica", runcard_and_replicas.items())
+def test_regression_fit(tmp_path, runcard_replica):
+    runcard, replica = runcard_replica
```
APJansen, line 33:
```suggestion
@pytest.mark.parametrize("runcard,replica", runcard_and_replicas.items())
def test_regression_fit(tmp_path, runcard, replica):
```

This fixes the test ids to be more informative.

scarlehoff: didn't realise you could do that with parametrize! Thanks!

scarlehoff:
> I ran the extra tests locally and failed 3, the first one because the stop epoch was different (actual 1663, desired 1667), and the other two because of some mismatched elements, only 7 / 2744, these are numbers 1 and 7.

One thing I just realized, you are probably running it in a Mac. I had to give the mac regression test (the one under `n3fit`) a much larger threshold in order to use the same number as linux used. I think these kind of numerical problems are expected for a change of systems.

it is still useful to run it locally in the sense that if you see that less than 1% of values change, you might still want to push and check whether in the CI is ok, but if they are all wrong then there's clearly something else going on!

.github/workflows/python_installation.yml
```diff
@@ -28,10 +32,8 @@ jobs:
         conda install -n base conda-libmamba-solver
         conda config --set solver libmamba
```
RoyStegeman, line 33: From 23.10 this should be the default behaviour: https://conda.org/blog/2023-11-06-conda-23-10-0-release/

I think we're just getting the latest version so this should no longer be necessary (same in other workflows, but maybe just test to be sure)

extra_tests/regression_fits/normal_fit.yml
```diff
@@ -0,0 +1,71 @@
...
+theory:
+  theoryid: 399        # database id
+
+############################################################
+genrep: True    # on = generate MC replicas, False = use real data
+trvlseed: 3
+nnseed: 2
+mcseed: 1  
+
+save: "weights.h5"
```
RoyStegeman, line 28: Does this run from the start or should it load?

extra_tests/regression_checks.py
```diff
@@ -0,0 +1,44 @@
...
+
+@pytest.mark.parametrize("runcard,replica", runcard_and_replicas.items())
+def test_regression_fit(tmp_path, runcard, replica):
+    runcard_name = f"{runcard}.yml"
+    runcard_file = REGRESSION_FOLDER / runcard_name
+    shutil.copy(runcard_file, tmp_path)
+
+    runcard_info = yaml.load(runcard_file.read_text())
+    if (wname := runcard_info.get("load")) is not None:
+        shutil.copy(REGRESSION_FOLDER / wname, tmp_path)
```
RoyStegeman, line 39:
```suggestion
        shutil.copy(REGRESSION_FOLDER / wname, tmp_path / "weights.h5")
```

extra_tests/readme.md
RoyStegeman, line 1: What is the reason to separate this from the other regression tests? I understand the naming of the file since we may not want to run all of these locally, but I don't see why they need to be outside the n3fit folder entirely.

APJansen:
I have tested the extra tests on top of #1905 and only 2 fail slightly, as they do for me in this branch, so looks good. Once this is merged I'll rebase again and update.

n3fit/src/n3fit/tests/regressions/quickcard_qed.yml
```diff
@@ -42,7 +33,7 @@ trvlseed: 3
 nnseed: 2
 mcseed: 1  
 
-load: "weights.h5"
+save: "weights.h5"
```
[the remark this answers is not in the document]
scarlehoff: Ups, good catch.

extra_tests/regression_fits/normal_fit.yml
scarlehoff: This one is intentional (that's why it's also longer) so that there's at least one that starts from the random initialisation.

extra_tests/regression_checks.py
scarlehoff: No, it needs to be copied as the original name, since different runcards might load different weights (unlike the regressions under `n3fit` where the runcard is the same but for different replicas)

RoyStegeman: Ah nvm, I didn't look close enough. I thought all runcards were loading from weights.h5 so I was expecting some to fail

extra_tests/regression_fits/normal_fit.yml
RoyStegeman: But why save it?

extra_tests/readme.md
scarlehoff: Because those two I want checked also when people do `pytest`. Those two (qed and normal, export grid and json, and replicas 1 and 3) should catch most of the situations we care about.

The new set (plus the validphys ones when I get to it) are hopefully catching edge cases but the thresholds are a bit more relaxed.

extra_tests/regression_fits/normal_fit.yml
scarlehoff: Mainly because I forgot to remove it but then I thought that way also tests that they can be saved :P

Also the weights.h5 in this folder corresponds to this runcard.

RoyStegeman (approved):

APJansen (approved):

PR #1926, opened 2024-02-02
Remove charm sumrule

tgiani:
This PR removes valence csr without affecting the other ones


tgiani:
still need to look into why tests are failing

scarlehoff:
Rebase on top of master, you are branching from an old version. There might be more problems, but the one I see is just that you are getting an old version of eko.

n3fit/src/n3fit/layers/msr_normalization.py
```diff
@@ -93,6 +114,14 @@ def call(self, pdf_integrated, photon_integral):
             ]
         if self._vsr_enabled:
             numerators += [self.vsr_factors]
+        if self._csr_enabled:
+            for i in range(0, len(CSR_COMPONENTS)):
+                numerators += [
+                    op.batchit(4.0 - 1./3. * y[IDX['v15']], batch_dimension=0)
+                ]
+            numerators += [self.vsr_factors]
```
scarlehoff, line 122: I'm confused about this part.
Why would this be different from simply doing:

```python
 numerators = len(CSR_COMPONENTS)*[op.batchit(4.0 - 1./3. * y[IDX['v15']], batch_dimension=0)]
```

scarlehoff (commented):
Thanks for this

Could you add an example for this?

This would be a great occasion to add a new test to the new regression tests:

For instance, make a copy of `no_msr` 
https://github.com/NNPDF/nnpdf/blob/master/extra_tests/regression_fits/no_msr.yml

for `allbutcrs`.

(it's great because it will also add a test for c-cbar which is not there yet!)

n3fit/src/n3fit/layers/msr_normalization.py
tgiani: yes, you re absolutely right

PR #1930, opened 2024-02-05
Update conda installation action

APJansen:
I happened to look at the failing tests in #1929 on my phone, and there I saw a hint which I don't see on my laptop, which was to update the action that sets up conda, hope this fixes the tests that I think will fail now on master as well when rerun.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v1
-    - uses: conda-incubator/setup-miniconda@v2
+    - uses: conda-incubator/setup-miniconda@v3
       with:
         python-version: ${{ matrix.python-version }}
     - name: Setup conda and install conda-build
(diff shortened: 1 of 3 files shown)


RoyStegeman:
You may want to have a look at https://github.com/NNPDF/nnpdf/pull/1928

Cmurilochem:
> [ ](/APJansen)

Just another hint: Maybe the `actions/checkout@v1` could also be updated. Currently we are in `actions/checkout@v4`.

APJansen:
I found someone else with the [same issue](https://github.com/mamba-org/boa/issues/392)
He suggests downgrading conda-build to 3.28.4, I'm not sure how to do that?
Also created a [minimal example](https://github.com/APJansen/citest) that reproduces it.

scarlehoff:
I also have no idea what is happening.
Since due to licence issues I cannot (or should not) use the anaconda repos I've tried to do at the same time an update to conda-forge _and_ updating the workflows.

That seems to work for linux but mac is still broken https://github.com/NNPDF/nnpdf/pull/1932

scarlehoff:
> I found someone else with the [same issue](https://github.com/mamba-org/boa/issues/392) He suggests downgrading conda-build to 3.28.4, I'm not sure how to do that? Also created a [minimal example](https://github.com/APJansen/citest) that reproduces it.

Thanks! I've done this in #1932 (only for mac) and it seemed to work :D 
Now mac is quite pinned while linux is freer than before, but as I said, hopefully most of this will be gone soon.

Issue #1908, opened 2024-01-15
Anticipate how POS FK tables are parsed in the new commondata reader


Radonirinaunimi:
I am not entirely sure yet how the positivity FK tables will be parsed in the new commondata reader. At one of the CMs, it was proposed to read the $(x, Q^2)$ info from the FK tables directly therefore removing the need for a commondata file. However, it might still be better (and more flexible) to parse the positivity exactly in the same way as experimental data (as in the old commondata way). This is especially crucial for the polarised fit in which a positivity dataset is composed of compound FK tables with non-`null` operation.

scarlehoff:
This is now implemented in https://github.com/NNPDF/nnpdf/pull/1678/commits/ea06b2d2c86fa25ea14ca4e85eafeb8049a69138

It is not necessary to define data or uncertainties, only kinematics and the number of datapoints.

Radonirinaunimi:
Thanks a lot for this! I've tested and the reading of the FK tables are fine, however, there is still one small issue remaining which is the passing of the `operation` into the positivity object. Currently, even if  `operation` is not `null` in the metadata, `operation_name` is still `NULL`. If you want, I could also try take care of this.

scarlehoff:
I don't underdstand what you mean? Operation name should be read directly. Can you send me the metadata failing?

Radonirinaunimi:
At the moment, if I catch the `operation_name` from the following line, for example:
https://github.com/NNPDF/nnpdf/blob/ec7640473d393df6db0ff3870e2bb9a21b350cd0/n3fit/src/n3fit/model_gen.py#L175
the value is always `NULL` for positivity dataset regardless of what is defined in the metadata.

> Can I reproduce it by simply adding an op?

Yes, you can reproduce the behavior by taking one of the POS data that you converted and add just `operation` under theory.

scarlehoff:
ok, found it. It is read ok but at some point vp drops operations from positivity and integrability. It will be fixed in the next commit in a few minutes.

Radonirinaunimi:
Awesome, thanks! Tested it and it works fine.

PR #1931, opened 2024-02-05
Automatic port old commondata to the new format

scarlehoff:
This is the port of all the old commondata into the new format. 

It is mostly working and it is mostly automated. However, before pushing the new set of data I need some input.

In the file [buildmaster/old_new_porting_map.yml](https://github.com/NNPDF/nnpdf/blob/automatic_port_old_commondata/buildmaster/old_new_porting_map.yml) you can find the mapping that I'm using. The mapping (maps) datasets to:

1. A new name
2. The arxiv and journal reference (thanks @enocera !). There's also hepdata for some since I already had some from pinecards. But it might be misleading. Feel free to point out when this is the case and I will remove it.

--------

:information_source: **HELP and FEEDBACK WANTED**: 

Please, review the file 
 [buildmaster/old_new_porting_map.yml](https://github.com/NNPDF/nnpdf/blob/automatic_port_old_commondata/buildmaster/old_new_porting_map.yml) 
it should be one of the ones at the top if you go to Files Changed.

----------
If people could already go through the naming that would very helpful. Specially the choice of name for the observable.

And, please, check the energies for CHORUS / SLAC / NUTEV / etc. I've put the average energy as @enocera suggested, but not sure what to put for SLAC there.

For the positivity datasets, all of them are now part of the `POS` process and the `NNPDF` experiment. As far as I am aware, experiment and process for positivity datasets should not mean anything, but if it does somehow please let me know.

Anything else you find funny or think that should be changed, please point it out!

Thank you very much!

(note that the mapping should not include datasets already implemented)

-----------

Todo:
- [x] In the next few days I will push also the actual data, together with reports separated by experiment to check that everything works with the reader. For now I've only checked I could read / compute covmat / compute predictions. I want to check that the plotting also works.
- [x]  I will add a `legacy` version to some of the dataset already implemented if necessary (or change the variant name to legacy, instead of `bugged`)
- [x] Still not sure how to deal with #1908, but I think I will add a fake dataset.
- [x] <s>Go through the _entire_ dataset, not only 4.0. The amount of metadata there will be of, err, lesser quality. But I think we can live with that.</s>

Once everything is working I will remove all old commondata _and_ the possibility of reading it up with validphys.


Radonirinaunimi:
Thanks @scarlehoff! I will have a look this afternoon.

Re the 3rd point on Positivity, why can't you just use the same script to also convert the positivity datasets? These should be exactly the same are regular datasets.

scarlehoff:
> why can't you just use the same script to also convert the positivity datasets?

They are not exactly the same. I need to change the reader to read it normally.
Also, I would like to avoid some of the vices of the previous implementation.

Radonirinaunimi:
> They are not exactly the same. I need to change the reader to read it normally. 

Indeed! For the time being, if positivity datasets are provided in the new format, everything inside `theory` of the `metadata.yaml` are ignored as these information are/were still read from the `yamldb`.

scarlehoff:
I think I'm planning to do this (let me know if it makes sense)

`NNPDF_POS_5GEV` as the dataset holding all metadata.

Kinematics are needed.
Data is needed (even if set to 0, but that allows for cuts)
Uncertainties not needed
Theory as normal.

Radonirinaunimi:
> I think I'm planning to do this (let me know if it makes sense)
> 
> `NNPDF_POS_5GEV` as the dataset holding all metadata.
> 
> Kinematics are needed. Data is needed (even if set to 0, but that allows for cuts) Uncertainties not needed Theory as normal.

Yes, I think this definitely makes sense! The main important thing is that the the value of `theory:operation` would later on (once parsed) be a part of the POS object.

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,382 @@
...
+#
+# <old_commondata_name>:
+#   new_name: <new_name>
+#   reference_arxiv: <http://etc>
+#   reference_hepdata: <http://etc>
+#
+
+# HERA
+HERACOMBCCEM:
+  new_name: HERA_CC_318GEV_EM-SIGMARED
```
felixhekhorn, line 11: Is there a difference between data and theory concerning naming? i.e. why is there a `-` instead of a `_` - compare https://github.com/NNPDF/pinecards/tree/master/HERA_CC_318GEV_EM_SIGMARED

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,382 @@
...
+HERACOMB_SIGMARED_C:
+  new_name: HERA_NC_318GEV_EAVG-SIGMARED-CHARM
+  reference_arxiv: '1804.01019'
+  reference_hepdata: ''
+  journal: Eur. Phys. J. C78 (2018) 473
+HERAF2CHARM:
+  new_name: HERA_NC_318GEV_F2_EAVG-SIGMARED
+  set_name: HERA_NC_318GEV_F2
+  reference_arxiv: '1804.01019'
+  reference_hepdata: ''
```
felixhekhorn, line 59: I don't recognize this dataset - what is it? The paper is the usual heavy pair production one (see entries above) and I think that has only the reduced cross sections for ccbar and bbar (also I just had a quick glance to the paper - i.e. we use table 2+3); moreover, if there would be a F2c measurement - that would go *instead* of `SIGMARED` (and not in parallel); as an aside: why this one does not have the journal?

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,382 @@
...
+  journal: Phys. Rev. D64 (2001) 052002
+DYE886P:
+  new_name: DYE866_Z0_800_PXSEC
+  reference_arxiv: hep-ex/0302019
+DYE906R_dw_ite:
+  new_name: DYE906_Z0_120GEV_DW_PDXSECRATIO
+  reference_arxiv: '2103.04024'
+  journal: Nature 590 (2021), no. 7847 561
+CHORUSNB:
+  new_name: CHORUS_CC_AVG98GEV_NBPB
```
felixhekhorn, line 194: Mmm clearly I'm missing some recent development on the naming, because I wouldn't have expected such a name ... (and again this does not match the theory side https://github.com/NNPDF/pinecards/tree/master/CHORUS_CC_NB_PB_SIGMARED ....)

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,382 @@
...
+  reference_arxiv: ''
+  reference_hepdata: ''
+  journal: Phys. Lett. B282 (1992) 475
+
+# BCDMS
+BCDMSD:
+  new_name: BCDMS_NC_AVG176GEV_DF2
+  reference_arxiv: ' Phys. Lett. B223 (1989) 485'
+BCDMSD_dw_ite:
+  new_name: BCDMS_NC_AVG176GEV_DW__DF2
```
felixhekhorn, line 275: surely `__` should never appear

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,382 @@
...
+ATLAS_SINGLETOP_TCH_R_8TEV:
+  new_name: ATLAS_SINGLETOP_8TEV_TCHAN-XSEC
+
+# FTDY (non dw/ missing)
+DYE605_dw_ite:
+  new_name: DYE605_Z0_38P8GEV_DW_PXSEC
+  reference_arxiv: ''
+  journal: Phys. Rev. D43 (1991) 2815
+DYE886R_dw_ite:
+  new_name: DYE866_Z0_800_RATIO_DW_PDXSECRATIO
```
felixhekhorn, line 182: I suspect `800` is missing a unit

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,382 @@
...
+NTVNUDMNFe_dw_ite:
+  new_name: NUTEV_CC_AVG176GEV_DW_NUFE
+  reference_arxiv: hep-ex/0102049
+  reference_hepdata: 10.17182/hepdata.42034.v1
+  journal: Phys. Rev. D64 (2001) 112006
+
+# SLAC
+# TODO: check energy
+SLACD:
+  new_name: SLAC_NC_AVGXXGEV_DF2
```
felixhekhorn, line 252: I guess we want to separate target `D` and obs `F2`, no?

buildmaster/old_new_porting_map.yml
scarlehoff: Yes. We decided at some point that the last `_` defined what the observable was and everything before was the set of data.
So I went for `HERA_CC_318GEV` for the set of data and `EM-SIGMARED` for the observable.

buildmaster/old_new_porting_map.yml
scarlehoff: Some datasets are added just because it existed before. These are those for which either there is a `metadata.txt` file or were in the file that @enocera sent me.

>why this one does not have the journal?

Because I didn't have the information.

buildmaster/old_new_porting_map.yml
scarlehoff: The naming was supposed to be `<experiment>_<process>_<energy>_<extrastuff>_<observable>`

Sadly the `<energy>` doesn't really work for all of the datasets so I had to move to "average energy" but... of course there might be better naming...

buildmaster/old_new_porting_map.yml
scarlehoff: No, that's a typo! Thanks.

buildmaster/old_new_porting_map.yml
scarlehoff: typo agai!

validphys2/src/validphys/datafiles/new_commondata/ATLAS_1JET_8TEV_R06/metadata.yaml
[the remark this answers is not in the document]
scarlehoff: Those in the diff are the ones that already existed but for which I added a legacy variant.
I "need them" (I should have only added the metadata) because I'd like to people who implemented them to check that as well.

buildmaster/old_new_porting_map.yml
scarlehoff: RE this, let's talk about it during the meeting today. This one (and everything with nuclear corrections really) need some choices to be made...

buildmaster/old_new_porting_map.yml
scarlehoff: This is something we should talk about... do we want to drop old data?

(if so that would make my life much easier...)

(I tried to do a lot of this automatically, there's a lot of chances for arxiv numbers and other stuff to be wrong!)

enocera: My recommendation would be to forget about all pre-NNPDF4.0 data.

buildmaster/old_new_porting_map.yml
felixhekhorn: Mmm ... average is an option, but if there is not a fixed collision energy (such as CC DIS) and the energy field is non-optional, but may contain non-numeric characters, I would suggest `DYNAMIC` or `NOTFIXED` or similar ... no need to hide that fact ...

buildmaster/old_new_porting_map.yml
enocera: I mean, all pre-NNPDF4.0 data sets that were not included in NNPDF4.0.

scarlehoff: Yes, of course. For now I've been checking things with theory 717. I'll also go back to theory 200 to make sure that I haven't missed anything that might have changed name on the meantime.

buildmaster/old_new_porting_map.yml
felixhekhorn: fine - but then I would suggest to slightly swap to `SIGMARED_EM` since we are clearly going "general->detail" ... this would make it also more similar to its sister dataset `SIGMARED_EP`

scarlehoff: That's a very good suggestion. Actually, given that all of them are `sigmared` we might want to even drop it?

buildmaster/old_new_porting_map.yml
felixhekhorn: nuclear corrections are a complete chapter on their own, but with the [discussion from above](https://github.com/NNPDF/nnpdf/pull/1931#discussion_r1481482698) this then should be something `F2-D`

buildmaster/old_new_porting_map.yml
felixhekhorn: > Actually, given that all of them are `sigmared` we might want to even drop it?

SLAC and BCDMS are actually F2

scarlehoff: All the HERA I meant. But it's true, it doesn't hurt to keep it.

Also I saw what you meant now with the journal, since it came from a different source I missed that they were coming from the same reference!

I'll push the mapping with your suggestions. I still haven't changed the energy, but I like the `NOTFIXED` suggestion.

buildmaster/old_new_porting_map.yml
scarlehoff: For now I have separated by target (so, `_D_F2` and `_P_F2` have independent metadata  but `_D_F2-DW` and `_D_F2` share it.

But we can consolidate it (by doing `_F2-D` instead)

buildmaster/old_new_porting_map.yml
```diff
@@ -8,54 +8,54 @@
 
 # HERA
 HERACOMBCCEM:
-  new_name: HERA_CC_318GEV_EM-SIGMARED
+  new_name: HERA_CC_318GEV_SIGMARED_EM
```
felixhekhorn, line 11: Now I'm confused ... just out of old habits I commented at first `SIGMARED_EM` - but then realized my mistake ... shouldn't it be `SIGMARED-EM`?

scarlehoff: Since they are all `sigmared` for these datasets the choice between `_` and `-` becomes arbitrary.

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,390 @@
...
+  new_name: DYE866_Z0_800GEV_PXSEC
+  reference_arxiv: hep-ex/0302019
+DYE906R_dw_ite:
+  new_name: DYE906_Z0_120GEV_PDXSECRATIO-DW
+  reference_arxiv: '2103.04024'
+  journal: Nature 590 (2021), no. 7847 561
+
+#Chorus
+CHORUSNB:
+  new_name: CHORUS_CC_AVG98GEV_NB_PB
```
Radonirinaunimi, line 196: Here, I think we should not use the target as an observable. And the same for some of the datasets below.

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,390 @@
...
+  new_name: ATLAS_Z0_7TEV_36PB_ETA
+  reference_arxiv: '1109.5141'
+  journal: Phys.Rev. D85 (2012) 072004
+ATLASWZRAP36PB:
+  new_name: ATLAS_DY_7TEV_36PB_ETA
+  reference_arxiv: '1109.5141'
+ATLAS_WCHARM_WM_DIFF_7TEV:
+  new_name: ATLAS_WCHARM_7TEV_WM-LEPTONETA
+  reference_arxiv: '1402.6263'
+  journal: JHEP 1405 (2014) 068
```
Radonirinaunimi, line 104: Should we even mention the distinction between the produced gauge boson and the lepton pair $\eta$? For sure, these shouldn't be part of the observable name.

buildmaster/old_new_porting_map.yml
Radonirinaunimi: What I'd propose goes along Felix's suggestion, ie doing `_EM-SIGMARED` instead.

felixhekhorn: Except I just [argued](https://github.com/NNPDF/nnpdf/pull/1931#discussion_r1481482698) for the opposite

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,390 @@
...
+  reference_arxiv: '1710.05167'
+  journal: JHEP 12 (2017) 059
+ATLAS_WM_JET_8TEV_PT:
+  new_name: ATLAS_WJ_JET_8TEV_WMPT
+  reference_arxiv: '1711.03296'
+  journal: JHEP 05 (2018) 077
+ATLAS_WP_JET_8TEV_PT:
+  new_name: ATLAS_WJ_JET_8TEV_WPPT
+  reference_arxiv: '1711.03296'
+  journal: JHEP 05 (2018) 077
```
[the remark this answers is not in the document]
scarlehoff: But in this case they are. It doesn't affect the process, the role of W here is the same as with EM-SIGMARED, EP-SIGMARED.

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,390 @@
...
+ATLAS_SINGLETOP_TCH_DIFF_8TEV_TBAR_RAP_NORM:
+  new_name: ATLAS_SINGLETOP_8TEV_TBAR-RAP-NORM
+  reference_arxiv: '1702.02859'
+  journal: Eur. Phys. J. C 77 (2017) 531
+ATLAS_SINGLETOP_TCH_DIFF_8TEV_T_RAP_NORM:
+  new_name: ATLAS_SINGLETOP_8TEV_T-RAP-NORM
+  reference_arxiv: '1702.02859'
+  journal: Eur. Phys. J. C 77 (2017) 531
+ATLAS_SINGLETOP_TCH_R_13TEV:
+  new_name: ATLAS_SINGLETOP_13TEV_TCHAN-XSEC
```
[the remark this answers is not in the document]
scarlehoff: I've changed it to TCHANNEL to make it clearer.

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,390 @@
...
+ATLAS_WM_JET_8TEV_PT:
+  new_name: ATLAS_WJ_JET_8TEV_WMPT
+  reference_arxiv: '1711.03296'
+  journal: JHEP 05 (2018) 077
+ATLAS_WP_JET_8TEV_PT:
+  new_name: ATLAS_WJ_JET_8TEV_WPPT
+  reference_arxiv: '1711.03296'
+  journal: JHEP 05 (2018) 077
+ATLASDY2D8TEV:
+  new_name: ATLAS_Z0_8TEV_2DMY
```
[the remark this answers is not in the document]
scarlehoff: It is the 2D distribution. I've changed it to M-Y (no need to say 2D if the two d's are explicit)

scarlehoff:
I have updated the names according to the discussions today.

@felixhekhorn @enocera @Radonirinaunimi @t7phy @comane 

Please have a look. If you could add suggestions under the names that you think that should be changed it would be much appreciated.
I've dealt with some of the comments that were here in the PR.
There's some inconsistencies between the names of the DY datasets between ATLAS and CMS. In order to resolve some of these I would need to go through all of these papers...

t7phy:
I know we discussed this in the code meeting and came to an agreement on this and therefore I apologize in advance for opening this box once more, but I realized I may have done a bad job explaining my point (partly because during our CM, I had some other things on my mind too but anyway), but it needs to be considered to ensure that we have a naming convention that is well defined and well understood, not arbitrary. 

There were a number of times the question came up on 'what do we mean by an observable?', which went unanswered so let me try to answer it based on what I had in my mind since the first time I actually proposed this format in last year's AMS meeting. Before looking at an observable, consider what I mean by a dataset: A dataset is a set of data points, i.e. in our case, a collection of events grouped together, and these events can have multiple observables that characterize them. This is precisely why in the PR where I first shared this new format, I called it 'observable name' and not 'distribution name', as a dataset was envisioned to be a collection of events with a collection of observables characterizing the entire set of events.
In this respect, just as there is no mixing of datasets with different process types, or different energy scales, there should not be a mixing of processes from different channels. I gave a bad example in the CM of jet vs dijet because these are ofc different process types, but let's consider just TTBAR, this could of course be produced in the 2L channel or the L+J channel, and we keep them as separate dataset because they are separate sets containing separate events. (there is added subtlety there in that these also originate from different papers, but nevertheless, the point of separate events holds, because experimental papers do not conform to a particular way of releasing their analysis but our commondata implementation must have a particular and well defined way)

In this respect, I strongly suggest that we have ****_EP_SIGMARED and ****_EM_SIGMARED where ****_EP and ****_EM are separate datasets. Sigmared becomes our observable name. One may argue that there is no difference it makes, and that is indeed technically true, but using ****_EP-SIGMARED/EM-SIGMARED sets a precedent whereby we do not have a well defined meaning of dataset and observable and it is upto the judgement of the person implementing the data to do as they please. 
And this ofc follows for every other type of dataset too. I write this at the risk of mildly annoying people by reopening a closed discussion, but I do so because I gave it more thought after our meeting and I truly believe consistency will make our lives simpler in the long run as we avoid unseen future problems. I propose to upvote or downvote this comment to simplify whether or not we should consider my comments on separating these datasets and having a well defined definition of dataset and observable.

buildmaster/old_new_porting_map.yml
```diff
@@ -8,63 +8,63 @@
...
   journal: Eur. Phys. J. C78 (2018) 473
 HERACOMB_SIGMARED_C:
-  new_name: HERA_NC_318GEV_EAVG_SIGMARED_CHARM
+  new_name: HERA_NC_318GEV_EAVG_CHARM-SIGMARED
   reference_arxiv: '1804.01019'
   reference_hepdata: ''
   journal: Eur. Phys. J. C78 (2018) 473
 HERAF2CHARM:
-  new_name: HERA_NC_318GEV_EAVG_SIGMARED_F2
+  new_name: HERAF2CHARM_NC_318GEV_EAVG_F2-SIGMARED
```
t7phy, line 56: I am not sure about this name... why is there F2CHARM in the experiment name? the naming format is <exp>_<process>_<energy or lack thereof (not fixed)>_<misc_details_like_lumi_etc.>_<obs-name-and-misc-additives>. If its imp to have f2charm in the name, it should be in the misc. section, i.e. it should be like HERA_NC_318GEV_F2CHARM_EAVG_<obs-name> or HERA_NC_318GEV_EAVG_F2CHARM_<obs-name> or even HERA_NC_318GEV_EAVG_CHARM_<obs-name>

scarlehoff: Ups. This is one of the ones coming from older times (not used in the 4.0 runcard dataset). It has the experiment set as HERAF2CHARM and so I need to either set a special case for it in my script to autofill this or remove it.

scarlehoff:
For now let's work under the assumption that the current version is the accepted one and review the names assuming everyone is happy with `EP-***`.

Until it is merged, this is really a question of doing a few replacement of `EP-` for `EP_` so we can discuss that point next week on Wednesday.

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,393 @@
...
+  reference_arxiv: '1804.01019'
+  reference_hepdata: ''
+  journal: Eur. Phys. J. C78 (2018) 473
+HERACOMB_SIGMARED_C:
+  new_name: HERA_NC_318GEV_EAVG_CHARM-SIGMARED
+  reference_arxiv: '1804.01019'
+  reference_hepdata: ''
+  journal: Eur. Phys. J. C78 (2018) 473
+HERAF2CHARM:
+  new_name: HERAF2CHARM_NC_318GEV_EAVG_F2-SIGMARED
```
scarlehoff, line 56:
```suggestion
  new_name: HERA_NC_318GEV_EAVG_F2CHARM_SIGMARED
```

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,393 @@
...
+HERACOMB_SIGMARED_C:
+  new_name: HERA_NC_318GEV_EAVG_CHARM-SIGMARED
+  reference_arxiv: '1804.01019'
+  reference_hepdata: ''
+  journal: Eur. Phys. J. C78 (2018) 473
+HERAF2CHARM:
+  new_name: HERA_NC_318GEV_EAVG_F2CHARM_SIGMARED
+  reference_arxiv: '1804.01019'
+  journal: Eur. Phys. J. C78 (2018) 473
+  reference_hepdata: ''
```
felixhekhorn, line 59:
```suggestion
```
right @enocera ?

scarlehoff: I thought we didn't want the older dataset in the sense of "don't bother" but if they were there for free we can keep them, right?

(happy to remove it however, one less to test)

felixhekhorn: even fine, but then a) I'm almost certain that the references are wrong (just checked the paper once more) and b) there should be no `SIGMARED` -> `HERA_NC_318GEV_EAVG-F2CHARM`

scarlehoff: Then it is not for free. Removing it.
Thanks for checking.

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,388 @@
...
+  reference_arxiv: '1506.06042'
+  reference_hepdata: 10.17182/hepdata.68951.v1/t2
+  journal: Eur. Phys. J. C75 (2015) 580
+HERACOMBNCEP920:
+  new_name: HERA_NC_318GEV_EP-SIGMARED
+  reference_arxiv: '1506.06042'
+  reference_hepdata: 10.17182/hepdata.68951.v1/t1
+  journal: Eur. Phys. J. C75 (2015) 580
+HERACOMB_SIGMARED_B:
+  new_name: HERA_NC_318GEV_EAVG_BOTTOM-SIGMARED
```
felixhekhorn, line 46: About this I'm not super happy, but I'm not sure if we can or want to do something ... the thing is this is very related to `HERA_NC_318GEV_*` ...
- here the name is however split in a different way, i.e. `EAVG` is not part of the observable name
- the physical events (in the collider) are the same as in the fully inclusive case, but this is a different analysis (and thus of course a different paper)

scarlehoff: At some point we have to accept that either we truly do one dataset - one folder (like in the old commondata) or there will always be some ambiguity.

felixhekhorn:
> In this respect, I strongly suggest that we have ****_EP_SIGMARED and ****_EM_SIGMARED where ****_EP and ****_EM are separate datasets.

let me play the advocate of the devil: we could also stuff all the Hera fully inclusive datasets into a single folder, based on the observation that they are all coming from a single paper (we did consider that policy already at some point, no?)

scarlehoff:
I'm afraid I will need to add legacy variants for all datasets, including the ones that have already been implemented and (in principle) work as expected, otherwise the naming of the systematics will screw the experimental covmat.

The only group of dataset that survives this is LHCB, because they have all been converted together (there was a typo I've fixed so all good). However all ATLAS and CMS will need this.

scarlehoff:
Ok, the datasets seem to work and the right chi2 can be reproduced per-experiment, which should mean that the fit would work well :) https://vp.nnpdf.science/XuSpuZbhSbytSQuJBEJ8Qw== 

The only dataset I was not able to included here is CMS_1JET_8TEV. I don't know what the problem is (unstable covmat probably) but even copying the data I am not able to reproduce the same results...

scarlehoff:
>  I don't know what the problem

Ok, I found the bug (there was a column ADD or MULT in the wrong place in the old datafile and that was throwing off my converter). Now 4.0 is perfectly reproducible :D

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,388 @@
...
+
+## ATLAS
+ATLASLOMASSDY11EXT:
+  new_name: ATLAS_Z0_7TEV_LOMASS_M
+  reference_arxiv: '1404.1212'
+  reference_hepdata: 10.17182/hepdata.64183.v1/t6
+  journal: JHEP 06 (2014) 112
+
+ATLASWRAP11CC:
+  new_name: ATLAS_WPWM_7TEV_CC
```
Radonirinaunimi, line 64:
```suggestion
  new_name: ATLAS_WPWM_7TEV_CC-ETA
```
Or even `new_name: ATLAS_WPWM_7TEV_CC_ETA` as you prefer.

buildmaster/old_new_porting_map.yml
```diff
@@ -0,0 +1,388 @@
...
+  reference_arxiv: '1612.03016'
+  journal: Eur. Phys. J. C77 (2017) 367
+ATLASWZRAP11CC:
+  new_name: ATLAS_DY_7TEV_46FB_CC
+  reference_arxiv: '1612.03016'
+  reference_hepdata: 10.17182/hepdata.76541.v1/t9
+ATLASWZRAP11CF:
+  new_name: ATLAS_Z0_7TEV_46FB_CF
+  reference_arxiv: '1612.03016'
+  journal: Eur. Phys. J. C77 (2017) 367
```
Radonirinaunimi, line 79: I assume these are all differential in pseudorapidity.

```suggestion
ATLASZRAP11CC:
  new_name: ATLAS_Z0_7TEV_46FB_CC-ETA
  reference_arxiv: '1612.03016'
  journal: Eur. Phys. J. C77 (2017) 367
ATLASWZRAP11CC:
  new_name: ATLAS_DY_7TEV_46FB_CC-ETA
  reference_arxiv: '1612.03016'
  reference_hepdata: 10.17182/hepdata.76541.v1/t9
ATLASWZRAP11CF:
  new_name: ATLAS_Z0_7TEV_46FB_CF-ETA
  reference_arxiv: '1612.03016'
  journal: Eur. Phys. J. C77 (2017) 367
```

Radonirinaunimi (commented):
These are final minor suggestions, please ignore them if they no longer apply.

buildmaster/old_new_porting_map.yml
scarlehoff: I think the distributions we have are in  rapidity (of the gauge boson / dilepton pair) rather than pseudorapidity. But the point still stands, I'll add `-Y` to these.

(I'll check first whether they are η or y)
thanks!

scarlehoff: For Z it's the rapidity of the leptonpair and for W the pseudorapidity of the visible lepton.

Yet another datapoint in favour of separating W and Z :P 

@enocera I'm going to put `-Y` for Z and `-ETA` for W and nothing for `DY`. Are you happy with this choice or you'd rather do something different?

enocera: Happy am I.

Radonirinaunimi: Ok, good. Thanks

scarlehoff:
This is ready for review.

I've rebased this branch on top of master and the reader on top of this branch

So, by checking out the reader branch https://github.com/NNPDF/nnpdf/pull/1678 you should be able to use the new commondata with everything that is in master right now.

```bash
git checkout final_reader_for_new_commondata_mk2
```

To test new commondata not included here (i.e., not part of the port) you can just copy it to `validphys2/src/validphys/datafiles/new_commondata` (remember to install the code in editable mode! (`pip install -e .`)


I've tried a few plots, chi2 comparisons and several fit runcards, but of course, the range of things that can go wrong is very wide.
For now the only thing I've found that it is broken is the kinematic coverage (x-q2) plot since some of the new commondata don't include this information https://github.com/NNPDF/nnpdf/issues/1934

Please let me know if you find any other problems.
Also, if you need me to port some data not already included here let me know. I will probably remove the old commondata from the code (so when you install you only get the new) so if you rely in some fringe dataset please let me know so I can add it to the map yaml file.

Radonirinaunimi:
There is this dataset `ATLAS_WMU_8TEV` for which we have old [commondata files](https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/datafiles/commondata/DATA_ATLAS_WMU_8TEV.dat) but no FK tables. If the old commondata will be totally scrapped, maybe we should also convert it (I couldn't find it in the current list).

In addition, there are also the ATLAS 3D distributions (`ATLAS_Z0_8TEV_3D_CRAP`, `ATLAS_Z0_8TEV_3D_FRAP`) for which we have the filters but no commondata files yet and no FK tables. Do we want to generate these old commondata files, convert them, and add them here?

PS: there is also the FPF datasets but I don't think we need to port them now. If absolutely needed, this can be done later.

scarlehoff:
> Do we want to generate these old commondata files

We might want to leave them so that when the corresponding fktables are generated, the (new) commondata is generated accordingly.

validphys2/src/validphys/datafiles/new_commondata/HERA_NC_225GEV/metadata.yaml
```diff
@@ -0,0 +1,285 @@
...
+        label: k1
+        units: ''
+      k2:
+        description: Variable k2
+        label: k2
+        units: ''
+      k3:
+        description: Variable k3
+        label: k3
+        units: ''
```
felixhekhorn, line 272: (just picked a random dataset so not sure about others) is this PR also supposed to get the correct meta data? like e.g. here k1=x, k2=Q2, k2=y

validphys2/src/validphys/datafiles/new_commondata/HERA_NC_225GEV/metadata.yaml
```diff
@@ -0,0 +1,285 @@
...
+    file: kinematics_EP-SIGMARED.yaml
+  theory:
+    conversion_factor: 1.0
+    operation: 'null'
+    FK_tables:
+    - - HERA_NC_225GEV_EP_SIGMARED
+  data_uncertainties: []
```
felixhekhorn, line 279: why is there no default uncertainties file? (like for the central?)

validphys2/src/validphys/datafiles/new_commondata/HERA_NC_225GEV/metadata.yaml
scarlehoff: No. This is an automatic port also in that sense, there were no labels in the old files.

validphys2/src/validphys/datafiles/new_commondata/HERA_NC_225GEV/metadata.yaml
scarlehoff: So unless people actually implement them manually, there will be no "default" dataset, only the legacy version.

And the goal was that variants are only used for uncertainties, while data and theory should be the same for all variants, I'm abusing the `legacy` variant in a few places though, for reproducibility.

validphys2/src/validphys/datafiles/new_commondata/HERA_NC_225GEV/metadata.yaml
felixhekhorn: I see - so we still need PRs to fine tune the outcome here, right?

validphys2/src/validphys/datafiles/new_commondata/HERA_NC_225GEV/metadata.yaml
felixhekhorn: I see - so this confirms my point [above](https://github.com/NNPDF/nnpdf/pull/1931#discussion_r1487629013)

validphys2/src/validphys/datafiles/new_commondata/HERA_NC_225GEV/metadata.yaml
scarlehoff: The original idea (and the reason why it has taken so long) was to manually reimplement all dataset so that we can correct the many bugs that might have accumulated.

However, I think that's an impossible ideal... so... don't count on it ^^U

scarlehoff:
I'm finding some performance issues that would actually be solved by just having a dataset-per-folder (sort of like in the old implementation was a dataset-per-file) since that would allow us to use the filesystem as a database.

There are many places in which it is assumed that reading the datasets names is cheap, while now it requires reading the metadata + observables. There are workarounds of course, but I feel that's adding complexity to a really simple problem... and would put an end to the naming discussion.
We can discuss more about this tomorrow.

Radonirinaunimi:
> I'm finding some performance issues that would actually be solved by just having a dataset-per-folder (sort of like in the old implementation was a dataset-per-file) since that would allow us to use the filesystem as a database.
> 
Just to get an idea, when you performance issues, how much worse is it?

scarlehoff:
A factor of almost infinite!
Not really, but the problem is that we are comparing with "asking the filesystem" with "asking the filesystem for the folder + the metadata file + reading it + looping over all observable until we find the one we want".
And because before the time was close to 0, there was no measure in place to avoid it, which is fair. So for simple things is very noticeable, for more complicated stuff it will of course be less noticeable (or not at all).

Anyway, the reasons are:
1. Because now we have the information in several `.yaml` file that needs to be reconstructed back, before the commondata dataframe was one single file. In particular in files with hundreds of systematics this is _bad_
2. There are several "datapoints" that before were fixed and now are derived quantities (such as the number of datapoints or the number of systematics). `ndata`, even though we agreed it is derive, it's in practice fixed in most metadata. And `nsys` could be fixed just the same (and updated through variants). See point 1.
3. Not having direct access to the dataset names (they need to be reconstructed from the metadata).

I'm thinking that a way of dealing with all three problems at once is to compile a cache of pickled commondata as they get read for the first time, with more or less the following algorithm:

1. Try to read one dataset
2. Look in ~/.local/share/NNPDF/vp-cache/ for <dataset>_<hash>
3. If <dataset>_<hash> exist, load the information you want from there, otherwise, recompute it and put it there
4. Before loading <dataset> check the hash, if it has changed, recompute it and put it there.

Whether this is a good solution or not will depend on how fast we can take the `hash` of the files https://docs.python.org/3/library/hashlib.html

Instead of <dataset>_<hash> we could even do <dataset>/<hash>.uncert <dataset>/<hash>.data etc so that each dataframe is by itself and one can inspect the dataframes.

I know I said I didn't like increasing the complexity but now I'm thinking it might be a net positive after all. Specially this last point I think would be very useful.

scarlehoff:
Proposal to unify the variables, to be discussed later today:

`pT_t`: pT of the top quark
`y_t`: rapidity of the top quark
`y_ttBar`: rapidity of the top-antitop pair
`m_ttBar`: mass of the top-antitop pair
`y` : rapidity
`eta`: pseudorapidity
`m_12`: mass of an invariant pair (of jets)
`sqrts`: square root of the center of mass energy
`pT`: transverse momementum
`M`: invariant mass of a gauge boson

DIS variables: `Q2`, `w2` , 'x'

Squared quantities will simply have a `2` at the end (i.e., mass squared `m_ttBar2`, or pt squared `pT2`).  

This covers all variables used, while keeping some conventions that are used across validphys (like `sqrts` or `2` for the squares) and it seems not to break anything (the most important big!)
There are a few more but are really variations or combinations of these (like the pt of the top: `pT_ttbar`)


Once agreed I'll simply do a search-and-replace.

scarlehoff:
agreed, I've updated the list. It will also make it consistent with `m_ll` currently used for leptons.

Radonirinaunimi:
> Proposal to unify the variables, to be discussed later today:
> 
> `pT_t`: pT of the top quark `y_t`: rapidity of the top quark `y_ttBar`: rapidity of the top-antitop pair `m_ttBar`: mass of the top-antitop pair `y` : rapidity `eta`: pseudorapidity `m_jj`: mass of an invariant pair (of jets) `m_ll`: mass of an invariant pair (of leptons) `sqrts`: square root of the center of mass energy `pT`: transverse momementum `M`: invariant mass of a gauge boson
> 
> DIS variables: `Q2`, `x`
> 
> Squared quantities will simply have a `2` at the end (i.e., mass squared `m_ttBar2`, or pt squared `pT2`).
> 
> This covers all variables used, while keeping some conventions that are used across validphys (like `sqrts` or `2` for the squares) and it seems not to break anything (the most important big!) There are a few more but are really variations or combinations of these (like the pt of the top: `pT_ttbar`)
> 
> Once agreed I'll simply do a search-and-replace.

Maybe it is better to choose a different name for the rapidity as it can easily be confused with the inelasticity in DIS.

t7phy:
But in principal wouldn't it be better to always have y_<some_particle_name> and so on for other variables like pT. This way we have a consistent name that also works for same process types but is also verbose for readers. That also avoids it's confusion with DIS y.

scarlehoff:
We should at the very least use `V` for both Z and W when relevant. 

Another option is to use `etay` since most of the time the distinction is not that relevant (and it is what was used before)

scarlehoff:
I've standardized now the variable names.

Regarding the problem with y/eta/etay problem, I've decided not to touch those. I would suggest we use `etay` moving forwards but I've preferred not to touch the ones that are already there.

If nobody has any complains, I will merge the new commondata files into master later today.
This does not include the reader (so the files will be in master, but still usable only with the reader) but it is a way of canonising the current implementation.

It would allo also people to start fixing the bugs directly as PR to master, or implementing new data as PR to master. I'll keep keeping the reader on top of master which hopefully be only for a few weeks more.

PR #1932, opened 2024-02-06
Change from defaults to forge, update workflows

scarlehoff:


APJansen:
Are all those other pins still necessary after downgrading conda-build?

scarlehoff:
Let me revert some of them, if they happen to be we keep them and merge this.

.github/workflows/tests.yml
```diff
@@ -26,39 +26,49 @@ jobs:
       NNPDF_SSH_KEY: ${{ secrets.NNPDF_SSH_KEY }}
       PYTHONHASHSEED: "0"
     steps:
-    - uses: actions/checkout@v1
-    - uses: conda-incubator/setup-miniconda@v2
+    - uses: actions/checkout@v4
+      with:
+        fetch-tags: true
+        fetch-depth: 0
```
RoyStegeman, line 32: Is this necessary?

scarlehoff: Yes. We can probably get away with something better than 0, but just fetch-tag doesn't do it.

RoyStegeman: But why? Don't we only need the last commit, which is the default behaviour?

scarlehoff: I don't know, but it was one of the problem here #1928
(and with `fetch-depth: 0` we recover the behaviour of version 1 in that sense)

Note that we use the tag but also the number of commits with respect to the tag. It might be enough with getting out the commit of the tag? I don't know...

scarlehoff:
Removing pins and trying to use conda-forge didn't work for mac. My appetite for testing is finished now.

PR #1933, opened 2024-02-08
Add Q0 varied theories

niclaurenti:
Project done with/by a bachelor student in Milan: to do fits varying Q0 fist of all with mc<Q0<mb.
Maybe afterwards we will move it also above mb or below mc


niclaurenti:
Closing and re-opening with a different ID since that bad person of @andreab1997 stole ID 804

PR #1936, opened 2024-02-12
Fk refactor

APJansen:
# Note

This builds on both #1905 and #1788, so it's better to wait for those to be merged before reviewing this, I'll rebase once that's done.

# The idea

It's a relatively small change (just look at the last commit only), only affecting the observable layers, changing a bit the order in which indices are contracted, and changing from a boolean mask to a float mask.

Preliminary testing on my macbook show significant improvements in:
1. Memory usage (can run 250 replicas, before this couldn't run 100)
2. Multi replica speed, getting more than a factor 2 improvement
3. Single replica speed (tested only on GPU, not CPU), close to factor 2 improvement

I'll do some more thorough tests on snellius.


n3fit/src/n3fit/layers/DIS.py
```diff
@@ -41,39 +45,19 @@ def gen_mask(self, basis):
...
+    def build(self, input_shape):
+        super().build(input_shape)
+        if self.num_replicas > 1:
 
-        Parameters
-        ----------
-            pdf:  backend tensor
-                rank 4 tensor (batch_size, replicas, xgrid, flavours)
+            def compute_observable(pdf, masked_fk):
+                return op.einsum('brxf, nxf -> brn', pdf, masked_fk)
```
scarlehoff, line 56: I don't understand these internal functions.

As far as I can see it is not taking anything from the outside scope so you could make a few of them module-level functions, say

`compute_observable_multireplica`
`compute_observable_one_replica`

etc, with a docstr explaining the rationale

and then according to the if conditional do `self.compute_observable = compute_observable`

n3fit/src/n3fit/layers/DY.py
```diff
@@ -19,48 +32,35 @@ def gen_mask(self, basis):
...
-        """
-        This function perform the fktable \otimes pdf \otimes pdf convolution.
+    def mask_fk(self, fk, mask):
+        if self.num_replicas > 1:
+            return op.einsum('fgF, nFxy -> nxfyg', mask, fk)
+        else:
+            mask = op.einsum('fgF -> Ffg', mask)
+            fk = op.einsum('nFxy -> nFyx', fk)
+            mask_and_fk = (mask, fk)
+            return mask_and_fk
```
scarlehoff, line 42:
```suggestion
        mask = op.einsum('fgF -> Ffg', mask)
        fk = op.einsum('nFxy -> nFyx', fk)
        return mask, fk
```

Please add a docstr and a comment on why this difference between multireplicas and single replicas is necessary.

n3fit/src/n3fit/layers/DY.py
```diff
@@ -19,48 +32,35 @@ def gen_mask(self, basis):
...
+                pdf_x_pdf = op.tensor_product(mask_x_pdf, pdf, axes=[(1,), (1,)])  # Ffy, xf -> Fyx
+                # nFyx, Fyx -> n
+                observable = op.tensor_product(unmasked_fk, pdf_x_pdf, axes=[(1, 2, 3), (0, 1, 2)])
+
+                return op.batchit(op.batchit(observable))  # brn
 
-        # the masked convolution removes the batch dimension
-        ret = op.transpose(self.operation(results))
-        return op.batchit(ret)
+        self.compute_observable = compute_observable
```
scarlehoff, line 66: same comment as before, having these function as module level functions would be great

I would also expect a speed up and memory reduction since they will be shared by different observables... You might want to put a `@tf.function` decorator on them.

... unless the speed up is coming from a memory-tradeoff by having these functions be observable specific? But I would hope not...
if that were the case you can still compile them when you attach them to the given layer by doing

`self.compute_observable = @tf.function(compute_observable)`

n3fit/src/n3fit/layers/observable.py
```diff
@@ -42,36 +43,104 @@ def __init__(self, fktable_data, fktable_arr, operation_name, nfl=14, **kwargs):
...
-            self.many_masks = True
-            self.all_masks = [self.gen_mask(i) for i in basis]
-
         self.operation = op.c_to_py_fun(operation_name)
         self.output_dim = self.fktables[0].shape[0]
 
+        if is_unique(all_bases) and is_unique(xgrids):
+            self.all_masks = [self.gen_mask(all_bases[0])]
+        else:
+            self.all_masks = [self.gen_mask(basis) for basis in all_bases]
```
scarlehoff, line 70: I guess `all_masks` is not needed to be saved anymore if you are doing `self.masks` below.

n3fit/src/n3fit/layers/observable.py
```diff
@@ -42,36 +43,104 @@ def __init__(self, fktable_data, fktable_arr, operation_name, nfl=14, **kwargs):
...
+        masks = self.masks * len(self.fktables)
+        self.masked_fktables = [self.mask_fk(fk, mask) for fk, mask in zip(self.fktables, masks)]
+
+        super().build(input_shape)
+
     def compute_output_shape(self, input_shape):
         return (self.output_dim, None)
 
-    # Overridables
+    def compute_float_mask(self, bool_mask):
```
scarlehoff, line 87: why is this part of the layer? `self` is not used i think

n3fit/src/n3fit/layers/observable.py
```diff
@@ -42,36 +43,104 @@ def __init__(self, fktable_data, fktable_arr, operation_name, nfl=14, **kwargs):
...
+            temp_matrix[tuple(idx)] = 1
+            masked_to_full.append(temp_matrix)
+        masked_to_full = np.stack(masked_to_full, axis=-1)
+        masked_to_full = op.numpy_to_tensor(masked_to_full)
+
+        return masked_to_full
+
+    def call(self, pdf):
+        """
+        This function perform the convolution with the fktable and one (DY) or two (DIS) pdfs.
```
scarlehoff, line 115:
```suggestion
        This function perform the convolution with the fktable and one (DIS) or two (DY-like) pdfs.
```

n3fit/src/n3fit/layers/observable.py
```diff
@@ -42,36 +43,104 @@ def __init__(self, fktable_data, fktable_arr, operation_name, nfl=14, **kwargs):
         super(MetaLayer, self).__init__(**kwargs)
 
         self.nfl = nfl
+        self.num_replicas = None  # set in build
+        self.compute_observable = None  # set in build
```
scarlehoff, line 47: Add a comment about `compute_observable` being a function with signature (pdf, masked_pdf) that need to be overwritten by any children of this class.

Actually, maybe it makes sense to make `compute_observable` be an abstract method and then in `DIS.py` and `DY.py` the choice of which (outside) function to use becomes

```python
def compute_observable(self, pdf, fk):
    if self._one_replica:
        return _compute_dis_observable_one_replica(pdf, fk)
    return _compute_dis_observable_many_replica(pdf, fk)

**Review by scarlehoff — COMMENTED** [review]
I guess that if you run 1 replica in a GPU it would be faster, right?

(otherwise it makes no sense that running 1 replica would be slower than running 100).

File: n3fit/src/n3fit/layers/DIS.py (hunk 1 above)
Comment (APJansen) [reply n3fit/src/n3fit/layers/DIS.py]: I agree that looks better, done!

File: n3fit/src/n3fit/layers/DY.py (hunk 2 above)
Comment (APJansen) [reply n3fit/src/n3fit/layers/DY.py]: Done!

File: n3fit/src/n3fit/layers/observable.py (hunk 5 above)
Comment (APJansen) [reply n3fit/src/n3fit/layers/observable.py]: Agreed, moved it out.

File: n3fit/src/n3fit/layers/observable.py (hunk 6 above)
Comment (APJansen) [reply n3fit/src/n3fit/layers/observable.py]: Fixed

File: n3fit/src/n3fit/layers/observable.py (hunk 7 above)
Comment (APJansen) [reply n3fit/src/n3fit/layers/observable.py]: I added the comment, wanted to avoid these if statements in the call. It probably doesn't matter but I thought it looked cleaner. But can make it an abstract method if you prefer.

**Comment by APJansen** [comment]
Thanks for the comments, I addressed them all, will do some timings after these changes, and with vs without a decorator on the compute_observables, as well as 1 replica on the GPU.

**Comment by APJansen** [comment]
I have no idea why, but the 1 replica case became significantly faster: 92s (vs 118 before, both on CPU).
100 replicas on GPU now takes 92 seconds as well, slightly more but most likely just random variations.

1 replica on the GPU takes 65 seconds, indeed faster than 100 of course, but the scaling is great.

The `@tf.function` decorators don't make a difference in performance.

**Comment by scarlehoff** [comment]
> I have no idea why, but the 1 replica case became significantly faster: 92s (vs 118 before, both on CPU).

It might be able to compile the functions differently now? Maybe playing with the `tf.function` options https://www.tensorflow.org/api_docs/python/tf/function you might be able to get it even faster.

> 1 replica on the GPU takes 65 seconds, indeed faster than 100 of course, but the scaling is great.

Indeed, a factor of 100 in exchange for a factor of 1.5!!!!

File: n3fit/src/n3fit/layers/DIS.py [hunk 8]
```diff
@@ -41,39 +45,55 @@ def gen_mask(self, basis):
...
+    return op.einsum('brxf, nxf -> brn', pdf, masked_fk)
 
-        return self.operation(results)
+
+def compute_dis_observable_one_replica(pdf, masked_fk):
+    """
+    Same operations as above but a specialized implementation that is more efficient for 1 replica,
+    masking the PDF rather than the fk table.
+    """
+    # TODO: check if that is actually true
```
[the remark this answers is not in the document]
APJansen: Good point, just tested using the many replica version in all the observables. On the GPU it's actually faster, 53 seconds. On the CPU it's way slower, 330 seconds.

There are 2 factors here, one is just einsum vs tensordot, another is the order of contractions for DIS.

Masking the PDF has the downside that it cannot be precomputed, but it has the upside that it reduces the number of flavours. Masking the fk table only has to be done once, but it enlarges it (perhaps masking is not the best word in this case, it's expanding it into all flavours with zeroes).

I've tested also the approach of masking the fk table but using tensordot for 1 replica on the CPU
```python
def compute_dy_observable_one_replica(pdf, masked_fk):
    pdf = pdf[0][0]  # yg
    fk_pdf = op.tensor_product(masked_fk, pdf, axes=[(3, 4), (0, 1)])  # nxfyg, yg -> nxf
    observable = op.tensor_product(fk_pdf, pdf, axes=[(1, 2), (0, 1)])  # nxf, xf -> n
    return op.batchit(op.batchit(observable))  # brn
```
This takes 177 seconds on CPU, and again 53 on the GPU.

I don't fully understand these timings, but unfortunately we'll have to keep the branching, unless you're prepared to completely do away with CPU runs, but I don't think that's the case.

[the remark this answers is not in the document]
APJansen: Added something, is that what you meant?
Perhaps I should change `masked_fk` to something like `padded_fk`?

Also did a quick check in multireplica. Using observable code as it is now (so tensordot and masking pdf), but using einsum in multidense, it comes out as 115 seconds, so also slower.

n3fit/src/n3fit/layers/DIS.py
```diff
@@ -4,6 +4,16 @@
...
+
+    Comment on the branching based on the number of replicas:
+        This is purely for performance, masking the PDF is more efficient than padding the fk table
+        for one replica, and so is tensordot over einsum.
+
+        Some timings done on snellius using tensorflow 2.15.0 and varying these 2 factors:
+            | CPU\GPU | einsum | tensordot |
+            | -- | -- | -- |
+            | mask pdf | -  | 92 \ 65 |
+            |mask fk | 330 \ 53 \  | 177 \ 53 |
```
scarlehoff, line 16:
```suggestion
            |mask fk | 330 \ 53 \  | 177 \ 53 |
            
            These timings are all for one replica.
            
            Crucially, `einsum` is a requirement of the multireplica case, while `tensordot` gives a benefit of a factor of 2x for the single replica case.
            Since this branching is required anyhow,
             by masking the PDF for 1 replica instead of padding the fktable we get an extra factor of x2
```

Adding some extra info otherwise in two months I'll see this code and I'll think "why... but the gpu with tensordot is perfect, why do we care about einsum" x)

APJansen:
I've rewritten it as `padded_fk`, it's clearer right?

scarlehoff:
Even better. imho mask was ok once the clarification was added

scarlehoff:
oh, I thought this was tracking master. is it fine for you if I rebase this one on master, and then the fix to 2.16 on this one?

APJansen:
Yep that's fine, I rebased on master a while ago, but not recently.

PR #1937, opened 2024-02-12
add nFONLL alphas varariations to theorydb

RoyStegeman:
This add the following theories to the theorydb
- 804: alphas=0.114
- 805: alphas=0.115
- 806: alphas=0.116
- 807: alphas=0.117
- 808: alphas=0.119
- 809: alphas=0.120
- 810: alphas=0.121
- 811: alphas=0.122


RoyStegeman:
@Radonirinaunimi FYI these are the theories we'll use for now. While we may need more alphas variations later, feel free to continue from 812, I don't think it's really worth the effort to try and group them together.

andreab1997:
> This add the following theories to the theorydb
> 
> * 804: alphas=0.114
> * 805: alphas=0.115
> * 806: alphas=0.116
> * 807: alphas=0.117
> * 808: alphas=0.118
> * 809: alphas=0.120
> * 810: alphas=0.121
> * 811: alphas=0.122

I believe 808 should be 0.119 no?

RoyStegeman:
Yes it should be correct in the db

Radonirinaunimi:
> @Radonirinaunimi FYI these are the theories we'll use for now. While we may need more alphas variations later, feel free to continue from 812, I don't think it's really worth the effort to try and group them together.

Thanks a lot for letting us know! Just a couple of minutes ago, we advised @toonhasenack to start from 820 (https://github.com/NNPDF/theories_slim/pull/33) (since I believe @niclaurenti also requires some IDs (?)) which then leaves 8 IDs in-between. If you are fine that there are these gaps for the time being which you/we could use later we'll keep 820 and 821. But if there are reasons against this, we can ofc start from 812.

PR #1944, opened 2024-02-16
Automate regression updates

APJansen:
# Aim

Be able to add a red-regressions label and have the regressions be recomputed in the CI and automatically committed.

# Status

It seems to be working for the extra regression tests, the result of the workflow applied to this PR as a test is the commit [d142744](https://github.com/NNPDF/nnpdf/pull/1944/commits/d142744692b9962e9ce60a6edfaab3cfec79db18)

I don't know about the fitbot, if we want to automate that too I guess it would be a different workflow but using the same label. And what would need to be updated exactly? The developing_weights are a starting point, so I assume those stay the same. Would you want to update the reference fit that it compares against?

Also, I expected the bot's push to trigger the other tests again, but it seems this doesn't happen. I don't know if that's necessary or not.

### DIFF
--- a/.github/workflows/redo_regressions.yml
+++ b/.github/workflows/redo_regressions.yml
@@ -27,6 +27,10 @@ jobs:
       NNPDF_SSH_KEY: ${{ secrets.NNPDF_SSH_KEY }}
     steps:
     - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        ref: ${{ github.event.pull_request.head.ref }}
+
     - uses: conda-incubator/setup-miniconda@v3
       with:
         python-version: ${{ matrix.python-version }}
--- a/.github/workflows/redo_regressions.yml
+++ b/.github/workflows/redo_regressions.yml
@@ -1,0 +1,60 @@
+name: Redo regressions
+
+# start job only for PRs when a label is added.
+on:
+  pull_request:
+    types: [labeled]
+
+# some general variables
+env:
+  # https://keras.io/getting_started/faq/#how-can-i-obtain-reproducible-results-using-keras-during-development
+  PYTHONHASHSEED: "0"
+
+jobs:
+  build:
+    if: contains(github.event.pull_request.labels.*.name, 'redo-regressions')
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: ["3.11"]
+        include:
+          - os: ubuntu-latest
+            CONDA_OS: linux-64
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
+    env:
+      NETRC_FILE: ${{ secrets.NETRC_FILE }}
+      NNPDF_SSH_KEY: ${{ secrets.NNPDF_SSH_KEY }}
+    steps:
+    - uses: actions/checkout@v4
+    - uses: conda-incubator/setup-miniconda@v3
+      with:
+        python-version: ${{ matrix.python-version }}
+        use-mamba: true
+        auto-update-conda: true
+        activate-environment: test
+    - name: Setup conda and install LHAPDF and pandoc
+      shell: bash -l {0}
+      run: |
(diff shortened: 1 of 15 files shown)


n3fit/src/n3fit/tests/test_fit.py
```diff
@@ -86,7 +86,7 @@ def check_fit_results(
     if regenerate:
         shutil.copy2(new_expgrid_file, old_expgrid_file)
         shutil.copy2(new_json_file, regression_json)
-        raise FileNotFoundError("Regression test regenerated, no files to check against")
+        return
```
scarlehoff, line 89:
```suggestion
```

If doing it in this way we might as well let it continue and check that the files are readable.

.github/workflows/redo_regressions.yml
```diff
@@ -0,0 +1,64 @@
...
+      shell: bash -l {0}
+      run: |
+        pytest extra_tests/regression_checks.py --regenerate True
+    - name: Commit
+      shell: bash -l {0}
+      run: |
+        git config user.name "Redo regressions bot"
+        git add extra_tests/regression_fits/*
+        git status
+        git commit -m "Automatically regenerated regressions."
```
scarlehoff, line 63: Please, add in the commit message the branch name, and PR

scarlehoff:
Thanks!

I think git bot commits cannot trigger again the checks. Let's see whether having the previous check is enough for the branch rules to consider the branch as ok.

Also, I like that we will basically have a `redo-regression` tag that would allow to quickly know whether they were regenerated in a particular branch!

.github/workflows/redo_regressions.yml
```diff
@@ -0,0 +1,64 @@
...
+        pip install .[qed,tests]
+    - name: Regenerate regressions
+      shell: bash -l {0}
+      run: |
+        pytest extra_tests/regression_checks.py --regenerate True
+    - name: Commit
+      shell: bash -l {0}
+      run: |
+        git config user.name "Redo regressions bot"
+        git add extra_tests/regression_fits/*
```
scarlehoff, line 61:
```suggestion
        git add extra_tests/regression_fits/*.json extra_tests/regression_fits/*.exportgrid 
```

Just to make the scope much more specific.

Interesting that the `.exportgrid` didn't change for any of the fits while the integrability and arc-lenghts did. arc-length changes are due to scipy's integrate module, fine, but the integrability surprises me.

APJansen:
> Interesting that the .exportgrid didn't change for any of the fits while the integrability and arc-lenghts did. arc-length changes are due to scipy's integrate module, fine, but the integrability surprises me.

They just changed again, so there is still some source of randomness somewhere. Not really an issue though I think, changes are very small.

scarlehoff:
Yes, sure. I'm just surprised that integrability changed while the `.exportgrid` (which contains hundreds of values) didn't. 

The integrability asks the NN for a bunch of points and sums them together, while the `.exportgrid` asks the NN for hundreds of values...

 (in any case I agree, it doesn't really matter. I just found it curious).

.github/workflows/redo_regressions.yml
APJansen: done!

APJansen:
> Yes, sure. I'm just surprised that integrability changed while the .exportgrid (which contains hundreds of values) didn't.

To be sure everythings ok I checked that if I run it locally, those do get changed.

What about the fit-bot?

APJansen:
> They just changed again, so there is still some source of randomness somewhere. Not really an issue though I think, changes are very small.

Actually that's not true, looking a bit more carefully, only the timings changed now.

scarlehoff:
What do you mean?

scarlehoff:
Ah, no, don't update the fitbot like this.
That's a more conscious choice (and it's really just changing the reference in the workflow, nothing needs to be rerun)

APJansen:
Ok, then this is done I think? I just rebased to master.

RoyStegeman:
Thanks! This is nice!

PR #1945, opened 2024-02-16
Added lru caches to speed up validphys multi-replica initalization

goord:
Added lru caches in validphys to accelerate multi-replica initialization

### DIFF
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -45,10 +45,12 @@ from validphys.loader import (
 )
 from validphys.paramfits.config import ParamfitsConfig
 from validphys.plotoptions import get_info
+from validphys.utils import freeze_args
+from frozendict import frozendict
 import validphys.scalevariations
 
-log = logging.getLogger(__name__)
 
+log = logging.getLogger(__name__)
 
 class Environment(Environment):
     """Container for information to be filled at run time"""
@@ -1247,6 +1249,8 @@ class CoreConfig(configparser.Config):
     def parse_added_filter_rules(self, rules: (list, type(None)) = None):
         return rules
 
+    @freeze_args
+    @functools.lru_cache
     def produce_rules(
         self,
         theoryid,
@@ -1286,7 +1290,7 @@ class CoreConfig(configparser.Config):
 
         if added_filter_rules:
             for i, rule in enumerate(added_filter_rules):
-                if not isinstance(rule, dict):
+                if not (isinstance(rule, dict) or isinstance(rule, frozendict)):
                     raise ConfigError(f"added rule {i} is not a dict")
                 try:
                     rule_list.append(
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -1249,6 +1249,9 @@ class CoreConfig(configparser.Config):
     def parse_added_filter_rules(self, rules: (list, type(None)) = None):
         return rules
 
+    # Every parallel replica triggers a series of calls to this function,
+    # which should not happen since the rules are identical among replicas.
+    # E.g for NNPDF4.0 with 2 parallel replicas 693 calls, 3 parallel replicas 1001 calls...
     @freeze_args
     @functools.lru_cache
     def produce_rules(
@@ -1290,7 +1293,7 @@ class CoreConfig(configparser.Config):
 
         if added_filter_rules:
             for i, rule in enumerate(added_filter_rules):
-                if not (isinstance(rule, (dict, frozendict)):
+                if not isinstance(rule, (dict, frozendict)):
                     raise ConfigError(f"added rule {i} is not a dict")
                 try:
                     rule_list.append(
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -1290,7 +1290,7 @@ class CoreConfig(configparser.Config):
 
         if added_filter_rules:
             for i, rule in enumerate(added_filter_rules):
-                if not (isinstance(rule, dict) or isinstance(rule, frozendict)):
+                if not (isinstance(rule, (dict, frozendict)):
                     raise ConfigError(f"added rule {i} is not a dict")
                 try:
                     rule_list.append(
--- a/validphys2/src/validphys/filters.py
+++ b/validphys2/src/validphys/filters.py
@@ -6,6 +6,7 @@ from collections.abc import Mapping
 from importlib.resources import read_text
 import logging
 import re
+import functools
 
 import numpy as np
 
@@ -13,6 +14,7 @@ from reportengine.checks import check, make_check
 from reportengine.compat import yaml
 from validphys.commondatawriter import write_commondata_to_file, write_systype_to_file
 import validphys.cuts
+from validphys.utils import freeze_args
 
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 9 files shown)


scarlehoff (commented):
Thanks for this.

I think we might be able to remove `freeze_args` depending on `produce_rules`.

validphys2/src/validphys/config.py
```diff
@@ -1247,6 +1249,8 @@ def parse_default_filter_rules_recorded_spec_(self, spec):
     def parse_added_filter_rules(self, rules: (list, type(None)) = None):
         return rules
 
+    @freeze_args
+    @functools.lru_cache
```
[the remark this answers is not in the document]
goord: I will check in master

goord:
Testing with NNPDF4.0 runcard (theorid 200) with 200 epochs and 3 replicas on my laptop yields no numerical differences with the master branch (16 digits)

validphys2/src/validphys/filters.py
```diff
@@ -561,6 +563,8 @@ def _make_point_namespace(self, dataset, idat) -> dict:
         return ns
 
 
+@freeze_args
+@functools.lru_cache
```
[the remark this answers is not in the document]
goord: For NNPDF4.0 with 3 replicas it is called 1001 times. For 2 replicas 693 times.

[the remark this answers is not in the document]
scarlehoff: This one yes, but I think it might be fixed now (in other PR).  The other one I'm more worried about.

validphys2/src/validphys/utils.py
```diff
@@ -5,12 +5,51 @@
...
+# have to try hashing it.
+def is_hashable(obj):
+    try:
+        hash(obj)
+        return True
+    except:
+        return False
+
+
+def immute(obj: Any):
```
RoyStegeman, line 29:
```suggestion
def make_hashable(obj: Any):
```

For symmetry reasongs, since you have `is_hashable`

validphys2/src/validphys/utils.py
```diff
@@ -5,12 +5,51 @@
...
 import tempfile
+from typing import Any, Sequence, Mapping
 
 import numpy as np
 from validobj import ValidationError, parse_input
+from frozendict import frozendict
+
+
+# Since typing.Hashable doesn't check recursively you actually
+# have to try hashing it.
```
RoyStegeman, line 20: This I don't really understand, what cases would we miss by doing `isinstance(obj,Hashable)`?

validphys2/src/validphys/n3fit_data_utils.py
```diff
@@ -7,6 +7,7 @@
 The ``validphys_group_extractor`` will loop over every dataset of a given group
 loading their fktables (and applying any necessary cuts).
 """
+import functools
```
RoyStegeman, line 10:
```suggestion
```

RoyStegeman (commented):
Thanks

validphys2/src/validphys/utils.py
RoyStegeman: Maybe even `to_hashable` for similarity with `pandas`

validphys2/src/validphys/utils.py
```diff
@@ -5,12 +5,51 @@
...
+def is_hashable(obj):
+    try:
+        hash(obj)
+        return True
+    except:
```
RoyStegeman, line 25:
```suggestion
    except TypeError:
```

validphys2/src/validphys/utils.py
goord: I agree, and I prefer the type check you propose. Will implement it

validphys2/src/validphys/config.py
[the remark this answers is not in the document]
goord: For NNPDF4.0 with 3 replicas it is called 1001 times. For 2 replicas 693 times.

validphys2/src/validphys/filters.py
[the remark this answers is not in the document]
goord: For NNPDF4.0 with 3 replicas it is called 1040 times. For 2 replicas 732 times.

validphys2/src/validphys/config.py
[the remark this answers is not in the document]
scarlehoff: I see. Then indeed, let's keep the freeze args for now, but please add a comment with that info on top. That should definitely not happen since the rules should be the same per replica, so I believe this is a signal of a problem somewhere else which hopefully can eventually be solved.

validphys2/src/validphys/utils.py
goord: done in 19771a027d0dacfc966c339180680bdeda06954c

validphys2/src/validphys/utils.py
goord: It works just as well, see 19771a027d0dacfc966c339180680bdeda06954c

validphys2/src/validphys/utils.py
goord: No longer relevant since 19771a027d0dacfc966c339180680bdeda06954c

validphys2/src/validphys/config.py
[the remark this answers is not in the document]
goord: Added comment in 19771a027d0dacfc966c339180680bdeda06954c

validphys2/src/validphys/utils.py
```diff
@@ -5,12 +5,42 @@
 @author: Zahari Kassabov
 """
 import contextlib
+import functools
 import pathlib
 import shutil
 import tempfile
+import typing
+from typing import Any, Sequence, Mapping
```
RoyStegeman, line 13: Hate to be difficult, but please make this consistent. I.e. just `import typing` or import `Hashable` to the local namespace.

goord: Black and isort still have to be run on these changes

RoyStegeman: I should have been clearer, what I meant was being consistent with the importing of Any, Sequence, Mapping, and Hashable, which is not something that isort will change. So doing the following
```suggestion
from typing import Any, Sequence, Mapping, Hashable
```

APJansen:
What is the status of this? Anything still needed? (Just rebased to master)

validphys2/src/validphys/utils.py
```diff
@@ -5,12 +5,41 @@
...
+from typing import Any, Sequence, Mapping, Hashable
 
 import numpy as np
 from validobj import ValidationError, parse_input
+from frozendict import frozendict
+
+
+def make_hashable(obj: Any):
+    # So that we don't infinitely recurse since frozenset and tuples
+    # are Sequences.
```
RoyStegeman, line 21: I guess the point of this comment is to justify why do `isinstance(obj, Hashable)` first. Namely, frozenset and tuples are both `Hashable` and `Sequence` (I don't think the second is true for frozenset), and by doing the checks in this order we avoid doing an unnecessary loop. 

Maybe this point can be clarified and generalised a bit
 
```suggestion
    # If the `Sequence` check came before the `Hashable` check we would end up unnecessarily, recursively checking all items in objects that are both `Sequence` and `Hashable`, such as e.g. a `tuple`.  To avoid this, the `Hashable` check comes first. 
```

RoyStegeman (approved):

PR #1946, opened 2024-02-16
Remove libNNPDF closure tests utilities

scarlehoff:
While preparing the parser for the new commondata one problem was the (very) old data used by the closure tests in libNNPDF and nnfit. This data could already not be read by n3fit and the `vp-rebuild-data` existed to update the data.

This PR removes `vp-rebuild-data` and associate utilities. Closure tests for nnfit will no longer be usable. This is very much a breaking change, if someone is still using them they will have to install a version of NNPDF 4.0.8 or lower, use `vp-rebuild-data` and convert them to the new format.

This is just a commit on top of the new commondata reader because I want to run the tests, fitbot, etc in isolation (and give a space for people to complain) but I will just merge on top of the new reader.


comane:
Hi @scarlehoff, I think this is fine.
I haven't used nnfit for closure testing, and I don't think we will ever need to perform any closure test using it.

PR #1951, opened 2024-02-20
Remove bugged raise Error when using multiple covmats

RoyStegeman:


scarlehoff:
I will continue complaining about the total covmat for my entire life. You cannot stop me.

RoyStegeman:
Counteroffer: please at least don't raise unnecessary errors anymore

scarlehoff:
I was seriously thinking about adding `raise Exception` everywhere inside the `plotoptions` module and then going on a long vacation immediately after.

validphys2/src/validphys/config.py
```diff
@@ -717,10 +711,11 @@ def produce_loaded_theory_covmat(
...
         paths.remove(str(output_path / "tables" / generic_path))
-        for f in files:
-            for path in paths:
-                if f == path:
-                    raise ValueError("More than one theory_covmat file in folder tables")
+        if not (use_user_uncertainties and use_scalevar_uncertainties):
+            for f in files:
+                for path in paths:
+                    if f == path:
+                        raise ValueError("More than one theory_covmat file in folder tables")
```
scarlehoff, line 718: We could also make this error into a warning.
If you have more than one theory covmat by now they should have different names and between higher twist, alphas, and normal mhou we might actually want to have more than one now, right?

validphys2/src/validphys/config.py
```diff
@@ -717,10 +711,11 @@ def produce_loaded_theory_covmat(
...
         ]
         paths.remove(str(output_path / "tables" / generic_path))
-        for f in files:
-            for path in paths:
-                if f == path:
-                    raise ValueError("More than one theory_covmat file in folder tables")
+        if not (use_user_uncertainties and use_scalevar_uncertainties):
+            for f in files:
+                for path in paths:
+                    if f == path:
```
scarlehoff, line 717:
```suggestion
            if any(f in paths for f in files):
                log.warning(f"More than one theory covmat file in {output_path}. Make sure you are using the right one! Using: {generic_path}")
```

validphys2/src/validphys/config.py
RoyStegeman: We could make it a warning, but since in the case of summing different covmats, there will be multiple files in the tables/ directory I think it would be odd to raise a warning in case of the expected behaviour. If anything we should raise a warning if there is only the total covmat but not it's consituent covmats. 

Higher twist and alphas are not in master, but since we may want to sum those together with a MHOU covmat they should indeed get a different name. 

Even currently in master there is the option to load a covmat from a file that's stored on the server at the same time as a MHOU covmat. For the alphas+mhou that's what I do, I store the mhou covmat in the server since it's shared by many fits anyway. This all just to say that even for the current master this is a bug.

validphys2/src/validphys/config.py
RoyStegeman: See my reply [here](https://github.com/NNPDF/nnpdf/pull/1951#discussion_r1496340305)

validphys2/src/validphys/config.py
scarlehoff: then if you confirm that you are indeed expecting to have more than one, just remove this check, the moment we add execeptions is just a false sense of security.

If someone messes up with the runcard after runing vp-setupfit it's their responsability.

RoyStegeman: > If someone messes up with the runcard after runing vp-setupfit it's their responsability

To be honest that might be the way to go. I'm not sure if this error message has saved @andreab1997 from wasting a lot of time before? If not, there's indeed no practical reason to keep it around.

andreab1997: No I never encountered this error so far since I never used more than one theory covmat.

RoyStegeman: Well the error is there specifically for when you want to use a single covmat but two exist in the fit folder, so if you never encountered it there is no point in keeping it.

scarlehoff (approved):
Thanks

PR #1953, opened 2024-02-22
drop figure_by when not required

t7phy:


scarlehoff:
do you want to drop them also for cms? Also the mass of the top would be constant there (or is it separated by mass of the top?)

t7phy:
> do you want to drop them also for cms? Also the mass of the top would be constant there (or is it separated by mass of the top?)

which cms one? because right now I have dropped them for both atlas and cms for all non-double-diff distributions

PR #1955, opened 2024-02-22
Run black and isort on entire repo

APJansen:
Feel free to delete, I don't oversee whether this may be more of a headache than it solves, but I just wanted to get rid of all the diffs that are just from applying black or isort.
So here I did only that, in the top level of the repo, `black .` and `isort .`.

There was one file that black couldn't reformat, due to [this line](https://github.com/NNPDF/nnpdf/blob/93667919fd7b914df35f1e0b3fd06dc6a7ed180b/buildmaster/scripts/commondata.py#L22) which apparently uses Python 2 syntax. I didn't touch that, literally only did the 2 commands above.


scarlehoff:
hi @APJansen this cannot be done now because it will create a mess for many branches.

Also, we already did it in the past and people kept on forgetting to add it so the only solution is to add a workflow that lints the code and reminds the user to do it (I'd rather not do it automatically)

RoyStegeman:
We could urge people to use [pre-commit](https://pre-commit.com/) and add our own dot file?

PR #1959, opened 2024-02-28
Add info about 4.0.9 tag + update fitbot

scarlehoff:


doc/sphinx/source/releases.rst
```diff
@@ -20,6 +20,9 @@ The code is tagged to contextualize the versioning, mark significant
 developments and to mark versions used to produce main results. The
 significant releases since the code was made public are:
 
+`Version 4.0.9 <https://github.com/NNPDF/nnpdf/releases/tag/4.0.9>`_
+    Final release of the 4.0 family for PDF fitting.
```
RoyStegeman, line 24: If the next tag should be 4.0.10 instead of 4.1.0, this statement is a bit confusing since it would refer to papers (but not e.g. alphas determination), but not code versions. 
```suggestion
```

RoyStegeman (approved):

doc/sphinx/source/releases.rst
scarlehoff: What about `4.0 (arxiv blabla)` ?

RoyStegeman: Ah yes, in line with what was said for 4.0.8, here we should say it was used for the N3LO paper.

scarlehoff, line 24:
```suggestion
    Release for 4.0 N3LO; last release fully backwards-compatible with 4.0 pipeline.
    4.0 runcards will still works but external tools, and data and theory not used in the 4.0 family of fits will no longer guaranteed to work from 4.0.10 onwards
```

PR #1960, opened 2024-02-28
 Unpin pandas version + update tag description

RoyStegeman:


scarlehoff:
I cannot see any problems with the report for the bot.

(I've added a specific commit of curio but only for the `pyproject.toml` since we are not doing anything beyond 3.11 with conda yet while the python installation can go beyond 3.12, but it will be also a problem eventually: https://github.com/conda-forge/curio-feedstock/pull/16)

We need to move away from curio but reportengine depends on it...

RoyStegeman:
I'm surprised the fits are not exactly the same, I directly branched this of #1959

scarlehoff:
oh no...

scarlehoff:
Between the two bots conda indexed the new eko which relaxes a bunch of dependencies. If you rebase it should be ok. Even just merging should be ok

RoyStegeman:
I had a look at the package versions used by the fitbot here an in #1959, and besides the different pandas version really the only difference is that here also `python-tzdata`, a timezone package that I'm pretty sure provides functionality in pandas that we don't use anyway, is being installed. 

Maybe 2.0 has some changes to rounding/numerical stability that can explain the difference?

scarlehoff:
We use pandas so heavily that I'm not particularly worried about this changes... as long as we consider this part of the 4.1-beta (so pandas 2 doesn't go in the tag).

RoyStegeman:
Okay if you're happy please approve this. 

> btw, I'll add another bot so that we can have an "exact fitbot" and "last stable version fitbot"

:+1:

.github/workflows/fitbot.yml
```diff
@@ -10,8 +10,8 @@ on:
 env:
   N3FIT_MAXNREP: 20 # total number of replicas to fit
   POSTFIT_NREP: 16 # requested replicas for postfit
-  REFERENCE_SET: NNBOT-c0f99b7b3-2024-02-28 # reference set for exact results
-  STABLE_REFERENCE_SET: NNBOT-c0f99b7b3-2024-02-28 # reference set for last tag
+  REFERENCE_SET: NNBOT-3e7c84220-2024-02-28 # reference set for exact results
+  STABLE_REFERENCE_SET: NNBOT-3e7c84220-2024-02-28 # reference set for last tag
```
scarlehoff, line 14:
```suggestion
  STABLE_REFERENCE_SET: NNBOT-c0f99b7b3-2024-02-28 # reference set for last tag
```

For the tag we should keep the previous one!

RoyStegeman: Ah right, I thought you wanted to tag after this PR

Issue #1942, opened 2024-02-14
Make a light data-only package


scarlehoff:
Given that some people have raised concern about the dependency hell that should be avoidable, it would be good to have a `nnpdf-data` package that includes only:

- data
- theory_db
- reader
- theory-reader

The only dependencies that would introduce are `yaml`, `numpy`, `pandas` and `validobj`.
With both `pandas` and `numpy` being used only to actually load the data/uncertainties/etc but I think that it would be good for such a package to include that as well.

felixhekhorn:
I would be happy with such a package - but I'm not sure I can do something on my own :upside_down_face: note that this would introduce a clear split between methodology (=n3fit) and the rest - which might come also handy to f3fit (cc @peterkrack ) and the GP project (cc @tgiani )

scarlehoff:
@comane a proposal for this:

Right now we have a `pyproject.toml` that installs the `nnpdf` package, which is basically this repository that has an structure of:

```bash
pyproject.toml
n3fit/<fitting code>
validphys/<report code>
validphys/datafiles/<all data>
```

A possible strategy to attempt this could be:

1. Separate the data into 
```bash
pyproject.toml
n3fit/<fitting code>
validphys/<report code>
datafiles/<all data>
```

and then add a second pyproject.toml inside `datafiles`
```bash
pyproject.toml
n3fit/<fitting code>
validphys/<report code>

datafiles/pyproject.toml
datafiles/<all data>
```

And then, perhaps, the repository level `pyproject.toml` can depend on the folder-level `pyproject.toml`. They probably need to have different names, they could be `nnpdf` and `nnpdf_data` (and then, instead of having to import `validphys.datafiles` like now you would need to import `nnpd_data`).

We can think about how to also move the reader to the `datafiles` later on.

We are using [poetry](https://python-poetry.org/docs/pyproject/#packages) because afaik is the only python building tool that allows the flexibility we want to separate the n3fit and validphys package (and the versioning pluging is quite nice) but we are not married to it. If some other build tool is more convenient for this, feel free to change (the other positive thing about poetry though is that eko/yadism/etc are using it too).

PR #1965, opened 2024-02-29
[WIP] NNPDF data package

comane:
This PR addresses #1942

In this draft, the `validphys/datafiles` are moved into a directory `nnpdf_data` with it's own `pyproject.toml`.


scarlehoff:
Since we are removing after the tag C++/libnnpdf/evolven3fit the problems in the tests might be resolves by that.

scarlehoff:
This looks good to me. The only thing is leave `disp_theory.py` in the vp/datafiles folder (we can deal with the theory later), let's do only data for now.
This is a good starting point tbh, if you apply the changes to current master we can have the stuff already in that folder in the root of the repository and then in a subsequent PR we create a specific package to install it separately from the rest of the code.

Eventually we can move the parser also there but that can be yet another pr.

scarlehoff:
You probably want also to set a minimum version of ruamel to avoid the need for the compat module

comane:
Taken care of in #2009

PR #1973, opened 2024-03-01
atlas 13tev ttb hadr implementation

t7phy:
~~https://vp.nnpdf.science/IsNUvaBuRLuWHFQ5CjPpVg==~~
https://vp.nnpdf.science/aMe4tajeTqG1pruPI5iI8g==


scarlehoff:
I think there might be some info missing or a bug somewhere in the reader since there's no label in the y axis https://vp.nnpdf.science/IsNUvaBuRLuWHFQ5CjPpVg==/#matched_datasets_from_dataspecs0_plot_fancy_dataspecs_0

(ex. of dataset with y label https://vp.nnpdf.science/oSZLrPg3Tyyf-i2mE33G-Q==/#dat_compare3_dataset_inputs0_plot_fancy_0)

Also, what does `impl` mean?

Radonirinaunimi:
Aren't you just missing the key `y_label` inside `plotting`?

t7phy:
> Aren't you just missing the key `y_label` inside `plotting`?

I didn't even know we had a y_label in plotting block. Shouldn't the y axis label come from observable label ?

Radonirinaunimi:
> > Aren't you just missing the key `y_label` inside `plotting`?
> 
> I didn't even know we had a y_label in plotting block. Shouldn't the y axis label come from observable label ?

You mean `dataset_label`?. They don't have to be necessarily the same. `dataset_label` is a more generic description and appears as title in the plot while `y_label` exactly specifies what's on the `y`-axis, as you can see on the second example above.

t7phy:
> > > Aren't you just missing the key `y_label` inside `plotting`?
> > 
> > 
> > I didn't even know we had a y_label in plotting block. Shouldn't the y axis label come from observable label ?
> 
> You mean `dataset_label`?. They don't have to be necessarily the same. `dataset_label` is a more generic description and appears as title in the plot while `y_label` exactly specifies what's on the `y`-axis, as you can see on the second example above.

No what I meant is as follows: in any of the implementation, I never added x_label or y_label in the plotting block. For x_label it always took from the label of the particular kinematic variable used, so for y_label, I assumed it would take from observable label (there is a observable key, which contains 3 more keys: description, label and units, so I am referring to that observable label). but anyway, I guess y_label will be needed in plotting block, but this will need a fix in all datasets that I implemented then.

t7phy:
> Yes, otherwise it'll never work.

yes, however I think this fix would be best done in a separate PR because it needs to be done for all the datasets, and I will do that asap, so for the time being, can we atleast merge this implementation?

validphys2/src/validphys/datafiles/new_commondata/ATLAS_TTBAR_13TEV_HADR_DIF/metadata.yaml
```diff
@@ -0,0 +1,183 @@
...
+    data_central: data_dSig_dyttBar.yaml
+    data_uncertainties:
+      - uncertainties_dSig_dyttBar.yaml
+    kinematic_coverage: [y_ttBar, m_t2, sqrts]
+    plotting:
+      dataset_label: 'ATLAS 13 TeV top quark pair in hadronic channel: $\frac{d\sigma}{d|y_{t\bar{t}}|}$'
+      kinematics_override: identity
+      x_scale: log
+      plot_x: y_ttBar
+      y_label: '$\frac{d\sigma}{d|y_{t\bar{t}}|}$ ($\pb$)'
```
Radonirinaunimi, line 93:
```suggestion
      y_label: '$\frac{d\sigma}{d|y_{t\bar{t}}|}$ ($pb$)'
```

Radonirinaunimi (changes requested):
Apart from the typo below, this looks good to me.

Later on, at some point, we should take some decisions regarding the utils from which more and more dataset implementations rely on.

Radonirinaunimi (approved):

Radonirinaunimi:
@t7phy, for future reference, before this PR gets merged, could you please update the report in the description? Thanks.

t7phy:
> @t7phy, for future reference, before this PR gets merged, could you please update the report in the description? Thanks.

done! I will merge once the tests complete.

PR #1975, opened 2024-03-04
Fix multidense compatibility with tf 2.16 and keras 3

APJansen:


scarlehoff:
I have added 3.12 to the tests. It might fail because of other reasons of course (if it runs!), but let's see whether this one is fixed and then revert that commit.

scarlehoff:
When trying it in my computer I got this error though:

```
/NNPDF/src/nnpdf/n3fit/src/n3fit/model_gen.py", line 789, in generate_nn
    pdfs = layer(pdfs)
           ^^^^^^^^^^^
   output_shape = output_shape[:1] + [self.replicas] + output_shape[1:]
                   ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
TypeError: Exception encountered when calling MultiDense.call().
can only concatenate tuple (not "list") to tuple

Arguments received by MultiDense.call():
  • args=('<KerasTensor shape=(1, None, 2), dtype=float32, sparse=None, name=xgrids_processed>',)
  • kwargs=<class 'inspect._empty'>

```

scarlehoff:
fixing that generates further errors I'm afraid

```
ValueError: In a nested call() argument, you cannot mix tensors and non-tensors. Received invalid mixed argument: inputs={'pdf_x': <KerasTensor shape=(1, 1, None, 14), dtype=float32, sparse=False, name=keras_tensor_25>, 'pdf_xgrid_integration': <KerasTensor shape=(1, 1, None, 14), dtype=float32, sparse=False, name=keras_tensor_36>, 'xgrid_integration': <KerasTensor shape=(1, None, 1), dtype=float32, sparse=None, name=integration_grid>, 'photon_integral': array([[[0.]]])}
```

APJansen:
Yes was just about to say that I would expect that...
We are doing many things which are not standard, particularly in `MetaModel` and `MetaLayer`, I would expect a lot of things to break.

I think transitioning to Keras 3 with multiple supported backends would be great, especially when all those backend wrappers would be removed (or at least moved down a level). But timing wise perhaps this is not the best moment..

One thing you can check is what happens with the dense-per-flavour layer, but I think the error you quote above would also arise there.

scarlehoff:
It is not that bad, the problem right now seem to be 

```
NotImplementedError: numpy() is only available when eager execution is enabled
```

I'm not sure where that's happening since the traceback is just the training so we are at some point calling something we are not allowed to call (but it should've failed earlier... it's, at least partially, a tensorflow bug)

APJansen:
numpy is used in `op.scatter_to_one` in `msr_normalization.py`, maybe that's it.

scarlehoff:
It seems to be inside tensorflow/keras. More specifically it is understanding the preprocessing factor variables as numpy arrays. Which it really seems like a bug in tensorflow in that if there's anything wrong with them they should've been caught some time ago.

scarlehoff:
Indeed. The problem is there only when the preprocessing is trainable.

scarlehoff:
In my computer it works for 3.12 but it breaks now for 3.11.

Which I'd say it is ok, now we just have to either have a conditional or fix it in some other way, but it's good :)

Let's see what happens with the rest of the tests.

scarlehoff:
The errors are in the `np.allclose` :)

n3fit/src/n3fit/tests/test_hyperopt.py
```diff
@@ -75,6 +75,7 @@ def test_restart_from_pickle(tmp_path):
 
     for i in range(n_trials_total):
         # check that the files share exactly the same hyperopt history
+        import ipdb; ipdb.set_trace()
```
scarlehoff, line 78:
```suggestion
```

scarlehoff:
Please @APJansen when you have time could you make sure that the changes I made are sensible?

@Cmurilochem I had to change the scope of the hyperopt pickle test to check that indeed when starting from a pickle the next first trial does give you the same parameters as if they had been run sequentially... however the results after the first trial change. It's a sub-% change but that's enough to change the following trials.
I've noticed that tf (or maybe keras) has changed the behaviour of the special case `seed=0` (which now basically means "get some random seed" and broke some of the multidense tests), I wonder whether this is related.

How important is that, after a restart, the hyperopt continues exactly on the same path after the first trial? If it is important it might be necessary to investigate where this difference is coming from before merging (or blocking hyperopt runs to only use tensorflow <2.16 / python <3.12 / numpy < X.YY... not sure which library is at fault here tbh)

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -414,6 +408,17 @@ def load_identical_replicas(self, model_file):
...
+    def save_weights(self, file, save_format="h5"):
+        """
+        Compatibility function for tf < 2.16
+        """
+        try:
+            super().save_weights(file, save_format=save_format)
+        except TypeError:
+            new_file = file.with_suffix(".weights.h5")
+            super().save_weights(new_file)
+            shutil.move(new_file, file)
```
APJansen, line 420: Maybe add a comment here which branch corresponds to which version. Also not sure what happens here, the second branch requires a particular suffix? Or doesn't allow you to overwrite an existing file?

n3fit/src/n3fit/backends/keras_backend/callbacks.py
```diff
@@ -75,9 +76,15 @@ def __init__(self, stopping_object, log_freq=100):
         super().__init__()
         self.log_freq = log_freq
         self.stopping_object = stopping_object
+        self._current_loss = None
+
+    def on_epoch_begin(self, epoch, logs=None):
+        # TODO This is an unnecessary performance hit, just for testing
+        self._current_loss = self.model.compute_losses()
```
APJansen, line 83: I guess this should go out again? Alongs with the `logs = self._current_loss` below?

n3fit/src/n3fit/backends/keras_backend/operations.py
```diff
@@ -249,11 +251,15 @@ def concatenate(tensor_list, axis=-1, target_shape=None, name=None):
...
     """
-    concatenated_tensor = tf.concat(tensor_list, axis, name=name)
-    if target_shape:
-        return K.reshape(concatenated_tensor, target_shape)
-    else:
+    try:
+        # For tensorflow >= 2.16, Keras >= 3
+        concatenated_tensor = keras.ops.concatenate(tensor_list, axis=axis)
+    except AttributeError:
+        concatenated_tensor = tf.concat(tensor_list, axis=axis)
```
APJansen, line 258: General comment about these try, except constructions: if they are used in the model, and you're in the version where the try fails, I can image it leading to some significant slowdown.

APJansen (commented):
Thanks for taking care of this, I left some comments. (can't approve as I started the PR)

Most importantly, have you checked that these try/except's don't occur inside the training loop? Because I imagine this can cause some significant slowdown, if you're in the except branch.

scarlehoff:
>  have you checked that these try/except's don't occur inside the training loop?

It should be compiled away upon first pass but it's true, it might be better to have an if condition to generate the function just in case...

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
scarlehoff: I'll add it as a comment but, the reason is that for keras 3 `save_format` doesn't exist anymore and instead the format is read directly from the extension [docs](https://keras.io/api/models/model_saving_apis/weights_saving_and_loading/#save_weights-method).

So you are forced to save it to a `.weights.h5` file.

goord:
> How important is that, after a restart, the hyperopt continues exactly on the same path after the first trial? If it is important it might be necessary to investigate where this difference is coming from before merging (or blocking hyperopt runs to only use tensorflow <2.16 / python <3.12 / numpy < X.YY... not sure which library is at fault here tbh)

I would say it's not super-important. The restart was mainly there to get around cluster allocation time limits, reproducability is more of a nice to have.

n3fit/src/n3fit/stopping.py
```diff
@@ -496,21 +490,22 @@ def _restore_best_weights(self):
...
+        Prints ``fitstate`` validation chi2 and the current total training loss
+        Note that the
         """
         epoch_index = epoch + 1
-        tr_chi2 = fitstate.total_tr_chi2()
         vl_chi2 = fitstate.total_vl_chi2()
-        total_str = f"At epoch {epoch_index}/{self.total_epochs}, total chi2: {tr_chi2}\n"
+        total_str = f"""Epoch {epoch_index}/{self.total_epochs}: loss: {fitstate.tr_loss:.7f}
+Validation loss after training step: {vl_chi2:.7f}.
+Validation chi2s: """
```
APJansen, line 500: Something went wrong here

APJansen (commented):
Yes ok by me, apart from the one comment.

n3fit/src/n3fit/stopping.py
scarlehoff: what do you mean? (the string is continued below)

scarlehoff: ah... yes... That one was continued only inside my head xD 
Thanks

Cmurilochem:
> Please @APJansen when you have time could you make sure that the changes I made are sensible?
> 
> @Cmurilochem I had to change the scope of the hyperopt pickle test to check that indeed when starting from a pickle the next first trial does give you the same parameters as if they had been run sequentially... however the results after the first trial change. It's a sub-% change but that's enough to change the following trials. I've noticed that tf (or maybe keras) has changed the behaviour of the special case `seed=0` (which now basically means "get some random seed" and broke some of the multidense tests), I wonder whether this is related.
> 
> How important is that, after a restart, the hyperopt continues exactly on the same path after the first trial? If it is important it might be necessary to investigate where this difference is coming from before merging (or blocking hyperopt runs to only use tensorflow <2.16 / python <3.12 / numpy < X.YY... not sure which library is at fault here tbh)

I agree with @goord. But remember that at the time we added some constraints to warrant this, namely:

https://github.com/NNPDF/nnpdf/blob/7babfe217bb40c938a71b55641622cb8adc5bfde/n3fit/src/n3fit/hyper_optimization/hyper_scan.py#L27

and 

https://github.com/NNPDF/nnpdf/blob/7babfe217bb40c938a71b55641622cb8adc5bfde/n3fit/src/n3fit/model_trainer.py#L876-L878

But please feel free to change the scope of the test and go over this.

scarlehoff:
This problem seems to only appear for python3.12 / tensorflow 2.16 so it might be safer to run hyperopt with <3.12 for the time being just in case then...

scarlehoff (approved):
I don't think we can have the conda package yet, but at least we can start using the code in py3.12 (and maybe finding other issues!)

PR #1979, opened 2024-03-04
Polarised proton fits

Radonirinaunimi:
This PR implements the necessary changes to perform polarised fits. It deprecates the intial implementation in https://github.com/NNPDF/nnpdf/pull/1893.

**Remaining Todos:**
- [X] add positivity bound plots to report
- [x] add some tests (Some tests, such as `test_fit`, requires https://github.com/NNPDF/nnpdf/pull/1816 to be merged and then rebased here.)
- [x] generate the files for regression tests (will be regenerated at the very end with CI)
- [X] ~fix positivity dataset plots~ no longer needed as superseded by BC plots


n3fit/src/n3fit/model_gen.py
```diff
@@ -130,13 +130,16 @@ def __call__(self, pdf_layer, mask=None):
...
+    fitbasis,
+    extern_lhapdf,
     mask_array=None,
     training_data=None,
     validation_data=None,
     invcovmat_tr=None,
     invcovmat_vl=None,
     positivity_initial=1.0,
     integrability=False,
+    n_replicas=1,
```
giacomomagni, line 142: This variable seems unused.
```suggestion
```

validphys2/src/validphys/utils.py
```diff
@@ -7,6 +7,7 @@
 
 from frozendict import frozendict
 import numpy as np
+from reportengine.compat import yaml
```
giacomomagni, line 10:
```suggestion
```
It' imported below.

validphys2/src/validphys/pseudodata.py
```diff
@@ -267,6 +268,9 @@ def make_replica(
         # positivity control
         if np.all(shifted_pseudodata[full_mask] >= 0):
             return shifted_pseudodata
+        elif "POL" in fitbasis and trials == (max_tries // 100):
```
giacomomagni, line 271: Why do you need the second condition and you are not simply retuning?

validphys2/src/validphys/utils.py
Radonirinaunimi: Thanks, addressed in [84c6ace](https://github.com/NNPDF/nnpdf/pull/1979/commits/84c6ace122b517379ba0c897f66f1a2fab8ed275)

n3fit/src/n3fit/model_gen.py
Radonirinaunimi: It should be used by the `Obs_layer` below but not passed correctly  due to it having a default. Now it is also fixed in [84c6ace](https://github.com/NNPDF/nnpdf/pull/1979/commits/84c6ace122b517379ba0c897f66f1a2fab8ed275) 
https://github.com/NNPDF/nnpdf/blob/84c6ace122b517379ba0c897f66f1a2fab8ed275/n3fit/src/n3fit/model_gen.py#L221

scarlehoff:
@Radonirinaunimi it would be good to rebase again since the changes to the convolution recently implemented will have an effect here (but they also simplify a lot what needs to be done, in that everything will be much more clear)

validphys2/src/validphys/pseudodata.py
Radonirinaunimi: The 2nd return is only valid in the polarized case in which observable can be negative. In the unpolarized (n)PDF case, given that the observable must always be positive, if is not caught by the first `if` it should fail.

Radonirinaunimi: But ofc, we also need to make sure that the negative obs are not due to random fluctuations, hence still the need for `trials == (max_tries // 100)`.

Radonirinaunimi:
> > @Radonirinaunimi it would be good to rebase again since the changes to the convolution recently implemented will have an effect here (but they also simplify a lot what needs to be done, in that everything will be much more clear)
> 
> Yep, I am doing so at the moment. Hopefully this will not be painful 🙃

It was slightly tricky but I believe it works. I will clean up a bit while adding some docs.

Radonirinaunimi:
@scarlehoff I think this is now ready if you want to have a look.

validphys2/src/validphys/pseudodata.py
scarlehoff, line 271: The "if POL" in fitbasis scattered around the code make me nervous.

Could you do instead something like `fitbasis.startwith("POL_")` or whatever so that it becomes a special key?

Otherwise at some point someone might do "POLYNOMIAL" or whatever and will be very confused ^^U

scarlehoff (commented):
a first batch of comments

I like a lot how the fit can change a lot with relatively very few changes to the actual fitting routines ^^

doc/sphinx/source/tutorials/polarized_fits.rst
```diff
@@ -0,0 +1,122 @@
...
+
+.. note::
+
+   While the treatment of integrability follows exactly the same concept as in the 
+   default unpolarized fits, the treatment of positivity in the polarized case 
+   requires specific treatment. Similar to the unpolarised fits, the cost function 
+   used to enforce the positivity is defined by the following observable:
+  
+   .. math::
+     \chi_{\mathrm{tot}}^2 \rightarrow \chi_{\mathrm{tot}}^2+ \Lambda_{\rm POS} \sum_{k=1}^8 \sum_{i=1}^{n_i} \operatorname{ReLU}\left(-\mathcal{C}_k\left(x_i, Q^2\right)\right)
```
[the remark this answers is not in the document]
Radonirinaunimi: It is really $n_i$ as it sums over the number of $x$-grid points. That equation has exactly the same structure as Eq. (3.10) of the NNPDF4.0 paper. But it is true it can be called anything ($n_x$ for example) but $n_k$.

Btw, thanks for the review @giacomomagni, the other comments have been addressed in [1eb11df](https://github.com/NNPDF/nnpdf/pull/1979/commits/1eb11dfca21395acac8736603bbc60aca2c35f56).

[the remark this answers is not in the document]
giacomomagni: Okay, thank for the clarification. I'd call it $n_x$, it's more clear and correct as $i$ is the variable of the sum.

n3fit/src/n3fit/backends/keras_backend/operations.py
```diff
@@ -290,6 +290,21 @@ def pow(tensor, power):
...
 
+@tf.function
+def absolute(tensor):
+    """
+    Compute the absolute value of a tensor
+    """
+    return K.abs(tensor)
+
+
+def multiply_minusone(tensor):
```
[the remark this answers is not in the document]
Radonirinaunimi: Thanks @scarlehoff for the comments! I will first try to reply to your questions and then will address the rest in some updated commits.

> You are already implementing a new operation, you can have the -abs() there if you really need to.

That was indeed what I tried at the very beginning but then got some errors because of type errors. But I agree that it will be great to have something like this and remove these functions. I will have another look again and try to fix the issues (maybe things have changed since).

n3fit/src/n3fit/model_gen.py
```diff
@@ -88,7 +89,8 @@ def _generate_loss(self, mask=None):
                 invcovmat_matrix, self.data, mask, covmat=covmat_matrix, name=self.name
             )
         elif self.positivity:
-            loss = losses.LossPositivity(name=self.name, c=self.multiplier)
+            alpha = 0.0 if self.polarized else 1e-7  # ELU -> RELU for Polarized Fits
```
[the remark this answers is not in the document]
Radonirinaunimi: Because what we want is $\mathcal{O}_{\mathrm{POS}} \equiv |\Delta  f_i(x,Q^2)| - f_i(x,Q^2)  \le 0$ and therefore negative values shouldn't contribute to the penalty.

n3fit/src/n3fit/layers/observable.py
```diff
@@ -101,13 +167,41 @@ def call(self, pdf):
...
 
         observables = self.operation(observables)
         return observables
 
+    def is_polarised_posdata(self):
+        """Check if the dataset is a Polarised Positivity dataset."""
+        # TODO: Better move the following to the new commondata input
+        if "POL" in self.fitbasis and "_POS_" in self.dataset_name:
+            # Polarised POS contains at least 2 FK tables
+            return self.nfks >= 2
```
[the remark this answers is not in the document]
Radonirinaunimi: Yes, the positivity attribute is contained in the FK tables. However, we always need to check two conditions: (a) that the positivity dataset is polarized and (b) that the FK table is also polarized. The second condition is required because a polarized positivity dataset also contains **non-** polarized FK tables.

This is why I think that having the information that the POS dataset is polarized in the metadata would simplify things a lot. If you think that can be a solution I will extend the new commondata parser to read an optional key from the metadata?

validphys2/src/validphys/pseudodata.py
Radonirinaunimi: Yep, I will do so xD

Radonirinaunimi: Maybe there is even a more elegant way to pass this information around.

n3fit/src/n3fit/model_gen.py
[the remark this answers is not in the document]
scarlehoff: But having ELU just avoids a sharp cut at 0.0.
If the difference is relevant I'd be worried... but in any case, the way to go would be making the positivity_alpha or whatever a runcard parameter.

n3fit/src/n3fit/layers/observable.py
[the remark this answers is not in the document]
scarlehoff: But that it is a positivity dataset is given by `self.dataset_name.startwith("NNPDF_POS")` and whether this is polarized should be `fktable.is_polarized()` (or whatever) but should not the fitbasis.

This will fail for instance if we happen to add non polarized composed positivities.

validphys2/src/validphys/pseudodata.py
scarlehoff: Yes!

The "right" way to do it, or, more precisely, the way we have been doing it until now, is to define some processes which allow for negative values in the data.

For instance, at the moment is only the `ASY` types. You can (you should, actually!) add a new process type to the library of process options for the polarized dataset that will allow for this. The check is done a bit above in `if ASY in commondataproc`.

n3fit/src/n3fit/backends/keras_backend/operations.py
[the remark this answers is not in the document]
Radonirinaunimi: Things have changed indeed and now this works!! Not only this removes the unneeded functions but also simplifies the `observable` a lot.

validphys2/src/validphys/convolution.py
```diff
@@ -70,6 +70,25 @@ def _asy(a, b):
     return (a - b) / (a + b)
 
 
+def _addp(a, b):
+    """Compute positivity constraints for polarized gluon. The unpolarized
+    and polarized predictions have to follow a particular order:
+
+    Parameters:
+    -----------
+    a: tf.tensor
```
scarlehoff, line 79: This should work with any type of array, not only tf (and indeed, it does, otherwise vp will not be able to use it!)

validphys2/src/validphys/convolution.py
```diff
@@ -91,6 +134,8 @@ def _id(a):
     "ASY": _asy,
     "ADD": operator.add,
     "SMN": _smn,
+    "SMP": _smp,
+    "ADDP": _addp,
```
scarlehoff, line 138: I hadn't looked at the data yet, but please use some other name.
In particular `ADDP` is a very weird name for `b - abs(a)`

Radonirinaunimi: Do you have some suggestions? I basically tried to follow what's there although admittedly inconsistently `SMN->SMP` and `ADD->ADDP` where `P` always stands for Polarized.

scarlehoff: `SUBTRACT_ABS` would be more correct that `ADDP` imho ^^U
the operation doesn't need to know anything about the polarized situation

Radonirinaunimi: Ok, do you like more the new names?
https://github.com/NNPDF/nnpdf/blob/043e974b803b74bcd4058eaf3e7e93015648df0d/validphys2/src/validphys/convolution.py#L140-L141

I've also re-ordered completely the FK tables here (#1816) such that we have `a - abs(b)` and `a + b - abs(c + d)` as the previous ones were odds as well.

Also, the arguments types are fixed. 

The other comments I will address in one push (with different commits ofc).

doc/sphinx/source/tutorials/polarized_fits.rst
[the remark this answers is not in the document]
Radonirinaunimi: Done now!

n3fit/src/n3fit/model_gen.py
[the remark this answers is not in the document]
Radonirinaunimi: But this is exactly what you want in the polarized case and seen to be empirically improve the convergences of the fits. But I agree that the best is to allow for such parameter to be defined in the runcard (this would requires a few modifications though).

n3fit/src/n3fit/layers/x_operations.py
```diff
@@ -22,23 +22,34 @@ class xDivide(MetaLayer):
...
+        output_dim: int = BASIS_SIZE,
+        fitbasis: str = "NN31IC",
+        div_list: Optional[List[int]] = None,
+        **kwargs
     ):
-        if div_list is None:
+        if div_list is None:  # Default Unpolarized Case
             div_list = [3, 4, 5, 6]
+        div_list = [9, 10] if "POL" in fitbasis else div_list
+
```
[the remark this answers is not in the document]
Radonirinaunimi: Maybe I am being very slow here but I don't think this would work though because atm `div_list` is always `None` and therefore an `elif` condition will always be skipped. And as you said, it should be always honored regardless of the basis. But we can add another condition:
```python
if div_list is None and fitbasis.startswith("POLARIZED_"):
    div_list = [9, 10]
```
A `match/case` statement would be nicer but then not available on py3.9 ...

validphys2/src/validphys/pseudodata.py
Radonirinaunimi: For this, should I cal the new process `DIS_POL` such that everything such as `_get_ported_kinlabel` and `_get_process_description` still works out of the box? In this way, the changes will be minimal.

Radonirinaunimi:
Ok, @scarlehoff, I believe I addressed all of your comments except for the ones I commented above. I have marked the obvious ones "resolved" but if you are happy with the changes on those without a comment could you also please mark them as resolved.

n3fit/src/n3fit/layers/x_operations.py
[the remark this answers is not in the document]
scarlehoff: Then just change the order:

```python
if div_list is None and "POL" in fitbasis:
    div_list = [9, 10]
elif div_list is None;
    div_list = [3,4,5,6]
```

This achieves the same thing and the user input is not modified silently.

n3fit/src/n3fit/model_gen.py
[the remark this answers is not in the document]
scarlehoff: If you do the fit with the default 1e-7 it doesn't work?

The idea of ELU and 1e-7 is that it was a numerical 0 so it doesn't affect the loss/chi2 but at the same time avoids a sharp cut at 0. We are using float32 for the fit so 1e-7 should be irrelevant.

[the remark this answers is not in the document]
Radonirinaunimi: This actually I should try again. Maybe things have changes with all the fixes and refactoring.

validphys2/src/validphys/pseudodata.py
Radonirinaunimi: This btw has been addressed here [0b0a925](https://github.com/NNPDF/nnpdf/pull/1816/commits/0b0a92522b0d79eb766d8dbd4764ce4e6c1f56c1) so I will remove that `elif` block.

n3fit/src/n3fit/model_gen.py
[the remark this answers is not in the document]
Radonirinaunimi: I turned out that this now works (there might have been some combinations of issues before). I have now removed it completely and the need to pass `fitbasis` everywhere.

extra_tests/regression_fits/polarized_evol.yml
```diff
@@ -0,0 +1,82 @@
...
+theory:
+  theoryid: 822
+
+############################################################
+trvlseed: 14613416
+nnseed: 1865683875
+mcseed: 746898422
+genrep: true
+
+save: "weights_polevol.h5"
```
scarlehoff, line 34:
```suggestion
save: "weights_polevol.weights.h5"
```

Are you sure you want to `save` here and not load a previously saved state? (note that there are already other two cards that test the save)

n3fit/src/n3fit/checks.py
```diff
@@ -448,3 +445,18 @@ def check_multireplica_qed(replicas, fiatlux):
...
+    if fitbasis.startswith("POLARIZED_"):
+        if positivity_bound is None:
+            raise CheckError(
+                "For polarized fits, the 'positivity_bound' key has to be defined in the runcard."
+            )
+        if positivity_bound.get("unpolarized_bc") is None:
+            raise CheckError(
+                "For polarized fits, the name of the PDF has to be defined in positivity_bound::unpolarized_bc."
+            )
+        if fitting.get("sum_rules") != "TSR":
```
scarlehoff, line 461:
```suggestion
        if fitting.get("sum_rules", True) and fitting.get("sum_rules") != "TSR":
```

`sum_rules: False` might be useful for testing.

extra_tests/regression_fits/polarized_evol.yml
Radonirinaunimi: I think this is already outdated. In the recent commits, this has been changed to load.

n3fit/src/n3fit/checks.py
Radonirinaunimi: Ah, that might be a solution indeed. Let me try.

validphys2/src/validphys/convolution.py
```diff
@@ -112,13 +112,13 @@ def _subtract_abspair(a, b, c, d):
...
+        Unpolarized array-like object
     b: np.ndarray
-        polarized tensor-like object
+        Unpolarized array-like object
     c: np.ndarray
-        Unpolarized tensor-like object
+        polarized array-like object
     d: np.ndarray
-        Unpolarized tensor-like object
+        polarized array-like object
```
scarlehoff, line 121: Please, remove the strong references to the polarize (I put a suggestion above as an example) in this and the dosctring above. In principle these functions, even if the target is polarized fits, are generally valid (and for sure, the input itself is not polarized).

(I think there is value on making it clear that the expectation is unp unp pol pol, but a "numpy unpolarized array" might a bit confusing!)

validphys2/src/validphys/dataplots.py
```diff
@@ -921,6 +923,106 @@ def plot_replica_sum_rules(pdf, sum_rules, Q):
...
+    """
+    Plot the correlated uncertainties for the truncated integrals of the polarized
+    gluon and singlet distributions.
+    """
+    from matplotlib.patches import Ellipse
+
+    def compute_hists(preds, nbins=10):
+        binning = np.linspace(preds.min(), preds.max(), num=nbins)
+        frequencies, bins = np.histogram(preds, bins=binning, density=True)
+        return {"x": bins[:-1], "bins": bins, "weights": frequencies}
```
scarlehoff, line 937: This function could (and should) be outside, as `_compute_hists` with a docstr explaining what it does.

The same is true for the function below, `_compute_ellipse`.

validphys2/src/validphys/dataplots.py
```diff
@@ -921,6 +923,106 @@ def plot_replica_sum_rules(pdf, sum_rules, Q):
     return fig
 
 
+@figure
+def plot_polarized_momentum(pdf, Q, xmin=0.001, angular_momentum=False):
+    """
+    Plot the correlated uncertainties for the truncated integrals of the polarized
+    gluon and singlet distributions.
+    """
+    from matplotlib.patches import Ellipse
```
scarlehoff, line 932: Why is this import here?

validphys2/src/validphys/dataplots.py
```diff
@@ -921,6 +923,106 @@ def plot_replica_sum_rules(pdf, sum_rules, Q):
...
+            "xy": (preds_x.mean(), preds_y.mean()),
+            "width": 2.0 * nstd * sqrt_eigval[0],
+            "height": 2.0 * nstd * sqrt_eigval[1],
+            "angle": np.degrees(np.arctan2(*eigvec[:, 0][::-1])),
+            "fill": False,
+            "linewidth": 2.0,
+            "color": "C0",
+        }
+
+    predictions = polarized_sum_rules(pdf, Q, lims=[(xmin, 1)])
```
scarlehoff, line 954: Mabye there is something I'm missing here, but shouldn't it be better if `polarized_sum_rules` where to return both the very low x and the normal values all the time? The computation is cheap enough that I think it would be worthwile, that way it can be lifted from here to the arguments of the function and leave reportengine deal with it.

(take it as a... separate suggestion with respect to the others because I'm not sure it would deal with it _correctly_ and this plotting function is self-contained enough that I'd be willing to accept these 2 calls to `polarized_sum_rules`)

validphys2/src/validphys/dataplots.py
```diff
@@ -921,6 +923,106 @@ def plot_replica_sum_rules(pdf, sum_rules, Q):
...
+    }  # Parameters that define the subplots
+    fig, ((ax_histx, ax_empty), (ax_scatr, ax_histy)) = plotutils.subplots(**params)
+    fig.subplots_adjust(wspace=1e-4, hspace=1e-1)
+
+    # Add the scatter plots
+    ax_scatr.scatter(xpreds, ypreds, label=pdf.label)
+    ax_scatr.scatter(xpreds.mean(), ypreds.mean(), marker="s", c="red")
+
+    # Add the Ellipse plot
+    ellipse = Ellipse(**compute_ellipse(xpreds, ypreds))
```
scarlehoff, line 980: If `compute_ellipse` is only used here, I think it would be better for the external function `_compute_ellipse` to return directly an `Ellipse` object.

validphys2/src/validphys/dataplots.py
```diff
@@ -921,6 +923,106 @@ def plot_replica_sum_rules(pdf, sum_rules, Q):
...
+            "linewidth": 2.0,
+            "color": "C0",
+        }
+
+    predictions = polarized_sum_rules(pdf, Q, lims=[(xmin, 1)])
+    if not angular_momentum:
+        xpreds = np.array(predictions["g"])
+        ypreds = np.array(predictions["singlet"]) / 2.0
+    else:
+        preds_low_x = polarized_sum_rules(pdf, Q, lims=[(1e-4, xmin)])
```
scarlehoff, line 959:
```suggestion
        preds_low_x = polarized_sum_rules(pdf, Q, lims=[(xmin*1e-1, xmin)])
```

Take into account that `xmin` can be controlled by the external runcard. You have to either write the lower limit here relative to `xmin` or just error out if the user asks for an `xmin` below a certain threshold.

validphys2/src/validphys/pdfplots.py
```diff
@@ -601,6 +602,25 @@ def legend(self, flstate):
...
+class BandPDFPlotterBC(BandPDFPlotter):
+    def __init__(self, *args, unpolarized_bcs, boundary_xplotting_grids, **kwargs):
+        self.unpolarized_bcs = unpolarized_bcs
+        self.boundary_xplotting_grids = boundary_xplotting_grids
+        super().__init__(*args, **kwargs)
+
+    def draw(self, pdf, grid, flstate):
+        xgrid = grid.xgrid
+        ax = flstate.ax
+        # TODO: Maybe not always select the first element in the list
```
scarlehoff, line 614: I'm a bit confused about this TODO. If the other members are not used, why are they there at all?

If `boundary_xplotting_grids` should be the same size as the list of PDF being plotted (it's not clear to me) you can make it an iterator in the constructor and get the next every time you go through draw (you should pass through here as many times as pdfs are there to be drawn if I'm not mistaken)

validphys2/src/validphys/pseudodata.py
```diff
@@ -129,6 +130,7 @@ def make_replica(
     replica_mcseed,
     dataset_inputs_sampling_covmat,
     sep_mult,
+    fitbasis="NN3IC",
```
scarlehoff, line 133: ?

n3fit/src/n3fit/layers/observable.py
```diff
@@ -15,6 +17,44 @@ def is_unique(list_of_arrays):
...
 
+def compute_pdf_boundary(pdf, q0_value, xgrid, n_std, n_replicas):
+    """
+    Computes the Unpolarized PDF set to be used as a Boundary Condition and
+    convert it into a Tensor object that can be understood by the convolution.
+
+    Parameters
+    ----------
+    pdf: validphys.core.PDF
+        a validphys instance of the Unpolarized PDF set
```
scarlehoff, line 28:
```suggestion
        a validphys PDF instance to be used as a boundary PDF set
```

I think this same function (maybe a generalized version) will be useful also going forward for the final universal qcd fit well beyond the polarized case

n3fit/src/n3fit/layers/observable.py
```diff
@@ -15,6 +17,44 @@ def is_unique(list_of_arrays):
...
+    xpdf_cvs = xpdf_obj.grid_values.central_value().T
+    xpdf_std = xpdf_obj.grid_values.std_error().T
+
+    # Computes the shifted Central Value as given by `n_std`
+    xpdf_bound = xpdf_cvs + n_std * xpdf_std
+
+    # Expand dimensions for multi-replicas and convert into tensor
+    mult_resx = np.repeat([xpdf_bound], n_replicas, axis=0)
+    add_batch_resx = np.expand_dims(mult_resx, axis=0)
+    return op.numpy_to_tensor(add_batch_resx)
```
scarlehoff, line 55: Please add a test for this function where you call it with a PDF and check on the other side that the result has the right shape with the values in the right positions.

n3fit/src/n3fit/layers/observable.py
```diff
@@ -39,20 +79,41 @@ class Observable(MetaLayer, ABC):
...
     """
 
-    def __init__(self, fktable_data, fktable_arr, operation_name, nfl=14, **kwargs):
+    def __init__(
+        self,
+        fktable_data,
+        fktable_arr,
+        dataset_name,
+        boundary_condition,
+        operation_name,
```
scarlehoff, line 88:
```suggestion
        boundary_condition=None,
        operation_name="NULL",
```

n3fit/src/n3fit/layers/observable.py
```diff
@@ -39,20 +79,41 @@ class Observable(MetaLayer, ABC):
...
         all_bases = []
         xgrids = []
         fktables = []
-        for fkdata, fk in zip(fktable_data, fktable_arr):
+        for idx, (fkdata, fk) in enumerate(zip(fktable_data, fktable_arr)):
             xgrids.append(fkdata.xgrid.reshape(1, -1))
             all_bases.append(fkdata.luminosity_mapping)
             fktables.append(op.numpy_to_tensor(fk))
+
+            if self.is_pos_polarized() and not fkdata.is_polarized:
```
scarlehoff, line 109:
```suggestion
            if self.is_pos_polarized() and not fkdata.is_polarized:
                if boundary_condition is None:
                    raise ValueError("Polarized FKTables require a boundary condition")
```

n3fit/src/n3fit/model_gen.py
```diff
@@ -129,13 +129,15 @@ def __call__(self, pdf_layer, mask=None):
 
 def observable_generator(
     spec_dict,
+    boundary_condition,
```
scarlehoff, line 132:
```suggestion
    boundary_condition=None,
```

scarlehoff (commented):
I think this is good to go. Thank you for the hard work! I like how they are changes that are relatively major for how the fit works but are small nudges to the code ^__^

The only thing left for me to be happy would be having a fit + report with this latest version of the code (after comments are applied or ignored in case-by-case basis xd) before merging

Radonirinaunimi:
> The only thing left for me to be happy would be having a fit + report with this latest version of the code (after comments are applied or ignored in case-by-case basis xd) before merging

I already read your mind earlier and the evolutions are running atm xd Will generate some reports tomorrow while addressing your suggestions.

Radonirinaunimi:
The reports are now failing because of some recent changes in the computation of stats that appear following lines:
https://github.com/NNPDF/nnpdf/blob/aeb2a96048200b02ca3a9bb75c00ed0906b73db2/validphys2/src/validphys/pdfplots.py#L617-L619

I am fixing this now.

scarlehoff:
What changes?
But in any case, that function is the only non-trivial so if you can deal also with the TODO that would be perfect.

Radonirinaunimi:
> What changes? But in any case, that function is the only non-trivial so if you can deal also with the TODO that would be perfect.

It is not actually related to the computation of stats but to the way the operation $+$ is now behaving. Now, doing `unpol_cv + unpol_stdv` concatenate the two arrays instead of adding them.

scarlehoff:
That should only happen if you are getting a list and not an array, and that should definitely not happen because the `data` is made into an array upon entering the `Stats` class.

https://github.com/NNPDF/nnpdf/blob/aeb2a96048200b02ca3a9bb75c00ed0906b73db2/validphys2/src/validphys/core.py#L828

So there might be something wrong somewhere else?

validphys2/src/validphys/pdfplots.py
Radonirinaunimi: This has been just taken into in [88a7b25](https://github.com/NNPDF/nnpdf/pull/1979/commits/88a7b25f3adede4b00111c0d008354693a73f8ad) but I could briefly explain how this ended up there. Usually, when comparing two fits, these two fits are produced with the same BC meaning that `unpolarized_bcs` will contain two elements with the exact same thing. But all the variations should now be taken into account in [88a7b25](https://github.com/NNPDF/nnpdf/pull/1979/commits/88a7b25f3adede4b00111c0d008354693a73f8ad).

validphys2/src/validphys/pseudodata.py
Radonirinaunimi: I guess, here, you are mainly referring to https://github.com/NNPDF/nnpdf/issues/2041?

validphys2/src/validphys/dataplots.py
Radonirinaunimi: Given that this is only used in polarized case, I thought it would be better if this is not imported at top level.

validphys2/src/validphys/pseudodata.py
scarlehoff: I don't understand why this change is necessary, it is not used in the function is it?

validphys2/src/validphys/dataplots.py
scarlehoff: It is better to import things at the top level, unless it is a specific library that you know can introduce some problems.

validphys2/src/validphys/pseudodata.py
Radonirinaunimi: Ah, sorry! That is indeed no longer needed. It was there before to allow negative pseudodata for POL but now it is encoded in the `process_type`. I will remove it in the next batch while taking care of the other comments.

validphys2/src/validphys/dataplots.py
scarlehoff: Sure.
As far as I can see you have two calls to `polarized_sum_rules(pdf, Q, lims=[(xmin, 1)])` and `polarized_sum_rules(pdf, Q, lims=[(1e-4, xmin)])`. 
I am wondering whether it makes sense to make `polarized_sum_rules` to always compute the sum rules for `(xmin, 1)` and `(xmin*1e-1, xmin)`. (maybe it doesn't if this is used in other places).

As a separate note, I think it would be better to have `polarized_sum_rules` as an argument to the function, so you will get already the value of `polarized_sum_rules(pdf, Q, lims=[(xmin, 1)])` as computed by validphys.
Then if you need to compute also `polarized_sum_rules(pdf, Q, lims=[(1e-4, xmin)])` that is a separate call.

But as I said, I'm willing to accept it as it is if that would complicate things, the only problematic thing is the issue I've mentioned about `xmin`.

Radonirinaunimi: I think there is a nicer compromise in [6edffdb](https://github.com/NNPDF/nnpdf/pull/1979/commits/6edffdbc201843a947a5f89ee854e07c03bb8599), let me know if you like this better xd

scarlehoff: If this works (namely, doesn't break things elsewhere) this is ok for me.

(if it does, a good middle way is to perform the call explicitly in the functions that call the plotting)

scarlehoff (approved):
The fitbot corresponding to the last commit https://github.com/NNPDF/nnpdf/pull/2057#issuecomment-2064766439

PR #1980, opened 2024-03-05
Remove cmake, make the conda recipe architecture independent

scarlehoff:
Is this works I'd be so happy.

This is a necessary step for #1965. _if the test pass here_ rebasing on top of this could (I hope) make your tests pass as well without further changes @comane


pyproject.toml
```diff
@@ -43,6 +43,8 @@ include = [
 n3fit = "n3fit.scripts.n3fit_exec:main"
 validphys = "validphys.scripts.main:main"
 # Fitting scripts
+evolven3fit = "n3fit.scripts.evolven3fit_new:main"
+# Keep the `new` suffix so people's scripts don't break
```
RoyStegeman, line 47: Is it possible to add a deprecation warning?

RoyStegeman: Maybe it makes more sense to do so in a different PR

RoyStegeman (approved):

pyproject.toml
comane, line 47: But does this work?
I have, until now, never used `evolven3fit_new`, but I tried running it now using the same convention as `evolven3fit` and it does not seem to work:

`evolven3fit_new <name_fit> <num_replicas>`

gives this error:
evolven3fit_new: error: argument actions: invalid choice: '240210_mnc_dis_ict_lam04' (choose from 'produce_eko', 'produce_eko_photon', 'evolve')

RoyStegeman: Yes because the command is `evolven3fit_new evolve <name_fit>`. See also `evolven3fit_new --help`

comane: Yes, what I mean is that there is this comment saying `keep suffix ...`
However, the scripts are going to be broken anyways since the old evolven3fit had a different convention right?

scarlehoff: With that comment I was thinking about people with scripts using `evolven3fit_new`. I didn't expect people still using the old one as this point...

RoyStegeman: Note that you may wish to use theories that have been generated with eko in combination with evolven3fit_new, i.e. theories 700+. Otherwise computation will be a bit slow, or you end up with the fktables being evolved with apfel but the PDF afterwards with eko

comane: Ah ok, thanks.
I mean for the inconsistent closure tests we have always used theory 200, so probably there we should continue using evolven3fit (old)?

PR #1982, opened 2024-03-05
Polish analysis of a multiclosure test

comane:
Here we collect new functions and template that allow the analysis of multiclosure tests.
@comane @giovannidecrescenzo @mariaubiali @andreab1997


validphys2/src/validphys/kinematics.py
```diff
@@ -96,6 +113,30 @@ def kinlimits(commondata, cuts, use_cuts, use_kinoverride: bool = True):
 
 all_kinlimits = collect(kinlimits, ('dataset_inputs',))
 
+def xq2_dataset_map(commondata, cuts,internal_multiclosure_dataset_loader,
+                        _internal_max_reps=None,
+                        _internal_min_reps=20):
+    """
+    Load in a dictionary all the specs of a dataset meaning:
```
comane, line 120: @giovannidecrescenzo, @andreab1997  I think that this stuff should stay in `multiclosure_inconsistent.py` or `multiclosure_inconsistent_output.py`

comane: With this stuff, I mean all of the single datapoint ratio bias variance analysis

comane (commented):
I think that the PR is ready for review.
The only things I would change as commented above is that the single data point ratio bias variance analysis does not reside in `kinematics.py` but rather in `multiclosure_inconsistent.py` / `multiclosure_inconsistent_output.py`

andreab1997:
@comane thanks for this. If you want, ask myself for the review (but also add one between @scarlehoff and @RoyStegeman given that part of this PR is based on my work and it does not make sense for me to review my own functions )

validphys2/src/validphys/scripts/vp_multiclosure.py
```diff
@@ -0,0 +1,159 @@
...
+import prompt_toolkit
+from prompt_toolkit.completion import WordCompleter
+
+from reportengine.compat import yaml
+from reportengine.colors import t
+
+from validphys.app import App
+from validphys.loader import RemoteLoader
+from validphys import compareinconsistentclosuretemplates
+from validphys.promptutils import confirm, KeywordsWithCache
```
comane, line 15:
```suggestion
import sys
import logging

from reportengine.compat import yaml

from validphys.app import App
from validphys import compareinconsistentclosuretemplates
```

These seem all unused

validphys2/src/validphys/compareinconsistentclosuretemplates/multiclosure_analysis.yaml
```diff
@@ -0,0 +1,242 @@
...
+    speclabel: "Current Fit"
+  
+  reference:
+    fit: {id: "240210_mnc_dis_ict_lam04"}
+    pdf: {id: "240210_mnc_dis_ict_lam04", label: "Reference Fit" }
+    theory:
+      from_: fit
+    theoryid:
+      from_: theory
+    speclabel: "Reference Fit"
```
comane, line 40: Hi @andreab1997 , a weird thing here is that we can basically compare as many fits (see `lambdavalues` below) as we want in terms of pca_template, single_point_template, etc..

However, correct me if I am wrong, the part of this script that does `vp-comparefits` only compares two.

So if the above is right, I think that it would be nicer to have a `multi vp-comparefits` (meaning that more than 2 fits can be compared in principle) feature for this script.

validphys2/src/validphys/kinematics.py
```diff
@@ -12,9 +12,16 @@
...
+
+
 from validphys.core import CutsPolicy
 from validphys.plotoptions import core as plotoptions_core
 
+
+
+
+
+
```
comane, line 24: this module should probably not be touched at all

validphys2/src/validphys/closuretest/multiclosure.py
```diff
@@ -147,6 +149,45 @@ def fits_dataset_bias_variance(
...
+        bias_diffs = np.mean(reps[i], axis = 1) - law_th.central_value
+        # bias diffs in the for loop should have dim = (n_obs)
+        sigmas = np.sqrt(np.var(reps[i], axis = 1))
+        # var diffs has shape (n_obs,reps)
+        bias_diffs = bias_diffs/sigmas
+        
+        deltas.append(bias_diffs.tolist())
+        # biases.shape = (n_fits, n_obs_cut/uncut)
+        # variances.shape = (n_fits, n_obs_cut/uncut, reps)
+    return np.asarray(deltas)
```
comane, line 184:
```suggestion
    closures_th, law_th, _, _ = internal_multiclosure_dataset_loader
    # The dimentions here are (fit, data point, replica)
    reps = np.asarray([th.error_members[:, :_internal_max_reps] for th in closures_th])
    # One could mask here some reps in order to avoid redundancy of information
    #TODO

    n_fits = np.shape(reps)[0]
    deltas = []
    # There are n_fits pdf_covariances
    # flag to see whether to eliminate dataset
    for i in range(n_fits):
        bias_diffs = np.mean(reps[i], axis = 1) - law_th.central_value
        # bias diffs in the for loop should have dim = (n_obs)
        sigmas = np.sqrt(np.var(reps[i], axis = 1))
        # var diffs has shape (n_obs,reps)
        bias_diffs = bias_diffs/sigmas
        
        deltas.append(bias_diffs.tolist())
        # biases.shape = (n_fits, n_obs_cut/uncut)
        # variances.shape = (n_fits, n_obs_cut/uncut, reps)
    return np.asarray(deltas)
```

These don't seem to be used anywhere

validphys2/src/validphys/closuretest/multiclosure.py
```diff
@@ -147,6 +149,45 @@ def fits_dataset_bias_variance(
...
+@check_multifit_replicas
+def fits_normed_dataset_central_delta(
+    internal_multiclosure_dataset_loader,
+    _internal_max_reps=None,
+    _internal_min_reps=20):
+    """
+    For each fit calculate the difference between central expectation value and true val. Normalize this
+    value by the variance of the differences between replicas and central expectation value (different
+    for each fit but expected to vary only a little). Each observable central exp value is 
+    expected to be gaussianly distributed around the true value set by the fakepdf
```
comane, line 161: would be good to have some numpydoc like docstrings

validphys2/src/validphys/closuretest/multiclosure_output.py
```diff
@@ -794,3 +853,21 @@ def plot_bias_variance_distributions(
     ax.legend()
     ax.set_title("Total bias and variance distributions.")
     yield fig
+
+@figuregen
+def xq2_data_prcs_maps(xq2_data_map):
+    keys = ["std_devs","xi"]
+    for elem in xq2_data_map:
```
comane, line 860: some small description explaining what this does could be nice to have

comane (commented):
Ciao @andreab1997 , I added some minor comments on some of the parts of the code that you and Giovanni wrote.

When you have the time, feel free to add a review to the `inconsistent_closuretest` folder

validphys2/examples/pca_bias_variance_report.yaml
```diff
@@ -0,0 +1,50 @@
+meta:
+  title: PCA report for bias and variance of multiclosure fits.
+  author: Mark N. Costantini
+  keywords: [multiclosure, PCA, bias, variance]
+
+
+dataset_inputs:
+  - {dataset: ATLAS_SINGLETOP_7TEV_TCHANNEL-XSEC, variant: legacy}
```
andreab1997, line 8: Why are you using the legacy versions? Is that necessary?

validphys2/src/validphys/closuretest/inconsistent_closuretest/multiclosure_inconsistent.py
```diff
@@ -0,0 +1,209 @@
...
+    Returns
+    -------
+    PCAInternalMulticlosureLoader
+    """
+    closures_th, law_th, _, _ = internal_multiclosure_dataset_loader
+
+    reps = np.asarray([th.error_members for th in closures_th])
+
+    # compute the covariance matrix of the theory predictions for each fit
+    _covmats = [np.cov(rep, rowvar=True, bias=True) for rep in reps]
```
andreab1997, line 95: I am wondering if it could be better to use `np.array` instead of lists

validphys2/src/validphys/closuretest/inconsistent_closuretest/multiclosure_inconsistent.py
```diff
@@ -0,0 +1,209 @@
...
+        return PCAInternalMulticlosureLoader(
+            closures_th=closures_th,
+            law_th=law_th,
+            pc_basis=1,
+            n_comp=1,
+            covmat_pca=_covmat_mean,
+            sqrt_covmat_pca=np.sqrt(_covmat_mean),
+        )
+
+    eighvals, eigvecs = np.linalg.eigh(_covmat_mean)
```
andreab1997, line 113: Maybe this block can be put in another function since I guess it could be reused

validphys2/src/validphys/closuretest/inconsistent_closuretest/multiclosure_inconsistent.py
```diff
@@ -0,0 +1,209 @@
...
+    closures_th: list
+    law_th: ThPredictionsResult
+    pc_basis: np.array
+    n_comp: int
+    covmat_pca: np.array
+    sqrt_covmat_pca: np.array
+
+
+@check_multifit_replicas
+def internal_multiclosure_dataset_loader_pca(
```
andreab1997, line 57: In general this function is too big for my taste, maybe it could be splitted in some more unit functions

validphys2/src/validphys/closuretest/inconsistent_closuretest/multiclosure_inconsistent_output.py
```diff
@@ -0,0 +1,227 @@
...
+
+@figuregen
+def plot_l2_condition_number(
+    each_dataset, internal_multiclosure_data_collected_loader, evr_min=0.90, evr_max=0.995, evr_n=20
+):
+    """
+    Plot the L2 condition number of the covariance matrix as a function of the explained variance ratio.
+    The plot gives an idea of the stability of the covariance matrix as a function of the
+    exaplained variance ratio and hence the number of principal components used to reduce the dimensionality.
+
```
andreab1997, line 158:
```suggestion

```

validphys2/src/validphys/closuretest/multiclosure_output.py
```diff
@@ -60,6 +60,65 @@ def plot_total_fits_bias_variance(fits_total_bias_variance):
...
+        mean_bias = np.mean(bias)
+        mean_var = np.mean(var)
+        rbv = mean_bias / mean_var
+
+        # compute uncertainy on rbv
+        delta_rbv = np.sqrt(
+            ((1 / mean_var) * np.std(bias)) ** 2 + (mean_bias / mean_var**2 * np.std(var)) ** 2
+        )
+        sqrt_rbv = np.sqrt(rbv)
+        delta_sqrt_rbv = 0.5 * delta_rbv / np.sqrt(rbv)
```
andreab1997, line 81: This block is used almost unchanged at least twice, maybe it would be sensible to have a unit function for that

validphys2/src/validphys/kinematics.py
andreab1997: I agree, could you remove the new lines?

andreab1997 (changes requested):

scarlehoff:
This is then ready to be merged? (@andreab1997 @comane )

andreab1997:
> This is then ready to be merged? (@andreab1997 @comane )

I don't think so, surely I need to review this again but in any case I would wait for the paper to be published.

andreab1997 (commented):
Ok I had a look and more or less nothing relevant changed since the last time I reviewed. @comane if you can answer the comments I left last time we can probably merge this soon.

validphys2/examples/pca_bias_variance_report.yaml
comane: The fits that I use in that runcard were done using th 200. I think this is the reason, why I added the legacy version

validphys2/src/validphys/closuretest/inconsistent_closuretest/multiclosure_inconsistent.py
comane: Yes this makes sense to have perhaps. I added a function called `eigendecomposition` that returns the eigenvectors and eigenvalues orderered in descendent way

comane:
Hi @andreab1997, I addressed most of the comments that you left, as well as those that I wrote myself.

There is one main issue with this PR at present, namely, the way in which the compare inconsistent closure test script works.

This script is problematic, because it's not possible to have a `dataset_inputs` which is different from `from_: fit`, note that this is a problem since we want to have out of sample datasets.

The problem, I think, can be solved by removing the vp-compare fits from `compareinconsistentclosuretemplates`.
We don't need to have another `vp-comparefits` anyways.

If you could take care of this that would be great.

andreab1997:
> Hi @andreab1997, I addressed most of the comments that you left, as well as those that I wrote myself.
> 
> There is one main issue with this PR at present, namely, the way in which the compare inconsistent closure test script works.
> 
> This script is problematic, because it's not possible to have a `dataset_inputs` which is different from `from_: fit`, note that this is a problem since we want to have out of sample datasets.
> 
> The problem, I think, can be solved by removing the vp-compare fits from `compareinconsistentclosuretemplates`. We don't need to have another `vp-comparefits` anyways.
> 
> If you could take care of this that would be great.

Just to understand, this issue is only there if you use the CLI or even if you write your own runcard and template?

comane:
> Just to understand, this issue is only there if you use the CLI or even if you write your own runcard and template?

I think it's there if I run `validphys multiclosure_analysis.yaml` (with dataset inputs that is not from the fit)

andreab1997:
> > Just to understand, this issue is only there if you use the CLI or even if you write your own runcard and template?
> 
> I think it's there if I run `validphys multiclosure_analysis.yaml` (with dataset inputs that is not from the fit)

Ok I need to test this, maybe it is something fixable

comane:
@scarlehoff, once the tests pass, this is ready to merge in my opinion

validphys2/src/validphys/closuretest/multiclosure.py
```diff
@@ -148,6 +149,42 @@ def fits_dataset_bias_variance(
...
+    # The dimentions here are (fit, data point, replica)
+    reps = np.asarray([th.error_members[:, :_internal_max_reps] for th in closures_th])
+    # One could mask here some reps in order to avoid redundancy of information
+    # TODO
+
+    n_fits = np.shape(reps)[0]
+    deltas = []
+    # There are n_fits pdf_covariances
+    # flag to see whether to eliminate dataset
+    for i in range(n_fits):
```
scarlehoff, line 172: Why don't you loop over `reps`, the `n_fits` variable doesn't seem necessary.

validphys2/src/validphys/closuretest/multiclosure.py
```diff
@@ -148,6 +149,42 @@ def fits_dataset_bias_variance(
...
+        bias_diffs = np.mean(reps[i], axis=1) - law_th.central_value
+        # bias diffs in the for loop should have dim = (n_obs)
+        sigmas = np.sqrt(np.var(reps[i], axis=1))
+        # var diffs has shape (n_obs,reps)
+        bias_diffs = bias_diffs / sigmas
```
scarlehoff, line 177:
```suggestion
        bias_diffs /= sigmas
```

validphys2/src/validphys/closuretest/multiclosure.py
```diff
@@ -148,6 +149,42 @@ def fits_dataset_bias_variance(
...
 
+@check_multifit_replicas
+def fits_normed_dataset_central_delta(
+    internal_multiclosure_dataset_loader, _internal_max_reps=None, _internal_min_reps=20
+):
+    """
+    For each fit calculate the difference between central expectation value and true val. Normalize this
+    value by the variance of the differences between replicas and central expectation value (different
+    for each fit but expected to vary only a little). Each observable central exp value is
+    expected to be gaussianly distributed around the true value set by the fakepdf
```
scarlehoff, line 160: Please add also the return variable to this docstr, smt like 

```
deltas: np.array
    array of shape (n_fits, nreps)
```
(or whatever the actual shape is)


If you add the input arguments as in the docstr of the function  below even better.

validphys2/src/validphys/closuretest/multiclosure.py
```diff
@@ -148,6 +149,42 @@ def fits_dataset_bias_variance(
...
+    # One could mask here some reps in order to avoid redundancy of information
+    # TODO
+
+    n_fits = np.shape(reps)[0]
+    deltas = []
+    # There are n_fits pdf_covariances
+    # flag to see whether to eliminate dataset
+    for i in range(n_fits):
+        bias_diffs = np.mean(reps[i], axis=1) - law_th.central_value
+        # bias diffs in the for loop should have dim = (n_obs)
```
scarlehoff, line 174:
```suggestion
        # bias diffs in the for loop should have shape = (n_obs,)
```

Do you mean the shape? It is relevant for the comment below

validphys2/src/validphys/closuretest/multiclosure.py
```diff
@@ -148,6 +149,42 @@ def fits_dataset_bias_variance(
...
+        sigmas = np.sqrt(np.var(reps[i], axis=1))
+        # var diffs has shape (n_obs,reps)
+        bias_diffs = bias_diffs / sigmas
+
+        deltas.append(bias_diffs.tolist())
+        # biases.shape = (n_fits, n_obs_cut/uncut)
+        # variances.shape = (n_fits, n_obs_cut/uncut, reps)
```
scarlehoff, line 181: I guess variances are the sigmas? 

Also, I don't understand how the division work if they have different shapes?

(I guess some of the comments are from previous versions of the function? I like having many comments just want to make sure they are correct and updated!)

validphys2/src/validphys/closuretest/multiclosure.py
```diff
@@ -817,17 +932,76 @@ def dataset_fits_bias_replicas_variance_samples(
...
+    _internal_min_reps=20,
+):
+    """
+    Load in a dictionary all the specs of a dataset meaning:
+    - ds name
+    - ds coords
+    - standard deviation (in multiclosure)
+    - mean (in multiclosure again)
+    - (x,Q^2) coords
+    """
```
scarlehoff, line 961: Could you add some more info here. It's not clear why is it necessary.

e.g., could you use `xq2map_with_cuts` as the input? Instead of calling `xq2map_with_cuts`?

(maybe not)

validphys2/src/validphys/closuretest/multiclosure.py
```diff
@@ -817,17 +932,76 @@ def dataset_fits_bias_replicas_variance_samples(
...
+    xq2_map_obj = xq2map_with_cuts(commondata, cuts)
+    coords = xq2_map_obj[2]
+    central_deltas = fits_normed_dataset_central_delta(internal_multiclosure_dataset_loader)
+    # std_devs = np.std(central_deltas, axis = 0)
+    std_devs = np.sqrt(np.mean(central_deltas**2, axis=0))
+    means = np.mean(central_deltas, axis=0)
+    xi = dataset_xi(dataset_replica_and_central_diff(internal_multiclosure_dataset_loader, False))
+
+    # for case of DY observables we have 2 (x,Q) for each experimental point
```
scarlehoff, line 970:
```suggestion
    # for the case of double-hadronic observables we have 2 (x,Q) for each experimental point
```

validphys2/src/validphys/closuretest/multiclosure_output.py
```diff
@@ -60,6 +60,65 @@ def plot_total_fits_bias_variance(fits_total_bias_variance):
...
+    )
+    df.columns = [
+        "ndata",
+        "bias",
+        "variance",
+        "ratio",
+        "error ratio",
+        "sqrt(ratio)",
+        "error sqrt(ratio)",
+    ]
```
scarlehoff, line 118: I don't understand why if you are creating the records in this same function you need to change the columns afterwards, can't you use the "right" names from the beginning? (instead of `dict` just use `{"error ratio" : <>}`, or does pandas complain if you do that?)

validphys2/src/validphys/closuretest/multiclosure_output.py
```diff
@@ -794,3 +853,25 @@ def plot_bias_variance_distributions(
     ax.legend()
     ax.set_title("Total bias and variance distributions.")
     yield fig
+
+@figuregen
+def xq2_data_prcs_maps(xq2_data_map,each_dataset):
```
scarlehoff, line 858: Please add a docstring, I would not know what this function is doing :_

scarlehoff (commented):
I haven't been able to test the code, but have had a look through and left a few comments.

validphys2/src/validphys/closuretest/multiclosure.py
comane: Again, a bit of code from @giovannidecrescenzo.

Here, I would rather have:
`delta = bias_diffs / sigma`

and then 

`delas.append(delta.tolist())`

validphys2/src/validphys/closuretest/multiclosure.py
comane: Yes, I agree, comments here make little sense. I now updated them.

validphys2/src/validphys/closuretest/multiclosure.py
comane: Thanks for the comment. 

xq2map_with_cuts is now taken in input and not loaded again

validphys2/src/validphys/closuretest/multiclosure_output.py
comane: I added a description. @giovannidecrescenzo just double check that it describes what the function is supposed to do please

validphys2/src/validphys/closuretest/runcards/old_rbv_definition.yaml
```diff
@@ -0,0 +1,56 @@
...
+  ---------------------------------------------------------------------
+
+  Table
+  =====
+
+  {@table_datasets_bias_variance_fits@}
+  
+
+actions_:
+  - report(main=True)
```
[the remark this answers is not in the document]
comane: It's supposed to allow one to reproduce the results from the paper https://arxiv.org/pdf/2111.05787

The paper above used a "wrong" definition of Ratio Bias Variance. After the new closure test paper will be published I think that we can get rid of the old functions that were used for rbv, as well as of this runcard.

Because of this, I don't think that this belongs to examples

validphys2/examples/pca_bias_variance_report.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+  ## Bootstrapped table for full dataset  
+  {@bootstrapped_table_bias_variance_data@}
+
+  ## L2 condition number
+  {@plot_l2_condition_number@}
+  
+
+actions_:
+  - report(main=true)
```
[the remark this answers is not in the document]
comane: smaller by reducing the number of fits that are included?
I think this would be reasonable yes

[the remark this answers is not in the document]
comane: I left only 3

validphys2/src/validphys/closuretest/multiclosure.py
comane: yes you are right. I updated it now, thanks!

validphys2/src/validphys/closuretest/inconsistent_closuretest/multiclosure_inconsistent_output.py
```diff
@@ -0,0 +1,557 @@
...
+    title: str
+        description of multiclosure
+
+    label_hist: str
+        summary description of multiclosure
+
+
+
+    Yields
```
scarlehoff, line 501:
```suggestion
    Returns
```

validphys2/src/validphys/closuretest/inconsistent_closuretest/multiclosure_inconsistent_output.py
```diff
@@ -0,0 +1,557 @@
...
+
+    Parameters
+    ----------
+    lambdavalues_bootstrapped_table_bias_variance_data: list
+        list of data frames computed as per table_bias_variance_data.
+
+    lambdavalues: list
+        list specified in multiclosure_analysis.yaml
+
+    Yields
```
scarlehoff, line 375:
```suggestion
    Returns
```

(or maybe the error is that you wanted to yield at the end?)

validphys2/src/validphys/closuretest/inconsistent_closuretest/multiclosure_inconsistent_output.py
```diff
@@ -0,0 +1,557 @@
...
+    mean_xi = np.mean(mean_boot_vals)
+    std_xi = np.std(mean_boot_vals)
+
+    records = [dict(data="full data", mean_dof=mean_dof, mean_xi=mean_xi, std_xi=std_xi)]
+
+    df = pd.DataFrame.from_records(
+        records, index="data", columns=("data", "mean_dof", "mean_xi", "std_xi")
+    )
+
+    df.columns = ["mean_dof", "mean_xi", "std_xi"]
```
scarlehoff, line 556: Why do you need `df.columns` in this case if the names are not changing?

scarlehoff (commented):
Thanks. I've now reviewed the part that I was missing (and ran a few reports).

I've tried to go through the entire thing but I haven't followed the project enough to be able to pick up any problems so take the review with that grain of salt, I haven't seen anything obvious though.

The only general question I have is, why do you need so many extra functionality under `inconsisent_closuretest`? I would have naively thought that many of the estimators could also be applied to the usual (multi)CT ?

As another request (maybe it is already in some of the other PRs), could you add a section in the docs for how to prepare and then analyze an inconsistent closure test? (hopefully together with the paper when sent to the arxiv)

validphys2/src/validphys/closuretest/inconsistent_closuretest/multiclosure_inconsistent_output.py
```diff
@@ -0,0 +1,557 @@
...
+    )
+    df.columns = [
+        "dof",
+        "bias",
+        "variance",
+        "ratio",
+        "error ratio",
+        "sqrt(ratio)",
+        "error sqrt(ratio)",
+    ]
```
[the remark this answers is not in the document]
comane: I am not sure that this works, as there is no `ds`

comane:
> The only general question I have is, why do you need so many extra functionality under `inconsisent_closuretest`? I would have naively thought that many of the estimators could also be applied to the usual (multi)CT ?

I agree with you, the reason to do this is that the multiclosure.py module is already very large and full of unused functions.
I have now put all of the general multiclosure functions in multiclosure.py. 
In a future PR I will then polish the multiclosure module from the old and unused multiclosure functions.

validphys2/src/validphys/closuretest/inconsistent_closuretest/multiclosure_inconsistent_output.py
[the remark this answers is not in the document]
comane: Sure, it should have been taken care of now.

scarlehoff (approved):
Thanks for this, let's wait for the tests to finish then merge!

PR #1983, opened 2024-03-05
add cuts for ttb and dis+j

t7phy:


Radonirinaunimi (approved):
This looks good! I have checked the dataset one-by-one and made sure that the correct bins are removed. For reference, below is the code that I used:

```python
from validphys.filters import (get_cuts_for_dataset, Rule, default_filter_settings_input, default_filter_rules_input)
from validphys.loader import Loader

l = Loader()
theory = l.check_theoryID(600)
params = theory.get_description()

cd = l.check_commondata("$dataset_name")
filter_defaults = default_filter_settings_input()
rule_list = [Rule(initial_data=i, defaults=filter_defaults, theory_parameters=params) for i in default_filter_rules_input()]
get_cuts_for_dataset(cd, rules=rule_list)
```

t7phy:
> This looks good to me!
> 

Thanks!

PR #1985, opened 2024-03-06
Test only 3.9 and 3.11 ; build only in 3.11

scarlehoff:
Since we are not compiling anymore, we only need the `noarch` package and so we only need to upload one package to the server every time we merge from master (instead of six).

For the tests, at the moment I've left 4: 3.9, 3.11 and ubuntu/mac. We are back to 4 tests with conda (+2 in python). We can get away with less of course. I just want to avoid the many false positives that we get when the server in Milan stops accepting connections because we are downloading too much stuff too quickly.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ["3.9", "3.11"]
+        python-version: ["3.9", "3.11"] # only test for the outer values for the range we support to prevent the cluster from refusing connections due to too many requests
       fail-fast: false
     runs-on: ${{ matrix.os }}
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -1,24 +1,18 @@
...
       matrix:
-        os: [ubuntu-latest] # , macos-latest]
-        python-version: ["3.11"] # ["3.9", "3.10", "3.11"]
-        include:
-          - os: ubuntu-latest
-            CONDA_OS: linux-64
-          - os: macos-latest
-            CONDA_OS: osx-64
+        os: [ubuntu-latest, macos-latest]
+        python-version: ["3.9", "3.11"]
```
RoyStegeman, line 15:
```suggestion
        python-version: ["3.9", "3.11"] # only test for the outer values for the range we support to prevent the cluster from refusing connections due to too many requests
```

RoyStegeman (approved):

Issue #1449, opened 2021-10-28
Run example runcards in tests


Zaharid:
It would be good to run some of the example runcards in tests. There is the question of whether that should be done always on every commit or e.g. like with the fit running when we set some label.

 In any case it should be done reasonably (e.g. downloading all the nine theories for theory covmats is out of the question) and reusing resources as much as possible.

PR #1989, opened 2024-03-08
Enable python 3.12

### LINKED ISSUE #1970: Changes needed to enable Python 3.12
We have a number of dependencies that would be blocking us from using Python 3.12:

- [x]  The `MultiDense` layer is not compatible with Tensorflow 2.16 (#1968) (fixed in #1975)
- [x] Recommonmark is no longer updated and it breaks for 3.12 due to newer versions of sphinx (https://github.com/NNPDF/nnpdf/issues/1967#issuecomment-1972957899) (in #1708 sphinx is pinned to 5)
- [x] Curio is no longer updated but the version of [reportengine](https://github.com/NNPDF/reportengine) we use depends on it.. Might be fixed by moving to dask but there are some small changes in reportengine needed for that https://github.com/NNPDF/reportengine/pull/61

Please add any other issues as you find them.

scarlehoff:
Now reportengine and pineappl are available for python 3.12
Let's see whether there's something else to be changed.

Closes #1970

edit: not sure why `boa<=0.14` is a problem when python <3.12 gets 0.16, but once the fit bot tests that everything run I'll check that

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,11 +72,8 @@ numpy = "*"
 validobj = "*"
 prompt_toolkit = "*"
 frozendict = "*"  # validphys: needed for caching of data loading
-# Reportengine (and its dependencies) need to be installed in a bit more manual way
-# the version of RE that we need requires curio, and the pypi package for curio is no longer updated
-curio = { git = "https://github.com/dabeaz/curio", rev = "8667999" }
-reportengine = { git = "https://github.com/NNPDF/reportengine", rev = "3bb2b1d"}
-ruamel_yaml = {version = "<0.18"}
+# Reportengine needs to be installed from git
+reportengine = { git = "https://github.com/NNPDF/reportengine" }
 # Fit
 psutil = "*"
 tensorflow = "*"
--- a/.github/workflows/fitbot.yml
+++ b/.github/workflows/fitbot.yml
@@ -12,7 +12,7 @@ env:
   POSTFIT_NREP: 16 # requested replicas for postfit
   REFERENCE_SET: NNBOT-80553d777-2024-03-07 # reference set for exact results
   STABLE_REFERENCE_SET: NNBOT-c0f99b7b3-2024-02-28 # reference set for last tag
-  CONDA_PY: 310
+  CONDA_PY: 312
   PYTHONHASHSEED: "0"
 
 jobs:
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.10"]
+        python-version: ["3.12"]
     runs-on: ${{ matrix.os }}
     env:
       NETRC_FILE: ${{ secrets.NETRC_FILE }}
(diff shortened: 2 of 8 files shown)


RoyStegeman:
I guess you read it as well in the boa docs: https://github.com/mamba-org/boa?tab=readme-ov-file
but it seems that now rattler is the recommended way to build a package: https://github.com/prefix-dev/rattler-build/tree/main

scarlehoff:
No I didn't, I hadn't look into this yet. I'll use conda-build then since I guess it is the only one guaranteed to stay...

RoyStegeman:
Indeed, if that doesn't increase the time beyond what is reasonable that's probably the way to go

scarlehoff:
I think the error (might) be coming from reportengine.

scarlehoff:
@comane chaged the generator to a list here https://github.com/NNPDF/reportengine/blob/93b94ade66e1139aafaf45ab3f57ecbfbd21dbed/src/reportengine/figure.py#L143 for parallelization reasons.
I've changed the tests but not sure whether it will have an effect somewhere else.

RoyStegeman:
I think that would affect all `figuregen` calls, so in a lot of places in vp. I guess we just test a subset

scarlehoff:
> I guess we just test a subset.

I think if the fitbot passes we can consider all the relevant places where figuregen is used are ok...
at good time to deal with this I guess https://github.com/NNPDF/nnpdf/issues/1449

RoyStegeman:
> if the fitbot passes we can consider all the relevant places where figuregen is used are ok

That's a big "if", but I agree. 

> good time to deal with this I guess https://github.com/NNPDF/nnpdf/issues/1449

Yes indeed, I suppose it's the vp analogue of the n3fit regression tests

scarlehoff:
tbh, for this we don't even need them to work as regression, just make sure that the runcards actually run

RoyStegeman (approved):

scarlehoff:
Sadly for reportengine we still need to install from git. Let's just point to the latest master for now there (since that should match what's in conda)

RoyStegeman:
Why do we need to install from git? What is stopping us from updating the conda package?

pyproject.toml
```diff
@@ -72,11 +72,8 @@ numpy = "*"
 validobj = "*"
 prompt_toolkit = "*"
 frozendict = "*"  # validphys: needed for caching of data loading
-# Reportengine (and its dependencies) need to be installed in a bit more manual way
-# the version of RE that we need requires curio, and the pypi package for curio is no longer updated
-curio = { git = "https://github.com/dabeaz/curio", rev = "8667999" }
-reportengine = { git = "https://github.com/NNPDF/reportengine", rev = "3bb2b1d"}
-ruamel_yaml = {version = "<0.18"}
+# Reportengine needs to be installed from git
+reportengine = { git = "https://github.com/NNPDF/reportengine" }
```
RoyStegeman, line 76: What's stopping us from releasing a package on pypi? Is Zahari the only person who can do this?

scarlehoff: Yes, it is under his account.

I can ask him to transfer it or free it or whatever once we have this working. But since nnpdf is also not in pypi this should be fine for now.

PR #1996, opened 2024-03-10
LHCb Z 13 TeV dimuon 2022

achiefa:
This PR implements the new measurements from LHCb for Z production at 13 TeV in the new commondata format. Please, observe the following:
- The keys in the metadata relative to the FK tables will be modified once FK tables will be produced.
- I'm not sure of the keys in "nnpdf_metadata", so please check if it is correct.
- Recall that the double distribution in the boson transverse momentum and rapidity has the issue with negative define correlation matrix. Hence, you won't find yaml file for this particular distribution.


scarlehoff:
Hi @achiefa coyld you cherry-pick or rebase the last few commits? (so that the branch contain only your changes).

RoyStegeman:
Should this be on top of `new_commodnata_utils` or `master`? Currently it contains commits that are in `master` but not in `new_commodnata_utils`.

achiefa:
My understanding from the last pheno meeting was to rebase on top of `master`, but I may be wrong. Should I rebase on top of `new_commondata_utils` ?

RoyStegeman:
I wasn't there so I don't know what was discussed, but rebasing on master indeed seems reasonable to me

RoyStegeman:
Is there a report (ideally comparefits) that includes this dataset?

achiefa:
No, there isn't. How can I produce such a report?

buildmaster/LHCB_Z_13TEV_DIMUON_2022/metadata.yaml
```diff
@@ -0,0 +1,104 @@
+setname: "LHCB_Z_13TEV_DIMUON_2022"
```
scarlehoff, line 1:
```suggestion
setname: "LHCB_Z_13TEV_DIMUON_2022"
```

This can be either `Z0` (for Z production) or `Z0J` (for Z + J production).
you cannot have both processes in the same dataset even if they are coming from the same paper. Please break it down into two distinct datasets.

Also, I'd avoid using `2022` in the name of the dataset unless it is necessary to disambiguate.

buildmaster/LHCB_Z_13TEV_DIMUON_2022/metadata.yaml
```diff
@@ -0,0 +1,104 @@
...
+
+nnpdf_metadata:
+  nnpdf31_process: "Z0PT"
```
scarlehoff, line 4:
```suggestion
  nnpdf31_process: "NC DY"
```

This is probably what you want here. But check the other `Z0` and `Z0J` datasets to check what to write here.
This information is used for the theory covariance matrix so it is important that it matches between datasets.

buildmaster/LHCB_Z_13TEV_DIMUON_2022/metadata.yaml
```diff
@@ -0,0 +1,104 @@
...
+    tables: {stat: table_8, eff:  table_11, data: table_15, sys:  table_20}
+    kinematics:
+      variables:
+        pT: {description: "transverse momentum", label: "$p_T^Z$", units: "$GeV$"}
+        sqrt_s: {description: "center of mass energy", label: '$\sqrt{s}$', units: "$GeV$"}
+      file: kinematics_pT.yaml
+    data_central: data_pT.yaml
+    data_uncertainties:
+      - uncertainties_pT.yaml
+    kinematic_coverage: [pT, sqrt_s]
```
scarlehoff, line 36: Please use `sqrts`

buildmaster/LHCB_Z_13TEV_DIMUON_2022/metadata.yaml
```diff
@@ -0,0 +1,104 @@
...
+version_comment: "First implementation of the data set"
+
+implemented_observables:
+
+  # pT data
+  - observable: { description: "Differential cross-section of Z-->µµ as a function of Z-transverse momentum",
+                  label: '$\frac{d\sigma}{dp_T^Z}$',
+                  units: "[pb]"}
+    observable_name: PT
+    process_type: Z0PT
```
scarlehoff, line 25: This process type doesn't exist yet.

You will need to add a new process type for this dataset https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/process_options.py (at the bottom of the module you can find the ones currently implemented)

buildmaster/LHCB_Z_13TEV_DIMUON_2022/metadata.yaml
```diff
@@ -0,0 +1,104 @@
...
+    tables: {data: table_17 ,sys: table_22 ,stat: figure_13_(left) ,eff: figure_15_(left)}
+    kinematics:
+      variables:
+        pT: {description: "transverse momentum", label: "$p_T^Z$", units: "$GeV$"}
+        yZ: {description: "boson rapidity", label: "$y^Z$", units: ""}
+        sqrt_s: {description: "center of mass energy", label: '$\sqrt{s}$', units: "$GeV$"}
+      file: kinematics_ZpT_y.yaml
+    data_central: data_ZpT_y.yaml
+    data_uncertainties: uncertainties_ZpT_y.yaml
+    kinematic_coverage: [pT, yZ, sqrt_s]
```
scarlehoff, line 93: I think (but again, check with the other `LHCB_Z` datasets that @Radonirinaunimi implemented) that for the rapidity the variable to use is `y`.

buildmaster/LHCB_Z_13TEV_DIMUON_2022/metadata.yaml
```diff
@@ -0,0 +1,104 @@
...
+    theory:
+      FK_tables: 
+        - - LHCB_Z_13TEV_DIMUON_2022_DIF_PT
+      operation: 'null' 
+
+# pT-y^Z data
+  - observable: { description: "Double differential cross-section of Z-->µµ as a function of Z-transverse momentum and Z-rapidity",
+                  label: '$\frac{d^2\sigma}{d p_T^Z y^Z}$',
+                  units: "[pb]"}
+    observable_name: PT_Y
```
scarlehoff, line 81: You cannot use underscore in the variable name, if you want a double differential you can use `PT-Y`.

scarlehoff (commented):
At the moment you won't be able to produce a report using this dataset since the theory is missing.

There's a few things to fix though.

RoyStegeman:
> At the moment you won't be able to produce a report using this dataset since the theory is missing.

Ah, I thought that came with the PR. I should have read the first message more carefully.

buildmaster/LHCB_Z_13TEV_DIMUON_2022/metadata.yaml
achiefa: The paper deals with Z production only. So I'll change it to `Z0`.

The `2022`label I thought was sort of necessary to tell it apart from the older `LHCB_Z0` datasets which date back to 2016 and that are already implemented.

scarlehoff: ZpT is Z + jet, that's why we have those as Z0J

achiefa: Yes, but legacy datasets fro the same experiment are `Z0`. See for instance [LHCB_Z0_13TEV](https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/datafiles/new_commondata/LHCB_Z0_13TEV/metadata.yaml). So I suppose I should do the same here, shouldn't I?

Moreover, since the same name is used for the implementation linked above, should I keep the `2022` label?

scarlehoff: With respect to the `2022` label, I guess we can live with that. But I'd prefer something that includes some information (like for instance the integrated luminosity, since is what we have in a few places already `XYPB`).

With respect to the process names, at the moment we have no Z+J process in for `LHCB` so this would be the first one.

Note that the process label `Z0` (or `Z0J`) refers in practice to the leading order process being considered.
For the Z rapidity this corresponds to $p \ p -> Z -> l^{+} \ l^{-}$ which we are calling `Z0` while for the Z pt is instead $p \ p -> Z + j$ where the measurement is the pT of the jet (and, if double differencial, pt and rapidity) and $j$ is a jet, and we are calling this one `Z0J`.

The difference is important since, for instance, this corresponds to distinct processes when computing predictions with matrix / nnlojet / madgraph.

achiefa: Ok, now I understand. It wasn't clear to me that `Z0` and `Z0J` must be broken down into two different datasets (i.e. two different folders), even if they come from the same measurement. I would suggest to specify that in the [documentation](https://docs.nnpdf.science/data/commondata.html), because it's not trivial for a novice (as I am).

RE the `2022` label, I can change it to the integrated luminosity if you wish. Note that the legacy implementation for `Z0` does not report this information in the name.

buildmaster/LHCB_Z_13TEV_DIMUON_2022/metadata.yaml
achiefa: I just checked [ATLAS_Z0J_8TEV](https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/datafiles/new_commondata/ATLAS_Z0J_8TEV/metadata.yaml) and [LHCB_Z0_13TEV](https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/datafiles/new_commondata/LHCB_Z0_13TEV/metadata.yaml) and it seems that `nnpdf31_process` is `DY NC` for both `Z0` and `Z0J`.

Anyway, I appreciate that you also provide an explanation for what these values actually do.

buildmaster/LHCB_Z_13TEV_DIMUON_2022/metadata.yaml
achiefa: This part is not really clear to me. In [LHCB_Z0_13TEV](https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/datafiles/new_commondata/LHCB_Z0_13TEV/metadata.yaml), `process_type` is set to `EWK_RAP` for the rapidity distribution. However, this does not appear at the bottom of the module that you mentioned. The same holds for [ATLAS_Z0J_8TEV](https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/datafiles/new_commondata/ATLAS_Z0J_8TEV/metadata.yaml), although here for the double distributions in pT. Am I missing something?

scarlehoff: Basically old datasets had the kinematic variables set to be always 3 variables and always the same one (for a given process, note that in this case "process" refer to both the physical process and the observable/measurement; so two datasets can have the same `nnpdf31_process`* and different `process_type`... it's confusing I know).

Anyway, the process_type basically defined how the variables are read by validphys and autogenerate some plotting labels. In the LHCB datasets this process_type still refers to the old process type because the kinematic variables are the same and written in the same order as before... 
I guess we can talk about this on Wednesday but basically the summary is that in order to add (new) DY data you will need to add a new process to that list that is able to understand the kinematic variables that you are using.

RoyStegeman:
Could you please move this from `buildmaster` to `validphys/datafiles/new_commondata`?

validphys2/src/validphys/process_options.py
```diff
@@ -252,6 +252,7 @@ def _displusjet_xq2map(kin_dict):
     "HQP_PTQ": HQP_PTQ,
     "HERAJET": HERAJET,
     "HERADIJET": dataclasses.replace(HERAJET, name="HERADIJET", description="DIS + jj production"),
+    "DY_Z_Y" : dataclasses.replace(DY_W_ETA, name="DY_Z_Y", description="DY Z -> ll rapidity")
```
achiefa, line 255: I made use of the process type implemented by MC, but I changed name and description. Moreover, note that the accepted kinematic variable in `DY_W_ETA` is `eta`. Should I use `y` instead, since LHCb data is in the boson rapidity?

scarlehoff: I think both are ok. Actually, we could call it also DY_RAP or something like that and unify both with a common description.

But changing these (internal) names are "as easy" as a search & replace.

validphys2/src/validphys/process_options.py
```diff
@@ -189,6 +189,17 @@ def _displusjet_xq2map(kin_dict):
...
 
+def _dptboson_xq2map(kin_dict):
+    """Compute x and q2 mapping for DY Z -> ll + jet process.
+    Here pT refers to the transverse momentum of the Z boson."""
+    pT = kin_dict[_Vars.pT]
+    m_Z2 = kin_dict[_Vars.m_Z2]
+    sqrts = kin_dict[_Vars.sqrts]
+    x = np.sqrt(m_Z2 + pT * pT) / sqrts
+    q2 = m_Z2 + pT * pT
+    return x, q2
```
achiefa, line 200: I took [kintransforms.py](https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/plotoptions/kintransforms.py) as a reference for the map. Is that correct?

validphys2/src/validphys/datafiles/new_commondata/LHCB_Z0J_13TEV_2022/metadata.yaml
```diff
@@ -50,18 +51,19 @@ implemented_observables:
                   label: '$\frac{d^2\sigma}{d p_T^Z y^Z}$',
                   units: "[pb]"}
     observable_name: DIMUON-PT-Y
-    process_type: EWK_PTRAP
+    process_type: _
```
achiefa, line 54: Here I left blank, because we are not going to use the double differential distribution. But if you wish I can implement the relative process type anyway.

validphys2/src/validphys/process_options.py
scarlehoff: There were some bugs there, you can try creating the x-q2 plot https://vp.nnpdf.science/vxS8mWGySsu-cehRFnhSOA==/ to see how it looks for these datasets.

validphys2/src/validphys/datafiles/new_commondata/LHCB_Z0J_13TEV_2022/metadata.yaml
scarlehoff: Why are we not going to use it?

(if we know already that we are not going to use it... why do we want to have it implemented?)

achiefa: Do you mean removing this distribution from the metadata? Observe that this part is not outputted in the yaml files exactly for that reason.

scarlehoff: yes, and don't include the kinematics/data/uncertainties files (which at least in the metadata are referenced)

validphys2/src/validphys/process_options.py
achiefa: Should I rebase on top of master again? Currently `DY_W_ETA` is not in `process_options`.

scarlehoff: Yes, if you need to use it yes.

validphys2/src/validphys/datafiles/new_commondata/LHCB_Z0_13TEV_2022/uncertainties_dimuon_y.yaml
```diff
@@ -0,0 +1,897 @@
...
+      calibration.(uncorrelated)
+    treatment: MULT
+    type: UNCORR
+  Alignment_correlated:
+    description: Correlated uncertainties from detector alignment and momentum scale
+      calibration.(correlated)
+    treatment: MULT
+    type: CORR
+bins:
+- Statistical_uncertainty:
```
achiefa, line 52: I have a feeling that I misunderstood how uncertainties must be listed in the file. Could please check if they are implemented correctly, @scarlehoff ?

scarlehoff: The uncertainties need to be listed one per bin:

```yaml
bins:
- stats: 0.10
   some_name: 11.0
   some_other_name: 143
- stats: 0.20 
   some_name: 143
```

and so on.
The statistical uncertainty should always be named `stat`.

validphys2/src/validphys/process_options.py
achiefa: This is what I get: https://vp.nnpdf.science/y1HWyAGhT0Sm1rubBeO5Xw==

nnpdf_data/nnpdf_data/new_commondata/LHCB_Z0J_13TEV_2022/kinematics_dimuon_pT.yaml
```diff
@@ -0,0 +1,169 @@
+bins:
+- pT:
+    min: 0.0
+    mid: 1.1
+    max: 2.2
+  sqrts:
+    min: null
+    mid: 13
```
scarlehoff, line 8:
```suggestion
    mid: 13
```

ah, gotcha! The problem you have is that you are using sqrts in TeV, if you write 13000 that should solve the problem you're seeing.

The same is probably true in The rapidity case as well. Make sure that the units are always GeV.

nnpdf_data/nnpdf_data/new_commondata/LHCB_Z0J_13TEV_2022/kinematics_dimuon_pT.yaml
```diff
@@ -0,0 +1,169 @@
+bins:
+- pT:
+    min: 0.0
+    mid: 1.1
+    max: 2.2
```
scarlehoff, line 5: @enocera I'm a bit worried about the kinematics here. For sure no fixed order calculation will be reliable at pT ~ 0, but I'm not even sure pT ~ 10 will be ok...

enocera: Indeed. The cut on pT that we have for the ATLAS and CMS 8 TeV data sets is at 30 GeV, therefore anything lower than that cannot be described by FO pert QCD (but needs resummation). So are you saying that the maximum pT in that measurement is 10 GeV? if this is the case, then we may forget about it until we do pT resummation.

enocera: Looking at the paper we may have, say, 3 bins above pT~30 GeV. Anyways, at this point I'd complete the implementation of the data set - that may be useful for the future.

nnpdf_data/nnpdf_data/new_commondata/LHCB_Z0J_13TEV_2022/kinematics_dimuon_pT.yaml
achiefa: Yes, really my bad. I've fixed that and now points are all within 1. Thanks for spotting it!

nnpdf_data/nnpdf_data/new_commondata/LHCB_Z0_13TEV_2022/metadata.yaml
```diff
@@ -0,0 +1,48 @@
...
+    kinematic_coverage: [y, m_Z2, sqrts]
+    plotting:
+      kinematics_override: identity
+      x_scale: log
+      dataset_label: "LHCb $Z\\to µµ$"
+      plot_x: y
+      y_label: '$d\sigma_{Z}/dy^Z$ (pb)'
+    theory:
+      FK_tables: 
+        - - LHCB_DY_13TEV_DIMUON
```
scarlehoff, line 47:
```suggestion
        - - LHCB_Z0_13TEV_DIMUON
```

nnpdf_data/nnpdf_data/new_commondata/LHCB_Z0_13TEV_2022/eigenvalues.yaml
```diff
@@ -0,0 +1,231 @@
+PT:
```
scarlehoff, line 1: Please put this file in the rawdata folder (that way it is easier to skip when distributing/installing the data)

scarlehoff:
Hi @achiefa, could you check the error:
```
FAILED tests/test_loader.py::test_kitable_with_cuts - validphys.loader.DataNotFoundError: No .dat file found for LHCB_Z0J_13TEV_2022_DIMUON and no new data translation found
```

Maybe there's some name problem? Or some mismatch for the ZJ dataset?

scarlehoff:
If they are the same just make sure that alm datasets are using the same one and remove the redundant one.

Since you are editing this dataset maybe it makes sense to use DY_NC, but it doesn't really matter.

achiefa:
Hi @scarlehoff, I've resolved the conflict and now all checks have passed. I just want to mention that I removed the jet data from `dataset_names.yaml` since we're not going to use them. 

Is there anything else that needs to be fixed or modified?

scarlehoff:
> Hi @scarlehoff, I've resolved the conflict and now all checks have passed. I just want to mention that I removed the jet data from dataset_names.yaml since we're not going to use them.

You should actually not add any new dataset to `dataset_names.yaml`. That's only a mapping for the legacy sets.

achiefa:
Ok, I've just removed it.

nnpdf_data/nnpdf_data/new_commondata/LHCB_Z0J_13TEV_2022/metadata.yaml
```diff
@@ -0,0 +1,48 @@
+setname: "LHCB_Z0J_13TEV_2022"
+
+nnpdf_metadata:
+  nnpdf31_process: "DY PT"
```
scarlehoff, line 4:
```suggestion
  nnpdf31_process: "DY NC"
```

nnpdf_data/nnpdf_data/new_commondata/LHCB_Z0J_13TEV_2022/metadata.yaml
```diff
@@ -0,0 +1,48 @@
...
+  - observable: { description: "Differential cross-section of Z-->µµ as a function of Z-transverse momentum",
+                  label: '$\frac{d\sigma}{dp_T^Z}$',
+                  units: "[pb]"}
+    observable_name: DIMUON-PT
+    process_type: DY_NC_PT
+    ndata: 14
+    tables: [table_15,  # data
+             table_20,  # sys
+             table_8,   # stat
+             table_11]  # eff
```
scarlehoff, line 29: Please, write the tables as integers ([15, 20, etc])

nnpdf_data/nnpdf_data/new_commondata/LHCB_Z0J_13TEV_2022/metadata.yaml
```diff
@@ -0,0 +1,48 @@
...
+    plotting:
+      kinematics_override: identity
+      x_scale: log
+      dataset_label: 'LHCB Z 13 TeV: $\frac{d\sigma}{dp_T^Z}$'
+      plot_x: pT
+      y_label: '$d\sigma_{Z}/dP_T^Z$ (pb)'
+    theory:
+      FK_tables: 
+        - -
+      operation: 'null'
```
scarlehoff, line 48:
```suggestion
```

validphys2/src/validphys/process_options.py
```diff
@@ -202,14 +202,14 @@ def _displusjet_xq2map(kin_info):
     x = q2 * q2 / s / (pt**2 - q2)
     return x, q2
 
-def _dywboson_xq2map(kin_dict):
+def _dyboson_xq2map(kin_dict):
```
scarlehoff, line 205:
```suggestion
def _dyboson_xq2map(kin_info):
```

Because what it gets is a `kinematics information` object and not a dictionary.

validphys2/src/validphys/process_options.py
```diff
@@ -202,14 +202,14 @@ def _displusjet_xq2map(kin_info):
...
     """
     Computes x and q2 mapping for pseudo rapidity observables
     originating from a W boson DY process.
     """
-    mass2 = kin_dict[_Vars.m_W2]    
+    mass2 = kin_dict[_Vars.m_W2] if _Vars.m_W2 in kin_dict._variables else kin_dict[_Vars.m_Z2]
```
scarlehoff, line 210:
```suggestion
    mass2 = kin_info.get_one_of(_Vars.m_W2, _Vars.m_Z2)
```

And the same below.

validphys2/src/validphys/process_options.py
```diff
@@ -229,6 +229,18 @@ def _dyncpt_xq2map(kin_info):
...
+def _dyncpt_xq2map(kin_dict):
+    """Compute x and q2 mapping for DY Z -> ll + jet process.
+    Here pT refers to the transverse momentum of the Z boson.
+    """
+    pT = kin_dict[_Vars.pT]
+    m_Z2 = kin_dict[_Vars.m_Z2]
+    sqrts = kin_dict[_Vars.sqrts]
+    x = (np.sqrt(m_Z2 + pT*pT) + pT) / sqrts
+    q2 = m_Z2 + pT * pT
+    return x, q2
```
scarlehoff, line 241: This is no longer necessary since it is already in master (note that the same function is defined twice)

validphys2/src/validphys/process_options.py
```diff
@@ -279,16 +291,18 @@ def _dyncpt_xq2map(kin_info):
...
 )
 
-DY_W_ETA = _Process(
-    "DY_W_ETA",
-    "DY W -> l nu pseudo rapidity",
+
+DY_NC = _Process(
+    "DY_RAP",
+    "DY W or Z -> final state lepton",
     accepted_variables=(_Vars.eta, _Vars.m_W2, _Vars.sqrts),
```
scarlehoff, line 298: It needs to accept also the variables for the Z

validphys2/src/validphys/process_options.py
```diff
@@ -279,16 +291,18 @@ def _dyncpt_xq2map(kin_info):
     xq2map_function=_displusjet_xq2map,
 )
 
-DY_W_ETA = _Process(
-    "DY_W_ETA",
-    "DY W -> l nu pseudo rapidity",
+
+DY_NC = _Process(
+    "DY_RAP",
+    "DY W or Z -> final state lepton",
```
scarlehoff, line 297:
```suggestion
    "DY W or Z -> 2 leptons ",
```

validphys2/src/validphys/process_options.py
```diff
@@ -279,16 +291,18 @@ def _dyncpt_xq2map(kin_info):
     xq2map_function=_displusjet_xq2map,
 )
 
-DY_W_ETA = _Process(
-    "DY_W_ETA",
-    "DY W -> l nu pseudo rapidity",
+
+DY_NC = _Process(
```
scarlehoff, line 295: I would call this `DY_2_LL` or `DY_2L` or something like that since the same kinematics can be used for both Z and W.

If you want to separate between `DY_NC` and `DY_CC` I would use those names directly. As this stands this is "wrong" since it says `DY_NC` but in the description it says that it can be either W or Z.

So I suggest you either use a generic name and then the process type is the same for both datasets, or if you want to differentiate, you can create `DY_NC` and `DY_CC` (both using the same function `_dyboson_xq2map`)

validphys2/src/validphys/process_options.py
```diff
@@ -306,8 +320,10 @@ def _dyncpt_xq2map(kin_info):
     "HQP_PTQ": HQP_PTQ,
     "HERAJET": HERAJET,
     "HERADIJET": dataclasses.replace(HERAJET, name="HERADIJET", description="DIS + jj production"),
-    "DY_W_ETA": DY_W_ETA,
-    "DY_NC_PT": DY_NC_PT,
+    "DY_W_ETA": dataclasses.replace(DY_NC, name="DY_W_ETA", description="DY W -> l nu pseudo rapidity"),
+    "DY_Z_Y": dataclasses.replace(DY_NC, name="DY_Z_Y", description="DY Z -> ll rapidity",
```
scarlehoff, line 324: See above.

scarlehoff (changes requested):

validphys2/src/validphys/process_options.py
achiefa: Yes you're right, but the two are slightly different. Should I go for the one implemented by Mark?

scarlehoff: Oh, I thought they were the same.

No, yours is more correct. The scale used in ZpT is ET (which is what you implemented) while mark's version uses mZ (which is what Z rapidity datasets use). Of course eT -> mZ in the limit of 0 pT

```
    x = (np.sqrt(m_Z2 + pT*pT) + pT) / sqrts
    q2 = m_Z2 + pT * pT
```

I would invert these two steps so you can do

```
    ET2 = m_Z2 + pT * pT
    x = (np.sqrt(ET2) + pT) / sqrts
```

and return ET2 so that is clearer.

I'm not sure which computation of `x` is correct though, @comane where did you get yours from ?

validphys2/src/validphys/process_options.py
achiefa: I'm slightly puzzled because in this branch I got rid of the process `DY_W_ETA` some commits ago. Now I only use `DY_NC`.

achiefa: Ok I understand.

validphys2/src/validphys/process_options.py
```diff
@@ -306,8 +310,9 @@ def _dyncpt_xq2map(kin_info):
     "HQP_PTQ": HQP_PTQ,
     "HERAJET": HERAJET,
     "HERADIJET": dataclasses.replace(HERAJET, name="HERADIJET", description="DIS + jj production"),
-    "DY_W_ETA": DY_W_ETA,
-    "DY_NC_PT": DY_NC_PT,
+    "DY_Z_Y": dataclasses.replace(DY_2L, name="DY_Z_Y", description="DY Z -> ll (pseudo)rapidity"),
+    "DY_W_ETA": dataclasses.replace(DY_2L, name="DY_W_ETA", description="DY W -> l nu (pseudo)rapidity"),
+    "DY_NC_PT": dataclasses.replace(DY_2L_PT, name="DY_NC_PT", description="DY Z -> ll + j")
```
achiefa, line 315: Could you please check that this makes sense? I'm still using `DY_W_ETA` because the previously implemented datasets for DY_CC adopts `DY_W_ETA` as process type in the metadata.

scarlehoff: Ok, this is fine. I don't mind that much having too many different types at the dataset level, which I'd like to avoid is the explosion of functions that compute said datasets.

One thing though, for the PT I don't think `DY_2L_PT` is correct. Just use `DY_PT`.

```suggestion
    "DY_NC_PT": dataclasses.replace(DY_2L_PT, name="DY_NC_PT", description="Z + j")
```

If you want to be more specific, `Z (ll) +j` but with `Z -> ll +j ` it looks like it is the Z that decays into two leptons and a jet.

achiefa: Ok, I've modified that.

validphys2/src/validphys/process_options.py
achiefa: Just to follow up, this PR now uses the following definition
```
    ET2 = m_Z2 + pT * pT
    x = (np.sqrt(ET2) + pT) / sqrts ,
```
which has been taken from [here](https://github.com/NNPDF/nnpdf/blob/9b1e436f5a36541e88f6e45ea27fe44012efebc7/validphys2/src/validphys/plotoptions/kintransforms.py#L207).

nnpdf_data/nnpdf_data/new_commondata/LHCB_Z0_13TEV_2022/filter.py
```diff
@@ -0,0 +1,333 @@
...
+    # Append efficiency uncertainties
+    with open("rawdata/table_" + str(observable['tables'][3]) + ".yaml", 'r') as file:
+     effcorr[observable['observable_name']] = yaml.safe_load(file)
+
+
+
+# Utility functions
+# ________________________________________________________
+def OuterDecomposition(matrix):
+    """ Compute the single value decomposition of the matrix A.
```
scarlehoff, line 45:
```suggestion
    r""" Compute the single value decomposition of the matrix A.
```

Because you are using latex in this docstr, tell python it is a raw string (otherwise python will interpret all the `\` as escape sequences).

(not that it matters for the filters, nobody is going to inspect these docstrings from within python and be surprised, but it's just good practice to do so)

The same needs to be done in the other `filter.py` file.

scarlehoff (approved):
I think this is ready to go.

Some small comments. Please link also some of the data-th comparisons you've already generated for this dataset for future references.

scarlehoff:
Thanks! I think this can be merged now.

> I made the report on the kinematic coverage for both LHCb and CMS (the latter was implemented by Mark). You can find the report here: [report](https://vp.nnpdf.science/dqX-_ADRTGyqfXqvB6zX2g==)

This looks good (in that it looks like what I would expect: CMS is central while LHCB accesses the extremes).

PR #1997, opened 2024-03-11
Make the sqlite theorydb into a folder of yaml files

scarlehoff:
This is my proposal for what we discussed in Amsterdam.

At the moment (for validphys) this doesn't change anything. All fields are mandatory as they were before. I think we should first change it to the layout we want (yaml file or other) and afterwards decide what fields are important and what are no.

All interactions with the theory database for validphys happened through the `validphys2/src/validphys/theorydbutils.py` module. All other changes to validphys are such that the database points to `theory_cards` instead of `theory.db`.

I've also modified the docs (I've searched for `theory.db` and changed stuff whenever they were mentioned).

I have to admit I am not fully convinced RE having a file per theory. Having them in a single file would allow us to do this:

```yaml
theory_41: &theory_41
  Comments: theory 41
  <other parameters>

theory_42:
  <<: *theory_41
  Comments: Theory inheriting from 41
```

40 000 000.yml --- baseline
40 000 001.yml

40 001 000.yml --- baseline
40 001 001.yml

```yaml
40 001 000: 40 000 000.yml
```

which might be very nice.

Another option is that we leave the ones we currently have as "one per file" (but with the syntax `id: <data>`) and going forward organize the theories as `41_012.yaml` and inside all the theories `41_012_000: <data>, 41_012_000: <data>` etc.

(edit: the regression test failed, maybe by casting some of the quantities as int or float changed slightly the result, but I'll deal with it at the end)


felixhekhorn:
> I have to admit I am not fully convinced RE having a file per theory. Having them in a single file would allow us to do this:

- Implementing (simple) inheritance should be easy enough for us to use
- having a huge theory file (~600*30 lines) is not much better then a sql database; it wouldn't be much readable either and we would couple (more or less) again several changes to the db together

so I would still prefer one unique theory per file

scarlehoff:
> Implementing (simple) inheritance should be easy enough for us to use

Actually, inheritance might also work across files with the default yaml loading.

I guess you are also against the mixture?
> Another option is that we leave the ones we currently have as "one per file" (but with the syntax id: <data>) and going forward organize the theories as 41_012.yaml and inside all the theories 41_012_000: <data>, 41_012_000: <data> etc.

giacomomagni (commented):
LGTM.

* Can you fix also:

https://github.com/NNPDF/nnpdf/blob/41471d854ca839ae7d8883e9d753aca5ded2683f/pyproject.toml#L30-L34

*<strike>And:

https://github.com/NNPDF/nnpdf/blob/theory_to_yaml_files/doc/sphinx/source/tutorials/reproduce.rst

RoyStegeman:
If we wish to search the files using grep as you suggested it would also help to have a single file per theory such. Also in that case we'll want to be very strict about formatting e.g. `ktThr: 1.0` should always be that and not `ktThr:   1.0`. Perhaps a formatting check in the tests can help with that?

validphys2/src/validphys/theorydbutils.py
```diff
@@ -5,80 +5,99 @@
...
+    PTO: int
+    FNS: str
+    DAMP: int
+    IC: int
+    ModEv: str
+    XIR: float
+    XIF: float
+    NfFF: int
+    MaxNfAs: int
+    MaxNfPdf: int
```
andreab1997, line 29: I would start removing parameters that we already know are useless. For example this `MaxNfPdf` but also `MaxNfAs` and `EScaleVar` (and most likely others...)

scarlehoff: I will do that in subsequent PRs.  The idea is that this leaves validphys essentially unchanged.

scarlehoff:
> should always be that and not ktThr:   1.0. Perhaps a formatting check in the tests can help with that?

This is important, yes. Maybe it is time to add a pre-commit hook to this repo...

RoyStegeman:
> Maybe it is time to add a pre-commit hook to this repo...

Something like that, but at least initially I'd be happy to have it check only this theory database and not the entire repository

doc/sphinx/source/theory/theoryparamsinfo.rst
```diff
@@ -0,0 +1,95 @@
...
+  'XIR': 1.0, 'XIF': 1.0, 'NfFF': 5, 'MaxNfAs': 5, 'MaxNfPdf': 5, 'Q0': 1.65,
+  'alphas': 0.118, 'Qref': 91.2, 'QED': 0, 'alphaqed': 0.007496252,
+  'Qedref': 1.777, 'SxRes': 0, 'SxOrd': 'LL', 'HQ': 'POLE', 'mc': 1.51,
+  'Qmc': 1.51, 'kcThr': 1.0, 'mb': 4.92, 'Qmb': 4.92, 'kbThr': 1.0, 'mt': 172.5,
+  'Qmt': 172.5, 'ktThr': 1.0, 'CKM': '0.97428 0.22530 0.003470 0.22520 0.97345 0.041000 0.00862 0.04030 0.999152',
+  'MZ': 91.1876, 'MW': 80.398, 'GF': 1.1663787e-05, 'SIN2TW': 0.23126, 'TMC': 1,
+  'MP': 0.938, 'Comments': 'NNPDF3.1 NNLO central', 'global_nx': 0, 'EScaleVar': 1}
+
+
+printing a dictionary in a python terminal is a bit cumbersome, the above
```
Radonirinaunimi, line 33:
```suggestion
Given that printing a dictionary in a python terminal is a bit cumbersome, the above
```
Not sure where does the sentence begin.

doc/sphinx/source/theory/theoryparamsinfo.rst
```diff
@@ -0,0 +1,95 @@
...
+Some of the entries of the table have been truncated to fit in the terminal, for
+example the CKM matrix elements. If you want to see the
+full output or wish to keep a permanent copy of the table then you can use the
+command line option ``-d``, which stands for ``--dumptable`` which will save the
+table in your current working directory as ``theory_<id>_info.csv``::
+
+  $vp-checktheory --fit FIT
+
+where ``FIT`` is a valid fit name. If the fit cannot be found locally, the script
+will attempt to download it.
```
Radonirinaunimi, line 93: I am not sure I understand this part. Do you want to highlight the use of `-d` or checking the theory settings for a given fit or both? For the last two, a sentence introducing it would be useful.

doc/sphinx/source/data/data-config.rst
```diff
@@ -30,14 +30,27 @@ all data files to be installed must have a ``.yaml`` extension.
...
-	``validphys/src/validphys2/datafiles/theory.db``
+	``validphys/src/validphys2/datafiles/theory_cards``
 
-This file should only be edited in order to add new theory options. It may be
-edited with any appropriate ``sqlite3``-supported software. A script is provided to
+in the form of ``yaml`` files. A new theory can be added by simply adding a new
+``yaml`` file with the desired theory id. The definition of the accepted and required parameters
+can be found at:
+
+	``validphys/src/validphys2/theorydbutils.py``
```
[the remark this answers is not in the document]
scarlehoff: That's better done once we know for sure which parameters are going to stay.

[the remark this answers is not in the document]
Radonirinaunimi: Okay, fine.

doc/sphinx/source/theory/theoryparamsinfo.rst
scarlehoff: I will not update these docs...  I just don't want to keep them too outdated but (specially taking into account the theory might change) I'd rather not touch them too much.

andreab1997 (approved):
As promised, I had a second look to the code part and it seems fine to me

PR #1998, opened 2024-03-11
Updates to the NNPDF docs

mariaubiali:
Hi all, 
I changed initial page, added new bib references and added instructions for installation from source on M3. This is just the beginning, any suggestions for extra things to modify are more than welcome.

### DIFF
--- a/doc/sphinx/source/get-started/installation.rst
+++ b/doc/sphinx/source/get-started/installation.rst
@@ -61,8 +61,6 @@ directory, by default this is ``~/miniconda3``, and run the command
        conda activate
        conda install nnpdf
 
-This will provide functioning C++ and Python executables.
-
 **Note:** The installer will set up its own version of the LHAPDF code,
 with its own path for storing PDFs, which can be seen running ``lhapdf –help``.
 If you have an existing directory with LHAPDF grids, you may want to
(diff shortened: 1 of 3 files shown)


RoyStegeman:
Thanks @mariaubiali. Since you're making changes to the people in the collaboration, could you please also update Zahari's position?

mariaubiali:
Thanks @RoyStegeman, well spotted, done! I have moved Zahari from current to past and added info on his current position

Radonirinaunimi:
Thanks @mariaubiali! Given that this installation now works for all M macs, could you also please remove the M1/M2 section?

scarlehoff:
Before merging this, let's test a bit the python 3.12 installation https://github.com/NNPDF/nnpdf/pull/1989
(I'd like to check that the change in reportengine won't have unintended consequences)

Also, since now we have a pineappl package also for Apple Silicon we can collapse the development installation for both mac and linux into just what's in this PR.

mariaubiali:
Thanks @Radonirinaunimi I have removed that paragraph and now have "Installation from source on M1-M2-M3 Mac". Could someone check if the section above "Installation from Source" is still up-to-date?

RoyStegeman:
> Could someone check if the section above "Installation from Source" is still up-to-date?

Thanks for asking. That's still fine.

RoyStegeman:
I updated the docs a little bit. I now mention the `nolha` option, but I'm not sure we should since if it's in the docs we may have to provide support for it.

RoyStegeman (approved):

PR #1999, opened 2024-03-12
Enable apple silicon

scarlehoff:
Given that we know now the code works in the M1/2/3 macs, that (almost) everyone in the collaboration with a mac uses one of those and that tensorflow will stop supporting the old macs from 2.17...
I think it is time to change the CI!

This is just a test to see whether it works.

It seems there is a problem with fiatlux in arm64. @niclaurenti this is probably the problem you mentioned the other day. Could you have a look to check whether the ci is installing the right version?

### DIFF
--- a/.github/workflows/python_installation.yml
+++ b/.github/workflows/python_installation.yml
@@ -10,8 +10,8 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-latest]
-        python-version: ["3.11", "3.12"]
+        os: [ubuntu-latest, macos-14]
+        python-version: ["3.12"]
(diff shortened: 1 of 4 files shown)


niclaurenti:
> It seems there is a problem with fiatlux in arm64. @niclaurenti this is probably the problem you mentioned the other day. Could you have a look to check whether the ci is installing the right version?

Hi @scarlehoff , yes the M1/M2 `fiatlux` (and also `adani`) packages have been fixed by me and @alecandido .
Regarding `fiatlux` I will try to do a new release in order to push it on pypi, but maybe I don't have the privileges.

PR #2004, opened 2024-03-14
don't compute predictions for all replicas if we only want replica0

RoyStegeman:


validphys2/src/validphys/results.py
```diff
@@ -307,6 +306,27 @@ def procs_data_values_experiment(proc_result_table_experiment):
...
+
+def group_result_central_table_no_table(groups_results_central, groups_index):
+    """Generate a table containing the data central value and the central prediction"""
+    result_records = []
+    for group_results in groups_results_central:
+        dt, th = group_results
+        for dt_central, th_central in zip(dt.central_value, th.central_value):
+            result_records.append(
+                dict([("data_central", dt_central), ("theory_central", th_central)])
+            )
```
scarlehoff, line 322:
```suggestion
            result_records.append({"data_central": dt_central, "theory_central": th_central})
```

I think you can get away with this since you don't need the extra item per replica of the total table (but ofc check please)

An easy test to add would be to check that indeed calling `group_result_central_table_no_table` returns the central value of `group_result_table_no_table`

andreab1997 (commented):
It seems fine but, remind me, what is it used in the thcovmat creation? I remember we faced this issue some months ago as well

RoyStegeman:
Indeed in the function that enables me to do this is the on created to make the thcovmat computation cheaper: https://github.com/NNPDF/nnpdf/pull/1863.

PR #2005, opened 2024-03-14
Fix a bunch of dataset names

scarlehoff:
The process for a few datasets is wrong:

```bash
CMS_Z0_8TEV_PT-Y: CMS_Z0J_8TEV_PT-Y
ATLAS_WMWP_13TEV_TOT: ATLAS_WPWM_13TEV_TOT
CMS_WP_7TEV_ELECTRON_ASY: CMS_WPWM_7TEV_ELECTRON_ASY
CMS_WP_7TEV_MUON_ASY: CMS_WPWM_7TEV_MUON_ASY
```

edit: regression tests will fail because the seed depends on the name


scarlehoff:
The question is... how should we deal with fixes to the names...

`dataset_names.yaml` can be abused for that, but it should also warn you when you are doing it so that it gets changed.

Radonirinaunimi (approved):
These were indeed an oversight. Thanks a lot for having checked and fixed these!

RoyStegeman:
Is there a baseline fit with these corrected that can be used for reports?

Radonirinaunimi:
> Is there a baseline fit with these corrected that can be used for reports?

I don't believe so, unless @scarlehoff already has one but not uploaded?

scarlehoff:
No. But anything with the old names will work.

PR #2011, opened 2024-03-18
Fix positivity names from 5 GeV to sqrt(5) GeV and DY -> Z0

scarlehoff:


RoyStegeman:
`GEV2` is not supported?

RoyStegeman:
`GEV2` is not supported?

scarlehoff:
No. And we already have the `P` syntax in a few places.

scarlehoff:
I haven't found any other problems...

 @RoyStegeman I might wait a few days in case I find more problems while removing the apfelcomb keys, and then I'll create a baseline with the new names

RoyStegeman:
> @RoyStegeman I might wait a few days in case I find more problems while removing the apfelcomb keys, and then I'll create a baseline with the new names

Thanks, there is not really a hurry. The reason I asked for such a fit is that I generally modify the runcard of the most recent baseline fit when I write runcards for anything else. So having such a fit helps with making the new dataset names the standard in fits from now on.

PR #2012, opened 2024-03-18
add pch NLO and NNLO theories

scarlehoff:
I never pushed these two...
but the fktables are in the server and the grids in https://github.com/NNPDF/theories_slim/pull/35

The report of 4.0 pch against this one: https://vp.nnpdf.science/I4yWhfTHTFy3UPg0gMM_IQ== (the biggest difference is the displaced charm threshold)


nnpdf_data/nnpdf_data/theory_cards/391.yaml
```diff
@@ -0,0 +1,48 @@
+ID: 391
+PTO: 1
+FNS: FONLL-B
+DAMP: 0
```
felixhekhorn, line 4: I think pch needs damping (here and below) cc @RoyStegeman

RoyStegeman: Yes, you're right. `pineko` takes as input the powers to the damping function so probably you want to add 
```
DAMPPOWERc: 2
DAMPPOWERb: 2
```

scarlehoff: But only at NLO ?

PR #2013, opened 2024-03-18
Remove apfelcomb flags

scarlehoff:
`repetition_flag` could be removed completely I think

`normalization` is used in silly ways in a couple of places but needed in others so it needs to stay anyway


RoyStegeman:
Rebased to get some broken imports from `nnpdf_data`

validphys2/src/validphys/commondataparser.py
```diff
@@ -48,11 +48,12 @@
 from validobj import ValidationError, parse_input
 from validobj.custom import Parser
 
+from nnpdf_data import new_to_legacy_map, path_commondata
```
RoyStegeman, line 51: Ah, `nnpdf_data` should now be added to the list of `known_first_party` in `pyproject.toml`. But then I tried and it doesn't seem to have any effect. I thought it would put it among reportengine and vp imports

validphys2/src/validphys/pineparser.py
```diff
@@ -300,11 +238,19 @@ def pineappl_reader(fkspec):
...
-    if fkspec.metadata.operation == "RATIO" and ndata == 1 and len(pines) == 1:
+    if fkspec.metadata.operation == "RATIO" and len(pines) == 1:
         # it _might_ be, check whether it is the divisor fktable
         divisor = fkspec.metadata.FK_tables[-1][0]
         name = fkspec.fkpath[0].name.replace(f".{EXT}", "")
-        protected = divisor == name
+
+        if np.allclose(sigma.loc[1:], 0.0):
+            # For old fktables, maybe the rest is filled with 0s
+            sigma = sigma.loc[0:0]
```
RoyStegeman, line 248: Could you add a comment about this syntax (i.e. making sure sigma remains a dataframe)

RoyStegeman (approved):

validphys2/src/validphys/commondataparser.py
scarlehoff: I think the blank space is due to the comment below, while the first party it autodiscovers (so we don't actually need to add vp/n3fit)

RoyStegeman: Ah yes, that might be it

PR #2014, opened 2024-03-20
Ignore 10% worst replicas in hyper loss

APJansen:
Basic implementation, not tested yet and the percentage is not configurable from the runcard.

### DIFF
--- a/n3fit/src/n3fit/hyper_optimization/rewards.py
+++ b/n3fit/src/n3fit/hyper_optimization/rewards.py
@@ -62,7 +62,8 @@ def _average_best(fold_losses: np.ndarray, percentage: float = 0.9, axis: int = 
         float: The average along the specified axis.
     """
     sorted_losses = np.sort(fold_losses, axis=axis)
-    best_losses = sorted_losses[: int(percentage * len(sorted_losses))]
+    num_best = int(np.ceil(percentage * len(sorted_losses)))
+    best_losses = sorted_losses[:num_best]
     return np.average(best_losses, axis=axis).item()
 
 
--- a/n3fit/src/n3fit/hyper_optimization/rewards.py
+++ b/n3fit/src/n3fit/hyper_optimization/rewards.py
@@ -44,17 +44,17 @@ from validphys.pdfgrids import distance_grids, xplotting_grid
 log = logging.getLogger(__name__)
 
 
-def _average_best(fold_losses: np.ndarray, percentage: float = 0.9, axis: int = 0) -> float:
+def _average_best(fold_losses: np.ndarray, proportion: float = 0.9, axis: int = 0) -> float:
     """
-    Compute the average of the input array along the specified axis, among the best `percentage`
+    Compute the average of the input array along the specified axis, among the best `proportion`
     of replicas.
 
     Parameters
     ----------
         fold_losses: np.ndarray
             Per replica losses for a single fold.
-        percentage: float
-            The percentage of best replicas to take into account (rounded up).
+        proportion: float
+            The proportion of best replicas to take into account (rounded up).
         axis: int, optional
             Axis along which the mean is computed. Default is 0.
 
@@ -62,13 +62,14 @@ def _average_best(fold_losses: np.ndarray, percentage: float = 0.9, axis: int = 
     -------
         float: The average along the specified axis.
     """
-    num_best = int(np.ceil(percentage * len(fold_losses)))
+    # TODO: use directly `validphys.fitveto.determine_vetoes`
+    num_best = int(np.ceil(proportion * len(fold_losses)))
 
     if np.isnan(fold_losses).any():
         log.warning(f"{np.isnan(fold_losses).sum()} replicas have NaNs losses")
     sorted_losses = np.sort(fold_losses, axis=axis)
     best_losses = sorted_losses[:num_best]
-    return np.average(best_losses, axis=axis).item()
+    return _average(best_losses, axis=axis)
 
 
 def _average(fold_losses: np.ndarray, axis: int = 0) -> float:
@@ -145,6 +146,13 @@ class HyperLoss:
 
     Computes the statistic over the replicas and then over the folds, both
     statistics default to the average.
+
+    The ``compute_loss`` method saves intermediate metrics such as the
+    chi2 of the folds or the phi regardless of the loss type that has been selected.
+    These metrics are saved in the properties
+        ``phi_vector``: list of phi per fold
+        ``chi2_matrix``: list of chi2 per fold, per replica
+
 
     Parameters
     ----------
@@ -304,12 +312,10 @@ class HyperLoss:
         if loss_type is None:
             loss_type = self._default_loss
             log.warning(f"No loss_type selected in HyperLoss, defaulting to {loss_type}")
-        else:
-            if loss_type not in IMPLEMENTED_LOSSES:
-                valid_options = ", ".join(IMPLEMENTED_LOSSES)
-                raise ValueError(
-                    f"Invalid loss type '{loss_type}'. Valid options are: {valid_options}"
-                )
+
+        if loss_type not in IMPLEMENTED_LOSSES:
+            valid_options = ", ".join(IMPLEMENTED_LOSSES)
+            raise ValueError(f"Invalid loss type '{loss_type}'. Valid options are: {valid_options}")
 
         log.info(f"Setting '{loss_type}' as the loss type for hyperoptimization")
 
@@ -341,12 +347,10 @@ class HyperLoss:
         if statistic is None:
             statistic = default
             log.warning(f"No {name} selected in HyperLoss, defaulting to {statistic}")
-        else:
-            if statistic not in IMPLEMENTED_STATS:
-                valid_options = ", ".join(IMPLEMENTED_STATS.keys())
-                raise ValueError(
-                    f"Invalid {name} '{statistic}'. Valid options are: {valid_options}"
-                )
+
+        if statistic not in IMPLEMENTED_STATS:
+            valid_options = ", ".join(IMPLEMENTED_STATS.keys())
+            raise ValueError(f"Invalid {name} '{statistic}'. Valid options are: {valid_options}")
 
         log.info(f"Using '{statistic}' as the {name} for hyperoptimization")
 
--- a/n3fit/src/n3fit/hyper_optimization/rewards.py
+++ b/n3fit/src/n3fit/hyper_optimization/rewards.py
@@ -52,8 +52,9 @@ def _average_best(fold_losses: np.ndarray, percentage: float = 0.9, axis: int = 
     Parameters
     ----------
         fold_losses: np.ndarray
-            Input array.
-        float: The percentage of best replicas to take into account.
+            Per replica losses for a single fold.
+        percentage: float
+            The percentage of best replicas to take into account (rounded up).
         axis: int, optional
             Axis along which the mean is computed. Default is 0.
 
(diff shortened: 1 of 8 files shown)


APJansen:
I have implemented what we discussed by removing the `passed` as it was created (unless there's no hyperopt, I left that case as is). Instead I've set passed just based on whether the hyper loss is bigger than the threshold, as if there are more than 10% replicas with infinite loss, this will automatically be violated.

I'm a bit confused now if that's what we want, as this check is already there of course. The difference is that the one I added does not include the penalties, and it sets the hyperopt status to failed, whereas the existing check just adds a penalty.


To make the tests pass I had to increase this hyper threshold, as it now fails the trial if it's violated. Can you check if that doesn't mess up other tests @Cmurilochem?


Also, it's running on the GPU again, without lhapdf and conda, see [here](https://github.com/LHC-NLeSC/nnpdf-scripts/tree/main/hyperopt-final) for the setup and slurm scripts.

Cmurilochem:
> I have implemented what we discussed by removing the `passed` as it was created (unless there's no hyperopt, I left that case as is). Instead I've set passed just based on whether the hyper loss is bigger than the threshold, as if there are more than 10% replicas with infinite loss, this will automatically be violated.
> 
> I'm a bit confused now if that's what we want, as this check is already there of course. The difference is that the one I added does not include the penalties, and it sets the hyperopt status to failed, whereas the existing check just adds a penalty.
> 
> To make the tests pass I had to increase this hyper threshold, as it now fails the trial if it's violated. Can you check if that doesn't mess up other tests @Cmurilochem?
> 
> Also, it's running on the GPU again, without lhapdf and conda, see [here](https://github.com/LHC-NLeSC/nnpdf-scripts/tree/main/hyperopt-final) for the setup and slurm scripts.

@APJansen. I do not expect any breaks on the other tests by increasing the hyper threshold. But they may take more time as you would be in principle fitting more folds than before. However, I am also a bit confused as to the reason behind this. 

If I understood well, in the past (not sure it still now), the `passed` and `not passed` had to do with [threshold_chi2](https://github.com/NNPDF/nnpdf/blob/8777835f3e9b2704c565f86cbede153723e6a0b7/n3fit/src/n3fit/model_trainer.py#L880) which was passed as an argument to instantiate the `Stopping` object; [here](https://github.com/NNPDF/nnpdf/blob/8777835f3e9b2704c565f86cbede153723e6a0b7/n3fit/src/n3fit/model_trainer.py#L965). This is used in its [monitor_chi2](https://github.com/NNPDF/nnpdf/blob/8777835f3e9b2704c565f86cbede153723e6a0b7/n3fit/src/n3fit/stopping.py#L405) method which is in turn used in [callback.py](https://github.com/NNPDF/nnpdf/blob/8777835f3e9b2704c565f86cbede153723e6a0b7/n3fit/src/n3fit/backends/keras_backend/callbacks.py#L88). But there is also [hyper_threshold](https://github.com/NNPDF/nnpdf/blob/8777835f3e9b2704c565f86cbede153723e6a0b7/n3fit/src/n3fit/model_trainer.py#L177C18-L177C33) which was compared with the final (statistic) `hyper_loss` (including penalties) of each fold and responsible to skip hyperparameter combination.  I see now that your `passed` is directly tied to `self.hyper_threshold`. I am quite sure that this is your intention that but even so I am also sure that this my discussion might help you in some way. Please, also have a look in [monitor_chi2](https://github.com/NNPDF/nnpdf/blob/8777835f3e9b2704c565f86cbede153723e6a0b7/n3fit/src/n3fit/stopping.py#L405). I see there the it is trying to deal with some `NaN`s.

n3fit/src/n3fit/hyper_optimization/rewards.py
```diff
@@ -43,6 +44,33 @@
...
+    -------
+        float: The average along the specified axis.
+    """
+    num_best = int(np.ceil(percentage * len(fold_losses)))
+
+    if np.isnan(fold_losses).any():
+        log.warning(f"{np.isnan(fold_losses).sum()} replicas have NaNs losses")
+    sorted_losses = np.sort(fold_losses, axis=axis)
+    best_losses = sorted_losses[:num_best]
+    return np.average(best_losses, axis=axis).item()
```
scarlehoff, line 71: We can do this for now, but I think ideally we should remove any replica more than X standard deviations from the average chi2 and if more than 10% are discarded then fail the test.

n3fit/src/n3fit/model_trainer.py
```diff
@@ -999,6 +991,15 @@ def hyperparametrizable(self, params):
                     for penalty in self.hyper_penalties
                 }
 
+                hyper_loss_per_replica = experimental_loss + sum(penalties.values())
+                fold_loss = self._hyper_loss.reduce_over_replicas(hyper_loss_per_replica)
+                passed = fold_loss < self.hyper_threshold
```
scarlehoff, line 996: I guess these should averaged over replicas and folds, to make sure that the threshold does not depend on the choice (of replicas/folds)

scarlehoff (commented):
I don't understand why the threshold has to go from 2 to 1000 ? Surely this has to do with some 1/N that has not been considered? (where N might be replicas, data points, fold, or a combination of the above)

n3fit/src/n3fit/hyper_optimization/rewards.py
RoyStegeman: Specifically, why don't we use the `validphys.fitveto.determine_vetoes` function with the same defaults as the `postfit` CLI for that? We know from nnpdf40 that only ~1% of replicas are discarded in postfit, so as an additional constraint we could decide to discard e.g. all setups that result in discarding more than 10% of replicas (or even fewer).

APJansen: > I don't understand why the threshold has to go from 2 to 1000 ? Surely this has to do with some 1/N that has not been considered? (where N might be replicas, data points, fold, or a combination of the above)

This is only in the hyperopt tests, what changes in this PR is not the value of the loss but what happens if it is violated, namely to fail the trial. The fact that it's violated is no surprise as it is run only for a small number of epochs in the test.

> Specifically, why don't we use the validphys.fitveto.determine_vetoes function with the same defaults as the postfit CLI for that?

You mean keep using the average in the loss as before, but in the passing criterion inside modeltrainer use this `determine_vetoes`? That's fine by me, not sure where to get these thresholds and the `fitinfos` from though.

n3fit/src/n3fit/model_trainer.py
APJansen: It's already reduced over replicas, in whatever way specified, so it shouldn't depend on the number of replicas. Regarding folds, I think there is also a normalization with respect to the number of datasets, no? So it should be fine?
You can also do it after all folds are done as you suggest, but then you won't be able to exit a trial early if the first fold was disastrous.

scarlehoff: Then you should not get to 1000...

Cmurilochem:
Hi @APJansen and @goord,
As I discussed before with you, I tried to make an hyperopt experiment and realized that most of the trials finish with {"status": "fail"} and  {"loss": NaN}. For instance, out of 85 trials calculated just one seems to have a loss lower than 10. Just in case, I attach below my partial `trial.json` file:
[tries.json](https://github.com/NNPDF/nnpdf/files/15036133/tries.json)

After looking at the [renew_hyperopt.yml](https://github.com/NNPDF/nnpdf/blob/hyper-selection/n3fit/runcards/hyperopt_studies/renew_hyperopt.yml#L164) runcard, I noticed that I should have added `average_best` as my `replica_statistic`. I overlooked it before as I am copying this runcard directly from the repo to my local directory before running the experiment in snellius.

1. @APJansen: to avoid confusion in the future, could we add `average_best` in `renew_hyperopt.yml`?:
https://github.com/NNPDF/nnpdf/blob/47a6898bd53bff165ca23f647f01bbf2938f2da7/n3fit/runcards/hyperopt_studies/renew_hyperopt.yml#L164
I could do that if you say so.

2. Turning back to the results of the experiments I have done using (without knowing) `average` as `replica_statistic`. I still cannot quite understand why so many trials are discarded. Our `passed` criterium seems to be much more restrictive than before. While looking at the code, I see
https://github.com/NNPDF/nnpdf/blob/47a6898bd53bff165ca23f647f01bbf2938f2da7/n3fit/src/n3fit/model_trainer.py#L996
and right after,
https://github.com/NNPDF/nnpdf/blob/47a6898bd53bff165ca23f647f01bbf2938f2da7/n3fit/src/n3fit/model_trainer.py#L1038
Both `fold_loss` and `hyper_loss` appear to represent reduced-over-replicas `hyper_loss`es for a specific fold.

n3fit/src/n3fit/model_trainer.py
Cmurilochem, line 1: Hi @APJansen. Still on the `{"loss": NaN}` subject. Now that we understand where this is coming from, would you agree to change it to `np.inf` in case `passed = fold_loss < self.hyper_threshold`?
If so, we could replace
https://github.com/NNPDF/nnpdf/blob/47a6898bd53bff165ca23f647f01bbf2938f2da7/n3fit/src/n3fit/model_trainer.py#L1058
by
```suggestion
if passed:
    # Compute the loss over all folds for hyperopt
    final_hyper_loss = self._hyper_loss.reduce_over_folds(l_hyper)
else:
    final_hyper_loss = np.inf
```

Cmurilochem: I believe that this would make more sense and eliminate this impression that something is "going wrong". If you agree, of course.

APJansen: But it can also fail to pass and still be finite, so I wouldn't do this. Note that `passed` also ends up in the trial's `"status"`, so you can see it from there.

Cmurilochem: Yes, I agree. But will this finite value be written into `tries.json`.
https://github.com/NNPDF/nnpdf/blob/1c3bb50828e4df655e72f70f39054468343bb352/n3fit/src/n3fit/model_trainer.py#L996-L1001
Here, if you `break`
https://github.com/NNPDF/nnpdf/blob/1c3bb50828e4df655e72f70f39054468343bb352/n3fit/src/n3fit/model_trainer.py#L1058
`final_hyper_loss` (the value which actually is written) will always be `NaN` anyways (`l_hyper` in this case is `[]`) ?

scarlehoff:
I'll rebase this on top of master to facilitate the review (it shouldn't make a big difference since the last rebase was not long ago)

n3fit/src/n3fit/hyper_optimization/rewards.py
```diff
@@ -192,10 +192,14 @@ def compute_loss(
         pdf_model: MetaModel,
         experimental_data: List[DataGroupSpec],
         fold_idx: int = 0,
+        include_penalties=False,
```
Cmurilochem, line 195: Hi @scarlehoff. Thank you for your amazing work in making this PR up. Would it be an option to `include_penalties=True` via runcard? I am asking this before pulling these changes to my snellius branch as the default `False` option would change the results I currently have.

scarlehoff: sure!
Since I'm at it, I'll also try to add the proportion of best replicas to the runcard so that it can be easily controlled.

scarlehoff: Added as `kfold:penalties_in_loss: True`

It defaults to `False`.

Please let me know if any of the changes I've done produce some effect somewhere else that I might have overlooked!

scarlehoff:
@Cmurilochem I've added a runcard option `penalties_in_loss`.

I'll add this to the docs. I've decided not to add the `proportion` to the runcard because it adds a bit of complexity (I need to add a custom parser in validphys) and thought that at this stage of the project it was better to skip it.

(However, since I've written already the code, if you would like to have access to the proportion or other arguments for the statistics let me know and I'll push it)

Cmurilochem:
> If @Cmurilochem confirms this PR works for him as it is I think it can be merged .

Hi @scarlehoff. Thanks for all your hard work. It looks great to me and it is working as expected. I just added a last commit in which I update our experiment's runcards adding a non-default `penalties_in_loss` for reproducibility. Please, let me know if you agree with that. For the moment I will take the lead and approving your changes.

Cmurilochem (approved):

PR #2016, opened 2024-03-21
Add unique log files for each mongo worker

Cmurilochem:
# Aim

Allow mongo workers to write output data to separate files.

# Idea

Redirect the output of each [subprocess.Popen](https://github.com/NNPDF/nnpdf/blob/760054c168848f8ff300cc1a5ef512093d825559/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py#L273-L275) to unique log files (`stdout`) for each mongo worker. These files are named `mongo-worker_n_output_path.log`, where `n=0, 1, 2, ..., n` is the unique id of the worker and `output_path` corresponds to the runcard name or `OUTPUT` if `n3fit -o OUTPUT` was used.

# Notes/Drawbacks

The `mongo-worker_n_output_path.log` files will only contain data output by any function/method called by `fmin`, *i.e.*, by [ModelTrainer.hyperparametrizable](https://github.com/NNPDF/nnpdf/blob/760054c168848f8ff300cc1a5ef512093d825559/n3fit/src/n3fit/model_trainer.py#L838) which consists of details of the hyperparameters being evaluated and info about epochs. Other infos and warnings are still printed to standard output (screen), **including trial numbers** which might not be ideal.

**Either way, it appears that with this simple change we can prevent the overwhelming/chaotic information that was previously displayed on the screen.**

# Tasks
- [x] Add `stdout` redirection files to `start_mongo_workers` Popen.
- [x] Add name of the runcard/output to the mongo_worker log file
- [x] Test in snellius

### DIFF
--- a/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
+++ b/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
@@ -155,7 +155,7 @@ class MongoFileTrials(MongoTrials):
             f"mongo://{self.db_host}:{self.db_port}/{self._process_db_name(self.db_name)}/jobs"
         )
         self.workers = []
-        self.job_output_name = replica_path.parts[-3]
+        self.output_folder_name = replica_path.parts[-3]
 
         self._store_trial = False
         self._json_file = replica_path / "tries.json"
--- a/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
+++ b/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
@@ -197,7 +197,6 @@ class MongoFileTrials(MongoTrials):
 
         # write json to disk
         if self._store_trial:
-            # log.info("Storing scan in %s", self._json_file)
             local_trials = []
             for idx, t in enumerate(self._dynamic_trials):
                 local_trials.append(t)
--- a/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
+++ b/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
@@ -266,7 +266,7 @@ class MongoFileTrials(MongoTrials):
                     # my_env["TF_GPU_ALLOCATOR"] = "cuda_malloc_async"
 
                 # create log files to redirect the mongo-workers output
-                mongo_workers_logfile = f"mongo-worker_{i+1}_{self.job_output_name}.log"
+                mongo_workers_logfile = f"mongo-worker_{i+1}_{self.output_folder_name}.log"
                 with open(mongo_workers_logfile, mode='w', encoding="utf-8") as log_file:
                     # run mongo workers
                     worker = subprocess.Popen(
--- a/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
+++ b/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
@@ -154,6 +154,7 @@ class MongoFileTrials(MongoTrials):
             f"mongo://{self.db_host}:{self.db_port}/{self._process_db_name(self.db_name)}/jobs"
         )
         self.workers = []
+        self.log_files = []
 
         self._store_trial = False
         self._json_file = replica_path / "tries.json"
@@ -195,7 +196,7 @@ class MongoFileTrials(MongoTrials):
 
         # write json to disk
         if self._store_trial:
-            log.info("Storing scan in %s", self._json_file)
+            # log.info("Storing scan in %s", self._json_file)
             local_trials = []
             for idx, t in enumerate(self._dynamic_trials):
                 local_trials.append(t)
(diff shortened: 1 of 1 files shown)


n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
```diff
@@ -264,21 +266,28 @@ def start_mongo_workers(
                     # avoid memory fragmentation issues?
                     # my_env["TF_GPU_ALLOCATOR"] = "cuda_malloc_async"
 
+                # create log files to redirect the mongo-workers output
+                mongo_workers_logfile = f"mongo-worker_{i+1}_{self.output_path}.log"
+                log_file = open(mongo_workers_logfile, 'w')
```
Radonirinaunimi, line 271: Can't you use a context manager with `with` here?

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
```diff
@@ -154,6 +154,8 @@ def __init__(
             f"mongo://{self.db_host}:{self.db_port}/{self._process_db_name(self.db_name)}/jobs"
         )
         self.workers = []
+        self.log_files = []
+        self.output_path = replica_path.parts[-3]
```
Radonirinaunimi, line 158: Why specifically at this location?

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
```diff
@@ -195,7 +197,7 @@ def refresh(self):
 
         # write json to disk
         if self._store_trial:
-            log.info("Storing scan in %s", self._json_file)
+            # log.info("Storing scan in %s", self._json_file)
```
Radonirinaunimi, line 200:
```suggestion

```

Radonirinaunimi (commented):
Thanks @Cmurilochem for this. This would be useful indeed to know what happened during a trial.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
Cmurilochem: Thanks @Radonirinaunimi for the very helpful comments. The reason is that these files cannot be closed until each mongo worker is done with its job. They are close here:
https://github.com/NNPDF/nnpdf/blob/4974bda5ec12dce1535b017fe6e06a993b2407b8/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py#L288-L290
after the mongo workers termininate.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
Cmurilochem: I define them together since these are related attributes used to define the workers list and corresponding output files:
https://github.com/NNPDF/nnpdf/blob/4974bda5ec12dce1535b017fe6e06a993b2407b8/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py#L270-L277

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
Radonirinaunimi: Yes, I indeed noticed that. And the reason why I was asking about the context manager was because I am a bit worried about the fact these files are only closed when the workers stopped. Since you are only flushing stdout, surely these can be appended.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
Radonirinaunimi: Sorry, my question wasn't clear. I was asking why the logs are saved in that particular path? Given that usually this will be run with more than one worker, why not collect them in some folder, for example?

Cmurilochem: Oh. Sorry. Now I see. `self.output_path ` is not a path but the name of the runcard or output name given by the user. So, to avoid this confusion I changed the name of this variable to `self.job_output_name`. Remember that `replica_path` is something like `path/RUNCARD_NAME/nnfit/replica_N`. So, `RUNCARD_NAME = replica_path.parts[-3]`.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
Cmurilochem: @Radonirinaunimi and @APJansen. It seems that the use of `with` while opening the mongo worker file just work. Although they are closed after the execution of the `start_mongo_workers` method, they are inherently tied to the subprocess and can be written as expected; see [here](https://github.com/NNPDF/nnpdf/blob/4bdcca4d7c04ef7140cdfc1178132a04f8e91845/n3fit/src/n3fit/hyper_optimization/mongofiletrials.py#L273). 

The only thing is that I would prefer opening them with `write` mode only as I foresee unexpected issues when using the `append` mode: *e.g.*: if you have the mongo logs in the same directory and want to rerun a previous calculation that gone wrong for any reason, the old files get appended with the new results. The drawback however it that if you want to restart your calculation, your mongo logs will be reset and only have info from the current run (not from the previous ones). But I think this is OK since all the info we need to make any post-processing of the data are contained in the `tries.json` and `hyperopt-db.tar.gz` files.

Radonirinaunimi: Thanks a lot @Cmurilochem! This is now much cleaner and (I'd argue) safer.

> The only thing is that I would prefer opening them with write mode only as I foresee unexpected issues when using the append mode: e.g.: if you have the mongo logs in the same directory and want to rerun a previous calculation that gone wrong for any reason, the old files get appended with the new results.

Yes, this is exactly the intended behavior and the fact that the logs are lost once restarting shouldn't be an issue at all (they can be copied elsewhere in-between) and as you said all the main information we need are saved correctly.

Radonirinaunimi (approved):
LGTM! Thanks again.

Cmurilochem:
> LGTM! Thanks again.

Thanks @Radonirinaunimi. Not sure if @RoyStegeman wants to have a look at it, but if not, could you merge it ?
It seems that I do not have permissions to so any more.

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
RoyStegeman, line 200:
```suggestion
```

n3fit/src/n3fit/hyper_optimization/mongofiletrials.py
```diff
@@ -154,6 +155,7 @@ def __init__(
             f"mongo://{self.db_host}:{self.db_port}/{self._process_db_name(self.db_name)}/jobs"
         )
         self.workers = []
+        self.job_output_name = replica_path.parts[-3]
```
RoyStegeman, line 158:
```suggestion
        self.output_folder_name = replica_path.parts[-3]
```

Perhaps calling it `output_folder_name` or `runcard_name` makes this a bit clearer.

RoyStegeman (approved):
Thanks @Cmurilochem! This looks good to me as well, feel free to merge after making these changes. If you agree with them of course

PR #2018, opened 2024-03-22
Separate the theorydb utilities into the data package

scarlehoff:
This is innocuous since the theorydb utils just define which entries the theory accepts.

This will make it easier for other programs (e.g., pineko) to use theories from here. The global definition of the theory can be done in the `_TheoryCard` dataclass. At the moment only the keys for NNPDF are there and they are all mandatory (even if it is not true).

### DIFF
--- a/doc/sphinx/source/data/data-config.rst
+++ b/doc/sphinx/source/data/data-config.rst
@@ -10,8 +10,8 @@ choices. In the code, every effort has been made to keep experimental and
 theoretical parameters strictly separate.
 In this section we shall specify the layout of the ``nnpdf`` data
 directory. It is in this directory that all of the read-only data to be used in
-the fit are accessed. The data directory is located in the ``nnpdf`` git
-repository, under the path ``validphys/src/validphys2/datafiles``.
+the fit are accessed. The data files is located in the ``nnpdf`` git
+repository, under the ``nnpdf_data`` package.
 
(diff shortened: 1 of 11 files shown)


felixhekhorn:
Do you want to do a further rename here `theorydbutils` -> `theory`?

scarlehoff:
No, at the moment externally these are just the utilities.
The proper `theory.py` can be born once someone creates the proper `Theory` class (at the moment there is a mock dataclass where I just declared all the fields that were in the sqlite database)

doc/sphinx/source/data/data-config.rst
```diff
@@ -47,7 +47,7 @@ The following lines will check whether a newly added theory can be read by valid
 ..  code-block:: python
     
     from nnpdf_data import theory_cards
-    from validphys.theorydbutils import fetch_theory
+    from nnpdf_data.theorydbutils import fetch_theory
     theory = fetch_theory(theory_cards, 700)
```
RoyStegeman, line 51: The full vp path is also mentioned a few lines above

scarlehoff: My sed-fu failed me :'(

thanks!

RoyStegeman (approved):
Other than the comment about the docs, the rest looks fine. Thanks!

PR #2021, opened 2024-03-25
add y labels, integrate utils in filters and remove comma in sys names v2

t7phy:
It seems there was an oversight on my part when opening #2020 that I hadn't noticed that nnpdf_data was already available as a standalone package that can be used while importing. This ofc allows (without any temporary breaking of filter files) to import utils from 1 place, as was agreed in CM. ATLAS_TTBAR_13TEV_HADR_DIF contains these utils and all the other ttb/jets/dis+j call whichever functions they need from here. This reduces code repetition and makes the review simpler.
To summarize:
- this PR adds y_label where we had discovered they were missing from my data implementations, here is a report to test the labels work: https://vp.nnpdf.science/MzdicX55Ti2M8PlOyp26Hg==/
- uncert. with names `sys,*` are changed to `syst_*`
- `from validphys.commondata_utils` is replaced with `from nnpdf_data.new_commondata.ATLAS_TTBAR_13TEV_HADR_DIF.utils` and I have explicitly checked rerunning the filters and they work.
- #1693 and #2020 are now closed.


nnpdf_data/nnpdf_data/new_commondata/ATLAS_TTBAR_8TEV_LJ_DIF/filter.py
```diff
@@ -444,4 +446,12 @@ def processData():
...
 
+def remove_commas():
+    pattern = "uncertainties*.yaml"
+    reg = re.compile(fr'({"sys,"})')
+    for file in Path(".").glob(pattern):
+        new_text = reg.sub("syst_", file.read_text())
+        file.write_text(new_text)
+
 processData()
+remove_commas()
```
RoyStegeman, line 457: Could you run black black/isort again on these files?

t7phy: is it fine if I run black and isort on the entire new commondata itself? Because I see that when I do run them, they end of changing reformatting almost every new data implementation filter.

RoyStegeman: Since no-one else is touching those files now, so you may as well. Makes life easier when reviewing those later since it keeps the diffs cleaner.

If that had not been the case, something like this will allow you to just run it on the files changed in a given branch `git diff --name-only origin/HEAD | xargs black`

Radonirinaunimi (approved):
Thanks! LGTM now.

Issue #1974, opened 2024-03-03
Removing traces of Old Commondata


Radonirinaunimi:
Once https://github.com/NNPDF/nnpdf/pull/1678 and https://github.com/NNPDF/nnpdf/pull/1708 are merged, we should remove all the traces of the old commondata. This includes:

- [ ] removing the `buildmaster` folder
- [ ] remnants of the old commondata in parsing/loading of `vp`
- [ ] ...

scarlehoff:
@RoyStegeman @Radonirinaunimi 

just to make sure that several of us don't end up working on the same stuff, if you want to tackle one of these issues either open an issue and assign yourself or put your name here in the list

in any case, next Wednesday I was planning to go through the list and assign tasks to everyone

RoyStegeman:
I assigned myself to updating the dataset names in the examples, so if I have time to work on any of these things, that's the one I'll address first.

Radonirinaunimi:
I will assign myself starting from the first one on this list and the two issues #1940 and #1990.

PR #2023, opened 2024-03-26
Fix kinematic vs kinematic_coverage check

Radonirinaunimi:
Currently, when `kinematic_coverage` does not contain 3 variables and `len(kinematics.variable)<=3`, `extra_` variables are added to comply with vp.
https://github.com/NNPDF/nnpdf/blob/8b9b29fbbb55a59d6eb2d459307bc30ede977cbb/validphys2/src/validphys/commondataparser.py#L399

This, however, could be inconsistent with the following check:
https://github.com/NNPDF/nnpdf/blob/8b9b29fbbb55a59d6eb2d459307bc30ede977cbb/validphys2/src/validphys/commondataparser.py#L435-L439

This is a minimal fix by basically skipping `extra_` whenever `kinematics` and `kinematic_coverage` are compared.

### DIFF
--- a/validphys2/src/validphys/commondataparser.py
+++ b/validphys2/src/validphys/commondataparser.py
@@ -433,9 +433,9 @@ class ObservableMetaData:
 
         # Ensure that all variables in the kinematic coverage exist
         for var in self.kinematic_coverage:
-            if var not in self.kinematics.variables:
+            if var not in self.kinematics.variables and not var.startswith("extra_"):
                 raise ValidationError(
-                    f"Variable {var} is in `kinematic_coverage` but not included in `kinematics` for {self.name}"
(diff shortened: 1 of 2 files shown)


scarlehoff (approved):
thanks!
I guess you were the first one trying to use a dataset with only two variables :__)

We need to get rid of this limitation of "3 and only 3" though...

Radonirinaunimi:
> thanks! I guess you were the first one trying to use a dataset with only two variables :__)
> 
> We need to get rid of this limitation of "3 and only 3" though...

Yes, I fully agree! In the meantime, I just added it to the holy list https://github.com/NNPDF/nnpdf/issues/1974 xd

PR #2026, opened 2024-03-27
add .pre-commit-config.yaml

RoyStegeman:
I did not keep `known_first_party`. I tested it on `model_trainer.py` and it put the `numpy` import inside the `n3fit` imports while separating `from n3fit import model_gen` (not sure what's going on there). I also realised that we missed e.g. `ekobox`. 

`isort` seems to recognise that `nnpdf_data`, `validphys` and `n3fit` are part of this repo so I think that's good enough (let's keep an eye on it though). 

I'm not if favour of a CI test


scarlehoff:
Why did you remove it from the global .toml, wouldn't it be better to have it there?

If black and isort from pre-commit don't recognize it (they should) you can explictly say `--config pyproject.toml` or similar.

That way one can use black on isort with your ide and make sure that it is using the same options.

(I guess we want to also check -but not autocommit- as part of the checks)

RoyStegeman:
> Why did you remove it from the global .toml, wouldn't it be better to have it there?

Because pre-commit cannot use the pyproject.toml file https://github.com/pre-commit/pre-commit/issues/1165

> If black and isort from pre-commit don't recognize it (they should) you can explictly say --config pyproject.toml or similar.

I thought there was no such option for isort and didn't really see the benefit. But I guess if you want to run it from the ide that's indeed not possible this way. I was wrong though, and isort does have an option `--settings-path` that I think do the job so I'll change it again.

> (I guess we want to also check -but not autocommit the fixes- as part of the checks)

I'm not so sure, there may be cases in which we don't want to do exactly what black or isort wants. Or maybe not. Either way, let's see, if people adopt pre-commit problems about forgetting to run black should become rare.

scarlehoff:
I don't know about pre-commit, but the tools you run inside certainly can (this is used in pineko or pinefarm)

RoyStegeman:
Not for black and isort

RoyStegeman:
I still need to update the mentions of black and isort in the docs, but I'll do that another time.

scarlehoff (approved):
thanks!
Since you are at it, it might be a good idea to remove the link to the old vp guide from https://github.com/NNPDF/nnpdf/tree/master/doc#currently-documented

(I think we already moved part of the stuff to the standard docs)

scarlehoff (approved):

Issue #2019, opened 2024-03-22
Create a single source of truth for all theories


scarlehoff:
In #2018 the theory db utils are separated into the data package, like this external programs can easily depend not only on the theory cards but also in the definition of the theory.

My suggestion is something very simple, just a dataclass declaring the field, with several which will mandatory for everyone and some that will be optional (and for which can either set a default here or set `None` and let the decision to each program).

```python
class _TheoryCard:
    ID: int
    PTO: int
    FNS: str
    DAMP: int
    IC: int
    ModEv: str
    XIR: float
    XIF: float
    NfFF: int
```

PR #2028, opened 2024-03-28
Provide an unique truth for all theory cards

andreab1997:
Here we address #2019.

- [x] Create a `dataclass` collecting all the possible parameters that can appear in a theory card, removing the ones that are not used anymore
- [ ] Add utilities function to read a theory which can be also used by `pineko`
- [ ] Provide a way to read and/or search in all theories

cc @scarlehoff


andreab1997:
At the moment if you see `#` it means that I am not sure if the parameter is still used or not. I am adding now comments and questions for all the problematic ones

nnpdf_data/theory.py
```diff
@@ -0,0 +1,59 @@
...
+class _TheoryCard:
+    ID: int
+    PTO: int
+    FNS: str
+    DAMP: int
+    DAMPPOWERc: None #Power of the damping factor in FONLL for the c
+    DAMPPOWERb: None #Power of the damping factor in FONLL for the b
+    IC: int
+    IB: int
+    ModEv: str
```
andreab1997, line 20: This I am not sure is used anymore given that EKO gets this through the operator card (@felixhekhorn please confirm this) and in pineko this field is not read at all. It might be still nice to have just to remember.

nnpdf_data/theory.py
```diff
@@ -0,0 +1,59 @@
...
+    ModSV: None#
```
andreab1997, line 21: This is actually used in `pineko` ([here](https://github.com/NNPDF/pineko/blob/4d1e3184038573b452c7addb48fd36ef85cafc7c/src/pineko/evolve.py#L26)).

nnpdf_data/theory.py
```diff
@@ -0,0 +1,59 @@
...
+    XIR: float
+    XIF: float
+    NfFF: int
+    #MaxNfAs: int
```
andreab1997, line 25: Both this and MaxNfPdf I believe are not actually used anymore in EKO (I remember a discussion in Amsterdam). Can you remind me? @giacomomagni @felixhekhorn

nnpdf_data/theory.py
```diff
@@ -0,0 +1,59 @@
...
+    #MaxNfPdf: int
+    nfref: 5
+    nf0: 3
+    Q0: float
+    alphas: float
+    Qref: float #TODO: understand if we want to enforce Qref = Qedref and remove Qedref. Note that this is mandatory in eko
+    QED: int
+    alphaqed: float
+    Qedref: float
```
andreab1997, line 34: Here, we probably should remove `Qedref` and assume always `Qedref=Qref` because (source @niclaurenti ) this is what is already enforced in EKO

nnpdf_data/theory.py
```diff
@@ -0,0 +1,59 @@
...
+    SxRes: int#
```
andreab1997, line 35: Both this and the following I believe are not used anymore (confirm this please @felixhekhorn @giacomomagni @alecandido )

nnpdf_data/theory.py
```diff
@@ -0,0 +1,59 @@
...
+    SxOrd: str#
+    HQ: str
+    mc: float
+    Qmc: float#
```
andreab1997, line 39: Can you remind me where `Qmc`, `Qmb` and `Qmt` are used? @felixhekhorn

nnpdf_data/theory.py
```diff
@@ -0,0 +1,59 @@
...
+    Qmb: float#
+    kbThr: float
+    mt: float
+    Qmt: float#
+    ktThr: float
+    CKM: list[float]
+    MZ: float
+    MW: float
+    GF: float
+    SIN2TW: float#
```
andreab1997, line 51: This I really do not know if it is used anymore or not...

nnpdf_data/theory.py
```diff
@@ -0,0 +1,59 @@
...
+    TMC: int
+    MP: float
+    Comments: str
+    global_nx: int#
+    #EScaleVar: int
```
andreab1997, line 56: As far as I remember this is not used anymore because it was something coming from APFEL (@felixhekhorn )

nnpdf_data/theory.py
```diff
@@ -0,0 +1,59 @@
...
+    ktThr: float
+    CKM: list[float]
+    MZ: float
+    MW: float
+    GF: float
+    SIN2TW: float#
+    TMC: int
+    MP: float
+    Comments: str
+    global_nx: int#
```
andreab1997, line 55: This I guess is in the same situation of SxRes

nnpdf_data/theory.py
felixhekhorn: this is used in yadism: https://github.com/NNPDF/yadism/blob/9db321d2f3496d8ba85ee9b7587b1b7231d8ecc3/src/yadism/coefficient_functions/coupling_constants.py#L186

nnpdf_data/theory.py
```diff
@@ -0,0 +1,59 @@
...
+    #EScaleVar: int
+    FactScaleVar: None #used in yadism to allow for Fact scale var
+    RenScaleVar: None #used in yadism to allow for Ren scale var
+    n3lo_cf_variation : None
```
felixhekhorn, line 59: we're at least missing `use_fhmruvv`, right @giacomomagni ?

nnpdf_data/theory.py
felixhekhorn: In eko [here](https://github.com/NNPDF/eko/blob/be97472bfe199fbf105d19d4b092aaa757488852/src/eko/io/runcards.py#L219), these are the reference scales for the MSbar masses

nnpdf_data/theory.py
felixhekhorn: both are still present in master [here](https://github.com/NNPDF/eko/blob/be97472bfe199fbf105d19d4b092aaa757488852/src/eko/io/runcards.py#L199), but are pending to be removed in https://github.com/NNPDF/eko/pull/295 (if ever we merge that PR :see_no_evil: )

nnpdf_data/theory.py
felixhekhorn: as we said: we should change pineko to propagate the information coming from here to the observable card (overwriting whatever is set there)

nnpdf_data/theory.py
```diff
@@ -0,0 +1,59 @@
...
+This module provides an unique source and definition for all the possible parameters
+that a theory card can contain. 
+It also implement some utilities function to manage the theory cards. 
+"""
+
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class _TheoryCard:
```
felixhekhorn, line 11:
```suggestion
class TheoryCard:
```
we want to expose this

nnpdf_data/theory.py
andreab1997: Ok so I'd say we remove them from here

nnpdf_data/theory.py
giacomomagni: yes

andreab1997: Ok thanks, I am adding it

nnpdf_data/theory.py
andreab1997: Do you agree @felixhekhorn @giacomomagni ?

giacomomagni: I think `Qedref` is not used anymore in eko.

andreab1997: Ok that is my point. I would remove it then.

scarlehoff:
Thanks @andreab1997, let's have next week as the target to merge this

With respect to the missing items:

> Add utilities function to read a theory which can be also used by pineko

I think this can be decoupled

> Provide a way to read and/or search in all theories

I'll add this to https://github.com/NNPDF/nnpdf/issues/2037 (but I think it can also be decoupled)

andreab1997:
> Thanks @andreab1997, let's have next week as the target to merge this
> 
> With respect to the missing items:
> 
> > Add utilities function to read a theory which can be also used by pineko
> 
> I think this can be decoupled
> 
> > Provide a way to read and/or search in all theories
> 
> I'll add this to #2037 (but I think it can also be decoupled)

Ok agreed. Do you want me to do something before merging it?

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,52 @@
...
+    CKM: list[float]
+    MZ: float
+    MW: float
+    GF: float
+    SIN2TW: float
+    TMC: int
+    MP: float
+    Comments: str
+    FactScaleVar: None #used in yadism to allow for Fact scale var
+    RenScaleVar: None #used in yadism to allow for Ren scale var
```
scarlehoff, line 50:
```suggestion
    FactScaleVar: bool = False #used in yadism to allow for Fact scale var
    RenScaleVar: bool = False #used in yadism to allow for Ren scale var
```

But if it is used only for yadism I don't think it should be part of the theory card, rather the program calling yadism (`pinefarm`) should be able to guess these from `xif/xir`? No?
(or... why is it not always set to True, we want to save that information when we create grids...)

scarlehoff (commented):
Some comments.

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,52 @@
...
+@dataclass(frozen=True)
+class TheoryCard:
+    ID: int
+    PTO: int
+    FNS: str
+    DAMP: int
+    DAMPPOWERc: None #Power of the damping factor in FONLL for the c
+    DAMPPOWERb: None #Power of the damping factor in FONLL for the b
+    IC: int
+    ModEv: str
```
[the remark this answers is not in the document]
andreab1997: Agreed

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+@dataclass(frozen=True)
+class TheoryCard:
+    ID: int
+    PTO: int
+    FNS: str
+    DAMP: int
+    DAMPPOWERc: None #Power of the damping factor in FONLL for the c
+    DAMPPOWERb: None #Power of the damping factor in FONLL for the b
```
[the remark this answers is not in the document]
andreab1997: It depends. Are we considering NNLO a default? (with respect to NLO for example). In that case, yes

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,52 @@
...
+    nf0: 3
+    Q0: float
+    alphas: float
+    Qref: float
+    QED: int
+    alphaqed: float
+    HQ: str
+    mc: float
+    Qmc: float
+    kcThr: float
```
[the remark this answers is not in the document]
andreab1997: I do not think so. From what I understood, `mc` is the pole mass, `Qmc` is the reference for the MS mass and `kcThr` is the threshold ratio. They are really three decoupled parameters

nnpdf_data/nnpdf_data/theory.py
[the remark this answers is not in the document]
scarlehoff: We can have different defaults according to `PTO` in `__post_init__`.
But if they depend maybe it is better if they default to `int = None` (and then the default can be applied later).

nnpdf_data/nnpdf_data/theory.py
andreab1997: In my opinion, this information should not be guessed from `xir` and `xif` because I agree with you that we should always ask for scale variations (i.e. they should be True by default). However, this is a default we should enforce in `yadism` or `pinefarm` and, once we do it there, we can remove these two parameters from the theory card

nnpdf_data/nnpdf_data/theory.py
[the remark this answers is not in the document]
scarlehoff: Ah, I thought Qmc = kcThr*mc. Then yes, we need to keep them separated.

(there should be a docstring or comment for every parameter I guess ^^U)

nnpdf_data/nnpdf_data/theory.py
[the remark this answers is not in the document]
andreab1997: I agree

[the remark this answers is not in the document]
andreab1997: [0655ba5](https://github.com/NNPDF/nnpdf/pull/2028/commits/0655ba5db3e823b6ecfe67bf5b43eb8bff470de0)

nnpdf_data/nnpdf_data/theory.py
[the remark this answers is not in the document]
andreab1997: In [6fb5b6d](https://github.com/NNPDF/nnpdf/pull/2028/commits/6fb5b6d1a8a3fc80f415fd3c52eae3b23100ff98) I added comments for almost all the parameters :)

nnpdf_data/nnpdf_data/theory.py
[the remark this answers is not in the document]
giacomomagni: `IterEv` shouldn't be placed in the operator card?

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,52 @@
...
+    XIF: float
+    NfFF: int
+    nfref: 5
+    nf0: 3
+    Q0: float
+    alphas: float
+    Qref: float
+    QED: int
+    alphaqed: float
+    HQ: str
```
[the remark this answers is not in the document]
giacomomagni: `HQ` is the heavy quark mass scheme, it can be `POLE` default or `MSBAR` (supported only in the evolution)

nnpdf_data/nnpdf_data/theory.py
giacomomagni: So this is used to save time. `RenScaleVar: true` can be used to generate all the scale variations even if `xir: 1.0`, 
so the 2 keys are decoupled. 
If you want to set it automatically, this should be true if the theory is associated to scale varied theories [here](https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/scalevariations/scalevariationtheoryids.yaml)

nnpdf_data/nnpdf_data/theory.py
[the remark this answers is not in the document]
alecandido: As `ModEv` (`evolution_method` in EKO) and `ModSV` (`scvar_method` in EKO). They are all part of the operator card.

But I have the feeling that they are ignoring the distinction, and flattening everything...

nnpdf_data/nnpdf_data/theory.py
alecandido: Well, as in the comment above, the theory parameters here are mixed with whatever program configurations (I would have limited theory to `MZ`, `alphas_ref`, and similar).

In any case, yadism is not special in any way, at least from the point of view of Pineko (and whatever else is downstream). Since you're working with PineAPPL grids, you only care about the structure.

However, in general I agree about the decoupling, and the option could be useful to include or not the scale variations in the grid, for whatever MC.
Not all the MC will have the option in their API, but it could also be implemented (and, if they are always including scale variations, they could be dropped from the grid in Pinefarm, after generation).

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,53 @@
...
+
+@dataclass(frozen=True)
+class TheoryCard:
+    ID: int # ID number of the theory
+    PTO: int # Perturbative order (0 = LO, 1 = NLO, 2 = NNLO ...)
+    FNS: str # Flavor number scheme (i.e. FONLL-C)
+    DAMP: int # Whether a damping function is applied or not for FONLL
+    DAMPPOWERc: int = None #Power of the damping factor in FONLL for the c
+    DAMPPOWERb: int = None #Power of the damping factor in FONLL for the b
+    IC: int # Intrinsic charm
```
giacomomagni, line 18:
```suggestion
    IC: int # 0 = perturbative charm only , 1 = intrinsic charm allowed
```

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,53 @@
...
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class TheoryCard:
+    ID: int # ID number of the theory
+    PTO: int # Perturbative order (0 = LO, 1 = NLO, 2 = NNLO ...)
+    FNS: str # Flavor number scheme (i.e. FONLL-C)
+    DAMP: int # Whether a damping function is applied or not for FONLL
+    DAMPPOWERc: int = None #Power of the damping factor in FONLL for the c
```
giacomomagni, line 16:
```suggestion
    DAMPPOWERc: int = None # Power of the damping factor in FONLL for the c
```

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,53 @@
...
+    DAMPPOWERb: int = None #Power of the damping factor in FONLL for the b
```
giacomomagni, line 17:
```suggestion
    DAMPPOWERb: int = None # Power of the damping factor in FONLL for the b
```

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,53 @@
...
+    IC: int # Intrinsic charm
+    ModEv: str # DGLAP evolution solution method (EXA or TRN)
+    IterEv: int = None # Number of iterations for the evolution of the PDF. Defaults to 40 when ModEv = EXA
+    ModSV: str = None # Scale variations method in EKO (expanded or exponentiated)
+    XIR: float # Renormalization scale ratio
```
giacomomagni, line 22:
```suggestion
    XIR: float # Renormalization scale over the hard scattering scale rato
```
Same for `XIF`. To be more clear for the future...

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,53 @@
...
+    XIF: float # Factorization scale ratio
+    NfFF: int # Number of active flavors
```
giacomomagni, line 24:
```suggestion
    NfFF: int # Number of active flavors, only for FFNS or FFN0 schemes
```

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,53 @@
...
+    nfref: 5 # Number of active flavors at Qref
+    nf0: 3 # Number of active flavors at the parametrization scale Q0
+    Q0: float # Parametrization scale
+    alphas: float # Value of alpha_s at the scale Qref
+    Qref: float # Reference scale for alphas and alphaqed
+    QED: int # Whether QED effects are taken into account
```
giacomomagni, line 30:
```suggestion
    QED: int # QED correction to strong coupling: 0 = disabled , 1 = allowed
```

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,53 @@
...
+    alphaqed: float # Values of alpha QED at the scale Qref
+    HQ: str
+    mc: float # Pole mass of the charm
```
giacomomagni, line 33:
```suggestion
    mc: float  # [GeV] charm mass
```
Maybe let's write down that all the quantities are in GeV?

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,53 @@
...
+    kbThr: float # Threshold ratio of the bottom
+    mt: float # Pole mass of the top
+    Qmt: float # MSbar mass reference scale of the top
+    ktThr: float # Threshold ratio of the top
+    CKM: list[float] # CKM matrix elements
+    MZ: float # Mass of Z
+    MW: float # Mass of W
+    GF: float # Fermi constant
+    SIN2TW: float
+    TMC: int # Time like
```
giacomomagni, line 47:
```suggestion
    TMC: int # Include target mass corrections: 0 = disabled, 1 = leading twist, 2 = higher twist approximated, 3 = higher twist exact
```

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,53 @@
...
+    MP: float # Mass of the proton
+    Comments: str # Comments on the theory
+    FactScaleVar: None #used in yadism to allow for Fact scale var
+    RenScaleVar: None #used in yadism to allow for Ren scale var
+    n3lo_cf_variation : None # Variation of the N3LO coefficient function approximation
```
giacomomagni, line 52:
```suggestion
    n3lo_cf_variation : int = 0 # N3LO coefficient functions variation: -1 = lower bound, 0 = central, 1 = upper bound
```

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,53 @@
...
+    use_fhmruvv: None
```
giacomomagni, line 53:
```suggestion
    use_fhmruvv: None # N3LO splitting functions approximation: if True use the FHMRUVV parametrization, otherwise use EKO parametrization.
```

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,53 @@
...
+    DAMPPOWERc: int = None #Power of the damping factor in FONLL for the c
+    DAMPPOWERb: int = None #Power of the damping factor in FONLL for the b
+    IC: int # Intrinsic charm
+    ModEv: str # DGLAP evolution solution method (EXA or TRN)
+    IterEv: int = None # Number of iterations for the evolution of the PDF. Defaults to 40 when ModEv = EXA
+    ModSV: str = None # Scale variations method in EKO (expanded or exponentiated)
+    XIR: float # Renormalization scale ratio
+    XIF: float # Factorization scale ratio
+    NfFF: int # Number of active flavors
+    nfref: 5 # Number of active flavors at Qref
```
giacomomagni, line 25: If we provide a default for the nf, we should give a default also for the scale.

nnpdf_data/nnpdf_data/theory.py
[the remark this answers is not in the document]
giacomomagni: Actually the idea was to keep here all the parameters that can actually distinguish a theory from another one, so if we have different theories differing only for `IterEv`, let's keep it here, although it's not fully consistent for instance with `polarized`.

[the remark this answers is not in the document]
scarlehoff: Maybe we should bring here polarized as well.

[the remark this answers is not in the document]
giacomomagni: Our idea for keeping it in the operator card was that at some point we'd like to have mixed pol, unpol theories and observables (as the g1/F1 ratio), same holds for timelike.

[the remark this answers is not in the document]
scarlehoff: The information on whether it is a mixed theory is also relevant. Note that these should be the same theory card for pineko / pinefarm and nnpdf.
The idea is that pineko / pinefarm (and nnpdf) use the information here to pass through to the interested parties what they need (so here it can be called `polarized: mixed/always/never` for instance, doesn't need to match -and shouldn't- the parameter in eko)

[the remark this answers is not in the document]
andreab1997: Indeed, even if a parameter belongs to the operator card, we should also keep it here and then `pineko` will take care of reading it from the theory card and writing it on the operator card. Otherwise we will keep having the same problem, that some of the parameters are hidden in a file which has no standards and which is not stored in NNPDF.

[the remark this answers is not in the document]
giacomomagni: I see your point, then polarized really depends on the observables, so I'd not add it here.

scarlehoff:
Are we happy with the current set of parameters? Should we move forward?

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@dataclass(frozen=True)
+class TheoryCard:
+    ID: int # ID number of the theory
+    PTO: int # Perturbative order (0 = LO, 1 = NLO, 2 = NNLO ...)
+    FNS: str # Flavor number scheme (i.e. FONLL-C)
+    DAMP: int # Whether a damping function is applied or not for FONLL
+    DAMPPOWERc: int = None # Power of the damping factor in FONLL for the c
+    DAMPPOWERb: int = None # Power of the damping factor in FONLL for the b
```
scarlehoff, line 17: All the optional ones need to go to the end (or, alternatively, we can add a default to every option)

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,51 @@
...
+    IC: int # 0 = perturbative charm only , 1 = intrinsic charm allowed
+    ModEv: str # DGLAP evolution solution method (EXA or TRN)
+    IterEv: int = None # Number of iterations for the evolution of the PDF. Defaults to 40 when ModEv = EXA
+    ModSV: str = None # Scale variations method in EKO (expanded or exponentiated)
+    XIR: float # Renormalization scale over the hard scattering scale ratio
+    XIF: float # Factorization scale over the hard scattering scale ratio
+    NfFF: int # Number of active flavors, only for FFNS or FFN0 schemes
+    nfref: int # Number of active flavors at Qref
+    nf0: 3 # Number of active flavors at the parametrization scale Q0
```
scarlehoff, line 26:
```suggestion
    nf0: int = 3 # Number of active flavors at the parametrization scale Q0
```
(and put at the end)

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,51 @@
...
+    CKM: list[float] # CKM matrix elements
+    MZ: float # [GeV] Mass of Z
+    MW: float # [GeV] Mass of W
+    GF: float # Fermi constant
+    SIN2TW: float
+    TMC: int # Include target mass corrections: 0 = disabled, 1 = leading twist, 2 = higher twist approximated, 3 = higher twist exact
+    MP: float # [GeV] Mass of the proton
+    Comments: str # Comments on the theory
+    n3lo_cf_variation : int = 0 # N3LO coefficient functions variation: -1 = lower bound, 0 = central, 1 = upper bound
+    use_fhmruvv: None # N3LO splitting functions approximation: if True use the FHMRUVV parametrization, otherwise use EKO parametrization.
```
scarlehoff, line 51:
```suggestion
    use_fhmruvv: bool = False # N3LO splitting functions approximation: if True use the FHMRUVV parametrization, otherwise use EKO parametrization.
```

scarlehoff:
Thanks.

I've tested this with pineko and just a few comments:

I need to add these keys so that old runcards can be read:
```
    ###### Keys for compatibility with old NNPDF theories
    MaxNfAs: int = 5
    SxRes: int = 0
    SxOrd: str = "LL"
    EScaleVar: int = 1
    Qedref: float = 91.0
    global_nx: int = 0
    MaxNfPdf: int = 5
    #### and this key needs a default 
    nfref: int = 5 # Number of active flavors at Qref
```

These are not used, but just so that we can load the old files with no changes. Since a default for `nfref` is dangerous, we might want to have a check in `__post_init__` so that it is correct/consistent.


Then, there are some problems with changes on how ruamel works in the new versions, so in the `utils.py` the yaml reading should be using the `CLoader`.

```utils.py
inp = yaml.load(input_yaml.read_text(encoding="utf-8"), Loader=yaml.CLoader)
```
instead of `inp = yaml.round_trip_load(input_yaml.open("r", encoding="utf-8"))`. Otherwise the float get read as a some ruamel_blabla_float (maybe there's another way to fix  that but I haven't looked more into it)

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,51 @@
...
+    TMC: int # Include target mass corrections: 0 = disabled, 1 = leading twist, 2 = higher twist approximated, 3 = higher twist exact
+    MP: float # [GeV] Mass of the proton
+    Comments: str # Comments on the theory
+    IterEv: int = None # Number of iterations for the evolution of the PDF. Defaults to 40 when ModEv = EXA
+    ModSV: str = None # Scale variations method in EKO (expanded or exponentiated)
+    DAMPPOWERc: int = None # Power of the damping factor in FONLL for the c
+    DAMPPOWERb: int = None # Power of the damping factor in FONLL for the b
+    n3lo_cf_variation : int = 0 # N3LO coefficient functions variation: -1 = lower bound, 0 = central, 1 = upper bound
+    use_fhmruvv: bool = False # N3LO splitting functions approximation: if True use the FHMRUVV parametrization, otherwise use EKO parametrization.
+    nf0: int = 3 # Number of active flavors at the parametrization scale Q0
```
giacomomagni, line 51: I'm not sure why we provide a default of `nf0` but not of `Q0` ...

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,51 @@
+"""
+This module provides an unique source and definition for all the possible parameters
```
giacomomagni, line 2:
```suggestion
This module provides an unique source and definition for some the possible parameters
```
These are not all the one that can be specified right?

nnpdf_data/nnpdf_data/theory.py
scarlehoff: While I agree that for consistency we should, there are some parameters for which we need a default just so the old runcards still work.

nnpdf_data/nnpdf_data/theory.py
scarlehoff: Let's say for "all the parameters that define a theory".

The importance people want to give to each parameter is up to people. And it is up to the interface between nnpdf and eko (i.e., pineko) to decide how the parameters are translated. So the parameters from eko don't need to have a duplicate here. But if you would need them to define and reproduce a theory, it would be better if it is here as well.

If `eko` decides to change how the parameter is called, here it will still be called `n3lo_ad_variation` and it would be up to `pineko` how to translate it.

nnpdf_data/nnpdf_data/theory.py
giacomomagni: Then I'd say, let's add a default also for `Q0`

scarlehoff: Ok, nf0 = 3, Q0= 1.65 ?

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,76 @@
...
+    Q0: float = 1.65  # [GeV] Parametrization scale
+    nfref: int = 5  # Number of active flavors at Qref
+    Qref: float = 91.2  # [GeV] Reference scale for alphas and alphaqed
+    alphas: float = 0.118  # Value of alpha_s at the scale Qref
+    alphaqed: float = 0.007496252  # Values of alpha QED at the scale Qref
+    # Evolution parameters
+    IterEv: int = None  # iterations for the evolution of the PDF. Defaults to 40 when ModEv = EXA
+    ModSV: str = None  # Scale variations method in EKO (expanded or exponentiated)
+    DAMPPOWERc: int = None  # Power of the damping factor in FONLL for the c
+    DAMPPOWERb: int = None  # Power of the damping factor in FONLL for the b
```
RoyStegeman, line 53: Maybe these can be moved below damp?

RoyStegeman: Or is it here because it has a default?

scarlehoff: Default. `damp` can be given an IC-dependent default as well.

(the conference finished before I could merge this :'( )

RoyStegeman:
I added the new baseline theorycard here since it's closely tied to what's happening in this PR. I'll delete it from https://github.com/NNPDF/theories_slim/issues/36

scarlehoff:
If nobody disagrees with this, I'll merge this.

@RoyStegeman maybe it makes sense to separate the 4.1 baseline as its own branch so that people can comment on the specific parameters.

felixhekhorn:
> If nobody disagrees with this, I'll merge this.
> 
> @RoyStegeman maybe it makes sense to separate the 4.1 baseline as its own branch so that people can comment on the specific parameters.

please give me a chance to have a second look, since I wasn't able last two weeks

felixhekhorn (commented):
just to repeat again what I discussed with @scarlehoff already in person:
- this "Theory" is a unique identifier for a folder in `theories_slim`, i.e. for a "theory" in the nnpdf sense (and as we all know the word "theory" may not be the best term for this object).
- As such "polarized" should not belong on this card (as we said we want to keep the possibility to have both states side-by-side), nor "RenScaleVar" (or Fact-) as this is a) dealt on a lower level (i.e. grid level) and b) here characterized by XIR (-F) and as such (with this value) baked in
- but `ev_op_iterations` should belong here as this can distinguish a "theory"
- importantly, this concept of "theory" may or may not be inherited by other programs, but they can define their own scheme. E.g. eko distinguishes really in theory setting (associated to the Langrangian) and operator settings - but this plays on a different level than nnpdf. It will be pinekos job to do the translation

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,99 @@
...
+    ID: int  # ID number of the theory
+    PTO: int  # Perturbative order (0 = LO, 1 = NLO, 2 = NNLO ...)
+    FNS: str  # Flavor number scheme (i.e. FONLL-C)
+    DAMP: int  # Whether a damping function is applied or not for FONLL
+    IC: int  # 0 = perturbative charm only , 1 = intrinsic charm allowed
+    ModEv: str  # DGLAP evolution solution method (EXA or TRN)
+    XIR: float  # Renormalization scale over the hard scattering scale ratio
+    XIF: float  # Factorization scale over the hard scattering scale ratio
+    NfFF: int  # Number of active flavors, only for FFNS or FFN0 schemes
+    QED: int  # QED correction to strong coupling: 0 = disabled , 1 = allowed
```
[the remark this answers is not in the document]
scarlehoff: Wait, now that I read this... why do we have "2" sometimes ?

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,99 @@
...
+    mc: float  # [GeV] charm mass
+    Qmc: float  # [GeV] MSbar mass reference scale of the charm
+    kcThr: float  # Threshold ratio of the charm
+    mb: float  # # [GeV] bottom mass
+    Qmb: float  # [GeV] MSbar mass reference scale of the bottom
+    kbThr: float  # Threshold ratio of the bottom
+    mt: float  # # [GeV] top mass
+    Qmt: float  # [GeV] MSbar mass reference scale of the top
+    ktThr: float  # Threshold ratio of the top
+    CKM: list[float]  # CKM matrix elements
```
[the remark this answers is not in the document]
andreab1997: Done in [73a7db6](https://github.com/NNPDF/nnpdf/pull/2028/commits/73a7db6261e67849a22803c1bcdc5a24f8273993)

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,99 @@
...
+    ID: int  # ID number of the theory
+    PTO: int  # Perturbative order (0 = LO, 1 = NLO, 2 = NNLO ...)
+    FNS: str  # Flavor number scheme (e.g. FONLL-C)
+    DAMP: int  # Whether a damping function is applied or not for FONLL
+    IC: int  # 0 = perturbative charm only , 1 = intrinsic charm allowed
+    ModEv: str  # DGLAP evolution solution method (e.g. EXA or TRN)
+    XIR: float  # Renormalization scale over the hard scattering scale ratio
+    XIF: float  # Factorization scale over the hard scattering scale ratio
+    NfFF: int  # Number of active flavors, only for FFNS or FFN0 schemes
+    QED: int  # Max order of alpha_qed in the splitting functions
```
felixhekhorn, line 29:
```suggestion
    QED: int  # Max order of alpha_qed in the evolution
```
this will also impact the couplings

RoyStegeman:
Since I believe this is almost final, don't forget to update (the `theory_params.csv` file in) the docs

nnpdf_data/nnpdf_data/theory.py
```diff
@@ -0,0 +1,99 @@
...
+    # Fit theory parameters default
+    nf0: int = 4  # Number of active flavors at the parametrization scale Q0
+    Q0: float = 1.65  # [GeV] Parametrization scale
+    nfref: int = 5  # Number of active flavors at Qref
+    Qref: float = 91.2  # [GeV] Reference scale for alphas and alphaqed
+    alphas: float = 0.118  # Value of alpha_s at the scale Qref
+    alphaqed: float = 0.007496252  # Values of alpha QED at the scale Qref
+    # Evolution parameters
+    HQ: str = "POLE" # Heavy quark mass scheme,  POLE for pole masses (default), MSBAR for running masses (used currently only in eko).
+    IterEv: int = None  # iterations for the evolution of the PDF. Defaults to 40 when ModEv = EXA
```
felixhekhorn, line 57: Where is this default enforced? because the new 4.1 needs this default

RoyStegeman:
One thing I just realised is that by having defaults we make it even harder to search within the theorydb based on theoryparameters. Of course I'll concede that since they are the defaults it's probably not particularly useful to use it as a filter

RoyStegeman:
But all old theories have an explicit entry for e.g. Q0 and alphas (and all other values that are given defaults in this PR), so why would those not be made mandatory now. There is no need to give defaults for older theories

> For now in https://github.com/NNPDF/nnpdf/pull/2062 Q0 is mandatory again

That's this PR, and it seems that in the class a default is defined

scarlehoff:
> But all old theories have an explicit entry for e.g. alphas (and all other values that are given defaults in this PR), so why would those not be made mandatory now. There is no need to give defaults for older theories.

The idea is that some of the values in the old theories were either redundant or not needed. But maybe we were overeager and there's some that have been moved to optional that shouldn't be.

(if so, please mention which ones are those and we can make them mandatory again!)

RoyStegeman:
wrt to this PR I would not add defaults for MaxNfPdf, Q0, Qref, alphaqed

scarlehoff:
If nobody is explicitly against this, I'll merge this PR on Thursday.

@RoyStegeman I'd appreciate a second look at the 41000000.yml file since I've made a few changes.

RoyStegeman:
> @RoyStegeman I'd appreciate a second look at the 41000000.yml file since I've made a few changes

As you know I'm happy with explicitly defining parameters in the runcard instead of relying on the defaults, so for me this is not a huge problem, but `IterEv` and `HQ` are set to the default values so they can also be removed.

scarlehoff:
I also prefer having explicit values even if defaults are defined.

RoyStegeman:
Good, other than that 41000000.yml seems fine

scarlehoff (approved):

PR #2029, opened 2024-03-28
add alphas variations NNLO and NLO

RoyStegeman:
Extend the range of alphas theories at NNLO 

Add the corresponding NLO theories


nnpdf_data/nnpdf_data/theory_cards/837.yaml
```diff
@@ -0,0 +1,48 @@
+ID: 837
+PTO: 2
+FNS: FONLL-B
```
andreab1997, line 3: I believe that this theory card for NLO should change a but, for example it is missing the DAMPOWERb and DAMPOWERc entries

andreab1997 (changes requested):

nnpdf_data/nnpdf_data/theory_cards/837.yaml
RoyStegeman: PTO is wrong :sweat_smile: 

But DAMPPOWER are not present in the nnpdf theorycards. They will only be after #2028

andreab1997: Yes indeed, I agree with that.

Sorry that I have not noticed the PTO 😂

RoyStegeman:
I should probably revert deleting the values for which there are defaults. Seems that #2028 is not stable yet

RoyStegeman:
@andreab1997 what should we do with these? I was waiting for the single truth, but maybe we should just revert to the old format, so same as all the others, and merge?

Issue #1969, opened 2024-03-01
Update to new commondata names in docs, examples and tests


RoyStegeman:
- [x] Remove reproduce 4.0
- [x] Update runcards in vp
- [x] Update runcards in n3fit
- [x] Update docstrings in n3fit
- [x] Update docstrings in vp
- [x] Update tests in vp
- [x] Update tests in n3fit
- [x] Update docs

enocera:
Dear @RoyStegeman, @peterkrack and I are willing to help change the names in this PR. I have a question regarding the docstrings in vp. Where do we exactly specify the dataset names?

RoyStegeman:
Thanks a lot! In the docstrings we regularly provide examples of how a function can be used (through the API). I put it separately because while the names in the tests can be updated and as long as the tests pass afterwords it's fine, the doctring examples may need to be tested. Feel free to do them simultaneously, but to me it made sense to separate them.

RoyStegeman:
Sorry this is not complete, if I just do a bruteforce search for the old names I still find plenty

enocera:
Really? My bad, then. I thought I had fixed them all.

RoyStegeman:
No worries, it doesn't look like there are many left, but it's a bit early to close this

enocera:
I'll have a second pass, then.

RoyStegeman:
Thanks, though if you don't have the time I'm sure we can include it in the list of tasks to be delegated in Morimondo

scarlehoff:
> list of tasks to be delegated in Morimondo

Perhaps you meant the illustrious list of missions that will be assigned to honoured researchers in Morimondo.

RoyStegeman:
Yes that's what I meant. Damn autocorrect!

kamillaurent:
I am doing the changes related to the first point:

- [ ] Update docstrings in vp

in this branch: https://github.com/NNPDF/nnpdf/tree/update_docstrings_vp-new_datasets_names

PR #2031, opened 2024-03-29
Update to new commondata names - n3fit

RoyStegeman:
Closes some of #1969

This PR makes changes to the dataset names in the following
- [x] Remove reproduce 4.0
- [x] Update runcards in n3fit
- [x] Update docstrings in n3fit
- [x] Update tests in n3fit


n3fit/runcards/examples/nnpdf40-like.yml
```diff
@@ -0,0 +1,167 @@
...
+datacuts:
+  t0pdfset: 240329-01-rs-nnpdf40like-baseline
+  q2min: 3.49
+  w2min: 12.5
+
+############################################################
+theory:
+  theoryid: 708
+sampling:
+  separate_multiplicative: false
```
scarlehoff, line 100: Just wondering whether we should remove this key by now or set it to true going forward (@andreab1997 ?)

n3fit/src/n3fit/tests/test_evolven3fit.py
```diff
@@ -111,7 +111,7 @@ def test_utils():
             np.testing.assert_allclose(my_PDF.xfxQ2(pid, x, q20), x * (1.0 - x))
     # Testing read_runcard
     runcard = utils.read_runcard(REGRESSION_FOLDER)
-    assert runcard["description"] == "n3fit regression test"
+    assert runcard["description"] == "evolven3fit utils test"
```
scarlehoff, line 114:
```suggestion
    assert isinstance(runcard["description"], str)
```

The exact string is probably not important

scarlehoff (commented):
Thank you very much for this. I guess there's still some to go, but as far as I'm concerned we can merge it when you wish (no need to have updated all of it before merging!)

RE the conda profiles for reproduction. Maybe those we want to keep in the repo, since reading them with conda should take care of having the code in the state decided by the .yaml file.

n3fit/runcards/examples/nnpdf40-like.yml
andreab1997: If we want to be backward compatible with NNPDF40 we cannot remove the key. Otherwise, if we no longer care about backward compatibility, given that we are approaching 4.1, I agree we should remove the key and just set it to False. Note that the NNPDF40 behaviour is recovered with true not with false.

scarlehoff: For 1-1 compatibility there is the conda environments provided. I think from 4.0.10 onwards we are happy with slowly moving away from 4.0 compatibility.

RoyStegeman:
> RE the conda profiles for reproduction. Maybe those we want to keep in the repo, since reading them with conda should take care of having the code in the state decided by the .yaml file.

Yes that's true, I removed them because I don't believe they have ever been used and for some of them the corresponding nnpdf version has not been uploaded to the server since we were uploading upon merge to master only and not upon tagging. Thinking about it again though, it's true that the dependencies _might_ come in handy at some point, so perhaps we can keep them around a bit longer.

RoyStegeman:
>  guess there's still some to go, but as far as I'm concerned we can merge it when you wish (no need to have updated all of it before merging!)

Yes maybe that's better, I'm not sure exactly when I'll have enough time to finish everything so we may as well merge what we have and I'll continue in a a new branch 

If you're happy please approve

scarlehoff (approved):

PR #2032, opened 2024-04-02
Remove buildmaster

RoyStegeman:
Since we now moved to the new commondata format, we no longer need the buildmaster folder. The datafiles we need to keep have been moved to the nnpdf_data folder. 

The first commit simply removes the buildmaster folder
The second commit removes some outdated parts from the docs

In general I noticed the docs have a lot of outdated info (mostly old pipeline) in them that we should either update or remove. That's not the point of this PR though.


comane (approved):
Hi @RoyStegeman thanks for this, this looks good to me.

I just noticed that in two of the jets datasets I had re implemented 
eg `nnpdf_data/nnpdf_data/new_commondata/ATLAS_2JET_7TEV_R06/filter.py`

there are still mentions of `buildmaster`. This was for the purpose of testing and has unfortunately not been removed.

In the `if __name__=="__main__"` of these filter files I think one should only keep the parts that generate the datasets yaml files and remove the rest. 

But I guess that this could in principle be done in a separate PR and I could take care of this.

PR #2034, opened 2024-04-03
make `pseudodata_table` correctly deal with multiple replicas

RoyStegeman:
For the thcovmat alphas stuff it makes a big difference whether I use the central data or the average over data replicas. While looking into it I want to use this function with multiple replicas.

### DIFF
--- a/validphys2/src/validphys/n3fit_data.py
+++ b/validphys2/src/validphys/n3fit_data.py
@@ -364,7 +364,7 @@ def pseudodata_table(groups_replicas_indexed_make_replica, replicas):
     # in that order. What this means is that groups_replicas_indexed_make_replica is a list of size
     # number_of_replicas x number_of_data_groups. Where the ordering inside the list is as follows:
     # [data1_rep1, data2_rep1, ..., datan_rep1, ..., data1_repn, data2_repn, ..., datan_repn].
-    #
+
     # To correctly put this into a single dataframe, we first need to know the number of
     # dataset_input groups there are for each replica
     groups_per_replica = int(len(groups_replicas_indexed_make_replica) / len(replicas))
--- a/validphys2/src/validphys/n3fit_data.py
+++ b/validphys2/src/validphys/n3fit_data.py
@@ -360,8 +360,12 @@ def pseudodata_table(groups_replicas_indexed_make_replica, replicas):
     replicas are fitted one at a time. The table can be found in the replica
     folder i.e. <fit dir>/nnfit/replica_*/
     """
-    # groups_replicas_indexed_make_replica is collected over both replicas and dataset_input groups
-    # to correctly put this into a single dataframe, we first need to know the number of
+    # groups_replicas_indexed_make_replica is collected over both replicas and dataset_input groups,
+    # in that order. What this means is that groups_replicas_indexed_make_replica is a list of size
+    # number_of_replicas x number_of_data_groups. Where the ordering inside the list is as follows:
+    # [data1_rep1, data2_rep1, ..., datan_rep1, ..., data1_repn, data2_repn, ..., datan_repn].
+    #
+    # To correctly put this into a single dataframe, we first need to know the number of
     # dataset_input groups there are for each replica
     groups_per_replica = int(len(groups_replicas_indexed_make_replica) / len(replicas))
     # then we make a list of pandas dataframes, each containing the pseudodata of all datasets
--- a/validphys2/src/validphys/n3fit_data.py
+++ b/validphys2/src/validphys/n3fit_data.py
@@ -4,6 +4,7 @@ n3fit_data.py
 Providers which prepare the data ready for
 :py:func:`n3fit.performfit.performfit`.
 """
+
 from collections import defaultdict
 import functools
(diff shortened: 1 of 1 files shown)


validphys2/src/validphys/n3fit_data.py
```diff
@@ -343,7 +343,7 @@ def replica_nnseed_fitting_data_dict(replica, exps_fitting_data_dict, replica_nn
 
 replicas_nnseed_fitting_data_dict = collect("replica_nnseed_fitting_data_dict", ("replicas",))
 groups_replicas_indexed_make_replica = collect(
-    "indexed_make_replica", ("group_dataset_inputs_by_experiment", "replicas")
+    "indexed_make_replica", ("replicas", "group_dataset_inputs_by_experiment")
```
scarlehoff, line 346: Why is the change of order necessary?

validphys2/src/validphys/n3fit_data.py
```diff
@@ -359,10 +359,20 @@ def pseudodata_table(groups_replicas_indexed_make_replica, replicas):
...
     replicas are fitted one at a time. The table can be found in the replica
     folder i.e. <fit dir>/nnfit/replica_*/
-
     """
-    # Concatenate over replicas
-    df = pd.concat(groups_replicas_indexed_make_replica)
+    # groups_replicas_indexed_make_replica is collected over both replicas and dataset_input groups
+    # to correctly put this into a single dataframe, we first need to know the number of
+    # dataset_input groups there are for each replica
+    groups_per_replica = int(len(groups_replicas_indexed_make_replica) / len(replicas))
```
scarlehoff, line 366:
```suggestion
    groups_per_replica = len(groups_replicas_indexed_make_replica) // len(replicas)
```

validphys2/src/validphys/n3fit_data.py
```diff
@@ -359,10 +359,20 @@ def pseudodata_table(groups_replicas_indexed_make_replica, replicas):
...
+    # then we make a list of pandas dataframes, each containing the pseudodata of all datasets
+    # generated for a singel replica
```
scarlehoff, line 368:
```suggestion
    # generated for a single replica
```

validphys2/src/validphys/n3fit_data.py
```diff
@@ -359,10 +359,20 @@ def pseudodata_table(groups_replicas_indexed_make_replica, replicas):
...
+    df = [
+        pd.concat(groups_replicas_indexed_make_replica[i : i + groups_per_replica])
+        for i in range(0, len(groups_replicas_indexed_make_replica), groups_per_replica)
+    ]
```
scarlehoff, line 372: Why can't you achieve this with a reshape or permutation of `groups_replicas_indexed_make_replica`

validphys2/src/validphys/n3fit_data.py
RoyStegeman: I thin it makes what we do in `pseudodata_table` more readable. There we group the entries to `groups_replicas_indexed_make_replica` corresponding to a given replica. I think that's easier to understand the way it's done now than if we had to take e.g. index 0 and then skip a number of indexes equal to the number of groups to get the second input corresponding to the same replica

validphys2/src/validphys/n3fit_data.py
RoyStegeman: It's used as a list index so it has to be `int`

validphys2/src/validphys/n3fit_data.py
RoyStegeman: because `groups_replicas_indexed_make_replica` is a list of `indexed_make_replica` containing a number of dataframes equal to number_of_replicas x number_of_datagroups. It's not a really clean input

validphys2/src/validphys/n3fit_data.py
scarlehoff: The output of `//` should be int.

scarlehoff: No, you are doing `len` vs `len` it should take the same type as the operands.

RoyStegeman: Ah! You're right!

validphys2/src/validphys/n3fit_data.py
scarlehoff: Just complicating your life here, but would it be possible to do something along the lines of 

`np.array(groups_replicas_indexed_make_replica).reshape(replicas, groups_per_replica)` ?

(or the other way around)

RoyStegeman: No because it's a list of dataframes and I do need to retain the information on the labels

validphys2/src/validphys/n3fit_data.py
scarlehoff: If it is because of taste I'm ok with the `int()`.

RoyStegeman: I also didn't realise it retained the original types, so it wasn't just taste. I already accepted this now

PR #2036, opened 2024-04-05
fix std_error function in MCStats class

RoyStegeman:
In #1522 we replaced self.data with self.error_members(), but did not update the `ddof` value accordingly.


scarlehoff:
Thanks. Let's see whether it has a visible effect anywhere.

RoyStegeman:
Hmm, I would have expected to at least be able to see the difference with only 16 replicas

scarlehoff:
I'm surprised not many regressions failed. I guess the PDF errors are almost only used for the plots themselves so not many chances of having an effect.

RoyStegeman:
It's just because our code is optimally tested. One and only one test per functionality 

Yes... That's why..

scarlehoff (approved):
I'm not going to write "I'm happy with it" because we have already proven scientifically that it brings misfortune.

PR #2038, opened 2024-04-08
Update QED parts in theories 391 392

niclaurenti:


scarlehoff (approved):
This seems to work :)

https://vp.nnpdf.science/LPIg3H86QIWofhm-LF7yZQ==

(the fits are different, current is both PC and flavour basis and the fktables are not qed-evolved)

PR #2039, opened 2024-04-08
Polish EIC data files

comane:
This PR aims at polishing the ATHENA data files.
Things that are done:

- moved all xlsx files into a `rawdata` folder
- added openpyxl to poetry extras. This is needed by `pd.read_excel`
- created an `athena_utils.py` file so as to minimise repeated code.


comane:
@scarlehoff  @Radonirinaunimi @RoyStegeman I think we should probably have a `nnpdf_data/filter_utils` folder that contains all filter utils modules for datasets. What do you think

RoyStegeman:
Thanks, I'll have a closer look later, but I agree that, if it makes sense to have multiple filter utils with e.g. some corresponding to only a single experiment, it would be good to have a `filter_utils` folder. That's exactly the sort of thing that's needed

Radonirinaunimi:
Thanks so much for this @comane!

> @scarlehoff @Radonirinaunimi @RoyStegeman I think we should probably have a nnpdf_data/filter_utils folder that contains all filter utils modules for datasets. What do you think

Yes, something along that line should definitely be adopted, and as a matter of fact I was planning to discuss with you (later this week) about the main utils that were/are required in the datasets you implemented in order to draft something.

In this respect, as far as this PR is concerned, you could already move `athena_ultis.py` in a folder `nnpdf_data/filter_utils`.

nnpdf_data/nnpdf_data/new_commondata/ATHENA_NC_105GEV_EP/filter.py
```diff
@@ -1,34 +1,11 @@
...
 import yaml
 
-np.random.seed(1234567890)
-
-
-def read_excel(path_xlsx: Path, beams: tuple) -> pd.DataFrame:
-    """Parse the xlsx file containing all the information regarding
-    the projections and returns the ones corresponding to the chosen
-    beam energies.
+from nnpdf_data.athena_utils import read_excel, fluctuate_data, write_data
```
Radonirinaunimi, line 6: Thanks again for this! This looks good to me. My only comment is to move the utils into a dedicated folder as you suggested.
```suggestion
from nnpdf_data.filter_utils.athena_utils import read_excel, fluctuate_data, write_data
```
And for the rest below.

nnpdf_data/nnpdf_data/athena_utils.py
```diff
@@ -0,0 +1,131 @@
+import pandas as pd
+from pathlib import Path
+from typing import Optional, Union
+
+import yaml
```
RoyStegeman, line 5:
```suggestion
from reportengine.compat import yaml
```

RoyStegeman: Ah no, reportengine is not a dependency and perhaps the standard library is good enough here

nnpdf_data/nnpdf_data/athena_utils.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Returns
+    -------
+    pd.DataFrame
+        returns a panda table corresponding to the chosen beams
+    """
+    xdf = pd.read_excel(path_xlsx)
+    el, ep = beams
+    df_beam = xdf[(xdf["El"] == el) & (xdf["Eh"] == ep)]
+    return df_beam[df_beam["y"] <= 1]
```
RoyStegeman, line 29: `df_beam["y"] <= 1` is always true so why put it here?

nnpdf_data/nnpdf_data/athena_utils.py
```diff
@@ -0,0 +1,131 @@
+import pandas as pd
```
RoyStegeman, line 1: Please make sure to use [pre-commit](https://pre-commit.com/)

nnpdf_data/nnpdf_data/filter_utils/athena_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    pd.DataFrame
+        returns a panda table corresponding to the chosen beams
+    """
+    xdf = pd.read_excel(path_xlsx)
+    el, ep = beams
+    df_beam = xdf[(xdf["El"] == el) & (xdf["Eh"] == ep)]
+    return df_beam[df_beam["y"] <= 1]
+
+
+def fluctuate_data(central: np.ndarray, abserr: np.ndarray) -> np.ndarray:
```
RoyStegeman, line 32: Why is the data fluctuated?

Radonirinaunimi: These are not actually real experimental data but rather pseudodata, so the central values are defined as fluctuations of some theory predictions provided with the projection uncertainties.

nnpdf_data/nnpdf_data/athena_utils.py
Radonirinaunimi: This was not the case for the previous rawdata and I'd leave it there for safety given that these datasets might be subjected to modifications (depending on updated projections).

nnpdf_data/nnpdf_data/athena_utils.py
comane: Thanks for this, I just updated.

nnpdf_data/nnpdf_data/filter_utils/athena_utils.py
Radonirinaunimi: Yes, this is indeed intended because we do not have predictions for this energy yet.

RoyStegeman: Ah right, I should've looked at the metadata.yaml and I could've found that out. Perhaps you can add to the docstring of the function that this is done because these datasets are pseudodata

nnpdf_data/nnpdf_data/filter_utils/athena_utils.py
```diff
@@ -0,0 +1,131 @@
+from pathlib import Path
+from typing import Optional, Union
+
+import numpy as np
+import pandas as pd
+import yaml
+
+
+def read_excel(path_xlsx: Path, beams: tuple) -> pd.DataFrame:
```
RoyStegeman, line 9: @Radonirinaunimi why is this stored in excel files? Since we're using git we should probably avoid that where possible.

nnpdf_data/nnpdf_data/athena_utils.py
Radonirinaunimi: If I remember correctly what was written in the ATHENA paper, this is an artefact of how the detector simulations are designed in the projections. What is happening is that when the detector simulations are combined with the MC, there are events (at large-$x$) in which the electrons receive larger energies rendering $y > 1$.  These are ofc unphysical events and therefore must be discarded. They nevertheless dump them in the rawdata and we have to remove them manually.

nnpdf_data/nnpdf_data/filter_utils/athena_utils.py
Radonirinaunimi: Unfortunately, that's the format in which these projections are provided. We can have another external script that transforms these into more trackable files and these are the ones used in the filters, but then I am not sure if we want that since the actual raw data and the external script will not be tracked (?)

nnpdf_data/nnpdf_data/filter_utils/athena_utils.py
Radonirinaunimi: I add some extra texts in the docstring to reflect this.

Radonirinaunimi:
@comane, @RoyStegeman, given that these Athena share quite a lot of the functions with the other EIC projections, I also fixed them here. I therefore renamed the module from `athena_utils` $\longrightarrow$ `eic_utils`.

nnpdf_data/nnpdf_data/filter_utils/athena_utils.py
RoyStegeman: > Unfortunately, that's the format in which these projections are provided

Where are they taken from? I agree that translating the rawdata to another format before then passing it through the filter.py into yet another format is not the cleanest thing to do. Fortunately, them being rawdata files we will never touch them. Unfortunately, it means we require some program to read the .xlsx files.

nnpdf_data/nnpdf_data/athena_utils.py
RoyStegeman: Thanks for the explanation. I think it's also worth adding that in a comment here

nnpdf_data/nnpdf_data/filter_utils/athena_utils.py
Radonirinaunimi: > Where are they taken from? I agree that translating the rawdata to another format before then passing it through the filter.py into yet another format is not the cleanest thing to do.

They were provided to us by the Athena collaboration during the FONLL POL studies.

> Fortunately, them being rawdata files we will never touch them. Unfortunately, it means we require some program to read the .xlsx files.
 
Indeed.

nnpdf_data/nnpdf_data/athena_utils.py
Radonirinaunimi: Addressed. I added some comments to reflect this.

nnpdf_data/nnpdf_data/filter_utils/README.md
RoyStegeman, line 1: I think this is self explanatory so this file can be removed

RoyStegeman:
Please rebase so the tests will pass

nnpdf_data/nnpdf_data/filter_utils/eic_utils.py
```diff
@@ -0,0 +1,206 @@
+from pathlib import Path
+from typing import Optional, Union
+
+import numpy as np
+import pandas as pd
+import yaml
+
+
+def read_cvs(path: Path) -> np.ndarray:
```
RoyStegeman, line 9:
```suggestion
def read_central_values(path: Path) -> np.ndarray:
```

I misread this as `read_csv`

nnpdf_data/nnpdf_data/filter_utils/eic_utils.py
```diff
@@ -0,0 +1,206 @@
...
+    -------
+    np.ndarray:
+        array containing the central predictions
+    """
+    cv_yaml = yaml.safe_load(path.read_text())
+    return np.array(cv_yaml["predictions_central"])
+
+
+def read_txt_data(path_txt: Path) -> pd.DataFrame:
+    """Return a Panda table in which the columns are ordered as
```
RoyStegeman, line 27:
```suggestion
    """Reads a rawdata `.txt` file and returns a pandas dataframe in which the columns are ordered as
```

RoyStegeman (approved):
Seems fine to me

Radonirinaunimi (approved):
@comane I think this is done now, please merge if you are fine with this.

comane:
thanks @Radonirinaunimi and @RoyStegeman for reviewing!

RoyStegeman:
You only force-pushed your local changes. I still have the commits from the review locally, would you like me to push those?

comane:
Yes please, I might have done a mess.
I had rebased on master but perhaps I didn't pull from the branch :(

PR #2042, opened 2024-04-10
Fix to weight loading/saving for compatibility with Keras 3

scarlehoff:
The tests failing elsewhere are due to a change in the `Dense` layer in keras. We are extending (and exploiting) the internal of their layer so changes there break things here.

Note, old weights still load fine (that's why only the saving was failing) since they have kept a legacy loader.

But, just in case, I'm updating also the weights to the new format. Once the regression pass I'll push the updated weights.

I've also re-enabled `single_dense` since it is very useful for debugging. I'll add a test for it. It also gives a fall-back that should work even if the internals of keras we rely upon change.

edit: a pity that I cannot change the description of the commit without force-pushing, the last one should obviously be "add check for single dense" :__


RoyStegeman (approved):
Thanks. In particular I like the green tick!

PR #2046, opened 2024-04-11
Correct bug in commondata loading for closure tests

andreab1997:
Here we correct two bugs affecting the loading of the commondata for a closure test that uses the new names for the datasets. 

The bug in `loader.py` is clearly a typo: `new_data_stream` is used instead of `new_unc_stream` while `new_data_stream` has already been closed. 

The other bug is due to the fact that `ds.commondata.datafile=None` for new commondata. Also, in the past `cd_df.iloc[cut_mask(ds.cuts), 5]` was necessary to select column 5 which, as stated in the comment, is the data column. Now, using `metadata.load_central()`, this is no longer necessary.


validphys2/src/validphys/closuretest/multiclosure_pseudodata.py
```diff
@@ -31,10 +31,10 @@ def fits_dataset_cvs(fits_dataset):
         # and then cut the central values manually.
         # TODO: Save central values in nice table like pseudodata
         # but this should be done beyond NNPDF4.0
-        cd_df = pd.read_csv(ds.commondata.datafile, sep=r'\s+', skiprows=1, header=None)
+        cd_df = ds.commondata.metadata.load_data_central()
```
scarlehoff, line 34: I don't think this will do what you actually want.
I don't like this function at all _but_ to recover the old behaviour you probably want to do `ds.commondata.with_modified_data(the fit data path)`

scarlehoff (changes requested):

validphys2/src/validphys/closuretest/multiclosure_pseudodata.py
scarlehoff: But to be honest, why don't you just get in the input the data loaded using use_fit_commondata.

This function looks like a convenient way of loading the data to avoid running over fits but I think it might have been hiding errors for a while...

scarlehoff (approved):
I've just added a test using a script that @andreab1997 sent me.

The fixes themselves are ok. I'm still not superhappy with the multiclosure module but at least with the test we should know immediately when it breaks again.

PR #2050, opened 2024-04-15
Restore cut of CMS_TTBAR_8TEV_2L_DIF_MTTBAR-YT-NORM

andreab1997:
We restore the old cut of the dataset CMS_TTBAR_8TEV_2L_DIF_MTTBAR-YT-NORM.  The reason for this is that closure test fits produced for ongoing studies have the old cuts.  

@scarlehoff 
@RoyStegeman

### DIFF
--- a/validphys2/src/validphys/cuts/filters.yaml
+++ b/validphys2/src/validphys/cuts/filters.yaml
@@ -2,7 +2,7 @@
   reason: |
    We remove one bin from the normalised distribution because it is
    linearly dependent on the others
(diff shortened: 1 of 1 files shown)


t7phy:
If I remember correctly, @RoyStegeman asked @enocera about this in one of the Pheno meetings and it was agreed that we want to stay with the new cuts whereby always the last point is removed, isn't that right @scarlehoff ?

t7phy:
If it helps in your study, why not? Fine for me 👍

t7phy (approved):

PR #2049, opened 2024-04-15
Add an example using `PineAPPL` to compute predictions

scarlehoff:
Added a notebook to the examples folder where validphys is monkeypatched to utilize pineappl grids instead of fktables.

The code for this will not be integrated in validphys (validphys will keep using only fktables because changing to grids can open many cans of worms that I don't want to open) but as you can see in the notebook the code is quite simple.

This deals (in a way) with parts of https://github.com/NNPDF/nnpdf/issues/1961

ps: If there is any function in validphys for which the monkey patching in the notebook doesn't produce the desired result it might be due to said function using directly convolute instead of results. Please open an issue / PR because the more we make the validphys API be consistent the easiest it becomes to do things like this in a simple manner!

PR #2051, opened 2024-04-15
Use `results` to create th uncertainties with pdf errors.

scarlehoff:
This is necessary to allow for things like #2049 but in general it also cleans up the code as more things are concentrated under `results`.

@achiefa you will need to rebase on top of this branch _if you want to add pdf errors_ but I think for the time being you don't need them so you can wait until it is in master.

### DIFF
--- a/validphys2/src/validphys/covmats.py
+++ b/validphys2/src/validphys/covmats.py
@@ -11,20 +11,15 @@ import scipy.linalg as la
 
 from reportengine import collect
 from reportengine.table import table
-from validphys.calcutils import get_df_block, regularize_covmat
+from validphys.calcutils import regularize_covmat
 from validphys.checks import (
     check_cuts_considered,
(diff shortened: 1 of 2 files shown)


RoyStegeman:
Could you be more explicit in your explanation? What is the example nb trying to do that needs this? It seems that all this does is replace the exp covmat with an identity matrix, but I don't believe it is ever read anyway. Is it to save the computational resources?

scarlehoff:
`pdferr_plus_dataset_inputs_covmat` needs to compute the predictions for a dataset in order to generate the covariance matrix with exp + pdf error.

Ideally we want to get everything in validphys through a result object* but this creates a chicken-and-egg problem because the results object includes the covmat, so this adds as well a result obect without a covmat.

*and the notebook is based on monkeypatching the results (and results_with_thcovmat) objects

RoyStegeman (approved):

PR #2053, opened 2024-04-16
Regenerate test pol

scarlehoff:
This fixes the polarized tests (by doing them a bit less aggressively _and_ in double precision) and adds a test for the TSR sum rules.

I’ve also made the polarized sum rules depend on the nn seed (I think it is fair to consider them part of the nn).

Now, I don’t like the double precision part, it means that there’s some instability affecting the polarized fit. I’ll try to figure out where is it coming from but I think it can be detached from the rest of the polarized fit (as it is just a numerical issue).


Radonirinaunimi (approved):
This looks perfect! Thanks!

PR #2054, opened 2024-04-16
Use groups or reorder tests

### LINKED ISSUE #2052: Ordering of groups_covmat
While testing #2051, I've noticed that groups_covmat is not following the ordering of the dataset inputs (it is not a problem for #2051 _I think_). You can check this quickly by simply putting an atlas dataset in the middle of cms.
This should not be a problem as long as we are always using it as a dataframe, but I thought it would be good to make you aware of this @RoyStegeman @andreab1997 

I'm combing the code to figure out if it is a problem somewhere, and if it is not, I'll explicitly change the ordering of the dataset in the tests so that we are testing the case in which a dataset is in the middle of the others.

scarlehoff:
Closes #2052 

I've put now the CMS dataset between the two datasets in `DATA` in `conftest.py`. This makes some test fail, some of them it is clear that the bug was in the test itself (e.g., in `test_regularized_covmat_generation` the covmat was collected over dataset inputs and the index over groups), while for some other (like the `make_replica` one) it is more subtle (I think the bug was also the test).

Some of the tests are quite redundant, in the sense that they made sense when there were two sides of the code (C++ and python) but now everything goes through the same functions. Better than nothing I guess.

Anyway, I've separated the fixes to the tests and pre-commit in case we need to trace them back.

### DIFF
--- a/validphys2/src/validphys/tests/test_pythonmakereplica.py
+++ b/validphys2/src/validphys/tests/test_pythonmakereplica.py
@@ -4,6 +4,7 @@ test_pythonmakereplica.py
 Module for testing the python implementation of make replica
 
 """
+
 from copy import deepcopy
 
 import numpy as np
@@ -14,10 +15,9 @@ from validphys.api import API
 from validphys.tests.conftest import DATA
 from validphys.tests.test_covmats import CORR_DATA
 
-
 SEED = 123456
 
-#Datasets to be tested 
+# Datasets to be tested
 SINGLE_SYS_DATASETS = [
     {"dataset": "DYE886R"},
     {"dataset": "D0ZRAP", "cfac": ["QCD"]},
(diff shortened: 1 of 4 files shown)


validphys2/src/validphys/tests/test_pythonmakereplica.py
```diff
@@ -102,7 +102,13 @@ def test_genrep_off(data_config, dataset_inputs, use_cuts):
...
     config["genrep"] = False
-    ld_cds = API.dataset_inputs_loaded_cd_with_cuts(**config)
+
+    # `make_replica` reorders the data according to the groups in the exp covmat
     not_replica = API.make_replica(**config)
+
+    # And so it needs to be tested against the central values loaded by group
+    ld_cds = API.groups_dataset_inputs_loaded_cd_with_cuts(**config)
+
+    central_data = []
```
RoyStegeman, line 112:
```suggestion
```

RoyStegeman (approved):

PR #2055, opened 2024-04-16
capitalize EicC dataset names

Radonirinaunimi:
As the title suggests, simply capitalizes EIcC dataset names to comply with convention.


nnpdf_data/nnpdf_data/new_commondata/EICC_NC_22GEV_EP/rawdata/EIcC_5_25_A1c_100fb-1.txt
felixhekhorn, line 1: Not sure if you want to change also these file names to the correct capitalization which is `EicC` - see e.g. https://inspirehep.net/literature/2683568

felixhekhorn: (or capitalize everything as you suggest in this PR)

scarlehoff (approved):
The dataset label and description you can keep with EicC if you want though

nnpdf_data/nnpdf_data/new_commondata/EICC_NC_22GEV_EP/rawdata/EIcC_5_25_A1c_100fb-1.txt
Radonirinaunimi: I think I can capitalize everything such that it is consistent everywhere.

Radonirinaunimi:
Everything capitalized except for the FK table names - which I think should be fine since there things are still not consistent yet.

RoyStegeman:
> Everything capitalized except for the FK table names - which I think should be fine since there things are still not consistent yet.

You mean that the fktable names are still in the old format? But wouldn't it make sense to be consistent with newly implemented data?

Radonirinaunimi:
> > Everything capitalized except for the FK table names - which I think should be fine since there things are still not consistent yet.
> 
> You mean that the fktable names are still in the old format? But wouldn't it make sense to be consistent with newly implemented data?

Ah, no. I simply meant that the naming is still kept with lower "c":
https://github.com/NNPDF/nnpdf/blob/14feef2d489114d08e5eb0f6ceb435bd76424f41/nnpdf_data/nnpdf_data/new_commondata/EICC_NC_15GEV_EP/metadata.yaml#L52

I can also manually rename the FK tables in the uploaded theory if you want before merging this.

RoyStegeman:
Yes that I understood. I didn't understand the justification for not updating those as well: 

> since there things are still not consistent yet

RoyStegeman:
Okay fair enough, that was my assumption as well. And that's why I said 

> But wouldn't it make sense to be consistent with newly implemented data?

Since you're changing things now, and I believe this fktable exisits only in a single theory, so I thought it might be worth fixing the fktable as well. If you think it's not worth the effort feel free to ignore though

Radonirinaunimi:
> Since you're changing things now, and I believe this fktable exisits only in a single theory, so I thought it might be worth fixing the fktable as well. If you think it's not worth the effort feel free to ignore though

If you are fine with this I'd keep this as it is for now, especially since renaming them manually is sort of hacky xd

RoyStegeman (approved):
I hope you weren't waiting for my green tick

PR #2056, opened 2024-04-17
Polish legacy jet datasets

comane:
The scope of this PR is to (try to) polish the new implementation of the legacy jet datasets, namely, CMS_2JET_7TEV, CMS_1JET_8TEV, ATLAS_1JET_8TEV_R06, ATLAS_2JET_7TEV_R06.

What the PR does:

1. Adds a `filter_utils.utils.py` module. This module should collect the functions that are generalisable and could, in principle, be used for all datasets.
2. Adds a `filter_utils.legacy_jets_utils.py` module that collects all of the utils functions for the filter files. Polishing these is going to be much harder given the nature of the raw data.


Radonirinaunimi (commented):
Thanks @comane for this! I will have a detailed look soon, but when skimming through I think there are at least two main issues with the CMS_1JET_8TEV_PTY dataset:
- the variable in the kinematic has changed from `p_T` to `p_T2` while in the metadata it's still kept as `p_T`
- the file `uncertainties_bugged.yaml` no longer exists but it is still referenced in the metadata

RoyStegeman:
I don't think we should start getting in the habit of suppressing warnings if there is another option

comane:
> This was not done in this PR, but I changed it since I don't think we should start getting in the habit of suppressing warnings if there is another option

Thanks @RoyStegeman, just to be sure I understand: removing the read "r" from `with open` avoids the pandas warnings?

RoyStegeman:
Since filter files will now share utilities, and we may need to change/fix those shared utilities at some point - do you think it's feasible to run all filters in the CI to test if the output remains unchanged?

RoyStegeman:
https://github.com/NNPDF/nnpdf/blob/eed00c75b4910606e967d1d0be644a45fd44b913/validphys2/src/validphys/process_options.py#L27

@scarlehoff what should be done in this PR?

scarlehoff:
> @scarlehoff what should be done in this PR?

Change to `pT2` I'd say.

RoyStegeman:
@comane I see you addressed the CI fails (as pointed out by @Radonirinaunimi), but could you rename p_T2 to pT2?

comane:
> @comane I see you addressed the CI fails (as pointed out by @Radonirinaunimi), but could you rename p_T2 to pT2?

Sure, I am not sure why though?
Who wrote that comment there and why is p_T2 wrong? The tests seem to be checking for p_T2 at the moment, not for pT2

So maybe I can simply remove that comment?

RoyStegeman:
I assume @scarlehoff wrote it. Either way I agree with it for consistency with PT and pT_t

nnpdf_data/nnpdf_data/new_commondata/CMS_1JET_8TEV/metadata.yaml
```diff
@@ -42,7 +42,7 @@ implemented_observables:
         description: jet rapidity
         label: $|y|$
         units: ''
-      pT:
+      pT2:
```
RoyStegeman, line 45: Why did was this change made (I see the corresponding change was also done in kinematics.yaml)?

nnpdf_data/nnpdf_data/filter_utils/legacy_jets_utils.py
```diff
@@ -0,0 +1,1146 @@
...
+    str_nums = str_range.split('-')
+    # Convert the parts to floats
+    min = float(str_nums[0])
+    max = float(str_nums[1])
+    mid = float(f"{0.5 * (min + max):.3f}")
+    # Return a dict
+    return {"min": min, "mid": mid, "max": max}
+
+
+# ==================================================================== CMS_2JET_7TEV ====================================================================#
```
RoyStegeman, line 27: If these are only used for a single dataset, why are they not inside the folder of the corresponding dataset?

nnpdf_data/nnpdf_data/filter_utils/legacy_jets_utils.py
```diff
@@ -0,0 +1,1146 @@
...
+            (scenario == 'nominal' and i == 67)
+            or (scenario == 'stronger' and i == 57)
+            or (scenario == 'weaker' and i == 69)
+        ):
+            df["lum"] = np.nan
+
+        elif err['label'] == 'sys':
+            df[f"{err['label']}_plus_{i}"] = np.nan
+            df[f"{err['label']}_minus_{i}"] = np.nan
+
```
RoyStegeman, line 1104: Could you do the same here as I did in [2e367b3](https://github.com/NNPDF/nnpdf/pull/2056/commits/2e367b39668ddf3384fca9f00470630b8533f4e2) for the 1jet case?

nnpdf_data/nnpdf_data/new_commondata/CMS_1JET_8TEV/metadata.yaml
comane: I am not completely sure, but I think this was discussed with @scarlehoff at some point.
The reason is probably that tanishq started using pT2 instead of pT, hence just to uniform to that.

nnpdf_data/nnpdf_data/filter_utils/legacy_jets_utils.py
comane: Yes, this is how it was before, where I had a `filter_utils.py` module in each of the datasets.

I guess that the idea is that in the dataset folders we only have a `filter.py` that produces the dataset, all of the utils instead go in `nnpdf_data/filter_utils`.
Otherwise we would have some utils in `nnpdf_data/filter_utils` and other within the dataset folder, which would be more confusing.

nnpdf_data/nnpdf_data/new_commondata/ATLAS_2JET_7TEV_R06/uncertainties.yaml
```diff
@@ -900,27 +900,27 @@ definitions:
...
-  art_sys_14: -1.9832179073202648e-08
-  art_sys_15: 5.042876412272401e-08
-  art_sys_16: -3.055464319753114e-08
-  art_sys_17: 8.299758199060853e-09
-  art_sys_18: -2.3278312338978964e-09
-  art_sys_19: 1.544485432184385e-09
-  art_sys_20: 5.301477069263894e-10
-  art_sys_21: -1.4029943620803375e-11
+- art_sys_1: -8666.003969584068
+  art_sys_2: 16.56039517543592
```
scarlehoff, line 904: Given that this many digits of precision are clearly wrong (because two runs of presumably the same code produce different results) I think the numbers should be truncated.

6-7 digits is probably more than the actual precision that we have.
The same is true for all other files where this happens.
I think it would be good (since this deals with utilities) to add an utility to truncate the outputs

nnpdf_data/nnpdf_data/filter_utils/legacy_jets_utils.py
```diff
@@ -0,0 +1,1148 @@
+"""
+Utils specifically needed to ease the implementation of the legacy jet datasets,
```
scarlehoff, line 2: why are these "legacy jet dataset".

These are simply the jet datasets for 7 and 8 TeV.

scarlehoff (commented):
Please add a `__init__.py` file to `filter_utils` as well.

In all honesty, I'm not entirely sure I see the point of having a `legacy_jet_utils.py` function where the functions inside are actually specific for each dataset.

However, I'm ok with concentrating all that stuff into `filter_utils` so that it can be easily skipped by the installation with the right rules in `pyproject.toml`.

nnpdf_data/nnpdf_data/new_commondata/CMS_1JET_8TEV/metadata.yaml
comane: I set it back to pT. I think this makes more sense as is the kin var for the process JET

Radonirinaunimi (approved):
This LGTM now! Thanks @comane.

@scarlehoff, are you fine with merging this now? Such that I can continue with https://github.com/NNPDF/nnpdf/pull/2100 and https://github.com/NNPDF/nnpdf/pull/2099.

scarlehoff:
Yes. I think it should be ok.

Issue #2058, opened 2024-04-19
Load positivity datasets with cuts


scarlehoff:
We always skip the cuts when loading positivity datasets. I'm not entirely sure why (it has always been done like that, maybe some parts of the code needs the positivity not to be cut)

https://github.com/NNPDF/nnpdf/blob/f1a0eb09f1d451663234ee5d47972334aa5ef5ee/validphys2/src/validphys/core.py#L590

However, now the positivity sets and fktables should be in the same footing as the others, so in principle we could also cut (in `x` for instance). Maybe removing this method (so that it goes to the parent, which applies the cut) is enough, but it should be tested.

RoyStegeman:
> However, now the positivity sets and fktables should be in the same footing as the others

Why? Positivity is quite special. I don't think we should e.g. not include positivity if we change the Q2 cut to above 5Gev2 if we want to test the stability for example

RoyStegeman:
Okay, as long as it's not by default I don't see any problem

comane:
Hi @scarlehoff, just to make sure I understand this correctly.
Would you like to have some rules in `cuts/filter.yml` that impose cuts in x for the positivity datasets?

Another options would be that of creating a new kinematics file eg for `XGL` -> `NNPDF_POS_2P24GEV_XGL_highX` where we simply remove the kinematic points that are below a certain treshold

comane:
Ok, I see.

Then I think that it's probably not enough to just remove the `load_commondata` from `LagrangeSetSpec`, it seems that positivity datasets are assigned a `Cuts` class whereas DataSetSpecs are assigned an `InternalCutsWrapper` which is the one aware of the rules.

PR #2059, opened 2024-04-19
Positivity cuts

comane:
This PR addresses #2058 

Main modifications:

-  removed `load_commondata` method from LagrangeSetSpec(DataSetSpec) so that it is directly inherited from `DataSetSpec` where cuts are loaded (before no cuts were loaded for LagrangeSetSpec type datasets)
- For a `LagrangeSetSpec` promoted `Cuts` to `InternalCutsWrapper` so as to allow for the proper loading of cuts. --> note: rules need now to be passed to `DataSetSpec`
- renamed kin of XGL and added new process for that positivity dataset
- Changed the names of some INTEG and POS processes

REPORTS
----------
- compare fits: https://vp.nnpdf.science/68xE8NcIQhu2i2hAwhII5A==/Scales0_pdf_report_report.html


n3fit/runcards/examples/nnpdf40-like_pos_cut.yaml
```diff
@@ -0,0 +1,189 @@
+#
```
scarlehoff, line 1: Please use a different name for this runcard. We should have only one `nnpdf40-like` to avoid generating confusion again.

Also, since this is just an example you can do with a subset of data.

validphys2/src/validphys/core.py
```diff
@@ -586,10 +594,6 @@ def to_unweighted(self):
         )
         return self
 
-    @functools.lru_cache
-    def load_commondata(self):
-        return self.commondata.load()
-
```
scarlehoff, line 592: Why has this been removed? Is it not used anywhere anymore?

n3fit/runcards/examples/nnpdf40-like_pos_cut.yaml
```diff
@@ -0,0 +1,189 @@
...
+    - dataset: NNPDF_POS_2P24GEV_XDQ
+      rule: "x > 0.1"
+
+    - dataset: NNPDF_POS_2P24GEV_XDB
+      rule: "x > 0.1"
+
+    - dataset: NNPDF_POS_2P24GEV_XSQ
+      rule: "x > 0.1"
+
+    - dataset: NNPDF_POS_2P24GEV_XSB
```
scarlehoff, line 178: Given that you are setting the `POS_XPDF` process type, you should be able to also do 

```suggestion
    - process_type: POS_XPDF
```

and have one single cut.

scarlehoff (commented):
Thanks! Only some minor comments. Could you also link in this PR some of the fits you've done using this branch?

nnpdf_data/nnpdf_data/new_commondata/NNPDF_POS_2P24GEV/metadata.yaml
```diff
@@ -56,7 +56,7 @@ implemented_observables:
     description: Deep Inelastic Scattering
     label: 'positivity dataset: DIS $c+\bar{c}$ structure function $F_2^c$'
     units: ''
-  process_type: DIS_F2C
+  process_type: POS_F2C
```
RoyStegeman, line 59: Why this change? The process is DIS_F2C, the only difference is that we use the dataset in a lagrange multiplier. Also, this change is not mirrored in `process_options.py`

n3fit/runcards/examples/nnpdf40-like_pos_cut.yaml
comane: Yes but at the same time the `Basic_runcard.yml` does not have any MS-bar POS sets.
So I would rather have something called:
`Basic_runcard_ms_bar_pos_cuts.yaml`, where the runcard is the same as the current nnpdf40-like_pos_cut.yaml apart from reduced datasets.
What do you think?

validphys2/src/validphys/core.py
comane: Yes exactly, this shouldn't be used anywhere else.

n3fit/runcards/examples/nnpdf40-like_pos_cut.yaml
scarlehoff: You can add a MS-bar POS set to the `Basic_runcard.yml`

n3fit/runcards/examples/Basic_runcard.yml
```diff
@@ -73,12 +73,17 @@ positivity:
...
   - {dataset: NNPDF_POS_2P24GEV_F2U, maxlambda: 1e6}        # Positivity Lagrange Multiplier
   - {dataset: NNPDF_POS_2P24GEV_FLL-19PTS, maxlambda: 1e6}
+  - {dataset: NNPDF_POS_2P24GEV_XGL, maxlambda: 1e6}
 
 ############################################################
 integrability:
   integdatasets:
   - {dataset: NNPDF_INTEG_3GEV_XT3, maxlambda: 1e2}
 
+added_filter_rules:
```
scarlehoff, line 83:
```suggestion
# Apply a cut to a dataset or process type, in this case to the positivity dataset `NNPDF_POS_2P24GEV_XGL` which has process type: POS_XPDF
added_filter_rules:
```

nnpdf_data/nnpdf_data/new_commondata/NNPDF_POS_2P24GEV/metadata.yaml
scarlehoff: I think it is better to keep it as a `POS_F2C`. The positivity are special. I would however call them all `POS_F2`?
The other reason why I think they should not be called `DIS_` is that there's a lot of hidden logic that treats `DIS_` processes differently, better  to keep the positivity ones differentiated.

Agree that a `POS_F2` would be needed in process type (I guess you are already doing it like that but positivity/integrability datasets don't need kinematic coverage so no need to implement one).

scarlehoff (approved):
Left some small comments, but other than that I think this is good to go.
Make sure you rebase the latests changes in master just to make sure nothing breaks but looks good to me.

nnpdf_data/nnpdf_data/new_commondata/NNPDF_POS_2P24GEV/metadata.yaml
comane: Just to be sure, would you change all process types starting with `POS_F2` with `POS_F2` (eg. `POS_F2C_CCP ` -> `POS_F2`)

What about other processes such as `POS_FLL` should we just leave them as they are?

comane: I did rename all of the `POS_F2*` process types to `POS_F2`. I have also added the corresponding process types in `process_options.py`

PR #2062, opened 2024-04-24
Add (and use) a default for IterEv

scarlehoff:
I wanted to deal with this comment by @felixhekhorn https://github.com/NNPDF/nnpdf/pull/2028#discussion_r1574778983 and it ended up growing so much that I thought it would be better to separate it in its own PR so it can be reviewed before merging it to #2028.

This does very little:

1. Adds some conditional default (and tests for them)
2. Ensures that `evolven3fit` will make use of that default and, in addition, will loudly fail if someone tries to mix stuff.

### DIFF
--- a/n3fit/src/evolven3fit/eko_utils.py
+++ b/n3fit/src/evolven3fit/eko_utils.py
@@ -7,7 +7,6 @@ import numpy as np
 from eko.io import runcards
 from eko.matchings import Atlas, nf_default
 from eko.quantities.heavy_quarks import MatchingScales
-from validphys.loader import Loader
 
 from . import utils
 
@@ -22,7 +21,6 @@ EVOLVEN3FIT_CONFIGS_DEFAULTS_TRN = {
 }
 
 EVOLVEN3FIT_CONFIGS_DEFAULTS_EXA = {
-    "ev_op_iterations": 30,
     "ev_op_max_order": (1, 0),
     "evolution_method": "iterate-exact",
     "inversion_method": "exact",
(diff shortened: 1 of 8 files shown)


nnpdf_data/nnpdf_data/theory.py
```diff
@@ -36,58 +44,78 @@ class TheoryCard:
     mt: float  # # [GeV] top mass
     Qmt: float  # [GeV] MSbar mass reference scale of the top
     ktThr: float  # Threshold ratio of the top
-    CKM: list[float]  # CKM matrix elements (running on the columns first, i.e. CKM_11 is CKM[0] and CKM_12 is CKM[1] and so on)
+    CKM: list[
```
andreab1997, line 47: I understand this comes from `black` but I would say that it is not the best formatting ever.

n3fit/src/evolven3fit/eko_utils.py
```diff
@@ -228,4 +221,3 @@ def find_nf(mu, theory, thresholds):
         nf = 5
     else:
         nf = 6
-    return nf
```
giacomomagni, line 231: is this a typo?

scarlehoff: Yes.

But I wonder whether the entire thing should go (since nf0 is now given so this is never going to be used).

giacomomagni: then we can drop this at all

scarlehoff:
@niclaurenti @giacomomagni please, have a look at my solution to avoid the need for `find_nf` in `evolven3fit`.
For the photon we are taking `Q0` instead of `Qref` and so it makes sense (I believe) to take `Q0` as well.

I've put `find_nf` as a method although the theory card is mostly accessed as a dictionary at the moment but if (eventually) the Theory class is passed around this can be used to find `nf` directly.

Issue #1925, opened 2024-02-01
Problem resolving explicit nodes


scarlehoff:
There's a few problems when resolving the explicit node of the covmat.

Some of these which involved the theory covmat have already been identified and will be solved in #1910 but some remain:

```yaml
meta:
    title: Example of bug
    keywords: [example]

pdfs:
    - id: NNPDF40_nnlo_lowprecision

theoryid: 162

use_cuts: "internal"

dataset_inputs:
      - { dataset: BCDMSP}

template_text: |
  {@ dataset_inputs::pdfs plot_fancy(normalize_to=data)@}

actions_:
  - report(main=true)
```

(the strange thing is that if I add `{@ dataset_inputs plot_fancy @}` in the `template_text`, the runcard works.

I don't know since when this bug has been around, but it is not new. I can reproduce it also with code from 2020 (nnpdf tag 4.0.0 and reportengine 0.29).

scarlehoff:
This is actually the same as #1496

PR #2066, opened 2024-04-30
add validphys.commondata to vp-setupfits providers list

RoyStegeman:
This is needed to do a MHOU fit. I guess somehow this broke while moving to the new commondata


scarlehoff:
I think this is due to some of the changes done when we found this problem: https://github.com/NNPDF/nnpdf/issues/1925

(it doesn't help that we don't have a test for MHOU fits...)

scarlehoff (approved):

PR #2068, opened 2024-05-03
update the closure test docs

RoyStegeman:


validphys2/src/validphys/closuretest/runcards/NNPDF40_closure_test.yml
```diff
@@ -92,16 +92,13 @@ datacuts:
 
 ############################################################
 theory:
-  theoryid: 53        # database id
+  theoryid: 700        # database id
```
scarlehoff, line 95: Do all the cuts above work at all?

(since you are at it... could you also change all the dataset names to the new version? Just remove the ones that don't have a matching one)

PR #2069, opened 2024-05-06
Implementation of additional systematic uncertainty in EIC data

toonhasenack:
Adding flat systematic uncertainties for EIC `g1_charm`. 
The value is provided by Yuxiang.


giacomomagni:
Do we want to do the same for EICc @Radonirinaunimi @enocera ?

Radonirinaunimi:
> Do we want to do the same for EICc @Radonirinaunimi @enocera ?

Thanks @toonhasenack and @giacomomagni for this! Could we have a theory vs data comparison here? As mentioned that'll provide us information on whether this normalization is sufficient. If that is not the case, we should not do the same for EIcC.

Radonirinaunimi:
@toonhasenack, this has now be moved here https://github.com/NNPDF/nnpdf/pull/2094.

PR #2071, opened 2024-05-07
ATLAS lepton rapidity 8 TeV data

comane:
Implements https://www.hepdata.net/record/149333

Data-Th comparison with NNPDF40
image

Kinematic coverage:
https://vp.nnpdf.science/9wx3nL2pTw6HKQQirZQn2g==


validphys2/src/validphys/process_options.py
```diff
@@ -217,6 +220,21 @@ def _dywboson_xq2map(kin_dict):
...
+    """
+    mass2 = kin_dict[_Vars.m_Z2]
+    sqrts = kin_dict[_Vars.sqrts]
+    abs_yll = kin_dict[_Vars.abs_yll]
+
+    x1 = np.sqrt(mass2) / sqrts * np.exp(-abs_yll)
+    x2 = np.sqrt(mass2) / sqrts * np.exp(abs_yll)
+    x = np.concatenate((x1, x2))
+    return np.clip(x, a_min=None, a_max=1, out=x), np.concatenate((mass2, mass2))
+
```
scarlehoff, line 236: Since (as it should be) Z and W share this logic I think it would be better to put them into one single function.

Whether to abstract the logic out and have two calls of the same function with different variables or a single function that uses `get_one_of` it's up to you.
I think I'd favour the `get_one_of` because I have the feeling that there might be datasets (for `Z`) that use rapidity instead of absolute rapidity and maybe even pseudorapidity...

nnpdf_data/nnpdf_data/new_commondata/ATLAS_Z0_8TEV/metadata.yaml
```diff
@@ -0,0 +1,59 @@
+# Generalia
+setname: "ATLAS_Z0_8TEV"
```
scarlehoff, line 2: We already have `ATLAS_Z0_8TEV_LOWMASS` and `ATLAS_Z0_8TEV_HIMASS`. I think that unless this dataset supersedes both of them, it should have some other qualifier @enocera

nnpdf_data/nnpdf_data/new_commondata/ATLAS_Z0_8TEV/metadata.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+    # Data 
+    data_central: data.yaml
+    data_uncertainties:
+      - uncertainties.yaml
+
+    # Theory
+    theory:
+      FK_tables:
+        - - TODO
```
scarlehoff, line 58: Please add the name of the grids.

nnpdf_data/nnpdf_data/new_commondata/ATLAS_Z0_8TEV/metadata.yaml
scarlehoff, line 2:
```suggestion
setname: "ATLAS_Z0_8TEV_ONPEAK"
```

scarlehoff (approved):

PR #2072, opened 2024-05-07
check also for closure test data in read_replica_pseudodata

RoyStegeman:
This adds functionality to the `read_replica_pseudodata` function to ensure it also checks for the existence of closure test pseuodata. 

Also, in the previous implementation the except statements were chained in a way that made the code act different from expected. Namely, both `except` statements correspond to the error raised under `try` (and not the second `except` corresponding to a potential error in the first `except`), as such the custom error was never raised. 

(yes I think the nesting is a bit ugly, but I'm not aware of a better way with try-except)

### DIFF
--- a/validphys2/src/validphys/pseudodata.py
+++ b/validphys2/src/validphys/pseudodata.py
@@ -18,8 +18,6 @@ from validphys.covmats import (
     sqrt_covmat,
 )
 
-FILE_PREFIX = "datacuts_theory_fitting_"
-
 log = logging.getLogger(__name__)
 
 DataTrValSpec = namedtuple('DataTrValSpec', ['pseudodata', 'tr_idx', 'val_idx'])
--- a/validphys2/src/validphys/pseudodata.py
+++ b/validphys2/src/validphys/pseudodata.py
@@ -80,25 +80,40 @@ def read_replica_pseudodata(fit, context_index, replica):
     log.debug(f"Reading pseudodata & training/validation splits from {fit.name}.")
     replica_path = fit.path / "nnfit" / f"replica_{replica}"
 
-    training_path = replica_path / (FILE_PREFIX + "training_pseudodata.csv")
-    validation_path = replica_path / (FILE_PREFIX + "validation_pseudodata.csv")
-
     try:
+        # The paths where sampled experimental data is stored
+        training_path = replica_path / "datacuts_theory_fitting_training_pseudodata.csv"
+        validation_path = replica_path / "datacuts_theory_fitting_validation_pseudodata.csv"
         tr = pd.read_csv(training_path, index_col=[0, 1, 2], sep="\t", header=0)
         val = pd.read_csv(validation_path, index_col=[0, 1, 2], sep="\t", header=0)
     except FileNotFoundError:
-        # Old 3.1 style fits had pseudodata called training.dat and validation.dat
-        training_path = replica_path / "training.dat"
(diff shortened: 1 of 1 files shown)


validphys2/src/validphys/pseudodata.py
```diff
@@ -80,25 +80,40 @@ def read_replica_pseudodata(fit, context_index, replica):
...
+            try:
+                # Old 3.1 style fits had pseudodata called training.dat and validation.dat
+                training_path = replica_path / "training.dat"
+                validation_path = replica_path / "validation.dat"
+                tr = pd.read_csv(
+                    training_path, index_col=[0, 1, 2], sep="\t", names=[f"replica {replica}"]
+                )
+                val = pd.read_csv(
+                    validation_path, index_col=[0, 1, 2], sep="\t", names=[f"replica {replica}"]
+                )
```
scarlehoff, line 110: So, I personally would go for this:

```python
trdata = Path(.../"datacuts_theory_fitting_training_pseudodata.csv")
trclosure = Path(.../"datacuts_theory_closuretest_fitting_training_pseudodata.csv")
tr31 = Path(../<3.1 version>)

if trdata.exists() and trclosure.exists():
    raise SomeError("Found both closure and normal data, please make up your mind and remove one of the two")
elif trdata.exists():
    tr = pd.read_csv
elif trclosure.exist():
    tr = pd.read_csv
elif tr1.exist(): # We reach 3.1 only as a fallback.
    tr = pd.read_csv
else:
    raise FileNotFound
```

RoyStegeman: Actually, do we still want to check for the 3.1 version? It has bee a long time since I last looked at such a fit, and this function is not called in any of the standard scripts for as far as I'm aware.

scarlehoff: If you want to remove it I certainly won't stop you.

achiefa: > So, I personally would go for this:

I think both approaches work fine and it's a matter of personal taste. I would go for @scarlehoff's implementation as well, since is more readable and coincise.

RE 3.1, I honestly don't know. Is there a case in which we might want to use 3.1 fits?

validphys2/src/validphys/pseudodata.py
```diff
@@ -80,25 +78,18 @@ def read_replica_pseudodata(fit, context_index, replica):
...
+    fakedata = fit.as_input().get("closuretest", {}).get("fakedata", False)
+    if fakedata:
+        tr_pseudodatafile = "datacuts_theory_closuretest_fitting_training_pseudodata.csv"
+        vl_pseudodatafile = "datacuts_theory_closuretest_fitting_validation_pseudodata.csv"
+    else:
+        tr_pseudodatafile = "datacuts_theory_fitting_training_pseudodata.csv"
+        vl_pseudodatafile = "datacuts_theory_fitting_validation_pseudodata.csv"
+
+    tr = pd.read_csv(replica_path / tr_pseudodatafile, index_col=[0, 1, 2], sep="\t", header=0)
+    val = pd.read_csv(replica_path / vl_pseudodatafile, index_col=[0, 1, 2], sep="\t", header=0)
```
scarlehoff, line 91: I would keep the try-except here to tell the user to use the flag.

scarlehoff: Yes.

scarlehoff (approved):
LGTM

PR #2074, opened 2024-05-09
Update to new commondata names - validphys 

### LINKED ISSUE #1969: Update to new commondata names in docs, examples and tests

RoyStegeman:
After dealing with the dataset names in n3fit in #2031, this PR continues to deal with the vp side. 

closes #1969


scarlehoff:
What's missing here? (I think the errors are just the fact that the plots have changed as well, right?)

RoyStegeman:
Yes, but so far I fixed far from all mentions of the old names so there is still a lot to do I'm afraid

Radonirinaunimi (commented):
@peterkrack for some of tests to pass, you need to rebase this on top of master. While for the figures (in `test_plots`), you need to re-generate them.

peterkrack:
I tried to replace ATLASTTBARTOT  with ATLAS_TTBAR_7TEV_TOT_X-SEC because I did not find any corresponding dataset among the new_commondata set names. test_plots.py fails with the new dataset. I am getting NotImplementedError: The process type INC for ATLAS_TTBAR_7TEV_TOT_X-SEC is not implemented. If I remove the ATLAS dataset from test_plots.py the test passes, at least when running it on stoomboot.

Radonirinaunimi:
> I am getting NotImplementedError: The process type INC for ATLAS_TTBAR_7TEV_TOT_X-SEC is not implemented. If I remove the ATLAS dataset from test_plots.py the test passes, at least when running it on stoomboot.

This one can be solved by simply rebasing, and indeed it is now.

> I tried to replace ATLASTTBARTOT with ATLAS_TTBAR_7TEV_TOT_X-SEC because I did not find any corresponding dataset among the new_commondata set names. test_plots.py fails with the new dataset.

It shouldn't be a problem which datasets you use to check the plotting. I you use `ATLAS_TTBAR_7TEV_TOT_X-SEC`, as indeed there is no correspondence to `ATLASTTBARTOT` in [dataset_names.yml](https://github.com/NNPDF/nnpdf/blob/master/nnpdf_data/nnpdf_data/new_commondata/dataset_names.yml), you just need to generate the plots and that'll become the new reference.

scarlehoff:
I've rebased on top of master and updated the regression tests.

validphys2/examples/pdf_flavour_plot.yaml
```diff
@@ -1,7 +1,7 @@
 meta:
   title: PDF flavours plot example
-  author: Rosalyn Pearson
-  keywords: [example]
+  author: Rosalyn L. Pearson
```
scarlehoff, line 3: I would honestly change the authors in the example from <any> to `Lazy Person` or `The Creator` or some other funny thing like that. Mainly so that if people uploads a report without changing these fields we know that we don't know who uploaded it instead of thinking it was Rosalyn :__

scarlehoff (approved):
I've tested the tests but haven't gone over every change of name in detail, so I'm just trusting they are sensible.

Just the one comment about the author in the tests (I've also opened the following issue #2130 related to that)

PR #2076, opened 2024-05-10
Hashable Rules

comane:
Rules (for cuts) are made to tuple instead of list.
This makes them hashable and allows eg them to contained within a `TupleComp` object


validphys2/src/validphys/config.py
```diff
@@ -1261,7 +1262,7 @@ def parse_filter_rules(self, filter_rules: (list, type(None))):
         """A list of filter rules. See https://docs.nnpdf.science/vp/filters.html
         for details on the syntax"""
         log.warning("Overwriting filter rules")
-        return filter_rules
+        return tuple(frozendict(rule) for rule in filter_rules) if filter_rules else None
```
scarlehoff, line 1265: Please  change the docstr from list to tuple. You might want to mention ad well that each rule is given as an inmutable dictionary.

validphys2/src/validphys/config.py
```diff
@@ -1286,17 +1286,16 @@ def produce_rules(
...
         """Produce filter rules based on the user defined input and defaults."""
-        from validphys.filters import Rule, RuleProcessingError, default_filter_rules_input
 
         theory_parameters = theoryid.get_description()
 
         if filter_rules is None:
             # Don't bother loading the rules if we are not using them.
             if use_cuts is not CutsPolicy.INTERNAL:
-                return None
+                return tuple()
```
scarlehoff, line 1298: Is this change necessary? `None` should be hashable and this could have collateral effects (like some code doing `if x is None`)

validphys2/src/validphys/config.py
```diff
@@ -1359,45 +1358,69 @@ def load_default_default_filter_settings(self, spec):
...
+        return parsed_filter_defaults
 
     def produce_defaults(
         self,
         q2min=None,
         w2min=None,
         maxTau=None,
         default_filter_settings=None,
-        filter_defaults={},
+        filter_defaults=FilterDefaults,
```
scarlehoff, line 1392:
```suggestion
        filter_defaults=None,
```

And check early whether it is None and change it to an empty dict.

validphys2/src/validphys/config.py
```diff
@@ -1359,45 +1358,69 @@ def load_default_default_filter_settings(self, spec):
...
 
-        if q2min is not None and "q2min" in filter_defaults and q2min != filter_defaults["q2min"]:
-            raise ConfigError("q2min defined multiple times with different values")
-        if w2min is not None and "w2min" in filter_defaults and w2min != filter_defaults["w2min"]:
-            raise ConfigError("w2min defined multiple times with different values")
+            if q2min is not None and filter_defaults["q2min"] != q2min:
+                raise ConfigError("q2min defined multiple times with different values")
+            
+            if w2min is not None and filter_defaults["w2min"] != w2min:
+                raise ConfigError("w2min defined multiple times with different values")
```
scarlehoff, line 1411: I think this is changing too much the behaviour of the function. I would leave this as close as possible to the previous behaviour.

validphys2/src/validphys/config.py
comane: I had some trouble at some point with a bit of the code saying `non Iterable` and therefore changed it to `tuple()`

I now changed it back, let's see after the tests

comane:
> Thanks for this, could you also remove frozenargs from the conda recipe / pyproject? It should not be needed anymore

I am not sure I understand, do you mean `frozendict`?  Because that one is actually still needed

scarlehoff:
> Because that one is actually still needed

I thought it was only needed by the filter rules?

validphys2/src/validphys/config.py
scarlehoff: No, before it would check whether I'm passing a `filter_defaults` that is incompatible with the values of `w2min`, `q2min` etc that I'm also passing.

Now this check is only done if the `filter_defaults` I'm passing is an instance of `FilterDefaults`. Note that regardless of the default input of the function I can call it with e.g.

`produce_defaults(qmin=1, filter_defaults={"q2min": 4})`

validphys2/src/validphys/config.py
```diff
@@ -1258,10 +1265,10 @@ def load_default_default_filter_rules(self, spec):
             )
 
     def parse_filter_rules(self, filter_rules: (list, type(None))):
-        """A list of filter rules. See https://docs.nnpdf.science/vp/filters.html
-        for details on the syntax"""
+        """A tuple of FilterRule objects. The rules are frozen dataclasses.
```
scarlehoff, line 1268:
```suggestion
        """A tuple of FilterRule objects. Rules are immutable after parsing.
```

validphys2/src/validphys/config.py
```diff
@@ -1271,12 +1278,15 @@ def parse_default_filter_rules_recorded_spec_(self, spec):
         return spec
 
     def parse_added_filter_rules(self, rules: (list, type(None)) = None):
-        return rules
+        """
+        Returns a tuple of AddedFilterRule objects. The rules are frozen dataclasses.
```
scarlehoff, line 1282:
```suggestion
        Returns a tuple of AddedFilterRule objects. Rules are immutable after parsing.
```

validphys2/src/validphys/config.py
```diff
@@ -1271,12 +1278,15 @@ def parse_default_filter_rules_recorded_spec_(self, spec):
...
+        AddedFilterRule objects inherit from FilterRule objects.
```
scarlehoff, line 1283: Why is `AddedFilterRule` not just a `FilterRule`?

validphys2/src/validphys/config.py
```diff
@@ -1305,7 +1314,7 @@ def produce_rules(
         try:
             rule_list = [
                 Rule(
-                    initial_data=rule,
+                    initial_data=rule.to_dict(),
```
scarlehoff, line 1317: I think that instead of passing `.to_dict` here, it is better to do, in the class `Rule`, 

```python
# For compatibility with legacy code that passed a dictionary
# TODO: make sure there are no calls left to `Rules` with a dictionary input
initial_data = initial_data.to_dict()
```

(you might want to cover this with a `isinstance` or `try-except`)

validphys2/src/validphys/config.py
```diff
@@ -1359,45 +1367,71 @@ def load_default_default_filter_settings(self, spec):
...
+            Currently these limits are ``q2min``, ``w2min``, and ``maxTau``.
         """
         log.warning("Overwriting filter defaults")
-        return filter_defaults
+
+        parsed_filter_defaults = FilterDefaults(
+                                            q2min = filter_defaults["q2min"] if "q2min" in filter_defaults else None, 
+                                            w2min = filter_defaults["w2min"] if "w2min" in filter_defaults else None,
+                                            maxTau = filter_defaults["maxTau"] if "maxTau" in filter_defaults else None,
+                                            )
```
scarlehoff, line 1392:
```suggestion
        parsed_filter_defaults = FilterDefaults(**filter_defaults)
```

And then you can initialise the attributes of the class `FilterDefaults` to `None`.

If for some reason that cannot be done, you can always do `filter_defaults.get("q2min")` which is equivalent to doing `filter_defaults["q2min"] if "q2min" in filter_defaults else None`

validphys2/src/validphys/filters.py
```diff
@@ -99,18 +101,63 @@ class FatalRuleError(Exception):
...
+        return filtered_dict
+
+
+@dataclasses.dataclass(frozen=True)
+class AddedFilterRule(FilterRule):
+    """
+    Dataclass which carries extra filter rule that is added to the
+    default rule.
+    """
+    pass
```
scarlehoff, line 144: as said above, I don't think we should separate AddedFilterRules and FilterRules, they should be at the same level once they are read.

validphys2/src/validphys/filters.py
```diff
@@ -511,7 +558,7 @@ def __init__(self, initial_data: dict, *, defaults: dict, theory_parameters: dic
         self.rule_string = self.rule
         self.defaults = defaults
         self.theory_params = theory_parameters
-        ns = {*self.numpy_functions, *self.defaults, *self.variables, "idat", "central_value"}
+        ns = {*self.numpy_functions, *self.defaults.to_dict().keys(), *self.variables, "idat", "central_value"}
```
scarlehoff, line 561:
```suggestion
        ns = {*self.numpy_functions, *self.defaults.to_dict(), *self.variables, "idat", "central_value"}
```

Not that it makes a difference though.

validphys2/src/validphys/loader.py
```diff
@@ -674,10 +674,10 @@ def check_default_filter_rules(self, theoryid, defaults=None):
         th_params = theoryid.get_description()
         if defaults is None:
             defaults = default_filter_settings_input()
-        return [
-            Rule(inp, defaults=defaults, theory_parameters=th_params, loader=self)
+        return tuple(
+            Rule(inp.to_dict(), defaults=defaults, theory_parameters=th_params, loader=self)
```
scarlehoff, line 678:
```suggestion
            Rule(inp, defaults=defaults, theory_parameters=th_params, loader=self)
```

(same as above, I think it'd be better to have `Rule` take the dataclass by default and convert it to a dict inside)

validphys2/src/validphys/tests/test_filter_rules.py
```diff
@@ -120,7 +120,7 @@ def test_added_rules():
             {
                 "speclabel": "fewer data",
                 "added_filter_rules": [
-                    {"dataset": "ATLAS_1JET_8TEV_R06_PTY", "rule": "pT < 1000", "reson": "pt cut"}
+                    {"dataset": "ATLAS_1JET_8TEV_R06_PTY", "rule": "pT < 1000", "reason": "pt cut"}
```
scarlehoff, line 123: We are already fixing bugs and naming problems thanks to these classes! ^__^

validphys2/src/validphys/config.py
```diff
@@ -1271,12 +1278,15 @@ def parse_default_filter_rules_recorded_spec_(self, spec):
...
+        """
+        Returns a tuple of AddedFilterRule objects. The rules are frozen dataclasses.
+        AddedFilterRule objects inherit from FilterRule objects.
+        """
+        return tuple(AddedFilterRule(**rule) for rule in rules) if rules else None
 
     # Every parallel replica triggers a series of calls to this function,
     # which should not happen since the rules are identical among replicas.
     # E.g for NNPDF4.0 with 2 parallel replicas 693 calls, 3 parallel replicas 1001 calls...
-    @freeze_args
```
scarlehoff, line 1279: I think now you can remove the comment above!

validphys2/src/validphys/config.py
comane: They are the same indeed (Added... inherits from FilterRule). But it might be useful to differentiate between the two for instance for error messages or don't know what.

So I think I prefer having two "different" objects for these

validphys2/src/validphys/filters.py
```diff
@@ -107,12 +107,14 @@ class FilterDefaults:
...
-    maxTau: float
+    q2min: float = None
+    w2min: float = None
+    maxTau: float = None
 
     def to_dict(self):
-        return dataclasses.asdict(self)
+        default_dict = dataclasses.asdict(self)
+        filtered_dict = {k: v for k, v in default_dict.items() if v is not None}
+        return filtered_dict
```
scarlehoff, line 117: Is this necessary? Before you were also returning `None` when there were no defaults.

validphys2/src/validphys/filters.py
```diff
@@ -457,10 +506,20 @@ class Rule:
 
     numpy_functions = {"sqrt": np.sqrt, "log": np.log, "fabs": np.fabs}
 
-    def __init__(self, initial_data: dict, *, defaults: dict, theory_parameters: dict, loader=None):
+    def __init__(self, initial_data: dataclasses.dataclass, *, defaults: dict, theory_parameters: dict, loader=None):
```
scarlehoff, line 509:
```suggestion
    def __init__(self, initial_data: FilterRule, *, defaults: dict, theory_parameters: dict, loader=None):
```

And I would be inclined to write in the error below that only `FilterRule`s are allowed.
The fact that we accept `Mapping`s is to make sure we don't break any new code yet but I don't think it should be advertised.

validphys2/src/validphys/filters.py
comane: yes, you are right. I don't think it is necessary

validphys2/src/validphys/config.py
scarlehoff: I think this is the only part left to be addressed to avoid having `filter_defaults` which are incompatible with the input values of `maxTau, q2min, w2min`

validphys2/src/validphys/config.py
```diff
@@ -1359,45 +1364,66 @@ def load_default_default_filter_settings(self, spec):
...
+        
         if w2min is not None and "w2min" in filter_defaults and w2min != filter_defaults["w2min"]:
             raise ConfigError("w2min defined multiple times with different values")
 
-        if (
-            maxTau is not None
-            and "maxTau" in filter_defaults
-            and maxTau != filter_defaults["maxTau"]
-        ):
+        if maxTau is not None  and filter_defaults["maxTau"] != maxTau:
```
scarlehoff, line 1416:
```suggestion
        if maxTau is not None and filter_defaults.get("maxTau", maxTau) != maxTau:
```

scarlehoff (approved):
Seems ok. I've done a few tests and I haven't seen anything obviously wrong.

validphys2/src/validphys/filters.py
```diff
@@ -99,18 +101,63 @@ class FatalRuleError(Exception):
...
+    IC: str = None
+
+    def to_dict(self):
+        rule_dict = dataclasses.asdict(self)
+        filtered_dict = {k: v for k, v in rule_dict.items() if v is not None}
+        return filtered_dict
+
+
+@dataclasses.dataclass(frozen=True)
+class AddedFilterRule(FilterRule):
```
RoyStegeman, line 139: What is the purpose of this class?

validphys2/src/validphys/config.py
comane: > What is the purpose of this class?
Hi @RoyStegeman  there was a little discussion here that might answer your question.

RoyStegeman: Okay thanks, I wasn't sure it was intentional or forgotten

PR #2080, opened 2024-05-15
Photon lo new

niclaurenti:
I just opened a new PR for #2040 since I rebased master but something was giving problems then I merged


scarlehoff:
Thanks. Once we have the NLO and NNLO QED variants I'll rerun the LO to test this branch.

Btw, could you have a look at #2067 ?

Basically to make sure that the theories there are what you used to generate the ekos and fktables!

scarlehoff (approved):
Seems to run ok

niclaurenti:
> Seems to run ok

Great. Do you already have a fit? Is it reasonable?

scarlehoff:
I've done only one replica.

It is LO so the results are by definition not reasonable at all..

PR #2082, opened 2024-05-16
update nnpdf40 runcard and add theory 40_000_000

### LINKED ISSUE #2075: Prepare a runcard with the exact 4.0 MHOU / QED / N3LO settings
In order to avoid any confusion, I will prepare a runcard that can be 1-to-1 compared to `NNPDF40_nnlo_as_01180_qcd`

However, since we want to avoid the usage of the old names and theories, I will also prepare theory `40_000_000` (as an exact copy of theory `708`) and will update the names of said runcard.

scarlehoff:
As promised in last week's PC.

Incidentally, this PR also solves a security issue we had in the code.

Fixes #2075 

I'd like to run a fit to check that everything is ok but don't have the resources available atm. Will come later today or tomorrow.


n3fit/runcards/examples/nnpdf40-like.yml
```diff
@@ -131,15 +129,16 @@ parameters: # This defines the parameter dictionary that is passed to the Model
 
 fitting:
   fitbasis: EVOL
+  savepseudodata: False
```
RoyStegeman, line 132: Why did you add this?

scarlehoff: Usually we don't need it... but yes maybe better to have it true by default...
```suggestion
  savepseudodata: True
```

RoyStegeman: The default is true if it's not set explicitly in the runcard, at some point it was changed to be stored by default because it's not that costly and reproducing it isn't stable at all. And I believe we had some problem because we didn't have this information for some fits

scarlehoff: I rather have the flag in the example even if set to the default value so that I can set to false without writing `savepsudodata` or something like that ^^U

scarlehoff:
Comparison https://vp.nnpdf.science/2Fw_aIEDT3amjUe8rDsBBA== 

This is ready now.

n3fit/runcards/examples/nnpdf40-like.yml
```diff
@@ -91,16 +91,14 @@ dataset_inputs:
...
   q2min: 3.49
   w2min: 12.5
 
 ################################################################################
 theory:
-  theoryid: 708
+  theoryid: 40000000
 
-sampling:
-  separate_multiplicative: false
```
scarlehoff, line 103: This is necessary to get the exact comparison of my comment but I believe that this should be the default now and I'll add it to this PR if it is fine for you 

@andreab1997 @RoyStegeman

(it should not be a big or relevant effect and I guess you are using it anyway, but it could have an effect in the closure tests)

andreab1997: Yes, I believe that we should remove this key now, and just enforcing the correct behaviour.

RoyStegeman: From this conversation I understood the default would be set to False such that it would allow us to remove these lines from the runcard, but instead just a comment was added

n3fit/runcards/examples/nnpdf40-like.yml
```diff
@@ -131,15 +133,16 @@ parameters: # This defines the parameter dictionary that is passed to the Model
...
-  - {fl: t8, trainable: false, smallx: [0.5878, 0.874], largex: [1.548, 3.265]}
-  - {fl: t15, trainable: false, smallx: [1.087, 1.146], largex: [1.493, 3.441]}
+  - {fl: sng, trainable: false, smallx: [1.091, 1.119], largex: [1.471, 3.021]}
+  - {fl: g, trainable: false, smallx: [0.7795, 1.095], largex: [2.742, 5.547]}
+  - {fl: v, trainable: false, smallx: [0.472, 0.7576], largex: [1.571, 3.559]}
+  - {fl: v3, trainable: false, smallx: [0.07483, 0.4501], largex: [1.714, 3.467]}
+  - {fl: v8, trainable: false, smallx: [0.5731, 0.779], largex: [1.555, 3.465]}
+  - {fl: t3, trainable: false, smallx: [-0.5498, 1.0], largex: [1.778, 3.5]}
+  - {fl: t8, trainable: false, smallx: [0.5469, 0.857], largex: [1.555, 3.391]}
+  - {fl: t15, trainable: false, smallx: [1.081, 1.142], largex: [1.491, 3.092]}
```
RoyStegeman, line 145: Did you take these from a particular fit?

scarlehoff: From `NNPDF40_nnlo_as_01180_qcd`

n3fit/runcards/examples/nnpdf40-like.yml
scarlehoff: I've added a comment in this PR and we take the default out in a different one (actually, @andreab1997 if you want to do the honors :P)

RoyStegeman (approved):

PR #2083, opened 2024-05-16
Unweighting implementation

toonhasenack:
Calculation of the unweighted weights in the vp framework. Module can be called with 
vp-unweight "path to chi2".csv "number of datapoints"

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,7 +107,7 @@ qed = ["fiatlux"]
 nolha = ["pdfflow", "lhapdf-management"]
 
 [tool.poetry-dynamic-versioning]
-enable = false
+enable = true
 vcs = "git"
 metadata = true
 dirty = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry_dynamic_versioning.backend"
 
 [tool.poetry]
 name = "nnpdf"
-version = "0.0.0"
+version = "4.0.9.post745.dev0+c7e687b31.dirty"
 description = "An open-source machine learning framework for global analyses of parton distributions."
 readme = "README.md"
 authors = [
@@ -47,6 +47,7 @@ postfit = "validphys.scripts.postfit:main"
 # validphys helpers and scripts
 vp-upload = "validphys.scripts.vp_upload:main"
 wiki-upload = "validphys.scripts.wiki_upload:main"
+vp-unweight= "validphys.scripts.vp_unweight:main"
 vp-get = "validphys.scripts.vp_get:main"
 vp-comparefits = "validphys.scripts.vp_comparefits:main"
 vp-fitrename = "validphys.scripts.vp_fitrename:main"
@@ -106,7 +107,7 @@ qed = ["fiatlux"]
 nolha = ["pdfflow", "lhapdf-management"]
 
 [tool.poetry-dynamic-versioning]
-enable = true
+enable = false
 vcs = "git"
 metadata = true
 dirty = true
--- a/validphys2/src/validphys/scripts/vp_unweight.py
+++ b/validphys2/src/validphys/scripts/vp_unweight.py
@@ -3,6 +3,7 @@ import pandas as pd
 from scipy.special import xlogy
 from typing import Tuple
 import argparse
+import matplotlib.pyplot as plt
 
 
 class Unweight:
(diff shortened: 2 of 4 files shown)


RoyStegeman:
Why is this a cli command and not just a vp action?

toonhasenack:
@RoyStegeman, I followed the suggestion of @scarlehoff to do it this way.

pyproject.toml
```diff
@@ -106,7 +107,7 @@ qed = ["fiatlux"]
 nolha = ["pdfflow", "lhapdf-management"]
 
 [tool.poetry-dynamic-versioning]
-enable = true
+enable = false
```
scarlehoff, line 110: Why did you disable versioning?

(it needs to be reenabled, but i'd like to know the exact problem that lead to beeding to disable it)

validphys2/src/validphys/scripts/vp_unweight.py
```diff
@@ -0,0 +1,154 @@
+import numpy as np
+import pandas as pd
+from scipy.special import xlogy
+from typing import Tuple
+from tqdm import tqdm
```
scarlehoff, line 5: I'm not too fond of tqdm because it usually makes logs and debugging harder to read.
But if you are set on using it you must add it to the dependencies as well.

validphys2/src/validphys/scripts/vp_unweight.py
```diff
@@ -0,0 +1,154 @@
...
+        weights.to_csv("weights.csv")
+
+    return weights
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Unweighting using chi squared values.")
+    parser.add_argument("chi2_name", help = "Add the filename of the chi2 dataset (.csv)")
+    parser.add_argument("N", help = "Add the amount of experimental datapoints that the chi2 is based on")
+    args = parser.parse_args()
+    chi2 = pd.read_csv(args.chi2_name).values
```
scarlehoff, line 152: How is the csv generated? This should be part of the script.

Ideally the script should have two steps

1) Check whether the csv exist and if It doesn't generate it

2) Do the actual computation

In the validphys language these would be two actions, where the second depends on the first, but we can make it into proper validphys actions at the end, it can remain a script for the time being.

For the first action you can use a lot of vp stuff (e.g., functions to compute the chi2, or load pdfs).

validphys2/src/validphys/scripts/vp_unweight.py
```diff
@@ -0,0 +1,154 @@
...
+                loc = np.where(entropies <= thresh)[0][0]
+            except:
+                print("Failed minimisation procedure! Defaulting to lowest entropy.")
+                loc = -1
+                
+        Nopt = Nps[loc]
+
+        return Nps, entropies, Nopt
+    
+def main(chi2, N, store = True):
```
scarlehoff, line 130: Please add also here some doctrings explaining what it does.

validphys2/src/validphys/scripts/vp_unweight.py
```diff
@@ -0,0 +1,154 @@
...
+from typing import Tuple
+from tqdm import tqdm
+import argparse
+
+class Unweight:
+    def __init__(self, weights: np.ndarray, *chis: Tuple[int, np.ndarray]) -> None:
+        """
+        Initialize the Unweight class.
+
+        Args:
```
scarlehoff, line 13: We are trying to follow the numpy conventions https://numpydoc.readthedocs.io/en/latest/format.html#parameters at least for new code (and specially for parameters, returns, etc). It helps consistency.

pyproject.toml
toonhasenack: I think this happened because I built NNPDF locally. I reverted it.

validphys2/src/validphys/scripts/vp_unweight.py
toonhasenack: This I could not find a better solution for. The calculation of the chi2 per replica is done in the nnpdf/external/ poljet code and if I would shift the calculation of the chi2 to validphys, I'm afraid I must migrate more functionality. However, I don't think the solution will be more elegant.

pyproject.toml
scarlehoff: But in principle you should be able to install with `pip install  -e .`  (or without `-e`) and it should not modify this file.

toonhasenack: Strange. It did happen in my case...

RoyStegeman: Is your environment managed by poetry?

Radonirinaunimi:
@toonhasenack Just to echo what has been above about the re/un-weighting being self-consistent. Everything except the generation of the theory predictions should be here, ie from the computations of the $\chi^2$ to the final dumping of the PDFs. Ideally, the theory predictions should also be stored in `../NNPDF/theories`.

Also agreed on this being vp actions instead of scripts.

If you want, I could take over this PR from this point.

PR #2085, opened 2024-05-17
Passes variant info to TupleComp for DataSetInput

RoyStegeman:
Passes variant info to TupleComp for DataSetInput

Also removes parenthesis from @lru_cache, see: https://bugs.python.org/issue36772


scarlehoff (approved):
Thinking more about this, given that `variant` is more general than `sys` if this creates a problem somewhere that would mean there was a bug.

So I'd be willing to merge this asap.

This also paves the way to this issue I just opened https://github.com/NNPDF/nnpdf/issues/2086

PR #2087, opened 2024-05-22
Make EKO into a separate resource

### LINKED ISSUE #2081: Separate `eko.tar` and fastkernel tables in the server
Given that the tables for our current NNLO theory are about ~700 MB and the eko for the postfit evolution instead ~1 GB, I think it would be very convenient to separate both.

I think this would bring only advantages:
1. ekos are not needed for validphys, so no need to have them in your computer to make plots
2. If you need to copy stuff to your nodes when doing a fit, you halve the size of what you are copying for the fit itself.
3. We have many theories without an eko, where the eko is actually shared (e.g. theory 200, 400, 700, 708, 600, 40000000). Making it a separate entity means you can have just one 1GB object in the server instead of many.

I would not change anything of the architecture once they are downloaded since they get used in the same manner. I would just create another resource which then gets downloaded to the right theory id and evolven3fit will depend on it.

scarlehoff:
Closes #2081 

It can be tested with `theory_40001002` 

From an user perspective there should be no differences. When you only need the fktables / theory (i.e., for reports or fits, when doing `check_theoryID`) only the fktables will be downloaded.
If the eko needs to be downloaded as well (only for evolution), then it will be downloaded into the same folder as the theory (done in this way so that all the theories that have the eko inside work just the same).

There's a separate `check_eko(thid)` which is used by `evolven3fit` and returns the path to the `eko.tar` of the theory, checking first the theory.

At the moment the ekos range between the 1.5  and 3.5 GB while the whole set of fktables used for 4.0 is about 600 MB so this can have a huge impact e.g. when doing reports in your laptop when the PDF has been generated elsewhere.

### DIFF
--- a/n3fit/src/evolven3fit/cli.py
+++ b/n3fit/src/evolven3fit/cli.py
@@ -2,7 +2,7 @@ from . import evolve, utils
 
 
 def cli_evolven3fit(
-    configuration_folder, q_fin, q_points, op_card_info, theory_card_info, dump, load, force
+    configuration_folder, q_fin, q_points, op_card_info, theory_card_info, force, load, dump
 ):
     """Evolves the fitted PDFs.
 
--- a/n3fit/src/evolven3fit/evolve.py
+++ b/n3fit/src/evolven3fit/evolve.py
@@ -10,7 +10,7 @@ import numpy as np
 import eko
 from eko import basis_rotation, runner
 from reportengine.compat import yaml
-from validphys.loader import Loader
+from validphys.loader import FallbackLoader as Loader
 
 from . import eko_utils, utils
 
@@ -63,12 +63,11 @@ def evolve_fit(
     stdout_log = logging.StreamHandler(sys.stdout)
     for log in [log_file, stdout_log]:
         log.setFormatter(LOGGING_SETTINGS["formatter"])
-    
+
     # The log file will get everything
     log_file.setLevel(LOGGING_SETTINGS["level"])
     # While the terminal only up to info
     stdout_log.setLevel(logging.INFO)
-
 
     for logger in (_logger, *[logging.getLogger("eko")]):
         logger.handlers = []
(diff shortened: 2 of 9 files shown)


RoyStegeman (approved):
The thing is dislike a bit about this is that as part of my workflow I will often (manually) run `vp-setupfit` before submitting many jobs to make sure that all the parallel jobs won't try to download the same thing. This PR means I'll have to be more careful about it and find some way to download the ekos as well.

scarlehoff:
> Ah no I just see evolven3fit is not on top of reportengine so there is no way to do --no-net which I think is import to prevent the issue of accidentally trying to download the same thing many times.

I can add a `--no-net` option to `evolven3fit` if you want.
However, would you like me to add (in that case) a `check_eko` option to `setupfit` so that it downloads _also_ the eko? So we can consider `setupfit` as a "download and check _everything_ that I will need for the fit". n3fit will not check the eko.

Putting `evolven3fit` on top of validphys would fix quite a few issues but I'm certainly not doing that...

scarlehoff:
>  would it be fine if it just downloads the eko along with theoryid from the theory key and not theoryids from the theorycovmatconfig (or whatever that key is called)?

Yes. I think it also makes sense since only the eko for the fit's theoryid is needed for setting up the fit.

RoyStegeman:
> Yes. I think it also makes sense since only the eko for the fit's theoryid is needed for setting up the fit.

Of course, but it immediatly adds 1.5GB more to the CI so I was just asking/wondering how much wiggle room we have. 

Anyway, downloading the eko with vp-setupfit would be my preferred solution

scarlehoff:
Done. Let's see what does the fitbot think.

scarlehoff:
"ok, everything works ok but it would be nice to add a nice small printout. FML" :)

scarlehoff:
Can I merge this @RoyStegeman / are you happy with the setupfit solution or do you want to have a second look?

PR #2088, opened 2024-05-22
add resample_pseudodata flag

RoyStegeman:
For parameter fits it is important not to distort the pseudodata sampling distribution. This PR adds a flag `resample_negative_pseudodata` that can be set to False to also accept negative pseudodata values instead of resampling.

### DIFF
--- a/validphys2/src/validphys/pseudodata.py
+++ b/validphys2/src/validphys/pseudodata.py
@@ -129,6 +129,7 @@ def make_replica(
     sep_mult,
     genrep=True,
     max_tries=int(1e6),
+    resample_pseudodata=True,
 ):
     """Function that takes in a list of :py:class:`validphys.coredata.CommonData`
     objects and returns a pseudodata replica accounting for
@@ -162,6 +163,10 @@ def make_replica(
         The stochastic nature of replica generation means one can obtain (unphysical) negative predictions.
         If after max_tries (default=1e6) no physical configuration is found,
         it will raise a :py:class:`ReplicaGenerationError`
+
+
+    resample_pseudodata: bool
+        Set to False to allow for negative predictions and avoid resampling
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


validphys2/src/validphys/pseudodata.py
```diff
@@ -263,7 +268,10 @@ def make_replica(
         # Shifting pseudodata
         shifted_pseudodata = (all_pseudodata + shifts) * mult_part
         # positivity control
-        if np.all(shifted_pseudodata[full_mask] >= 0):
+        if resample_pseudodata:
+            if np.all(shifted_pseudodata[full_mask] >= 0):
+                return shifted_pseudodata
+        else:
```
scarlehoff, line 274:
```suggestion
        if np.all(shifted_pseudodata[full_mask] >= 0) or not resample_negative_pseudodata:
```

validphys2/src/validphys/pseudodata.py
```diff
@@ -129,6 +129,7 @@ def make_replica(
     sep_mult,
     genrep=True,
     max_tries=int(1e6),
+    resample_pseudodata=True,
```
scarlehoff, line 132:
```suggestion
    resample_negative_pseudodata=True,
```

I think this way it is clearer what is happening.

validphys2/src/validphys/pseudodata.py
```diff
@@ -163,6 +164,10 @@ def make_replica(
         If after max_tries (default=1e6) no physical configuration is found,
         it will raise a :py:class:`ReplicaGenerationError`
 
+
+    resample_pseudodata: bool
+        Set to False to allow for negative predictions and avoid resampling
+
```
scarlehoff, line 170:
```suggestion
    resample_negative_pseudodata: bool
        When True, replicas that produce negative predictions will be resampled for ``max_tries`` until all points are positive (default: True)
```

scarlehoff (commented):
I guess this will be a runcard parameter?

Could you add it either to docs or to some runcard example (even with the default value) just to signal that it exists.

RoyStegeman:
This doesn't do any resampling and just accepts also negative xsecs

scarlehoff (approved):

andreab1997:
> This doesn't do any resampling and just accepts also negative xsecs

Ok yes but I understood you did try the resampling no?

RoyStegeman:
Yes I did do MHOU fits with resampling (so the same as in master now), but I didn't track how much time that took

Issue #1991, opened 2024-03-08
Processes `HQP_MQQ` and `INC` are not implemented in `process_options`


scarlehoff:
The process `HQP_MQQ` is currently being used by a number of datasets (e.g., https://github.com/NNPDF/nnpdf/blob/9777baf5a96187e3d3d17f595bdfe521c70e9eb1/nnpdf_data/nnpdf_data/new_commondata/ATLAS_TTBAR_13TEV_LJ_DIF/metadata.yaml#L22)

but it is not implemented in
https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/process_options.py

Also `INC` (used for instance in: https://github.com/NNPDF/nnpdf/blob/9777baf5a96187e3d3d17f595bdfe521c70e9eb1/nnpdf_data/nnpdf_data/new_commondata/ATLAS_TTBAR_13TEV_TOT/metadata.yaml#L20 ) should be added, otherwise the results won't probably be correct.

The exception corresponding to this issue in `test_all_datasets` should be removed before this is closed

PR #2092, opened 2024-05-26
Add ttb processes xq2 maps

t7phy:
fix top mass, add x-q2 map for mttbar, fix docstrings for other ttb vars, apply black

### DIFF
--- a/validphys2/src/validphys/process_options.py
+++ b/validphys2/src/validphys/process_options.py
@@ -191,10 +191,8 @@ def _hqp_mqq_xq2map(kin_info):
 
 def _inc_xq2map(kin_info):
     # Compute x, Q2
-    if {"k1", "k2", "k3"} <= kin_info.keys():
-        mass2 = kin_info["k2"]
-    else:
-        mass2 = kin_info.get_one_of("m_W2", "m_Z2", "m_t2")
+    # k2 necessary to take the mass for DY inclusive cross sections still not migrated
+    mass2 = kin_info.get_one_of(_Vars.m_W2, _Vars.m_Z2, _Vars.m_t2, "k2")
 
     return np.sqrt(mass2) / kin_info[_Vars.sqrts], mass2
 
--- a/validphys2/src/validphys/process_options.py
+++ b/validphys2/src/validphys/process_options.py
@@ -24,7 +24,7 @@ class _Vars:
     ystar = "ystar"
     ydiff = "ydiff"
     m_jj = "m_jj"
-    p_T2 = "p_T2" # This one is wrong, should be pT2
+    p_T2 = "p_T2"  # This one is wrong, should be pT2
     y_t = "y_t"
     y_ttBar = "y_ttBar"
     m_t2 = "m_t2"
@@ -38,7 +38,7 @@ class _Vars:
 class _KinematicsInformation:
     """Read the 3 columns dataframe corresponding to the values set in the
     ``kinematic_coverage`` field into a dictionary defining the name of the variables.
-    
+
     Adds the special "sqrts" key unless it is already part of the kinematic coverage.
 
     Provides a ``.get_one_of`` method that accepts any number of variables
@@ -189,10 +189,14 @@ def _hqp_ptq_xq2map(kin_info):
     Q = np.sqrt(QMASS2 + kin_info[_Vars.pT_t] * kin_info[_Vars.pT_t]) + kin_info[_Vars.pT_t]
     return Q / kin_info[_Vars.sqrts], Q * Q
 
+
 def _hqp_mqq_xq2map(kin_info):
     # Compute x, Q2
     QQMASS2 = (2 * TMASS) * (2 * TMASS)
-    Q = np.sqrt(QQMASS2 + kin_info[_Vars.m_ttBar] * kin_info[_Vars.m_ttBar]) + kin_info[_Vars.m_ttBar]
+    Q = (
+        np.sqrt(QQMASS2 + kin_info[_Vars.m_ttBar] * kin_info[_Vars.m_ttBar])
+        + kin_info[_Vars.m_ttBar]
+    )
     return Q / kin_info[_Vars.sqrts], Q * Q
 
 
@@ -208,12 +212,13 @@ def _displusjet_xq2map(kin_info):
     x = q2 * q2 / s / (pt**2 - q2)
     return x, q2
 
+
 def _dywboson_xq2map(kin_dict):
     """
     Computes x and q2 mapping for pseudo rapidity observables
     originating from a W boson DY process.
     """
-    mass2 = kin_dict[_Vars.m_W2]    
+    mass2 = kin_dict[_Vars.m_W2]
     sqrts = kin_dict[_Vars.sqrts]
     eta = kin_dict[_Vars.eta]
 
@@ -223,6 +228,7 @@ def _dywboson_xq2map(kin_dict):
     x = np.concatenate((x1, x2))
     return np.clip(x, a_min=None, a_max=1, out=x), np.concatenate((mass2, mass2))
 
+
 def _dyncpt_xq2map(kin_info):
     """
     Computes x and q2 mapping for DY NC dilepton
@@ -230,7 +236,7 @@ def _dyncpt_xq2map(kin_info):
     """
     q2 = kin_info[_Vars.m_Z2]
     pt = kin_info[_Vars.pT]
-    s = kin_info[_Vars.sqrts]**2
+    s = kin_info[_Vars.sqrts] ** 2
     x = q2 * q2 / s / (pt**2 - q2)
     return x, q2
 
--- a/validphys2/src/validphys/process_options.py
+++ b/validphys2/src/validphys/process_options.py
@@ -193,7 +193,6 @@ def _inc_xq2map(kin_info):
     # Compute x, Q2
     if {"k1", "k2", "k3"} <= kin_info.keys():
         mass2 = kin_info["k2"]
-        kin_info[_Vars.sqrts] = kin_info["k3"]
     else:
         mass2 = kin_info.get_one_of("m_W2", "m_Z2", "m_t2")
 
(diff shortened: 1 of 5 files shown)


validphys2/src/validphys/process_options.py
```diff
@@ -10,7 +10,7 @@
 import numpy as np
 from validobj.custom import Parser
 
-TMASS = 173.3
+TMASS = 172.5
```
scarlehoff, line 13: Please don't change this because this was the value that was there in the old version so until _all_ processes are migrated this should not be touched.
If you need the value of the mass of the top add it to the kinematic information of the relevant dataset.

validphys2/src/validphys/process_options.py
```diff
@@ -190,6 +190,16 @@ def _hqp_ptq_xq2map(kin_info):
     return Q / kin_info[_Vars.sqrts], Q * Q
 
 
+def _hqp_mqq_xq2map(kin_info):
+    # Compute x, Q2
+    QQMASS2 = (2 * TMASS) * (2 * TMASS)
```
scarlehoff, line 195: Since in many of the datasets you also have `m_t2` please use that instead of `TMASS`.

scarlehoff (commented):
Please add x-q2 plots for a few datasets as well to make sure they work as expected.

validphys2/src/validphys/process_options.py
```diff
@@ -185,21 +183,31 @@ def _hqp_yqq_xq2map(kin_info):
...
 
 
+def _inc_xq2map(kin_info):
+    # Compute x, Q2
+    if {"k1", "k2", "k3"} <= kin_info.keys():
+        kin_info[_Vars.m_X2] = kin_info["k2"]
+        kin_info[_Vars.sqrts] = kin_info["k3"]
+
+    mass2 = kin_info.get_one_of(m_W2, m_Z2, m_t2, m_X2)
+    return np.sqrt(mass2) / kin_info[_Vars.sqrts], mass2
```
scarlehoff, line 208: `m_X2` is not a variable.
In this case it would be better to do something like

```
if {"k1 ... 
    m2 = kin_info["k2"]
else:
    m2 = kin_info.get_one_of(...)

and same for sqrts.
```

And then use the formula referring to the two variables you just got directly.

t7phy:
@scarlehoff https://vp.nnpdf.science/Utit4-sfQqe5aZeY76h0dg==/ is based on 4 1d and 1 2d distributions. All work. 
to make INC work, it becomes necessary to comment out the if condition for k1,k2,k3 and then it also works fine.

validphys2/src/validphys/process_options.py
```diff
@@ -157,39 +155,51 @@ def _dijets_xq2map(kin_info):
...
+    # Compute x, Q2
+    Q = kin_info[_Vars.m_ttBar] / 4
     return Q / kin_info[_Vars.sqrts], Q * Q
 
 
+def _inc_xq2map(kin_info):
+    # Compute x, Q2
+    if {"k1", "k2", "k3"} <= kin_info.keys():
+        mass2 = kin_info["k2"]
+        kin_info[_Vars.sqrts] = kin_info["k3"]
```
scarlehoff, line 196:
```suggestion
```

scarlehoff (commented):
What is wrong with the inc here? I don't see a problem, at least in the function?

validphys2/src/validphys/process_options.py
t7phy: As I told you yesterday on slack, that leads to an error. If i comment out the if condition, then it works.

t7phy: Also I don't understand the suggested change. We need to sqrts for x.

scarlehoff: sqrts is included automatically (or should be) from the energy.

> As I told you yesterday on slack, that leads to an error.

What is the error?

t7phy:
> What is wrong with the inc here? I don't see a problem, at least in the function?

NotImplementedError: The process type INC for CMS_TTBAR_13TEV_TOT_X-SEC is not implemented

RoyStegeman:
https://github.com/NNPDF/nnpdf/issues/1991

t7phy:
> #1991

?

scarlehoff:
> NotImplementedError: The process type INC for CMS_TTBAR_13TEV_TOT_X-SEC is not implemented

Ah, ok.

This is not the actual error. The actual error will be a bit above if you scroll up, this is after several layers of redirections. 
The problem is on `kin_info.keys()`, do instead of the `if` condition:

```
mass2 = kin_info.get_one_of("m_W2", "m_Z2", "m_t2", "k2")
```

The if statement in the other one is because it is working with a dictionary instead of the kin_info file.

validphys2/src/validphys/process_options.py
```diff
@@ -157,39 +155,50 @@ def _dijets_xq2map(kin_info):
...
+    Q = kin_info[_Vars.m_ttBar] / 4
     return Q / kin_info[_Vars.sqrts], Q * Q
 
 
+def _inc_xq2map(kin_info):
+    # Compute x, Q2
+    if {"k1", "k2", "k3"} <= kin_info.keys():
+        mass2 = kin_info["k2"]
+    else:
+        mass2 = kin_info.get_one_of("m_W2", "m_Z2", "m_t2")
```
scarlehoff, line 197:
```suggestion
    # k2 necessary to take the mass for DY inclusive cross sections still not migrated
    mass2 = kin_info.get_one_of(_Vars.m_W2, _Vars.m_Z2, _Vars.m_t2, "k2")
```

Just to make the suggestion more explicit.

t7phy: yes yes it was clear, I just hadn't gotten to correcting it yet xD

t7phy:
@scarlehoff and here is a report showing INC also works fine: https://vp.nnpdf.science/Nu5Z9OADRF-SRf8E2D3ufg==

validphys2/src/validphys/process_options.py
```diff
@@ -157,39 +155,48 @@ def _dijets_xq2map(kin_info):
...
 def _hqp_yq_xq2map(kin_info):
     # Compute x, Q2
-    if {"k1", "k2", "k3"} <= kin_info.keys():
-        kin_info[_Vars.y_t] = kin_info["k1"]
-        kin_info[_Vars.m_t2] = kin_info["k2"]
-        kin_info[_Vars.sqrts] = kin_info["k3"]
-
-    mass2 = kin_info.get_one_of(_Vars.m_t2, _Vars.m_ttBar)
-
+    # Theory predictions computed with HT/4 see 1906.06535
```
scarlehoff, line 158:
```suggestion
    # Theory predictions computed with HT/4 ~ mt/2 for rapidity distr.
    # see section 3 from 1906.06535
```

so that if someone wants to source where this comes from is easier to find

validphys2/src/validphys/process_options.py
```diff
@@ -157,39 +155,48 @@ def _dijets_xq2map(kin_info):
...
     x = np.concatenate((x1, x2))
-    return np.clip(x, a_min=None, a_max=1, out=x), np.concatenate((q2, q2))
+    return np.clip(x, a_min=None, a_max=1, out=x), np.concatenate((q2, q2)) / 4
 
 
 def _hqp_ptq_xq2map(kin_info):
     # Compute x, Q2
-    QMASS2 = TMASS * TMASS
-    Q = np.sqrt(QMASS2 + kin_info[_Vars.pT_t] * kin_info[_Vars.pT_t]) + kin_info[_Vars.pT_t]
+    Q = (kin_info[_Vars.m_t2] + kin_info[_Vars.pT_t] * kin_info[_Vars.pT_t]) ** 0.5 / 2
```
scarlehoff, line 182: Why divided by 2? Wasn't it divided by 4?

validphys2/src/validphys/process_options.py
```diff
@@ -157,39 +155,48 @@ def _dijets_xq2map(kin_info):
...
-    mass2 = kin_info.get_one_of(_Vars.m_t2, _Vars.m_ttBar)
+    # Theory predictions computed with HT/4 see 1906.06535
+    mass2 = kin_info[_Vars.m_t2]
     ratio = np.sqrt(mass2) / kin_info[_Vars.sqrts]
     x1 = ratio * np.exp(kin_info[_Vars.y_ttBar])
     x2 = ratio * np.exp(-kin_info[_Vars.y_ttBar])
     q2 = kin_info[_Vars.m_t2]
     x = np.concatenate((x1, x2))
-    return np.clip(x, a_min=None, a_max=1, out=x), np.concatenate((q2, q2))
+    return np.clip(x, a_min=None, a_max=1, out=x), np.concatenate((q2, q2)) / 4
```
scarlehoff, line 177: Given that this function and the one above are so similar, I think it would be better to have `rapidity = kin_info.get_one_of(_Vars.y_ttBar, _Vars.y_t)` and use the same function for both.

(unless there's some important difference I'm not missing)

scarlehoff:
Thanks for this. Just a few minor comment and this will be good to go.

validphys2/src/validphys/process_options.py
t7phy: The paper gives a different expression for pT

scarlehoff: I thought you were always using HT/4?

t7phy: > I thought you were always using HT/4?

true dat! but I could do that in MATRIX, I can't do that here as I don't have the pT of tbar

t7phy: for ref, this is HT/4: 
image

scarlehoff: At LO pT_t == pT_tb since they are back-to-back.

But I see, HT/4 ~ sqrt(mt^2 + pt^2) / 2

Could you also write the expression for HT in one of the functions as well please?

t7phy:
> Thanks for this. Just a few minor comment and this will be good to go.

done!

scarlehoff (approved):
I've removed the special case in the tests for `HQP_MQQ` and `INC`, once the tests pass this can be merged.

Thanks!

Issue #2025, opened 2024-03-27
Update the kinematic part of all ported commondata files so that they are not just k1, k2, k3 but have instead variable names.


scarlehoff:

scarlehoff:
Labels need to be beautified a bit, but they are no longer simply k1-k2-k3

PR #2093, opened 2024-05-28
kinematic labels

tgiani:
addressing #2025


tgiani:
regarding `ATLAS_WJ_JET_8TEV`: is there a reason because of which `k2` is set to a placeholder rather than `m_W^2`?
The original buildmaster implementation is here https://github.com/NNPDF/buildmaster/blob/master/filters/ATLAS_WJET_8TEV.cc

nnpdf_data/nnpdf_data/new_commondata/ATLAS_DY_13TEV/kinematics_TOT.yaml
```diff
@@ -1,37 +1,37 @@
 bins:
-- k1:
+- zero:
     min: null
     mid: 0.0
     max: null
-  k2:
+  MWZ:
```
scarlehoff, line 6: Please use the names from https://github.com/NNPDF/nnpdf/blob/37f48752cc362291c06df2f3990f815276bf12ee/validphys2/src/validphys/process_options.py#L16

(so `m_W2` instead of `MW2`...or... in this case, `m_V` and we will need to add a new variable for that)

scarlehoff:
I took the values from the commondata that were available, so I guess at some point it became a placeholder.

If you want to change it feel free. If not... we've lived with that for a while so ¯\_(ツ)_/¯

nnpdf_data/nnpdf_data/new_commondata/ATLAS_DY_13TEV/kinematics_TOT.yaml
tgiani: what s the procedure to add a new variable? In the case above we need `m_V`, for `ATLAS_PH_13TEV` we need the transverese ewnergy square `ET2` and I guess there will be other cases. Is it enought to add these new variables to the `_Vars` class?

tgiani:
just to keep track of the new variables:
- `ATLAS_DY_13TEV` : `m_V`
-  `ATLAS_PH_13TEV` : `ET2`
- `ATLAS_WPWM_13TEV` : `m_W`
- `ATLAS_Z0_7TEV_49FB`: `m_ll`, `m_ll2`

nnpdf_data/nnpdf_data/new_commondata/ATLAS_DY_13TEV/kinematics_TOT.yaml
scarlehoff: Yes. And then for the processes for which you want to add it, include it in the "accepted variables".

(if the process still doesn't exist you can either create it or leave it for later, it should not matter)

scarlehoff:
Closed since most of them have now been done in the specific PRs.

PR #2096, opened 2024-05-31
Allow for underscore in theoryfiles

RoyStegeman:


nnpdf_data/nnpdf_data/theorydbutils.py
```diff
@@ -51,7 +51,17 @@ def fetch_theory(theory_database: Path, theoryID: int):
...
+    # Since theoryfile names may contain underscores, the theoryfile name cannot uniquely be
+    # determined from the theoryID. Therefore we create a mapping between the theoryid as python
+    # int and the corresponding file
+    available_theories = {
+        int(theoryfile.name.removesuffix('.yaml')): theoryfile.name
+        for theoryfile in theory_database.glob("*.yaml")
+    }
+    theoryfile = available_theories[theoryID]
+
+    filepath = theory_database / theoryfile
```
scarlehoff, line 64: You can just save the full path in the dictionary and skip this step.

nnpdf_data/nnpdf_data/theorydbutils.py
```diff
@@ -51,7 +51,17 @@ def fetch_theory(theory_database: Path, theoryID: int):
...
-    filepath = theory_database / f"{theoryID}.yaml"
+
+    # Since theoryfile names may contain underscores, the theoryfile name cannot uniquely be
+    # determined from the theoryID. Therefore we create a mapping between the theoryid as python
+    # int and the corresponding file
+    available_theories = {
+        int(theoryfile.name.removesuffix('.yaml')): theoryfile.name
+        for theoryfile in theory_database.glob("*.yaml")
+    }
+    theoryfile = available_theories[theoryID]
```
scarlehoff, line 62:
```suggestion
    try:
        theoryfile = available_theories[theoryID]
    except KeyError:
         raise TheoryNotFoundInDatabase(f"Theory {theoryID} not found in database") 
```

(this fixes one of the errors in the tests, not sure about the others)

nnpdf_data/nnpdf_data/theorydbutils.py
```diff
@@ -51,7 +51,17 @@ def fetch_theory(theory_database: Path, theoryID: int):
...
     """
-    filepath = theory_database / f"{theoryID}.yaml"
+
+    # Since theoryfile names may contain underscores, the theoryfile name cannot uniquely be
+    # determined from the theoryID. Therefore we create a mapping between the theoryid as python
+    # int and the corresponding file
+    available_theories = {
+        int(theoryfile.name.removesuffix('.yaml')): theoryfile.name
+        for theoryfile in theory_database.glob("*.yaml")
+    }
```
scarlehoff, line 61: I would cover this as well with a try-except block in case someone left a file "not_a_theory.yaml" in this folder with "file whatever doesn't conform to the .yaml convention"

It would be also good to check that not two files are the same.

Something like:
```suggestion
    available_theories = {}
    for theoryfile in theory_database.glob("*.yaml"):
        try:
            tmp_id = int(theoryfile.name.removesuffix('.yaml'))
        except ValueError as e:
             return InvalidTheoryName(f"Bad theory name for{theoryfile}") from e
         if tmp_id in available_theories:
             another = available_theories[tmp_id]
             raise ValueError("Two theory files with same id: {theoryfile} and {another}")
         available_theories[tmp_id] = theoryfile
```

And I would make this into its own function `get_available_theory_cards(path)` with a `lru_cache` so that for a given path you don't do all this every time but then a test that simply loops over all the theories would catch a few possible bugs that the underscores may create (for instance, having theory 400.yaml and 40_0.yaml and 4_00.yaml)

nnpdf_data/nnpdf_data/theorydbutils.py
```diff
@@ -51,13 +49,35 @@ def fetch_theory(theory_database: Path, theoryID: int):
...
 
+@lru_cache
+def get_available_theory_cards(path):
+    """Since theoryfile names may contain underscores, the theoryfile name cannot uniquely be
+    determined from the theoryID. Therefore we create a mapping between the theoryid as python
+    int and the corresponding file.
+    """
+    available_theories = {}
+    for theoryfile in path.glob("*.yaml"):
+        tmp_id = int(theoryfile.name.removesuffix('.yaml'))
```
scarlehoff, line 73:
```suggestion
        tmp_id = int(theoryfile.stem)
```

would this work? (nitpicking here)

scarlehoff (approved):
LGTM, just one comment, since  you added a `get_available_theory_cards` function could you use that function also in `fetch_all` (at the end of the `theorydbutils.py` file).
It doesn't change anything practically, but it makes sure that there's only one place where available theories are listed.

PR #2097, opened 2024-06-03
Small fixes for TF

scarlehoff:
These are some small fixes to TF. It fixes a few warnings but afaics they were not bugs.

The only one that made a bit of a difference (in that otherwise some tensorflow-cuda combinations fail) is adding the output shape to the boolean mask with a `tf.function` decorator.
Without that, (at least the Arch builds of) tensorflow 2.16.1 with Cuda 12.4 fail since it compiles away the boolean mask but then it calls the cuda kernel with the wrong shape.

### DIFF
--- a/n3fit/src/n3fit/backends/keras_backend/MetaLayer.py
+++ b/n3fit/src/n3fit/backends/keras_backend/MetaLayer.py
@@ -8,8 +8,8 @@
     For instance: np_to_tensor is just a call to K.constant
 """
 
-from tensorflow.keras.layers import Layer
-from tensorflow.keras.initializers import Constant, RandomUniform, glorot_normal, glorot_uniform
+from keras.initializers import Constant, RandomUniform, glorot_normal, glorot_uniform
+from keras.layers import Layer
 
 # Define in this dictionary new initializers as well as the arguments they accept (with default values if needed be)
 initializers = {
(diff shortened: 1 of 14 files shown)


scarlehoff:
Everything seems ok. The bot used TF 2.15, I'm using 2.16 (which is the version that raised the warnings that are addressed in this PR).

n3fit/src/n3fit/backends/keras_backend/MetaLayer.py
```diff
@@ -8,8 +8,8 @@
     For instance: np_to_tensor is just a call to K.constant
 """
 
-from tensorflow.keras.layers import Layer
-from tensorflow.keras.initializers import Constant, RandomUniform, glorot_normal, glorot_uniform
+from keras.initializers import Constant, RandomUniform, glorot_normal, glorot_uniform
+from keras.layers import Layer
```
RoyStegeman, line 12: I'm not such a fan of this. Yes, keras 3 now supports multiple backends but we're still only using tensorflow (and in various places inside `keras_backend` we explicitly use tensorflow, not just keras) so to avoid possible sources of conflict I'd rather just keep using the integrated `tf.keras`

RoyStegeman (approved):

PR #2099, opened 2024-06-04
Execute new commondata generation in CI

Radonirinaunimi:
Addresses https://github.com/NNPDF/nnpdf/issues/1457 but also:

- [x] makes sure that all the `filter.py` works properly
- [x] makes sure that a default formatter is used when dumping floating numbers

Remaining to do:
- [x] fail if the exit status of even one single filter is non-zero (which currently happens a lot)
- [x] fail if there are modified and/or unstaged files in `nnpdf_data/nnpdf_data/new_commondata`
- [x] list all the changed commondata files in comments

PR #2100, opened 2024-06-05
Fix new commondata filters

Radonirinaunimi:
This does not attempt to polish the new commondata implementation, but simply to make sure that the filters run properly. The main purpose of this PR is to make https://github.com/NNPDF/nnpdf/pull/2099 works.

This has a minimal conflict with https://github.com/NNPDF/nnpdf/pull/2056, but that's something that can be solved easily.


Radonirinaunimi:
@comane what should be done about the following hard-coded paths?
https://github.com/NNPDF/nnpdf/blob/83f9af90c91cc4463e94ee2347e310a0e1268d3c/nnpdf_data/nnpdf_data/new_commondata/ATLAS_2JET_7TEV_R06/filter.py#L192-L193

Radonirinaunimi:
> Hi @Radonirinaunimi, I think that this is already taken care of in #2056

Ah, okok, thanks. I think it would then be better if that one is merged first. Do you want me to address the remaining comments there?

Radonirinaunimi:
@comane This should be ready to merged, now all the filters run, everything else could addressed in #2099.

Issue #1978, opened 2024-03-04
Get rid of `recommonmark` dependency 


RoyStegeman:
As discussed in https://github.com/NNPDF/nnpdf/issues/1970 recommonmark is no longer being maintained, and we should no longer depend on it. @achiefa has agreed to solve this by updating all `.md` files in our docs to `.rst` format. 

See https://github.com/NNPDF/nnpdf/pull/1597 for an earlier attempt

Issue #1360, opened 2021-08-20
Sphinx warnings


Zaharid:
Currently if I build the docs I get

```
build succeeded, 2756 warnings.
```

which makes it very difficult to act on any one of them, including legitimate ones such as bad links. It would be good to figure out how to suppress the irrelevant ones and  fix some of the relevant ones.

PR #2101, opened 2024-06-05
Get rid of `recommonmark` dependency 

achiefa:
This PR addresses #1978.


RoyStegeman:
Thanks a lot for starting this! Don't forget to also delete the markdown file you're replacing.

achiefa:
> Don't forget to also delete the markdown file you're replacing.

Yes, I was't sure whether I should've deleted them or not. I was waiting for your feedback, many thanks!

achiefa:
I'm currently working on this PR. I just want to mention that nested inline markups are currently not possible in reStructuredText, as explained [here](https://docutils.sourceforge.io/FAQ.html#is-nested-inline-markup-possible). There are some workarounds, but they are really ugly and are not worth the effort. Also, the documentation discourages to adopt them. This just to say that, in covering *md* to *rst*, I'm not using nested inline markups when dealing with cross-references.

RoyStegeman:
thanks, it's a bit ugly anyway so that's fine.

achiefa:
Ok, I think I'm done. Please, @RoyStegeman and @scarlehoff could you please check that I didn't forget anything?

RoyStegeman:
Thanks! I'll have a look next week.

RoyStegeman:
Could you also remove it from the dependencies?

achiefa:
I only see dependencies on `recommonmark`  in `pyproject.toml` and in `meta.yaml` in the conda-recipe. Should I just remove  any `recommonmark` key from these files?

RoyStegeman:
Yes exactly, I suspect you will also have to remove the recommonmark settings from some config file if you haven't already.

doc/sphinx/source/conf.py
achiefa, line 1: Should I also remove the `setup` function at the very end of this?

RoyStegeman: I think you can, at least the recommmonmark parts but I think the entire function. If you delete it and can still build the docs than its fine

achiefa:
Ok, the doc compiles even after that. Here you can find the log of the build (both stdout and stderr).
[build.log](https://github.com/user-attachments/files/17365054/build.log)

RoyStegeman (approved):
Thanks, looks good! 

At some point we should also take care of the (important) warning such as non-existent labels https://github.com/NNPDF/nnpdf/issues/1360

PR #2102, opened 2024-06-05
Shifts due to correlated uncertainties

enocera:
This PR implements a function to compute the shifts due to correlated ucnertainties.


RoyStegeman:
@enocera if you would like to rebase you can press the down arrow next to "update branch" (assuming there are no conflicts). The difference is that with rebase you put the new commits after the commits in master (it does this by rewriting the commit which involves changing the timestamp), while a simple merge combines the new commits and master like a zipper in such a way that commits from this PR may be place before the latest commits in master.

PR #2103, opened 2024-06-06
fix bug in _hqp_yq_xq2map process

comane:
Fixes a bug for the HQP_YQ process


scarlehoff (approved):
Thanks for noticing and fixing it!
I've also added an extra check to avoid this problem in the future.

Issue #2104, opened 2024-06-06
Kinematic coverage for polarized datasets


scarlehoff:
While in principle a process type was added, the implementation of the (unpolarized) DIS datasets are slightly different from the polarized ones. Mainly because of the missing kinematic variables (DIS still use k1/k2/k3).

https://github.com/NNPDF/nnpdf/blob/6691f56f975f27732a9102b8ef58b01b04f2b810/validphys2/src/validphys/process_options.py#L124

It might be enough to modify this function to

```python
    x = kin_info.get_one_of("k1", _Vars.x)
    q = kin_info.get_one_of("k2", _Vars.Q)
    return x, q * q
```

but it requires some testing to make sure that the results are as expected.

Radonirinaunimi:
I have already looked into this and fixed it before to generate kinematic plots for a talk ([slack-slides](https://lhcfitnikhef.slack.com/archives/C033U9SS612/p1711092191883969)). But I couldn't find anymore nor remember what has done back then.

But your suggestion (with a very tiny tweak) works in https://github.com/NNPDF/nnpdf/pull/2109 and gives the same result as back then ([report](https://vp.nnpdf.science/vbI9vzsRRnCGhnUyhT440g==/)).

PR #2109, opened 2024-06-07
Fix kinematic coverage for polarized DIS processes

Radonirinaunimi:
Addresses https://github.com/NNPDF/nnpdf/issues/2104.


validphys2/src/validphys/process_options.py
```diff
@@ -125,8 +125,8 @@ def _dis_xq2map(kin_info):
     """In the old style commondata, the variables in the dataframe were ``x, Q2, y``
     but due to the transformations that happen inside validphys they become ``x, Q, y``
     """
-    x = kin_info["k1"]
-    q = kin_info["k2"]
+    x = kin_info.get_one_of("k1", _Vars.x)
+    q = kin_info.get_one_of("k2", _Vars.Q2)  # this is really Q not Q2
```
scarlehoff, line 129: If it is Q why do you use Q2?

Radonirinaunimi: The name of the passed variable is Q2 but if one looks at the values they are really Q not Q2. Maybe something is still missing somewhere.

scarlehoff: You should be using, in the polarized datasets, `identity` so no kinematic override or transformation. When you do that, what goes into validphys is whatever you included in the data file, raw.

It is possible that other DIS datasets (the ones that have k1, k2, k3) use "x" "Q2", "y" but, after going through validphys, the value of k2 gets a `sqrt`.

But for new data we should just use the variables we want. If you need `sqrt(q2)` you can do that explicitly.

Radonirinaunimi: Could you check that [296b23b](https://github.com/NNPDF/nnpdf/pull/2109/commits/296b23b0523107cdc59776c7e5f5903d552c7f99) is now fine. If so, I will propagate the `kinematics_override` into all the polDIS commondata.

validphys2/src/validphys/commondataparser.py
```diff
@@ -761,13 +761,15 @@ def folder(self):
     def cm_energy(self):
         """Return the center of mass energy as GeV if it can be understood from the name
         otherwise return None"""
-        energy_string = self.setname.split("_")[2]
-        if energy_string == "NOTFIXED":
+
+        if "NOTFIXED" in self.setname:
```
scarlehoff, line 765: I think this is a bad idea. We need to change the `E<int>_NC` datasets to be `E<int>-NC`, otherwise we are already breaking the convention.

https://docs.nnpdf.science/data/dataset-naming-convention.html

Luckily we can use the same strategy that is used for the legacy versions to ensure old runcards still work.

validphys2/src/validphys/plotoptions/kintransforms.py
```diff
@@ -267,6 +267,15 @@ def new_labels(self, *old_labels):
...
 
+class dis_q2_scale(DISXQ2MapMixin):
+    def __call__(self, k1, k2, k3):
+        ecm = np.sqrt(k2 / (k1 * k3))
+        return k1, k2, np.ceil(ecm)
+
+    def new_labels(self, *old_labels):
+        return ('$x$', '$Q^2$ (GeV)', r'$\sqrt{s}$ (GeV)')
+
+
```
scarlehoff, line 278:
```suggestion
```

We want to remove this module!
You should not need a kinematic override. Instead of doing this the `kinematics.yaml` file should be changed.

validphys2/src/validphys/process_options.py
```diff
@@ -68,6 +68,12 @@ def get_one_of(self, *variables):
                 return self._kins[var]
         raise KeyError(f"Need one of the following variables {self._variables} to continue")
 
+    def check_variable(self, variable):
+        """Check if a given variable as defined in `_Vars` is present in the
+        `self._kins`.
+        """
+        return variable in self._kins
+
```
scarlehoff, line 76:
```suggestion
```

validphys2/src/validphys/process_options.py
```diff
@@ -127,32 +133,35 @@ def __str__(self):
...
-    x = kin_info["k1"]
-    q = kin_info["k2"]
-    return x, q * q
+    x = kin_info.get_one_of("k1", _Vars.x)
+    if kin_info.check_variable(_Vars.Q2):
+        q2 = kin_info.get_one_of(_Vars.Q2)
+        return x, q2
+    else:
+        q = kin_info.get_one_of("k2")
+        return x, q * q
```
scarlehoff, line 142:
```suggestion
    """Variables in the dataframe should be x, Q2, y
    TODO: Once old variables are removed, remove if condition
    """
    x = kin_info.get_one_of("k1", _Vars.x)
    if "k2" in kin_info._kins:
        q2 = kin_info.get_one_of("k2")**2
    else:
        q2 = kin_info.get_one_of(_Vars.Q2)
    return x, q2
```

validphys2/src/validphys/commondataparser.py
giacomomagni: solved in https://github.com/NNPDF/nnpdf/pull/2142 and 697470ceb

validphys2/src/validphys/plotoptions/kintransforms.py
giacomomagni: done in 5e5208feb

validphys2/src/validphys/process_options.py
giacomomagni: Resolved in [e5208f](https://github.com/NNPDF/nnpdf/pull/2109/commits/5e5208feb4bcfec226116f0df5447e73cc9ad528), [9e34271](https://github.com/NNPDF/nnpdf/pull/2109/commits/9e34271580fbddf3115a8e77c1016bd73afc4cd4), [0d680d0](https://github.com/NNPDF/nnpdf/pull/2109/commits/0d680d05e7971b11712bb97df2a21c279bda9217)

scarlehoff (commented):
Thanks. Did you already created by any chance kinematic coverage plots with the latest version of this branch?

scarlehoff:
Actually that's a good point... I've created the plot with one of @Radonirinaunimi fits, please check that it is (relatively) sensible. This way this is also testing that the translation in `dataset_names.yaml` is working as expected (I'm using the datasets directly from the fit).

https://vp.nnpdf.science/7F2RTYKkQZW5TUT4HqECWw==

scarlehoff:
If the plot in the previous msg looks ok, I think this can be merged.

Radonirinaunimi:
Sorry for having forgotten about this. This is  ok, and the kinematic plots are good.

We can merge this!

giacomomagni (approved):

PR #2110, opened 2024-06-07
Extend n3fit and vp to deal with two different initial state distributions

Radonirinaunimi:
The following PR modifies both n3fit and vp to deal with the case in which (SI)DIS/DY processes involve convolutions between two different distributions (PDF, pPDF, FF, etc.). A simple example of such a scenario is the case in which one of the protons is polarized and the other not:

$$\Delta \mathcal{R} = \frac{\sum \Delta f \otimes f \otimes \Delta \Sigma}{\sum f \otimes f \otimes  \Sigma}$$

This PR introduces a feature available since pineappl `<v0.8.0` that specifies the type of convolution required by an initial state with the `convolution_type_` key (whose value can be `UnpolPDF`, `PolPDF`, `FF`).

The main modifications affect:
- [X] the new convolution that includes the $\Delta V$ and $\Delta V_3$ distributions
- [X] the convolution during the fits in which one of the distributions is computed from a boundary condition
- [x] the computation of the theory predictions in vp 
    - [x] $\chi^2$-computations
    - [x] data vs theory comparisons
- [x] allow for computations of theory uncertainties (computing of the theory covariance matrix)

Minimal non-congruent modifications:
- [x] deprecate the `is_polarized` logical
- [x] modify the polarized Positivity commondata to impose the BCs on the individual pPDF flavors

To do at the very last stage:
- [x] re-generate the regression tests
- [x] update the documentations

Rely on https://github.com/NNPDF/pineappl/pull/289, https://github.com/NNPDF/pineko/pull/181.


validphys2/src/validphys/convolution.py
```diff
@@ -400,22 +423,24 @@ def appl(df):
...
 
 
-def hadron_predictions(loaded_fk, pdf):
+def hadron_predictions(loaded_fk, pdf_list):
     """Implementation of :py:func:`fk_predictions` for hadronic observables."""
-    gv = functools.partial(evolution.grid_values, pdf=pdf)
-    res = _gv_hadron_predictions(loaded_fk, gv)
-    res.columns = range(pdf.get_members())
+    gv1 = functools.partial(evolution.grid_values, pdf=pdf_list[0])
+    gv2 = functools.partial(evolution.grid_values, pdf=pdf_list[1])
```
scarlehoff, line 429: This generates an ambiguity, when `pdf_list[0] == pdf_list[1]` it doesn't matter, but if they are different, whether one of the pdfs is just the central values can change things a lot.

I would add a check for that in `determine_pdfs` ( I put it in the template)

validphys2/src/validphys/fkparser.py
```diff
@@ -312,13 +313,18 @@ def parse_fktable(f):
             hadronic = res['GridInfo'].hadronic
             ndata = res['GridInfo'].ndata
             xgrid = res.pop('xGrid')
+            # There are no non-unpolarized FK tables with the old theory,
+            # so everything is `UnpolPDF`.
+            nb_convolutions = 2 if hadronic else 1
+            conv_types = tuple("UnpolPDF" for _ in range(nb_convolutions))
```
scarlehoff, line 319: A nitpick, but I dislike intermediate variables that just hide what's really happening.

```suggestion
            if hadronic:
                conv_types = ("UnpolPDF", "UnpolPDF")
            else:
                 conv_types = ("UnpolPDF",)
```

validphys2/src/validphys/pineparser.py
```diff
@@ -155,13 +155,26 @@ def pineappl_reader(fkspec):
...
 
-    # Check if it is Polarized FK table
-    is_polarized = pine_rep.key_values().get("polarized") == "True"
-
     # Is it hadronic? (at the moment only hadronic and DIS are considered)
-    hadronic = pine_rep.key_values()["initial_state_1"] == pine_rep.key_values()["initial_state_2"]
+    try:
+        parton1 = pine_rep.key_values()["convolution_particle_1"]
+        parton2 = pine_rep.key_values()["convolution_particle_2"]
+    except KeyError:
```
scarlehoff, line 162:
```suggestion
    except KeyError:
        # Old pineappl FKTables used  `initial_state` instead of `convolution_particle`
```

validphys2/src/validphys/results.py
```diff
@@ -656,14 +677,21 @@ def one_or_more_results(
     sqrt_covmat,
     pdfs: (type(None), Sequence) = None,
     pdf: (type(None), PDF) = None,
+    unpolarized_bcs: (type(None), Sequence) = None,
```
scarlehoff, line 680: This is the first time that more than one unpolarized boundary condition appear, and would be inconsistent with other functions. It is preferable to remove it from existing imho

```suggestion
```

validphys2/src/validphys/results.py
```diff
@@ -656,14 +677,21 @@ def one_or_more_results(
...
 ):
     """Generate a list of results, where the first element is the data values,
     and the next is either the prediction for pdf or for each of the pdfs.
     Which of the two is selected intelligently depending on the namespace,
     when executing as an action."""
+    if unpolarized_bc or unpolarized_bcs:
+        # NOTE: Boundary conditions should always be unique for a Prediction
+        boundary_pdf = unpolarized_bcs[0] if unpolarized_bcs else unpolarized_bc
+    else:
+        boundary_pdf = None
```
scarlehoff, line 691:
```suggestion
```

validphys2/src/validphys/results.py
```diff
@@ -656,14 +677,21 @@ def one_or_more_results(
...
     if pdf is not None:
-        return results(dataset, pdf, covariance_matrix, sqrt_covmat)
-    return pdf_results(dataset, pdfs, covariance_matrix, sqrt_covmat)
+        return results(dataset, pdf, covariance_matrix, sqrt_covmat, boundary_pdf)
+    return pdf_results(dataset, pdfs, covariance_matrix, sqrt_covmat, boundary_pdf)
```
scarlehoff, line 694:
```suggestion
        return results(dataset, pdf, covariance_matrix, sqrt_covmat, unpolarized_bc)
    return pdf_results(dataset, pdfs, covariance_matrix, sqrt_covmat, unpolarized_bc)
```

scarlehoff (commented):
Thanks for this. I've made a few comments, my thinking for all comments is about making it more general where I felt it could be a bit too restrictive  in the future.

After you apply this comment I can take over for the stuff very deep in validphys that needs to be changed.

n3fit/src/n3fit/layers/observable.py
```diff
@@ -164,7 +169,7 @@ def call(self, pdf):
 
         observables = []
         for idx, (pdf, padded_fk) in enumerate(zip(pdfs, self.padded_fk_tables)):
-            pdf_to_convolute = pdf if self.boundary_pdf[idx] is None else self.boundary_pdf[idx]
+            pdf_to_convolute = [pdf if p is None else p for p in self.boundary_pdf[idx]]
```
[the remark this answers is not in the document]
Radonirinaunimi: Yes, there are (will be) situations in which we have multiple boundary conditions. So this way we are as general as possible, this can even already accommodate for Fragmentation Function fits (which we'll be able to do very soon).

n3fit/src/n3fit/tests/regressions/quickcard_pol.yml
```diff
@@ -23,7 +23,7 @@ positivity_bound:
 
 ############################################################
 theory:
-  theoryid: 822
+  theoryid: 20_000_001
```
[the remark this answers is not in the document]
Radonirinaunimi: Will update as soon as #2129 is merged.

n3fit/src/n3fit/layers/observable.py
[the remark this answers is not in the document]
scarlehoff: Then the `determine_pdf` function should be in use also here. That way adding FF means just adding them in that function (instead of adding them differently in different parts of the code).

It also makes it easier to eventually lift the function to be "something else" (maybe another PDF, or vp object coming from n3fit, etc)

validphys2/src/validphys/convolution.py
```diff
@@ -248,7 +250,23 @@ def linear_predictions(dataset, pdf):
...
+    """
+    if unpolarized_bc is not None:
+        map_pdf_to_conv_types = {
+            "PolPDF": pdf,
+            "UnpolPDF": unpolarized_bc,
+        }  # NOTE: add additional PDF type here
+        pdf_list = [map_pdf_to_conv_types.get(i, pdf) for i in conv_types]
+    else:
+        pdf_list = [pdf for _ in conv_types]
+    return pdf_list
```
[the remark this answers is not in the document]
Radonirinaunimi: Regarding this, I have two comments:
- actually, we always have to load the entire `unpolarized_bc` replicas set because we need to shift the cv by one standard deviation when imposing the positivity BC  (forgot this myself). And therefore we cannot just load the set as a `t0` PDF.
https://github.com/NNPDF/nnpdf/blob/840ac3e2775929353c7e44be20fd494ff7fb1c10/n3fit/src/n3fit/layers/observable.py#L47-L51

- Isn't the old way actually more modular? In that it is not bounded by `pdf_1` and `pdf_2`. But I agree that this function could be greatly improved. For instance, I don't really like coding manually the mapping of the pdfs from the convolution types.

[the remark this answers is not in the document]
scarlehoff: > actually, we always have to load the entire unpolarized_bc replicas set because we need to shift the cv by one standard deviation when imposing the positivity BC (forgot this myself). And therefore we cannot just load the set as a t0 PDF.

That's fine. The `determine_pdf` function can also manipulate the PDF if needed. I would concentrate these manipulations in this function.

The "check that one is central" can also be done at the end "check that one of the two is manipulated". The point is that (the way it is done right now) all these functions work _only_ if the number of members in the unpolarized PDF and the polarized one is the same.

> Isn't the old way actually more modular? In that it is not bounded by pdf_1 and pdf_2. But I agree that this function could be greatly improved. For instance, I don't really like coding manually the mapping of the pdfs from the convolution types.

But it is basically the same. I'm just passing the fktable instead of one of the variables of the fktable, with the idea that eventually there will be also fragmentation functions and that information will be included in the fktable.

validphys2/src/validphys/results.py
scarlehoff: But you drop inmediately the list and say `Boundary conditions should always be unique for a Prediction`. Therefore, don't allow more than one and you don't have to drop it.

Radonirinaunimi: Maybe I am missing something, but the thing is each polarized fit card will always contain `unpolarized_bc` and during fits comparison these are passed as `unpolarized_bcs`:
https://github.com/NNPDF/nnpdf/blob/840ac3e2775929353c7e44be20fd494ff7fb1c10/validphys2/src/validphys/comparefittemplates/comparecard_polarized.yaml#L37-L39

which technically could be different but we require it to be always the same.

scarlehoff: Since you require them to be always the same, there should only be one. In practice here you are always taking the one from current and dropping reference, at that point is better just to do

```yaml
unpolarized_bcs:
    _from: current
```

n3fit/src/n3fit/layers/observable.py
```diff
@@ -107,16 +109,19 @@ def __init__(
...
                     q0_value=fkdata.Q0,
                     xgrid=fkdata.xgrid,
-                    n_std=boundary_condition.get("n_std", 0.0),
+                    n_std=boundary_condition.get("n_std", n_std),
                     n_replicas=n_replicas,
                 )
+                self.boundary_pdf[idx] = [
+                    set_boundary if conv_type == 'UnpolPDF' else None
+                    for conv_type in fkdata.convolution_types
+                ]
```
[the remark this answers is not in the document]
Radonirinaunimi: Addressed in [0adedad](https://github.com/NNPDF/nnpdf/pull/2110/commits/0adedada52fc69b2a638153c5436d6543822d04f). Now it is simpler.

n3fit/src/n3fit/tests/regressions/quickcard_pol.yml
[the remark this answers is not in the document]
Radonirinaunimi: Done in [4bffa43](https://github.com/NNPDF/nnpdf/pull/2110/commits/4bffa434b355afa929b3401b31b7587a9068d1c8) (and for other runcards as well.

scarlehoff (approved):
Waiting for the reports / fits to be ready to make sure, but lgtm

validphys2/src/validphys/coredata.py
```diff
@@ -189,6 +189,40 @@ def get_np_fktable(self):
...
+                        )
+
+                    conv_pdfs.append(pdf.unpolarized_bc.make_only_cv())
+                else:
+                    conv_pdfs.append(pdf)
+
+            elif convolution_type == "PolPDF":
+                if not pdf.is_polarized:
+                    raise ValueError(
+                        "The FKTable asked for a polarized PDF, but the PDF received is unpolarized"
```
scarlehoff, line 220:
```suggestion
                        """The FKTable asked for a polarized PDF, but the PDF received cannot be understood as polarized.
                        When using a polarized PDF make sure to include a boundary condition `unpolarized_bc: <pdf name>` whenever needed (`t0`, `dataspecs`...)."
```

Radonirinaunimi: Updated the docs the reflect the changes here. Please have a look if you are fine with the additional modifications.

Radonirinaunimi:
Everything now works as expected here (and consistent with previous results). For the record, here are some extensive fit comparisons: https://vp.nnpdf.science/KSerNDeFSiGIwdgA81RaGw==/

doc/sphinx/source/tutorials/polarized_fits.rst
```diff
@@ -33,6 +33,45 @@ as a boundary condition while ``n_std`` specifies the shift in terms of the stan
...
 PDF central values will be used to constrain their polarized counterparts.
 
+.. note::
+
+   This will also be the PDF set to be used to compute observable predictions
+   during the fit.
+
+Instead, the unpolarized PDF set to be used to compute observable predictions during
+data vs. theory comparisons or :math:`\chi^2` computations need to be specified under
+the ``datacuts`` entry:
```
scarlehoff, line 43:
```suggestion
Instead, the unpolarized PDF set to be used to compute the `t0` covariance matrix need to be specified under the ``datacuts`` entry:
```

Note that you can also do `unpolarized_bc` in the outer namespace and let everything use it. The important point is that every polarized pdf (doesn't matter what is it used for) needs to have an `unpolarized_bc` associated in the validphys runcard.

doc/sphinx/source/tutorials/polarized_fits.rst
```diff
@@ -33,6 +33,45 @@ as a boundary condition while ``n_std`` specifies the shift in terms of the stan
...
+
+.. code-block:: yaml
+
+  # Define the unpolarized PDF set to be used to compute predictions with validphys actions
```
scarlehoff, line 47:
```suggestion
```
This is just for t0

doc/sphinx/source/tutorials/polarized_fits.rst
```diff
@@ -33,6 +33,45 @@ as a boundary condition while ``n_std`` specifies the shift in terms of the stan
...
+    t0pdfset: NNPDFpol11_100   # PDF set to generate t0 covmat
+    unpolarized_bc: NNPDF40_nnlo_pch_as_01180
+    q2min: 1.00                # Q2 minimum
+    w2min: 4.00                # W2 minimum
+  ...
+
+And if a theory covariance matrix is included in the fit
+(see :ref:`how to include a theory covariance matrix in a fit <_thcov_tutorial>`
+for more details), then the unpolarized PDF set also needs to be specified in the
+``theorycovmatconfig``:
```
scarlehoff, line 59:
```suggestion
``theorycovmatconfig`` together with the polarized PDF to be used:
```

doc/sphinx/source/tutorials/polarized_fits.rst
Radonirinaunimi: Ah yes, that's right!

n3fit/src/n3fit/layers/DY.py
```diff
@@ -84,8 +84,9 @@ def compute_dy_observable_many_replica(pdf, padded_fk):
         tensor
             observable of shape (batch=1, replicas, ndata)
     """
-    temp = op.einsum('nxfyg, bryg -> brnxf', padded_fk, pdf)
-    return op.einsum('brnxf, brxf -> brn', temp, pdf)
+    # TODO: check which PDF must go first in case of different initial states!!!
+    temp = op.einsum('nxfyg, bryg -> brnxf', padded_fk, pdf[0])
+    return op.einsum('brnxf, brxf -> brn', temp, pdf[1])
```
scarlehoff, line 89:
```suggestion
    pdfa = pdf[1]
    pdfb = pdf[0]
    
    temp = op.einsum('nxfyg, bryg -> brnxf', padded_fk, pdfa)
    return op.einsum('brnxf, brxf -> brn', temp, pdfb)
```

I still need to check but I think this is the right ordering.

n3fit/src/n3fit/layers/DY.py
```diff
@@ -94,10 +95,12 @@ def compute_dy_observable_one_replica(pdf, mask_and_fk):
     masking the PDF rather than the fk table.
     """
     mask, fk = mask_and_fk
-    pdf = pdf[0][0]  # yg
+    pdfb = pdf[0][0][0]  # yg
+    pdfa = pdf[1][0][0]  # yg
```
scarlehoff, line 99:
```suggestion
    # Retrieve the two PDFs (which may potentially be coming from different initial states)
    # Since this is the one-replica function, remove the batch and replica dimension
    pdfb = pdf[0][0][0]  # xf
    pdfa = pdf[1][0][0]  # yg
```

scarlehoff (approved):
For completeness, I'm going to try and run a parallel fit with this branch.

Looking through the changes in n3fit I noticed a potential bug running polarized fits in parallel. I'll report back once the fits are finished.
For the rest (docs, -sequential- fits, reports) everything looks ok.

n3fit/src/n3fit/layers/DY.py
Radonirinaunimi: Good catch!!! Will push this.

scarlehoff:
Here's a report comparing one of the fits in @Radonirinaunimi's report, with the fit run in a GPU.

https://vp.nnpdf.science/KV97c7yfTRCt5PEOZ6Dh1g==


Everything seems to work well :)

(given past experiences I want to make sure we have a lot of granularity around these commits that can potentially change things in the fit behavior!)

Radonirinaunimi:
> Here's a report comparing one of the fits in @Radonirinaunimi's report, with the fit run in a GPU.
> 
> https://vp.nnpdf.science/KV97c7yfTRCt5PEOZ6Dh1g==
> 
> Everything seems to work well :)
> 
> (given past experiences I want to make sure we have a lot of granularity around these commits that can potentially change things in the fit behavior!)

Perfect! Thanks a lot for this check @scarlehoff!

PR #2112, opened 2024-06-12
Fix metadata of ATLAS Z pt 13 TeV 

comane:
use the correct table for uncertainties.

@Radonirinaunimi I think we have two options here, either changing the number of the table in the metadata or hard code  line 83 in  `get_systematics` function to 

`hepdata_table = f"rawdata/HEPData-ins1768911-v{version}-Table_5.yaml"`


Radonirinaunimi (approved):
Thanks a lot @comane for fixing this! Once the still running tests are done, this can be merged.

PR #2114, opened 2024-06-20
make sure tempfile is alive when it's used

RoyStegeman:
The first commit is the relevant one. 

We create a tempfile to download files such as ekos to, and then move that file to eko.tar (or whatever the target path is). However, the moving is done outside the context manager, so if `delete_on_failure` is true, the tempfile that should be moved to eko.tar has already been deleted before it's moved.


andreab1997 (approved):

andreab1997:
Looks good to me, thanks!

PR #2115, opened 2024-06-20
take n3lo_ad_variations from the corresponding key in theorycard instead of 'Comments'

RoyStegeman:
As the title suggests. After updating the theorycards in #2113 the `Comments` key was changed so the trick of taking the variations from the comment was broken.


giacomomagni (approved):
LGTM, thanks

PR #2116, opened 2024-06-24
replace POS_F2 process with more inclusive POS_DIS

RoyStegeman:
Currently only `POS_F2` exists in the `PROCESSES` dict in process_options.py, so `POS_FLL` would need to be added as well, that this is missing is simply a bug. 

Then, it doesn't make much sense to make `POS_FLL` exclusive in  "heaviness/flavour", while using the more inclusive definition for `POS_F2`. However, they are both just DIS structure functions so there is no need to separate them by type. Because of this I introduced a single `POS_DIS` (which only has to supports structure functions and not xsecs).

### DIFF
--- a/nnpdf_data/nnpdf_data/new_commondata/NNPDF_POS_2P24GEV/metadata.yaml
+++ b/nnpdf_data/nnpdf_data/new_commondata/NNPDF_POS_2P24GEV/metadata.yaml
@@ -207,22 +207,17 @@ implemented_observables:
       $F_L^l$'
     plot_x: k1
   kinematic_coverage:
-  - k1
-  - k2
-  - k3
+  - x
+  - Q2
   kinematics:
     variables:
-      k1:
-        description: Variable k1
-        label: k1
+      x:
+        description: Bjorken x
+        label: x
         units: ''
-      k2:
-        description: Variable k2
-        label: k2
-        units: ''
-      k3:
-        description: Variable k3
-        label: k3
+      Q2:
+        description: Factorization scale
+        label: Q2
         units: ''
     file: kinematics_FLL.yaml
   theory:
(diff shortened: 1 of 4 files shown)


RoyStegeman:
@comane could you please have a look? The change is very minimal, but perhaps you disagree with it

nnpdf_data/nnpdf_data/new_commondata/NNPDF_POS_2P24GEV/metadata.yaml
```diff
@@ -559,7 +559,7 @@ implemented_observables:
     description: Deep Inelastic Scattering
     label: 'positivity dataset: DIS light quark longitudinal structure function $F_L^l$'
     units: ''
-  process_type: POS_FLL
+  process_type: POS_DIS
   tables: []
```
comane, line 563: Hi @RoyStegeman, I think it would be good to change the `kinematic_coverage` as well as `kinematics` for `POS_FLL` as it was done for the other `POS_F2` sets so as to avoid any bug in processes

comane (approved):
hi @RoyStegeman, thanks for taking care of this!
PR looks good to me now

nnpdf_data/nnpdf_data/new_commondata/NNPDF_POS_2P24GEV/metadata.yaml
```diff
@@ -207,22 +207,17 @@ implemented_observables:
...
+        description: Bjorken x
+        label: x
         units: ''
-      k3:
-        description: Variable k3
-        label: k3
+      Q2:
+        description: Factorization scale
+        label: Q2
         units: ''
```
felixhekhorn, line 221: I guess this should be `GeV^2`

PR #2120, opened 2024-06-27
add pos cuts through added_filter_rules instead of independent dataset

RoyStegeman:
Now that we can apply cuts to pos sets, there is no longer a use for the separate datasets. 

This has no significant impact on the fit: https://vp.nnpdf.science/XQE-J30TTg-z3D56Gnxg0w==

### DIFF
--- a/n3fit/runcards/examples/nnpdf40-like.yml
+++ b/n3fit/runcards/examples/nnpdf40-like.yml
@@ -169,9 +169,22 @@ positivity:
 added_filter_rules:
   - dataset: NNPDF_POS_2P24GEV_FLL
     rule: "x > 5.0e-7"
-
   - dataset: NNPDF_POS_2P24GEV_F2C
     rule: "x < 0.74"
+  - dataset: NNPDF_POS_2P24GEV_XGL
+    rule: "x > 0.1"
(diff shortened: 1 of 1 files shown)


comane (approved):
Thanks @RoyStegeman, looks good to me as soon as tests pass

n3fit/runcards/examples/nnpdf40-like.yml
```diff
@@ -136,34 +136,56 @@ fitting:
...
+  - dataset: NNPDF_POS_2P24GEV_XUB
+    rule: "x > 0.1"
+  - dataset: NNPDF_POS_2P24GEV_XDQ
+    rule: "x > 0.1"
+  - dataset: NNPDF_POS_2P24GEV_XDB
+    rule: "x > 0.1"
+  - dataset: NNPDF_POS_2P24GEV_XSQ
+    rule: "x > 0.1"
+  - dataset: NNPDF_POS_2P24GEV_XSB
+    rule: "x > 0.1"
```
comane, line 187: Hi @RoyStegeman , why didn't you use the process here?

RoyStegeman: I didn't think about it and just copied it from your runcard

PR #2125, opened 2024-07-13
Allow for a 4th optional kinematic variable.

giacomomagni:
To properly descrive some `dijets` measurements I'd like to allow for a 4th kinematic variable.
This would only be needed to produce the kinematic coverage and can be drop everywhere else. 

The new commodata format allows to have more than 3 kinematic variables, however the parse still 
parser to load such data. 

This would be a very minimal fix that allows to store as many kinematic variables you want,
be able to load the data correctly (and eventually access the full kinematics), but keep using 
in `CommonData` only the first 3 variables.

Would be this trick be sufficient ? Or am I going to break the entire machine?

### DIFF
--- a/validphys2/src/validphys/commondataparser.py
+++ b/validphys2/src/validphys/commondataparser.py
@@ -445,8 +445,10 @@ class ObservableMetaData:
                 )
 
         if len(self.kinematic_coverage) > 3:
-            raise ValidationError(
-                "Only a maximum of 3 variables can be used for `kinematic_coverage`"
+            logging.warning(
+                """Only a maximum of 3 variables can be used in `CommonData`
+                discarding the others. Access the full kinematic with
+                `kinematic_coverage`"""
             )
 
     def apply_variant(self, variant_name):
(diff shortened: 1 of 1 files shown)


scarlehoff:
In principle the new commondata is prepared to use more than 3 variables, however, before adding hacks to do so within validphys we need to first finish removing the old commondata so that the stuff that fixes the number of kinematic variables to 3 can be removed.

I don't want to change things until the old commondata is removed, partially to pressure people into finishing the removal of the old commondata.

That said, for this problem in particular, which 4 variables do you need for the kinematic coverage plot?

giacomomagni:
@scarlehoff  for your comment. 
So some of the polarized dijets measurements have asymmetric ranges in the `eta_j1` and eta_j2`. 
So I need to store `sqrts, m_jj,eta_j1,eta_j2`. The actual observable depends only on `m_jj` but there are different `eta` bins.

Lireally what we need now it's just to be able to load the dataset, nothing more.

scarlehoff:
I see. `sqrts` doesn't need to be in the list of variables since it is part of the name (and the kincov plot can read it from there) so you can leave it out

giacomomagni:
Okay I'm adopting that suggestion. Thanks

PR #2126, opened 2024-07-15
New ht thcovmat

achiefa:
This PR allows for the inclusion of theory uncertainties due to the effect of power corrections. The theory covariance matrix is constructed by computing the shifts for the theoretical predictions as done for the MHOUs. The shift is computed at the level of the structure functions. Then, the shifts for the structure functions are combined to reconstruct the shift for the xsec. For this reason, the calculation of the shift depends on the dataset. Currently, only 1-JET and DIS (NC and CC) are supported.

At the level of the runcard, power corrections are specified as follows

```yaml
theorycovmatconfig:
  point_prescriptions: ["power corrections"]
  pc_parameters:
  - {ht: H2p, yshift: [0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.0], nodes: [0.0, 0.001, 0.01, 0.1, 0.3, 0.5, 0.7, 0.9, 1]}
  - {ht: H2d, yshift: [0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.0], nodes: [0.0, 0.001, 0.01, 0.1, 0.3, 0.5, 0.7, 0.9, 1]}
  - {ht: HLp, yshift: [0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.0], nodes: [0.0, 0.001, 0.01, 0.1, 0.3, 0.5, 0.7, 0.9, 1]}
  - {ht: HLd, yshift: [0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.0], nodes: [0.0, 0.001, 0.01, 0.1, 0.3, 0.5, 0.7, 0.9, 1]}
  - {ht: H3p, yshift: [0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.0], nodes: [0.0, 0.001, 0.01, 0.1, 0.3, 0.5, 0.7, 0.9, 1]}
  - {ht: H3d, yshift: [0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.0], nodes: [0.0, 0.001, 0.01, 0.1, 0.3, 0.5, 0.7, 0.9, 1]}
  - {ht: Hj, yshift: [2.0, 2.0, 2.0, 2.0, 2.0, 2.0], nodes: [0.25, 0.75, 1.25, 1.75, 2.25, 2.75]}
  pc_included_procs: ["JETS", "DIS NC", "DIS CC"]
  pc_excluded_exps: [HERA_NC_318GEV_EAVG_CHARM-SIGMARED,
                     HERA_NC_318GEV_EAVG_BOTTOM-SIGMARED]
  pdf: 240701-02-rs-nnpdf40-baseline
  use_thcovmat_in_fitting: true
  use_thcovmat_in_sampling: true
```
For each process/sf implemented, we need to specify a series of information:
1. The name of the power correction, under the key `ht`. This is useful to identify the type of power correction, in particular when computing the posterior.
2. The values in `yshift` are the magnitudes of the prior.
3. `nodes` contains the points where the prior is shifted
 
The array `pc_included_procs` specifies the processes for which the shifts are computed. I've also implemented the possibility to exclude some particular datasets within the selected processes, and this can be done by specifying the names in `pc_excluded_exps`.

~~The key `func_type` is temporary and will be deleted once we decide which function to use to construct the prior.~~

All the relevant details for this PR are contained in the module `higher_twist_functions.py`. For each observable for which the shift has to be computed, I implemented a factory that constructs a function which will then compute the shift. I thought it was kind of necessary to make the shift dependent on the shift parameters (`yshift` and `nodes`) and on the prescription according to which we vary the parameters (which for now is fixed), and not on the kinematics (I think this is known as _currying_ in computer science).

TO DO
=====
- [x] Add documentation
- ~~[x] Remove `func_type`~~
- [x] What happens if `power_corrections` is specified but the parameters are not given?
- [x] Change the name `ht` to `name` (?)
- [x] Something else?

### DIFF
--- a/doc/sphinx/source/vp/theorycov/power_corrections.rst
+++ b/doc/sphinx/source/vp/theorycov/power_corrections.rst
@@ -1,0 +1,238 @@
+.. _vptheorycov-pc:
+
+Power corrections
+=================
+
+Power corrections (also referred to as higher twist corrections for DIS-like
+processes) model contributions from non-perturbative effects that scale as
+inverse powers of the hard scale. They are implemented in the
+``theorycovariance`` module and can be included as a theory covariance matrix in
+a fit. Power corrections for jets and higher twists for DIS data have been
+determined in :cite:p:`Ball:2025xtj`, based on NNPDF4.0, where the reader can
+find further details on the methodology and phenomenological implications.
+
+The implementation is in
+`higher_twist_functions.py <https://github.com/NNPDF/nnpdf/tree/master/validphys2/src/validphys/theorycovariance/higher_twist_functions.py>`_
+and
+`construction.py <https://github.com/NNPDF/nnpdf/tree/master/validphys2/src/validphys/theorycovariance/construction.py>`_.
+
+
+Overview
+--------
+
+In NNPDF, power corrections modify theoretical predictions by introducing multiplicative
+shifts. For a generic observable :math:`O`, the corrected prediction is
+
+.. math:: O \to O \times (1 + \mathrm{PC}),
+
+where :math:`\mathrm{PC}` is the power correction. The shift to the prediction
+is therefore
+
+.. math:: \Delta O = O \times \mathrm{PC}.
+
+Different functional forms for the power correction are used depending on the
+process type:
+
+- **DIS** (neutral current and charged current): the correction depends on
+  Bjorken-:math:`x` and :math:`Q^2`, and scales as :math:`1/Q^2`.
+- **Single-inclusive jets**: the correction depends on rapidity and transverse
+  momentum :math:`p_T`, and scales as :math:`1/p_T`.
+- **Dijets**: the correction depends on a rapidity variable and the dijet
+  invariant mass :math:`m_{jj}`, and scales as :math:`1/m_{jj}`.
+
+
+Parametrisation
+---------------
+
+Power corrections are parametrised using a piecewise-linear interpolation
+between a set of nodes. The node positions (``nodes``) and the function values
+at each node (``yshift``) are specified in the runcard.
+
+The interpolation is constructed as a sum of triangular basis functions: each
+node :math:`i` is associated with a triangle that peaks at the node position
+with value ``yshift[i]`` and drops linearly to zero at the two neighbouring
+nodes. The resulting function is continuous and piecewise-linear.
+
+For DIS processes, the nodes are placed in Bjorken-:math:`x` and the power
+correction for a data point at :math:`(x, Q^2)` is
+
+.. math:: \mathrm{PC}(x, Q^2) = \frac{h(x)}{Q^2},
+
+where :math:`h(x)` is the piecewise-linear interpolation.
+
+For jet processes, the nodes are placed in rapidity and the correction at
+:math:`(y, p_T)` is
+
+.. math:: \mathrm{PC}(y, p_T) = \frac{h(y)}{p_T}.
+
+For dijets, the same functional form is used but the suppression scale is the
+dijet invariant mass :math:`m_{jj}`.
+
+
+Dataset routing
+---------------
+
+Each dataset is mapped to one or more power correction parameter keys via the
+function ``get_pc_type``. The mapping depends on the process type and dataset
+name:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 30 30 40
+
+   * - Process type
+     - PC type key
+     - Datasets
+   * - DIS NC (proton :math:`F_2`)
+     - ``f2p``
+     - SLAC, BCDMS proton :math:`F_2`; NMC, HERA :math:`\sigma_{\mathrm{red}}`
+   * - DIS NC (deuteron :math:`F_2`)
+     - ``f2d``
+     - SLAC, BCDMS deuteron :math:`F_2`
+   * - DIS NC (NMC ratio :math:`F_2^d / F_2^p`)
+     - ``(f2p, f2d)``
+     - NMC ratio dataset
+   * - DIS CC
+     - ``dis_cc``
+     - CHORUS, NuTeV, HERA CC
+   * - Jets
+     - ``Hj``
+     - Single-inclusive jet datasets
+   * - Dijets (ATLAS)
+     - ``H2j_ATLAS``
+     - ATLAS dijet datasets (falls back to ``H2j`` if key absent)
+   * - Dijets (CMS)
+     - ``H2j_CMS``
+     - CMS dijet datasets (falls back to ``H2j`` if key absent)
+
+
+Special case: NMC ratio
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The NMC ratio dataset :math:`F_2^d / F_2^p` receives contributions from both
+the proton and deuteron power corrections. The corrected ratio is
+
+.. math::
+
+   \frac{F_2^d}{F_2^p} \to \frac{F_2^d \,(1 + \mathrm{PC}_d)}{F_2^p \,(1 + \mathrm{PC}_p)},
+
+and the shift is
+
+.. math::
+
+   \Delta\!\left(\frac{F_2^d}{F_2^p}\right) = \frac{F_2^d}{F_2^p} \,
+   \frac{\mathrm{PC}_d - \mathrm{PC}_p}{1 + \mathrm{PC}_p}.
+
+
+Covariance matrix construction
+------------------------------
+
+The theory covariance matrix is constructed from the shifts :math:`\Delta O` by
+taking outer products. For each combination of power correction parameters, a
+shift vector is computed per dataset. The sub-matrix between datasets :math:`i`
+and :math:`j` is then
+
+.. math::
+
+   S_{ij} = \sum_k \Delta_i^{(k)} \otimes \Delta_j^{(k)},
+
+where :math:`k` runs over all parameter combinations (one non-zero ``yshift``
+entry at a time, with all others set to zero). This corresponds to the
+``covmat_power_corrections`` function in ``construction.py``.
+
+
+Runcard configuration
+---------------------
+
+Power corrections are included via the ``theorycovmatconfig`` section of the
+runcard. The key ``"power corrections"`` must be added to the
+``point_prescriptions`` list, alongside any scale variation prescriptions.
+
+The following keys are used:
+
+- ``pc_parameters``: a dictionary mapping PC type keys to their parametrisation
+  (``yshift`` and ``nodes`` arrays). The length of ``yshift`` must match the
+  length of ``nodes``.
+- ``pc_included_procs``: list of process types to which power corrections
+  are applied (e.g. ``["DIS NC", "DIS CC", "JETS", "DIJET"]``).
+- ``pc_excluded_datasets``: list of dataset names to exclude from power corrections
+  even if their process type is included.
+- ``pdf``: the PDF used for computing the theory predictions that enter the
+  multiplicative shifts.
+
+Example
+~~~~~~~
+
+.. code:: yaml
+
+    theorycovmatconfig:
+      point_prescriptions: ["9 point", "power corrections"]
+      pc_parameters:
+        f2p:
+          yshift: [0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.0]
+          nodes:  [0.0, 0.001, 0.01, 0.1, 0.3, 0.5, 0.7, 0.9, 1.0]
+        f2d:
+          yshift: [0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.0]
+          nodes:  [0.0, 0.001, 0.01, 0.1, 0.3, 0.5, 0.7, 0.9, 1.0]
+        dis_cc:
+          yshift: [0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.0]
+          nodes:  [0.0, 0.001, 0.01, 0.1, 0.3, 0.5, 0.7, 0.9, 1.0]
+        Hj:
+          yshift: [2.0, 2.0, 2.0, 2.0, 2.0, 2.0]
+          nodes:  [0.25, 0.75, 1.25, 1.75, 2.25, 2.75]
+        H2j_ATLAS:
+          yshift: [2.0, 2.0, 2.0, 2.0, 2.0, 2.0]
+          nodes:  [0.25, 0.75, 1.25, 1.75, 2.25, 2.75]
+        H2j_CMS:
+          yshift: [2.0, 2.0, 2.0, 2.0, 2.0]
+          nodes:  [0.25, 0.75, 1.25, 1.75, 2.25]
+      pc_included_procs: ["JETS", "DIJET", "DIS NC", "DIS CC"]
+      pc_excluded_datasets:
+        - HERA_NC_318GEV_EAVG_CHARM-SIGMARED
+        - HERA_NC_318GEV_EAVG_BOTTOM-SIGMARED
+      pdf: NNPDF40_nnlo_as_01180
+      use_thcovmat_in_fitting: true
+      use_thcovmat_in_sampling: true
+
+.. warning::
+   The lengths of ``yshift`` and ``nodes`` must be equal for each PC type.
+   A mismatch will raise an error at initialisation time.
+
+.. note::
+   Power corrections can be combined with scale variation prescriptions.
+   Both contributions are summed into a single theory covariance matrix.
+   See the tutorial on :ref:`including a theory covmat in a fit <thcov_tutorial>`.
+
+
+Module reference
+----------------
+
+``higher_twist_functions.py`` provides the following public functions:
+
+- ``get_pc_type(exp_name, process_type, experiment, pc_dict)``:
+  determines which PC type key(s) apply to a given dataset.
+- ``linear_bin_function(a, y_shift, bin_edges)``:
+  evaluates the piecewise-linear triangular interpolation at points ``a``.
+- ``dis_pc_func(delta_h, nodes, x, Q2)``:
+  computes the DIS power correction :math:`h(x)/Q^2`.
+- ``jets_pc_func(delta_h, nodes, pT, rap)``:
+  computes the jet power correction :math:`h(y)/p_T`.
+- ``mult_dis_pc(nodes, x, q2, dataset_sp, pdf)``:
+  returns a function that computes the multiplicative DIS shift given node values.
+- ``mult_dis_ratio_pc(p_nodes, d_nodes, x, q2, dataset_sp, pdf)``:
+  returns a function that computes the shift for the :math:`F_2^d/F_2^p` ratio.
+- ``mult_jet_pc(nodes, pT, rap, dataset_sp, pdf)``:
+  returns a function that computes the multiplicative jet shift given node values.
+- ``construct_pars_combs(parameters_dict)``:
+  builds the list of one-at-a-time parameter combinations used to construct
+  the covariance matrix.
+- ``compute_deltas_pc(dataset_sp, pdf, pc_dict)``:
+  computes the full set of shifts for a single dataset.
+
+``construction.py`` provides:
+
+- ``covmat_power_corrections(deltas1, deltas2)``:
+  computes the theory covariance sub-matrix between two datasets from their
+  shift dictionaries.
+- ``covs_pt_prescrip_pc(combine_by_type, point_prescription, pdf, pc_parameters, pc_included_procs, pc_excluded_datasets)``:
+  assembles the full power correction covariance matrix across all datasets.
--- a/nnpdf_data/nnpdf_data/commondata/CMS_2JET_7TEV/metadata.yaml
+++ b/nnpdf_data/nnpdf_data/commondata/CMS_2JET_7TEV/metadata.yaml
@@ -26,16 +26,16 @@ implemented_observables:
   tables: [6, 7, 8, 9, 10]
   process_type: DIJET
   # Plotting
-  kinematic_coverage: [ydiff, m_jj, sqrts]
+  kinematic_coverage: [ymax, m_jj, sqrts]
   plotting:
     dataset_label: "CMS dijets 7 TeV"
     x_scale: log
     plot_x: m_jj
     figure_by:
-    - ydiff
+    - ymax
   kinematics:
     variables:
-      ydiff: {description: "rapidity separation", label: "$y$", units: ""}
+      ymax: {description: "rapidity separation", label: "$y$", units: ""}
       m_jj: {description: "dijet mass", label: "$m_{jj}$", units: "GeV"}
       sqrts: {description: "center of mass energy", label: r"$\sqrt(s)$", units: "GeV"}
     file: kinematics.yaml
--- a/nnpdf_data/nnpdf_data/filter_utils/legacy_jets_utils.py
+++ b/nnpdf_data/nnpdf_data/filter_utils/legacy_jets_utils.py
@@ -98,11 +98,11 @@ def get_kinematics_CMS_2JET_7TEV(tables, version):
             input = yaml.safe_load(file)
 
         rapidity_interval = input['dependent_variables'][0]['qualifiers'][0]['value']
-        ydiff = {}
+        ymax = {}
         if rapidity_interval == '< 0.5':
-            ydiff['min'], ydiff['max'], ydiff['mid'] = 0.0, 0.5, 0.25
+            ymax['min'], ymax['max'], ymax['mid'] = 0.0, 0.5, 0.25
         else:
-            ydiff = range_str_to_floats(rapidity_interval)
+            ymax = range_str_to_floats(rapidity_interval)
 
         sqrts = float(input['dependent_variables'][0]['qualifiers'][2]['value'])
 
@@ -115,7 +115,7 @@ def get_kinematics_CMS_2JET_7TEV(tables, version):
             m_jj['mid'] = float(f"{0.5 * (m_jj['low']+m_jj['high']):.3f}")
 
             kin_value = {
-                'ydiff': {'min': ydiff['min'], 'mid': ydiff['mid'], 'max': ydiff['max']},
+                'ymax': {'min': ymax['min'], 'mid': ymax['mid'], 'max': ymax['max']},
                 'm_jj': {'min': m_jj['low'], 'mid': m_jj['mid'], 'max': m_jj['high']},
                 'sqrts': {'min': None, 'mid': sqrts, 'max': None},
             }
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -1387,6 +1387,25 @@ class CoreConfig(configparser.Config):
 
         return f
 
+    def produce_mult_factor_user_covmat(
+        self, mult_factor: float = 1.0, user_covmat_path: str = None
+    ):
+        """
+        Multiplicative factors for the user covmat provided by mult_factors_user_covmat in the runcard.
+        If no factors are provided, returns None.
+        For use in theorycovariance.construction.user_covmat.
+        """
+        # Check that if mult_factors is provided, user_covmat_paths is also provided
+        if mult_factor is not None and user_covmat_path is None:
+            raise ConfigError(
+                "If mult_factors is provided, user_covmat_paths must also be provided."
+            )
+
+        if mult_factor is None:
+            return 1.0 if user_covmat_path is not None else None
+        else:
+            return mult_factor
+
     def produce_fitthcovmat(
         self, use_thcovmat_if_present: bool = False, fit: (str, type(None)) = None
     ):
@@ -1909,6 +1928,8 @@ class CoreConfig(configparser.Config):
         prescription. The options for the latter are defined in pointprescriptions.yaml.
         This hard codes the theories needed for each prescription to avoid user error."""
         th = t0id.id
+        if point_prescription == 'power corrections':
+            return NSList([t0id], nskey="theoryid")
 
         lsv = yaml_safe.load(read_text(validphys.scalevariations, "scalevariationtheoryids.yaml"))
 
@@ -2018,6 +2039,19 @@ class CoreConfig(configparser.Config):
         if fitthcovmat is None:
             return validphys.results.total_phi_data_from_experiments
         return validphys.results.dataset_inputs_phi_data
+
+    @configparser.explicit_node
+    def produce_covs_pt_prescrip(self, point_prescription):
+        if point_prescription == 'power corrections':
+            from validphys.theorycovariance.construction import covs_pt_prescrip_pc
+
+            f = covs_pt_prescrip_pc
+        else:
+            from validphys.theorycovariance.construction import covs_pt_prescrip_mhou
+
+            f = covs_pt_prescrip_mhou
+
+        return f
 
 
 class Config(report.Config, CoreConfig):
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -1402,7 +1402,7 @@ class CoreConfig(configparser.Config):
             )
 
         if mult_factor is None:
-            return 1.0 if user_covmat_path is not None else None
+            return 1.0
         else:
             return mult_factor
 
--- a/validphys2/src/validphys/filters.py
+++ b/validphys2/src/validphys/filters.py
@@ -125,7 +125,6 @@ class FilterDefaults:
 class FilterRule:
     """
     Dataclass which carries the filter rule information.
-
     """
 
     dataset: str = None
@@ -172,6 +171,9 @@ def default_filter_rules_input():
     are unique, i.d. if there are no multiple rules for the same dataset of
     process with the same rule (`reason` and `local_variables` are not hashed).
     """
+    # TODO: This should be done using a more sophisticated comparison
+    # that checks if two rules are actually the same, regardless of the
+    # order in which the cuts are defined.
     list_rules = yaml_safe.load(read_text(validphys.cuts, "filters.yaml"))
     unique_rules = set(FilterRule(**rule) for rule in list_rules)
     if len(unique_rules) != len(list_rules):
@@ -482,9 +484,9 @@ def check_luxset(luxset):
 def check_luxset_exists(fiatlux):
     """Check that the Photon QED set for this theoryid and luxset exists"""
     if fiatlux is not None:
-      luxset = fiatlux['luxset']
-      luxset.load()
-      log.info(f'{luxset} Lux pdf checked.')
+        luxset = fiatlux['luxset']
+        luxset.load()
+        log.info(f'{luxset} Lux pdf checked.')
 
 
 def check_unpolarized_bc(unpolarized_bc):
--- a/validphys2/src/validphys/theorycovariance/construction.py
+++ b/validphys2/src/validphys/theorycovariance/construction.py
@@ -3,7 +3,8 @@ construction.py
 Tools for constructing theory covariance matrices and computing their chi2s.
 """
 
-from collections import defaultdict, namedtuple
+from collections import defaultdict
+import dataclasses
 import logging
 
 import numpy as np
@@ -11,9 +12,9 @@ import pandas as pd
 
 from reportengine import collect
 from reportengine.table import table
-
-pass
+from validphys.checks import check_pc_parameters
 from validphys.results import results, results_central
+from validphys.theorycovariance.higher_twist_functions import compute_deltas_pc
 from validphys.theorycovariance.theorycovarianceutils import (
     check_correct_theory_combination,
     check_fit_dataset_order_matches_grouped,
@@ -47,11 +48,18 @@ def theory_covmat_dataset(results, results_central_bytheoryids, point_prescripti
     return thcovmat
 
 
-ProcessInfo = namedtuple("ProcessInfo", ("preds", "namelist", "sizes"))
+@dataclasses.dataclass(frozen=True)
+class ProcessInfo:
+    """Dataclass containing the information needed to construct the theory covariance matrix."""
+
+    preds: dict
+    namelist: dict
+    sizes: dict
+    data_spec: dict
 
 
-def combine_by_type(each_dataset_results_central_bytheory):
-    """Groups the datasets bu process and returns an instance of the ProcessInfo class
+def combine_by_type(each_dataset_results_central_bytheory, groups_data_by_process):
+    """Groups the datasets by process and returns an instance of the ProcessInfo class
 
     Parameters
     ----------
@@ -68,6 +76,7 @@ def combine_by_type(each_dataset_results_central_bytheory):
     dataset_size = defaultdict(list)
     theories_by_process = defaultdict(list)
     ordered_names = defaultdict(list)
+    data_spec = {}
     for dataset in each_dataset_results_central_bytheory:
         name = dataset[0][0].name
         theory_centrals = [x[1].central_value for x in dataset]
@@ -77,8 +86,14 @@ def combine_by_type(each_dataset_results_central_bytheory):
         theories_by_process[proc_type].append(theory_centrals)
     for key, item in theories_by_process.items():
         theories_by_process[key] = np.concatenate(item, axis=1)
+
+    # Store DataGroupSpecs instances
+    for group_proc in groups_data_by_process:
+        for ds in group_proc.datasets:
+            data_spec[ds.name] = ds
+
     process_info = ProcessInfo(
-        preds=theories_by_process, namelist=ordered_names, sizes=dataset_size
+        preds=theories_by_process, namelist=ordered_names, sizes=dataset_size, data_spec=data_spec
     )
     return process_info
 
@@ -206,6 +221,35 @@ def covmat_n3lo_ad(name1, name2, deltas1, deltas2):
     return 1 / norm * s
 
 
+def covmat_power_corrections(deltas1, deltas2):
+    """Returns the the theory covariance sub-matrix for power
+    corrections. The two arguments ``deltas1`` and ``deltas2`` contain the
+    shifts for the first and second experiment, respectively.
+
+    The shifts are given in this form: ``` deltas1 = [array1_of_shifts1,
+               array1_of_shifts2, array1_of_shifts3, ...]
+    deltas2 = [array2_of_shifts1,
+               array2_of_shifts2, array2_of_shifts3, ...]
+    ``` The sub-matrix is computed as
+
+      s = array1_of_shifts1 X array2_of_shifts1 + array1_of_shifts2 X
+      array2_of_shifts2 + ...
+
+    where ``X`` is the outer product. This is equivalent to a generalised 5
+    point prescription. Since corrections are linearised, plus and minus
+    variations are not distinguished and all variations are treated as
+    independent, hence the sum over all variations without any additional
+    factors.
+    """
+
+    size1 = deltas1[0].size
+    size2 = deltas2[0].size
+    s = np.zeros(shape=(size1, size2))
+    for shift1, shift2 in zip(deltas1, deltas2):
+        s += np.outer(shift1, shift2)
+    return s
+
+
 def compute_covs_pt_prescrip(point_prescription, name1, deltas1, name2=None, deltas2=None):
     """Utility to compute the covariance matrix by prescription given the
     shifts with respect to the central value for a pair of processes.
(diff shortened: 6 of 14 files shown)


n3fit/src/n3fit/layers/DIS.py
```diff
@@ -39,6 +40,122 @@ class DIS(Observable):
...
+        def H_L(x):
+            return (H_2(x) - np.power(x, 0.05) * H_T(x))
+
+        H_2 = np.vectorize(H_2)
+        H_L = np.vectorize(H_L)
+
+        x = self.exp_kinematics['kin1']
+        y = self.exp_kinematics['kin3']
+        Q2 = self.exp_kinematics['kin2']
+        N2, NL = 1#compute_normalisation_by_experiment(self.dataname, x, y, Q2)
```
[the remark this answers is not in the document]
achiefa: The DIS.py is not in sync with the new way shifts are computed. Hence, that function that adds the shift in the theory predictions can be removed. I'll work on that later on, as we discussed. I should have deleted them.

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -88,11 +93,197 @@ def combine_by_type(each_dataset_results_central_bytheory):
...
+                s += np.outer(deltas[shifts][exp_idx_1: exp_idx_1 + ndata_by_exp[exp_name_1]],
+                              deltas[shifts][exp_idx_2: exp_idx_2 + ndata_by_exp[exp_name_2]])
+
+              start_locs = (exp_idx_1, exp_idx_2)
+              covmats[start_locs] = s
+
+      return covmats, deltas
+
+
+def thcov_ht(thcov_shifts_ht, table_ht_deltas):
```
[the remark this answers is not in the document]
achiefa: Yes, you're right. But I did that in order to store the table of shifts. I know is a bit naive.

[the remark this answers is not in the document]
RoyStegeman: Yes I assumed it was as a quick way to avoid having to call it explicilty. Fine for now

achiefa:
Hi @Roy, I've updated the code so that now the covmat for power corrections can be constructed as in the case of scale variations. Please, look and let me know if something is unclear. I'd like if you could double-check the functions that compute the shifts, in particular when normalisation factors and conversion factors are used.

There are few things that I still don't like here and there, but you're free to propose modifications.

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -276,37 +323,73 @@ def compute_covs_pt_prescrip(point_prescription, name1, deltas1, name2=None, del
...
         s = covmat_3fpt(deltas1, deltas2)
+    elif point_prescription == 'power corrections':
+        # Shifts computed from power corrected predictions
+        s = covmat_power_corrections(deltas1, deltas2)
     return s
 
 
 @check_correct_theory_combination
-def covs_pt_prescrip(combine_by_type, point_prescription):
+def covs_pt_prescrip(
```
achiefa, line 333: I had to add additional dependencies here for the power correction case. I don't know if there is a clever solution, maybe extracting these quantities from the context?

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -276,37 +323,73 @@ def compute_covs_pt_prescrip(point_prescription, name1, deltas1, name2=None, del
...
+    combine_by_type,
+    point_prescription,
+    pdf: PDF,
+    power_corr_dict,
+    pc_included_prosc,
```
achiefa, line 338: I think the tests fail because there is not default for `pc_included_procs` (typo) and `pc_excluded_exps`. What do you suggest to do? Should I add two functions in config.py that default these two values when not specified in the runcard?

scarlehoff (commented):
Hi @achiefa thanks for this.

I've left a few comments on the code. The 1000line `higher_twist_functions` I haven't look into in much detail, but one thing that I'd like to mention is please avoid using `commondata_table`. This is basically an object containing all commondata information like in the old days which was necessary to add for some of validphys functions but the idea is to instead get only the information you want.

For instance, the process type is part of the metadata now (no need to load the data and uncertainties!)
Same for the kinematics (although by the time you need the kinematics you often have already read the whole commondata). You can get the kinematics directly with `cd.metadata.load_kinematics()` and then you already have `x`, `Q`, and `y` for DIS as god intended!

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
```diff
@@ -0,0 +1,1072 @@
...
+import numpy.typing as npt
+
+from validphys.convolution import central_fk_predictions
+from validphys.core import PDF, DataSetSpec
+from validphys.process_options import _Process
+
+GEV_CM2_CONV = 3.893793e10
+GF = 1.1663787e-05  # Fermi's constant [GeV^-2]
+Mh = 0.938  # Proton's mass in GeV/c^2
+MW = 80.398  # W boson mass in GeV/c^2
```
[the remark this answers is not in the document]
scarlehoff:
```
from validphys.loader import Loader
tid = Loader().check_theoryID(40_000_000)
mw = tid.get_description()["MW"]
```

You can also load the theories directly without checking them (which will download the full theory) but I guess you have already some theory locally to use these functions?

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -276,37 +323,82 @@ def compute_covs_pt_prescrip(point_prescription, name1, deltas1, name2=None, del
...
-    If 5 theories are provided, a scheme 'bar' or 'nobar' must be
     chosen in the runcard in order to specify the prescription. Sub-matrices
     correspond to applying the scale variation prescription to each pair of
     processes in turn, using a different procedure for the case where the
     processes are the same relative to when they are different."""
+    process_info = combine_by_type
+    running_index = 0
 
+    covmats = defaultdict(list)
+    if point_prescription != 'power corrections':
```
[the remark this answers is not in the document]
achiefa: You're right, this is already sorted in `config.py`.

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -47,11 +49,11 @@ def theory_covmat_dataset(results, results_central_bytheoryids, point_prescripti
     return thcovmat
 
 
-ProcessInfo = namedtuple("ProcessInfo", ("preds", "namelist", "sizes"))
+ProcessInfo = namedtuple("ProcessInfo", ("preds", "namelist", "sizes", "data", "data_spec"))
```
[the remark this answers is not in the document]
achiefa: Ok, I'll make it a dataclass. What's the advantage though?

validphys2/src/validphys/config.py
```diff
@@ -1830,6 +1832,55 @@ def produce_total_phi_data(self, fitthcovmat):
             return validphys.results.total_phi_data_from_experiments
         return validphys.results.dataset_inputs_phi_data
 
+    def produce_pc_func_type(self, theorycovmatconfig=None):
+        return theorycovmatconfig.get('func_type', 'step')
```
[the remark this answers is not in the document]
achiefa: Yes, you're right. However, `pc_func_type` is just for development and I don't plan to maintain it in the final version. I can put a docstring or a comment to clarify that.

validphys2/src/validphys/theorycovariance/construction.py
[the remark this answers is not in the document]
scarlehoff: Here, flexibility, a bit more of control and the auto discovery rules from validobj.

Since you are changing the number of arguments, to make sure nothing breaks data and data_spec (I still think we should not need both) should have some default, perhaps none, so you need to start complicating the namedtuple. Not all of them need defaults.

(actually, I'm not even sure you can get a namedtuple where only some of the arguments have a default)

validphys2/src/validphys/config.py
```diff
@@ -1830,6 +1832,55 @@ def produce_total_phi_data(self, fitthcovmat):
...
+
+    # @configparser.explicit_node
+    def produce_power_corr_dict(self, pc_parameters=None, pc_func_type=None):
+        """The parameters for the power corrections are given as a list.
+        This function converts this list into a dictionary with the keys
+        being the names of the types of power corrections (e.g. `H2p`, `H2d`,...).
+        """
```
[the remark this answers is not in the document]
achiefa: Probably yes, but I will still need to check if the number of nodes matches the numbers of shifts. Do you suggest to do it here or somewhere else?

validphys2/src/validphys/theorycovariance/construction.py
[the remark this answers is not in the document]
achiefa: Ok, I understand. Thank you for the explanation. You're actually right - I don't need both `data_spec` and `data`. I'll keep the former and discard the latter (also because `data` is not used anywhere).

validphys2/src/validphys/config.py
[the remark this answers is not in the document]
scarlehoff: In `checks.py` you can create checks so that a function needed to use this dictionary will first to run through the check.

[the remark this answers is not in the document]
achiefa: Thank you :)

validphys2/src/validphys/commondata.py
```diff
@@ -35,3 +38,7 @@ def loaded_commondata_with_cuts(commondata, cuts):
 groups_dataset_inputs_loaded_cd_with_cuts = collect(
     "loaded_commondata_with_cuts", ("group_dataset_inputs_by_metadata", "data_input")
 )
+
+groups_dataset_inputs_loaded_cd_with_cuts_byprocess = collect(
+    "loaded_commondata_with_cuts", ("group_dataset_inputs_by_process", "data")
+)
```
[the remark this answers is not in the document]
achiefa: This was a leftover from a previous implementation of the power corrections. I will remove it

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -206,6 +224,40 @@ def covmat_n3lo_ad(name1, name2, deltas1, deltas2):
     return 1 / norm * s
 
 
+def covmat_power_corrections(deltas1, deltas2):
+    """Returns the the theory covariance sub-matrix for power
+    corrections. The two arguments `deltas1` and `deltas2` contain
+    the shifts for the firs and second experiment, respectively.
```
scarlehoff:
```suggestion
    the shifts for the firts and second experiment, respectively.
```

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -206,6 +224,40 @@ def covmat_n3lo_ad(name1, name2, deltas1, deltas2):
...
+    deltas2 = {shift1_label: array2_of_shifts1,
+               shift2_label: array2_of_shifts2,
+               shift3_label: array2_of_shifts3,
+               ...}
+    ```
+    The sub-matrix is computed using the 5-point prescription, thus
+
+      s = array1_of_shifts1 X array2_of_shifts1 + array1_of_shifts2 X array2_of_shifts2 + ...
+
+    where `X` is the outer product.
```
scarlehoff:
```suggestion
    where X is the outer product.
```

Or two of ` `` ` if you wanted teletype text

validphys2/src/validphys/config.py
```diff
@@ -1323,6 +1323,21 @@ def produce_nnfit_theory_covmat(
...
+    
+    def produce_mult_factor_user_covmat(self, mult_factor: float = None, user_covmat_path: str = None):
+        """
+        Multiplicative factors for the user covmat provided by mult_factors_user_covmat in the runcard.
+        If no factors are provided, returns None.
+        For use in theorycovariance.construction.user_covmat.
+        """
+        # Check that if mult_factors is provided, user_covmat_paths is also provided
+        if mult_factor is not None and user_covmat_path is None:
+            raise ConfigError("If mult_factors is provided, user_covmat_paths must also be provided.")
```
scarlehoff: This also could be in the check.

validphys2/src/validphys/config.py
```diff
@@ -1928,7 +1945,18 @@ def produce_filter_data(
...
+            # but we do want to use the theory covmat for L2 data
+            if theorycovmatconfig is not None and theorycovmatconfig.get(
+                "use_thcovmat_in_fakedata_sampling"
+            ):
+                # NOTE: By the time we run theory covmat closure tests,
+                # hopefully the generation of pseudodata will be done in python.
+                raise ConfigError(
+                    "Generating L1 closure test data which samples from the theory "
+                    "covariance matrix has not been implemented yet."
+                )
```
scarlehoff: ?????

This has appeared due to a wrong rebasing maybe? It seems to come from the inconsistent closure tests? (and we do now the generation of pseudodata in python)

scarlehoff (commented):
Thanks. We can talk on Tuesday to have someone go through it in more detail.

I've just looked at the changes very quickly and left a few comments, but since you have been rebasing this can probably be merged relatively quickly.

Btw, I think it might be a good idea to add a trimmed down version of the runcard you have added to the regression tests folder (in the `root`), without the `"9 point"` only with the power corrections.
That way we'll be testing at least partially the theory covmat part (which we are not doing right now to avoid downloading too many fktables).

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -206,6 +224,40 @@ def covmat_n3lo_ad(name1, name2, deltas1, deltas2):
...
+    """
+    # Check that `deltas1` and `deltas2` have the same shifts
+    if deltas1.keys() != deltas2.keys():
+        raise RuntimeError('The two dictionaries do not contain the same shifts.')
```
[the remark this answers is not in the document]
achiefa: OK!

RoyStegeman:
In the end, have the theories on the storage server been updated with the k-factors?

nnpdf_data/nnpdf_data/commondata/CMS_2JET_7TEV/metadata.yaml
```diff
@@ -26,16 +26,16 @@ implemented_observables:
   tables: [6, 7, 8, 9, 10]
   process_type: DIJET
   # Plotting
-  kinematic_coverage: [ydiff, m_jj, sqrts]
+  kinematic_coverage: [ymax, m_jj, sqrts]
```
scarlehoff, line 29: Are these changes necessary / should they be in this PR?

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -206,6 +223,37 @@ def covmat_n3lo_ad(name1, name2, deltas1, deltas2):
     return 1 / norm * s
 
 
+def covmat_power_corrections(deltas1, deltas2):
+    """Returns the the theory covariance sub-matrix for power
+    corrections. The two arguments `deltas1` and `deltas2` contain
```
scarlehoff:
```suggestion
    corrections. The two arguments ``deltas1`` and ``deltas2`` contain
```

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -206,6 +223,37 @@ def covmat_n3lo_ad(name1, name2, deltas1, deltas2):
...
+      s = array1_of_shifts1 X array2_of_shifts1 + array1_of_shifts2 X array2_of_shifts2 + ...
+
+    where ``X`` is the outer product.
+    """
+
+    size1 = deltas1[0].size
+    size2 = deltas2[0].size
+    s = np.zeros(shape=(size1, size2))
+    for shift1, shift2 in zip(deltas1, deltas2):
+        s += np.outer(shift1, shift2)
```
scarlehoff, line 249: It says that it is using the 5 point prescription, but this is different from the `covmat_5pt` function. Why?

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -276,28 +324,30 @@ def compute_covs_pt_prescrip(point_prescription, name1, deltas1, name2=None, del
...
+        s = covmat_power_corrections(deltas1, deltas2)
     return s
 
 
 @check_correct_theory_combination
-def covs_pt_prescrip(combine_by_type, point_prescription):
+def covs_pt_prescrip_mhou(combine_by_type, point_prescription):
     """Produces the sub-matrices of the theory covariance matrix according
     to a point prescription which matches the number of input theories.
-    If 5 theories are provided, a scheme 'bar' or 'nobar' must be
```
scarlehoff, line 286: Why has this single line fallen?

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -307,6 +357,49 @@ def covs_pt_prescrip(combine_by_type, point_prescription):
...
+    to each pair of `datasets`."""
+    process_info = combine_by_type
+    datagroup_spec = process_info.data_spec
+    running_index = 0
+
+    covmats = defaultdict(list)
+    start_proc_by_exp = defaultdict(list)
+    for exp_name, data_spec in datagroup_spec.items():
+        start_proc_by_exp[exp_name] = running_index
+        running_index += data_spec.load_commondata().ndata
```
scarlehoff: I am a bit worried here, you are loading the commondata and taking the number of data point.
But then in the loop below, inside `compute_deltas_pc` you are cutting the data...

are you sure this is being done consistently? (in other words, are you sure that `load_commmondata` here is happening after cuts?

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
```diff
@@ -0,0 +1,848 @@
...
+    deltas = defaultdict(list)
+
+    if "H2j_ystar" in pc_dict and "H2j_yb" in pc_dict:
+        pars_combs = pars_comb_func(pc_dict, ("H2j_ystar", "H2j_yb"))
+    else:
+        pars_combs = pars_comb_func(pc_dict)
+
+    pc_type = get_pc_type(exp_name, process_type, experiment=experiment, pc_dict=pc_dict)
+
+    if process_type.startswith('DIS'):
```
scarlehoff, line 732: Counting the days until someone call a process `DISASTER` for some reason 🙈 

I would perhaps also guard against processes that have no `x` or no `Q2`.

validphys2/src/validphys/config.py
```diff
@@ -1869,6 +1888,8 @@ def produce_theoryids(self, t0id, point_prescription):
         prescription. The options for the latter are defined in pointprescriptions.yaml.
         This hard codes the theories needed for each prescription to avoid user error."""
         th = t0id.id
+        if point_prescription == 'power corrections':
+            return NSList([t0id], nskey="theoryid")
```
scarlehoff, line 1932: I wonder how many things across the code will assume that a theory covmat needs more than one theory...

scarlehoff (commented):
Left a few comments but haven't run the code yet. I've also look at higher_twist_functions very superficially for now, just got curious about the cuts.

I'm a bit worried about all the if conditions that this one prescription adds and whether it breaks some assumptions in the code. In particular the one where theoryid(s) is a one-item list...

nnpdf_data/nnpdf_data/commondata/CMS_2JET_7TEV/metadata.yaml
achiefa: I checked this with Emanuele, and we concluded that `ydiff` is equivalent to `ymax`. Since the other dijet datasets adopt `ymax`, and the name of the variables is processed in the construction of the theory covmat, I changed it to `ymax`

validphys2/src/validphys/theorycovariance/construction.py
achiefa: To be fear, I don't know what to call this. Richard called this generalised 5-pt. The point is that we have a shift for each ht parameter. Moreover, since each shift is linear in the parameter, the outer products of positive and negative shifts are equivalent. That's way there is no 0.5 factor. In few words, is a generalised 5-pt prescription where positive and negative shifts are equivalent. Happy to change the name.

scarlehoff: No problem with the name, with a mention to the difference in the docstr I'd be happy.

validphys2/src/validphys/theorycovariance/construction.py
achiefa: I don't know what happened here. I'll restore this line.

validphys2/src/validphys/theorycovariance/construction.py
achiefa: I think this might give me mercy, right?
 https://github.com/NNPDF/nnpdf/blob/73ad16a0e249da9a51761886360ba81ea4d5febd/validphys2/src/validphys/core.py#L504-L512

scarlehoff: I see. I guess never load the data without cuts now. That's good then :)

Experience is a prison sometimes.

achiefa: I can add a comment next to `load_commondata` to clarify this.

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
achiefa: > Counting the days until someone call a process DISASTER for some reason 🙈

Highly likely. Do you have a better idea?

> I would perhaps also guard against processes that have no x or no Q2.

Why would this happen?

validphys2/src/validphys/config.py
achiefa: Eheh, we can ask our AI companions...

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
scarlehoff: That's what I meant, check both that `startswith('DIS')` and that the processes have `x` and `Q2` and we agree that everything that complies with both is a DIS process even if it called `DISNOTREALLY`

validphys2/src/validphys/config.py
scarlehoff: I wouldn't even know how to start prompting it. Perhaps starting with "write a test for everything that uses this function" :__)

doc/sphinx/source/vp/theorycov/power_corrections.rst
```diff
@@ -0,0 +1,238 @@
...
+``point_prescriptions`` list, alongside any scale variation prescriptions.
+
+The following keys are used:
+
+- ``pc_parameters``: a dictionary mapping PC type keys to their parametrisation
+  (``yshift`` and ``nodes`` arrays). The length of ``yshift`` must match the
+  length of ``nodes``.
+- ``pc_included_procs``: list of process types to which power corrections
+  are applied (e.g. ``["DIS NC", "DIS CC", "JETS", "DIJET"]``).
+- ``pc_excluded_exps``: list of dataset names to exclude from power corrections
```
scarlehoff: A nitpick but it should be excluded observables or excluded datasets, not excluded experiments.

doc/sphinx/source/vp/theorycov/power_corrections.rst
```diff
@@ -0,0 +1,238 @@
...
+  even if their process type is included.
+- ``pdf``: the PDF used for computing the theory predictions that enter the
```
scarlehoff, line 160: A question for my own education, these multiplicative shift will be performed using FkTables.
How important is it that the PDF used for the multiplicative shift has been determined using the same evolution settings ad the FkTables that will be used for the shifts?

(in the case of theory variations, since the evolution settings are supposed to be at a higher orders, these differences may be too, but we know that there are regions in which the evolution settings do have an impact, e.g., low-x for truncated vs exact)

nnpdf_data/nnpdf_data/filter_utils/legacy_jets_utils.py
```diff
@@ -115,7 +115,7 @@ def get_kinematics_CMS_2JET_7TEV(tables, version):
             m_jj['mid'] = float(f"{0.5 * (m_jj['low']+m_jj['high']):.3f}")
 
             kin_value = {
-                'ydiff': {'min': ydiff['min'], 'mid': ydiff['mid'], 'max': ydiff['max']},
+                'ymax': {'min': ymax['min'], 'mid': ymax['mid'], 'max': ymax['max']},
```
scarlehoff, line 118: Make sure to apply the label regenerate data so that all datasets that might be affected by this change are regenerated with the right kinematics name.

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -13,7 +14,9 @@
 from reportengine.table import table
 
 pass
+from validphys.checks import check_pc_parameters
```
scarlehoff, line 15: why is there a `pass` here? Apparently is there already in master but could you remove it please?

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -77,8 +88,14 @@ def combine_by_type(each_dataset_results_central_bytheory):
         theories_by_process[proc_type].append(theory_centrals)
     for key, item in theories_by_process.items():
         theories_by_process[key] = np.concatenate(item, axis=1)
+
+    # Store DataGroupSpecs instances
+    for group_proc in groups_data_by_process:
+        for exp_set in group_proc.datasets:
+            data_spec[exp_set.name] = exp_set
```
scarlehoff: I am a bit confused about the variable names here.

These are not experiments but rather separate datasets, right? So this is a dictionary of `{dataset_name:  dataset}`. Given that "experiment" has a very specific name within the codebase it would be better to change that name.

Also, since you are setting with `=` you don't need to use a `defaultdict` here.

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -307,6 +356,49 @@ def covs_pt_prescrip(combine_by_type, point_prescription):
...
+    """Produces the sub-matrices of the theory covariance matrix for power
+    corrections. Sub-matrices correspond to applying power corrected shifts
+    to each pair of `datasets`."""
+    process_info = combine_by_type
+    datagroup_spec = process_info.data_spec
+    running_index = 0
+
+    covmats = defaultdict(list)
+    start_proc_by_exp = defaultdict(list)
+    for exp_name, data_spec in datagroup_spec.items():
```
scarlehoff: Given the comments above, are these experiments or datasets?

Also, since you are simply looping over them, why don't you simply keep a list?

```python
for dataset in process_info.datasets:
    start_proc_by_exp[dataset.name] = running_index
    running_index += dataset.load_commondata().ndata
```

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -307,6 +356,49 @@ def covs_pt_prescrip(combine_by_type, point_prescription):
...
+
+    for exp_name1, data_spec1 in datagroup_spec.items():
+        for exp_name2, data_spec2 in datagroup_spec.items():
+            process_type1 = process_lookup(exp_name1)
+            process_type2 = process_lookup(exp_name2)
+
+            is_excluded_exp = any(name in pc_excluded_exps for name in [exp_name1, exp_name2])
+            is_included_proc = any(
+                proc not in pc_included_procs for proc in [process_type1, process_type2]
+            )
```
scarlehoff, line 386: nitpick, but I think it would be easier to understand:

```python
if is_excluded_exp or not is_included_proc:
    continue
```

otherwise go with the rest of the logic

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
```diff
@@ -0,0 +1,848 @@
...
+    elif process_type == "DIJET_3D":
+        if not (pc_dict.get("H2j_ystar") and pc_dict.get("H2j_yb")):
+            raise ValueError(
+                "For DIJET_3D, both 'H2j_ystar' and 'H2j_yb' keys must be present in pc_dict."
+            )
+        return ("H2j_ystar", "H2j_yb")
+
+    else:
+        raise RuntimeError(f"{process_type} has not been implemented.")
+
```
scarlehoff, line 122:
```suggestion
    raise RuntimeError(f"{process_type} has not been implemented.")
```

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
```diff
@@ -0,0 +1,848 @@
...
+        return ("H2j_ystar", "H2j_yb")
+    else:
+        if experiment == 'ATLAS':
+            specific_key = "H2j_ystar"
+        elif experiment == 'CMS':
+            specific_key = "H2j_ymax"
+        else:
+            raise ValueError(f"{experiment} is not implemented for DIJET.")
+
+    return specific_key
```
scarlehoff, line 153:
```suggestion
    if experiment == 'ATLAS':
        specific_key = "H2j_ystar"
    elif experiment == 'CMS':
        specific_key = "H2j_ymax"
    else:
        raise ValueError(f"{experiment} is not implemented for DIJET.")

    return specific_key
```

Also, this function has a `pc_dict = None` default argument, but it should be an empty dictionary `dict()`, otherwise the first line might crash when `None` is passed.

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
```diff
@@ -0,0 +1,848 @@
...
+        cond_low = np.multiply(
+            a >= bin_low, a < bin_mid if shift_pos != len(y_shift) - 1 else a <= bin_mid
+        )
+        cond_high = np.multiply(
+            a >= bin_mid, a < bin_high if shift_pos != len(y_shift) - 1 else a <= bin_high
+        )
+        res = np.add(res, [m1 * (val - bin_low) if cond else 0.0 for val, cond in zip(a, cond_low)])
+        res = np.add(
+            res, [-m2 * (val - bin_high) if cond else 0.0 for val, cond in zip(a, cond_high)]
+        )
```
scarlehoff: Why `np.add` instead of simply `res += ` ?

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
```diff
@@ -0,0 +1,848 @@
...
+
+                def _F(tx, ty):
+                    return (hx0 * (1.0 - tx) + hx1 * tx) * (hy0 * (1.0 - ty) + hy1 * ty)
+
+                # Simpson's rule (exact for the quadratic product of two linears)
+                integral += (b - a) / 6.0 * (_F(tx_a, ty_a) + 4.0 * _F(tx_m, ty_m) + _F(tx_b, ty_b))
+
+        H_z[z_idx] = integral / Delta_x_total
+
+    return H_z / m_jj
```
scarlehoff, line 412: In general in this module I guess performance is not an issue right?
(There's a few things that could make a difference but I guess that if you are using python loops it means it's not really relevant)

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
```diff
@@ -0,0 +1,848 @@
...
+    F2D = np.concatenate(F2D.values)
+    F2P = np.concatenate(F2P.values)
+    F2_ratio = operator.truediv(F2D, F2P)
+
+    def func(y_values_p, y_values_d):
+        h2d = dis_pc_func(y_values_d, d_nodes, x, q2)
+        h2p = dis_pc_func(y_values_p, p_nodes, x, q2)
+        if cfactors:
+            # If this function is used to compute c-factors,
+            # no need to multiply by the observable
```
scarlehoff: Put this comment above the whole group, I was going to write a comment asking exactly about this (where you mention `# Power correction callable factories`)

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
```diff
@@ -0,0 +1,848 @@
...
+) -> callable:
+    cuts = dataset_sp.cuts
+    (fkspec,) = dataset_sp.fkspecs
+    fk = fkspec.load_with_cuts(cuts)
+    th_preds = central_fk_predictions(fk, pdf)
+
+    def func(ystar_shifts, y_b_shifts):
+        if distrb == '3D':
+            if len(ystar) == 0 or len(yb) == 0:
+                raise ValueError("For 3D distributions, ystar and yb must be provided.")
```
scarlehoff: I think these errors should be before the function, since `dstrib`, `ystar` and `yb` are not arguments of `func` it menas that by the time the function is generated you know already whether it is going to error out.

validphys2/src/validphys/config.py
```diff
@@ -1347,6 +1347,25 @@ def produce_nnfit_theory_covmat(
...
+        if mult_factor is not None and user_covmat_path is None:
+            raise ConfigError(
+                "If mult_factors is provided, user_covmat_paths must also be provided."
+            )
+
+        if mult_factor is None:
+            return 1.0 if user_covmat_path is not None else None
+        else:
+            return mult_factor
+
```
scarlehoff, line 1408: Wouldn't it make sense to return just `1.0` instead of `None`? That way you don't need to guard later against `None`.

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -307,6 +356,49 @@ def covs_pt_prescrip(combine_by_type, point_prescription):
...
+            if not (is_excluded_exp or is_included_proc):
+                deltas1 = compute_deltas_pc(data_spec1, pdf, pc_parameters)
+                deltas2 = compute_deltas_pc(data_spec2, pdf, pc_parameters)
+
+                # Convert deltas1 and deltas2 to the format expected by compute_covs_pt_prescrip
+                deltas1 = [np.array(deltas1[shift]) for shift in sorted(deltas1.keys())]
+                deltas2 = [np.array(deltas2[shift]) for shift in sorted(deltas2.keys())]
```
scarlehoff: I think it would make sense for `compute_deltas_pc` to return automatically the format you need (why do you need the dictionary if you drop it later?)

But otherwise I think it could be a bit clearer to do:

```suggestion
                deltas1_dict = compute_deltas_pc(data_spec1, pdf, pc_parameters)
                deltas2_dict = compute_deltas_pc(data_spec2, pdf, pc_parameters)
                deltas1 = []
                deltas2 = []
                
                # Convert deltas1 and deltas2 to the format expected by compute_covs_pt_prescrip
                for shift in sorted(deltas1_dict.keys()):
                    deltas1.append(np.array(deltas1_dict[shift])
                    deltas2.append(np.array(deltas2_dict[shift])
```

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
```diff
@@ -0,0 +1,848 @@
...
+        pc_func = mult_jet_pc(nodes, pT, eta, dataset_sp, pdf, cfactors=cfactors)
+        deltas.update(_apply_pars_combs(pars_combs, pc_type, pc_func))
+
+    elif process_type == 'DIJET':
+        rap_var = DIJET_RAPIDITY_VAR.get(experiment)
+        if rap_var is None:
+            raise ValueError(f"{experiment} is not implemented for DIJET.")
+
+        kinematics = dataset_sp.commondata.metadata.load_kinematics()
+        # TODO Excpetion rule for CMS_2JET_7TEV_M12-Y should be removed
```
scarlehoff: ?

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
```diff
@@ -0,0 +1,848 @@
...
+                ymax=rap if rap_var == 'ymax' else np.array([]),
+                distrb=rap_var,
+                cfactors=cfactors,
+            )
+            # Both ystar and yb parameters contribute for all 2D distributions
+            deltas.update(_apply_pars_combs(pars_combs, ["H2j_ystar", "H2j_yb"], pc_func))
+
+        else:
+            nodes = pc_dict[pc_type]['nodes']
+            pc_func = mult_jet_pc(nodes, m_jj, rap, dataset_sp, pdf, cfactors=cfactors)
```
scarlehoff, line 796: So, if we enter this `lese` we call the jet function but passing down the `m_jj` as `pt`. Is this correct? (why is this correct)

Perhaps a comment above (I guess there's a reason) so the next person looking through this doesn't have the same question I had

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
```diff
@@ -0,0 +1,848 @@
...
+    Parameters
+    ----------
+    delta_h: ArrayLike
+      One-dimensional array containing the shifts for each bin.
+    nodes: ArrayLike
+     One-dimensional array containing the edges of the bins in rapidity.
+    rap: ArrayLike
+      List of rapidity points at which the power correction is evaluated.
+    pT: ArrayLike
+      List of pT points at which the power correction is evaluated.
```
scarlehoff: The two are the wrong way around

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
```diff
@@ -0,0 +1,848 @@
...
+            bin_mid = bin_edges[shift_pos]
+            bin_high = bin_mid
+            m1 = shift / (bin_mid - bin_low)
+            m2 = 0.0
+        cond_low = np.multiply(
+            a >= bin_low, a < bin_mid if shift_pos != len(y_shift) - 1 else a <= bin_mid
+        )
+        cond_high = np.multiply(
+            a >= bin_mid, a < bin_high if shift_pos != len(y_shift) - 1 else a <= bin_high
+        )
```
scarlehoff: I'm trying to wrap my head around these two statements but I'm a bit confused. I think you cannot get a  `True * True` as long as `shift_pos != len(y_shift) - 1` 

But if `not shift_pos != len(y_shift) - 1` then you can get `True` in both if `a == bin_mid` (because the second one is `<=` while I think it should be just `<`?

But maybe I'm understanding this wrong.

(also, why instead of multiplying Trues and Falses don't you use an `&` here?)

scarlehoff (commented):
I tried to run the exmaple runcard and got:


```
~ vp-setupfit Basic_runcard_pc_covmat.yml  
ValueError: Shape of passed values is (3148, 3148), indices imply (3418, 3418)
```

This does look like a problem with the cuts (with the funny coincidence of 14 -> 41)

Also, a lot of errors like this one before crashing:

```
[INFO]: Summary: 3238/3850 datapoints passed kinematic cuts.
/Volumes/csDisk/Repositories/NNPDF/nnpdf/validphys2/src/validphys/covmats_utils.py:88: RuntimeWarning: divide by zero encountered in matmul
  return np.diag(diagonal) + corr_sys_mat @ corr_sys_mat.T
```

For the rest, I've had a look at the `higher_twist_functions.py‎` module and I've flagged a few things. It's probably not comprehensive enough, it is very long 😅 but anyway it can probably be merged once the issue above gets solved.

doc/sphinx/source/vp/theorycov/power_corrections.rst
achiefa: Agreed, I'll change it to `pc_excluded_datasets` here and in the code.

doc/sphinx/source/vp/theorycov/power_corrections.rst
achiefa: That's a good question, for which I don't have an answer. We could investigate this quantitatively, but I'm afraid I won't have time. Nevertheless, I think there's an analogy between your question and the t0 prescription. Before iteration, the pdf set used to compute the predictions (for t0 or power corrections) is certainly different from the one resulting from the fit. Therefore, the effects you mentioned may be substantial. However, when the fit is iterated and the pdf in the runcard is updated accordingly, I'd expect these differences to be mitigated as the pdf used for t0/pc  is much closer this time compared to the first fit. Could this be a good explanation?

validphys2/src/validphys/theorycovariance/construction.py
achiefa: Yeah you're absolutely right. I think this goes back to my first days when my understanding of the codebase and nomenclature was a bit limited (though slightly improved now). I will replace the names here accordingly.

validphys2/src/validphys/theorycovariance/construction.py
achiefa: These are dataset names.

validphys2/src/validphys/theorycovariance/construction.py
achiefa: I prefer to keep the dictionary, because it turns out to be convenient for debugging and for keeping track of the different shifts. Indeed, it becomes useful when using the tcm. However, I'll use different names, as you suggested.

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
achiefa: Idk, maybe back then I was into C-style syntax.

doc/sphinx/source/vp/theorycov/power_corrections.rst
scarlehoff: I was thinking more about the fact of how a mismatch on settings (for instance, using 4.0 with 4.1 fktables) would affect the result.
But it is true that, as long as there is at least one iteration, the settings then will be, by definition, the same.

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
achiefa: As long as the construction of the theory covmat is concerned, this module is called only in setupfit. Therefore the fit is not affected. Also these loops are typically short (nodes have three elements for dijets), so I don't think it would make any difference.

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
achiefa: Tbg, I wrote this function almost two years ago and I detested it ever since. I'm pretty sure I can come up with a shorter version that does exactly the same thing.

achiefa: Something like this looks much better and cleaner.
```python
def linear_bin_function_proposed(
     a: npt.ArrayLike, y_shift: npt.ArrayLike, nodes: npt.ArrayLike
) -> np.ndarray:
    a = np.asarray(a, dtype=float)
    res = np.interp(a, nodes, y_shift)
    res[(a < nodes[0]) | (a > nodes[-1])] = 0.0
    return res

```

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
achiefa: You're right.

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
achiefa: Forget about this: was a workaround that I came up with when the name of the dataset changed.

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
achiefa: Yes, this both correct and expected. The reason is that for datasets the end up in the else, the structure of the power corrections is identical to the single-inclusive jet, provided that pT is replaced with mjj. I'll add a comment.

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
scarlehoff: I'm not that worried about the complexity of the function or the length, I'm just not convinced it is correct ^^U

achiefa:
Hi @scarlehoff, thanks for your comments and suggestions. I went through them all and I'll push the new fixes. I also fixed the basic runcard: now it works without crashing. I just don't understand how to reproduce this warning
```
[INFO]: Summary: 3238/3850 datapoints passed kinematic cuts.
/Volumes/csDisk/Repositories/NNPDF/nnpdf/validphys2/src/validphys/covmats_utils.py:88: RuntimeWarning: divide by zero encountered in matmul
  return np.diag(diagonal) + corr_sys_mat @ corr_sys_mat.T
```
which btw is the same as #2455.

scarlehoff:
If you are not getting the warning perhaps it is an issue with my version of numpy...

achiefa:
> If you are not getting the warning perhaps it is an issue with my version of numpy...

I'm not able to reproduce this warning.

scarlehoff:
Indeed, it seems it was my version of numpy.

validphys2/src/validphys/filters.py
```diff
@@ -482,9 +484,9 @@ def check_luxset(luxset):
 def check_luxset_exists(fiatlux):
     """Check that the Photon QED set for this theoryid and luxset exists"""
     if fiatlux is not None:
-      luxset = fiatlux['luxset']
-      luxset.load()
-      log.info(f'{luxset} Lux pdf checked.')
+        luxset = fiatlux['luxset']
+        luxset.load()
+        log.info(f'{luxset} Lux pdf checked.')
```
scarlehoff, line 489: I'm a bit worried about these files having been touched. They were not in the previous version I reviewed were there?

scarlehoff (approved):
Thanks, I'm not going to go through the whole of `higher_twist_functions.py` again (also because I think you played god with git's history so it is even more difficult to know what is the different with respect to the previous version) but I'll trust the changes are basically those suggested 😅

`vp-setupfit` at least now works

validphys2/src/validphys/filters.py
achiefa: I've checked on a backup that I did before changing the commit history, and these changes in filters.py where already there. And the rest is as expected after my hard rebase. SO this is good to go.

achiefa:
Should I be worried about the regression test failing?

scarlehoff:
This _is_ one of the random failures that will be fixed by https://github.com/NNPDF/nnpdf/issues/2464

PR #2107, opened 2024-06-06
Set `separate_multiplicative` as `False` by default

scarlehoff:
As agreed.

As far as I am aware most projects had already `sampling::separate_multiplicative = False` in their runcards anyway.

PR #2128, opened 2024-07-17
move `separate_multiplicative` key outside `sampling` key in runcard

### LINKED ISSUE #2106: Add `resample_negative_pseudodata` to the `sampling` namespace
Or make them both top-level variables. But it doesn't make sense to have them separated.

https://github.com/NNPDF/nnpdf/blob/e9f27634acecbd7629eca2caa9dcac9ea473933e/n3fit/runcards/examples/Basic_runcard.yml#L26

I'm perfectly ok with having both as top-level variables tbh

RoyStegeman:
This fixes vp-setupfit since there `separate_multiplicative` was not parsed from the `sampling` key and thus it was not found. This was broken in https://github.com/NNPDF/nnpdf/pull/2107

This also resolves #2106 by removing the sampling namespace all together, though since we never use it we may also want to drop the `separate_multiplicative` keyword all together.

### DIFF
--- a/extra_tests/regression_fits/trainable_prepro.yml
+++ b/extra_tests/regression_fits/trainable_prepro.yml
@@ -35,7 +35,7 @@ mcseed: 1
 
 load: "weights.weights.h5"
 
-sep_mult: True
+separate_multiplicative: True
 
 parameters: # This defines the parameter dictionary that is passed to the Model Trainer
   nodes_per_layer: [15, 10, 8]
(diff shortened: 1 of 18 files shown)


scarlehoff:
> This also resolves https://github.com/NNPDF/nnpdf/issues/2106 by removing the sampling namespace all together,

This solution works for me.

> though since we never use it we may also want to drop the separate_multiplicative keyword all together.

But given we already have the option to be able to change between the two modes, we might as well keep it. It might be useful to test (and to compare with 4.0).

andreab1997 (approved):
Looks good to me, thanks for this!

RoyStegeman:
I don't think you can use the same name for user-input and for the result of a parse_ or produce_ function

PR #2129, opened 2024-07-18
Bump pineappl version

Radonirinaunimi:
Now that pineappl [v0.8.1](https://github.com/NNPDF/pineappl/releases/tag/v0.8.1) is out we can update here. In principle, these changes are also introduced in https://github.com/NNPDF/nnpdf/pull/2110 but given that this is needed to make pineko work (https://github.com/NNPDF/pineko/pull/181) it would be better to have this in master first.

Tests relying on conda are still failing because of https://github.com/NNPDF/pineappl/issues/300.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,7 +62,7 @@ vp-deltachi2 = "validphys.scripts.vp_deltachi2:main"
 # Generic dependencies (i.e., validphys)
 python = "^3.9"
 matplotlib = ">=3.3.0,<3.8"
-pineappl = "^0.7.0"
+pineappl = "^0.8.1"
 pandas = "*"
 numpy = "*"
 validobj = "*"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,7 +62,7 @@ vp-deltachi2 = "validphys.scripts.vp_deltachi2:main"
 # Generic dependencies (i.e., validphys)
 python = "^3.9"
 matplotlib = ">=3.3.0,<3.8"
-pineappl = "^0.8.1"
+pineappl = "^0.8.2"
 pandas = "*"
 numpy = "*"
 validobj = "*"
--- a/validphys2/src/validphys/pineparser.py
+++ b/validphys2/src/validphys/pineparser.py
@@ -159,9 +159,16 @@ def pineappl_reader(fkspec):
     is_polarized = pine_rep.key_values().get("polarized") == "True"
 
     # Is it hadronic? (at the moment only hadronic and DIS are considered)
-    hadronic = pine_rep.key_values()["initial_state_1"] == pine_rep.key_values()["initial_state_2"]
+    try:
+        parton1 = pine_rep.key_values()["convolution_particle_1"]
+        parton2 = pine_rep.key_values()["convolution_particle_2"]
+    except KeyError:
+        parton1 = pine_rep.key_values()["initial_state_1"]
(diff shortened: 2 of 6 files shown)


giacomomagni:
@Radonirinaunimi shouldn't we update also the `conda-recipe` to make all the test passing?

Radonirinaunimi:
> @Radonirinaunimi shouldn't we update also the `conda-recipe` to make all the test passing?

Yes, I will do so once this https://github.com/NNPDF/pineappl/issues/300 is resolved. I will have time to look into it more tomorrow (if not today).

scarlehoff:
The problem is that the conda package for pineappl seems to not be working at the moment https://github.com/conda-forge/pineappl-feedstock/pull/41

pyproject.toml
```diff
@@ -62,7 +62,7 @@ vp-deltachi2 = "validphys.scripts.vp_deltachi2:main"
 # Generic dependencies (i.e., validphys)
 python = "^3.9"
 matplotlib = ">=3.3.0,<3.8"
-pineappl = "^0.7.0"
+pineappl = "^0.8.2"
```
scarlehoff, line 65: I'm not a friend of setting very strict requirements, `0.8.0` should work as well. Given 0-1-2 were released almost together it's ok

(also, I hope going forward, there's not many more changes of api, having to change code just because channels is more correct than lumi is a bit annoying...)

scarlehoff (approved):

pyproject.toml
Radonirinaunimi: > I'm not a friend of setting very strict requirements, 0.8.0 should work as well.

Likewise! Although I'd keep it this way because for whatever reason `pineko` (given that `pineko` now depends on `nnpdf`) would pick versions between  `<=v0.8,0<v0.8.2` which will not work.

> (also, I hope going forward, there's not many more changes of api, having to change code just because channels is more correct than lumi is a bit annoying...)

I have also always advocated for this...

validphys2/src/validphys/pineparser.py
```diff
@@ -159,9 +159,16 @@ def pineappl_reader(fkspec):
...
     # Is it hadronic? (at the moment only hadronic and DIS are considered)
-    hadronic = pine_rep.key_values()["initial_state_1"] == pine_rep.key_values()["initial_state_2"]
+    try:
+        parton1 = pine_rep.key_values()["convolution_particle_1"]
+        parton2 = pine_rep.key_values()["convolution_particle_2"]
+    except KeyError:
+        parton1 = pine_rep.key_values()["initial_state_1"]
+        parton2 = pine_rep.key_values()["initial_state_2"]
+    hadronic = parton1 == parton2
+
```
cschwan, line 169: This should probably changed in the future to use the `Grid::convolutions` method to make it future proof; starting with the new file format the convolutions will no longer be stored in metadata.

pyproject.toml
scarlehoff: > given that pineko now depends on nnpdf

The dependency of pineko in nnpdf is optional. But yes, the problem with being very restrictive and changing the api without necessity (in _this_ case I guess there was necessity) is that the different projects will start clashing with each other.

validphys2/src/validphys/pineparser.py
Radonirinaunimi: That'd be great, especially since the metadata might not always be as reliable.

PR #2132, opened 2024-07-24
nsigma dataspecs tables

comane:
Allows one to do comparisons of nsigma between different fits

https://vp.nnpdf.science/92_y8rL4TMeXDGMlOK9tOg==

```
meta:
  title: chi2 inconsistent DIS fits
  author: Mark N. Costantini
  keywords: [closure test, inconsistencies]


dataset_inputs:
# in sample DIS
- {dataset: NMCPD_dw_ite, variant: legacy}
- {dataset: SLACP_dwsh, variant: legacy}
- {dataset: SLACD_dw_ite, variant: legacy}
- {dataset: BCDMSP_dwsh, variant: legacy}
- {dataset: BCDMSD_dw_ite, variant: legacy}
- {dataset: CHORUSNUPb_dw_ite, variant: legacy}
- {dataset: CHORUSNBPb_dw_ite, variant: legacy}
- {dataset: NTVNUDMNFe_dw_ite, variant: legacy, cfac: ['MAS']}
- {dataset: HERACOMBNCEM, variant: legacy}
- {dataset: HERACOMBNCEP575, variant: legacy}
- {dataset: HERACOMBNCEP820, variant: legacy}
- {dataset: HERACOMBNCEP920, variant: legacy}
- {dataset: HERACOMBCCEP, variant: legacy}
- {dataset: HERACOMB_SIGMARED_C, variant: legacy}
- {dataset: NTVNBDMNFe_dw_ite, variant: legacy}


pdf:
    from_: fit

use_cuts: internal

use_fitcommondata: True

theory:
    from_: fit

datacuts:
    from_: fit

theoryid:
    from_: theory

use_t0: False

dataspecs:
    - fit: 240312_mnc_dis_cct_fs2
      speclabel: "DIS, consistent"

    - fit: 240312_mnc_dis_ict_lam0_fs2
      speclabel: "DIS, inconsistent lambda = 0"

template_text: |
    
    ## Chi2 table
    {@dataspecs_chi2_table(show_total=true)@}
    
    ## Nsigma table
    {@dataspecs_nsigma_table(show_total=True)@}


actions_:
    - report(main=True)

```


validphys2/src/validphys/results.py
```diff
@@ -949,6 +949,26 @@ def fits_groups_chi2_table(
     return res
 
 
+@table
+def fits_groups_nsigma_table(
+    fits_name_with_covmat_label, fits_groups, fits_groups_chi2_data, per_point_data: bool = True
+):
```
scarlehoff, line 955: Please add docs

validphys2/src/validphys/results.py
```diff
@@ -949,6 +949,26 @@ def fits_groups_chi2_table(
...
+    df = fits_groups_chi2_table(
+        fits_name_with_covmat_label, fits_groups, fits_groups_chi2_data, per_point_data
+    )
```
scarlehoff, line 958: Can't you have this function directly as an input?

validphys2/src/validphys/results.py
```diff
@@ -949,6 +949,26 @@ def fits_groups_chi2_table(
...
+    df = df.rename(columns={'ndata': 'ndata', '$\chi^2/ndata$': 'nsigma'})
+
+    for level_0 in df.columns.levels[0]:
+        ndata_col = (level_0, 'ndata')
+        nsigma_ndata_col = (level_0, 'nsigma')
+
+        # Apply the transformation and add the new column to the DataFrame
+        df[nsigma_ndata_col] = df.apply(
+            lambda row: (row[nsigma_ndata_col] - 1) / np.sqrt(2 / row[ndata_col]), axis=1
+        )
```
scarlehoff, line 968: I think this requires more comments or explanation but probably the docstring suffices

validphys2/src/validphys/results.py
```diff
@@ -1098,6 +1165,51 @@ def fits_chi2_table(
...
+        dfs.append(pd.concat((edf.loc[lv], ddf.loc[lv]), copy=False, axis=0))
+    if show_total:
+        total_points = np.array([total_chi2_data.ndata for total_chi2_data in fits_total_chi2_data])
+        total_chi = np.array(
+            [
+                (total_chi2_data.central_result - total_chi2_data.ndata)
+                / np.sqrt(2.0 * total_chi2_data.ndata)
+                for total_chi2_data in fits_total_chi2_data
+            ]
+        )
```
scarlehoff, line 1197: If a list comprehesion needs more than a line, I think it is brtter to write the loop explictly.

In particular here you are doing twice the same loop (here and above)

validphys2/src/validphys/results.py
comane: No I don't think it works in this case as the `fits_groups_chi2_table` is not really used as a node of the graph in this case but as a normal function. See `dataspecs_groups_chi2_table` function for instance

validphys2/src/validphys/results.py
comane: Ok, I added a single for loop, let me know what you think.

validphys2/src/validphys/results.py
scarlehoff: But for `dataspecs_` the input changes, here it doesn't you should have the entire input available when you call this function. Maybe you need a "notable" version without the decorator?

comane: Ah yes, you are right. Thanks for this, I should have added it now.

validphys2/src/validphys/results.py
```diff
@@ -989,6 +1009,13 @@ def dataspecs_groups_chi2_table(
     )
 
 
+@table
+@check_speclabels_different
+def dataspecs_groups_nsigma_table(dataspecs_groups_chi2_table):
+    """Same as fits_groups_chi2_table but for an arbitrary list of dataspecs."""
```
scarlehoff, line 1015:
```suggestion
    """Same as fits_groups_nsigma_table but for an arbitrary list of dataspecs."""
```

:monocle_face:

validphys2/src/validphys/results.py
```diff
@@ -1054,6 +1100,15 @@ def dataspecs_datasets_chi2_table(
     )
 
 
+@table
+@check_speclabels_different
+def dataspecs_datasets_nsigma_table(dataspecs_datasets_chi2_table):
+    """
+    Same as fits_datasets_chi2_table but for nsigma.
```
scarlehoff, line 1107:
```suggestion
    Same as dataspecs_datasets_chi2_table but for nsigma.
```

validphys2/src/validphys/results.py
```diff
@@ -1098,6 +1153,53 @@ def fits_chi2_table(
...
+def fits_nsigma_table(
+    fits_total_chi2_data,
+    fits_datasets_nsigma_table,
+    fits_groups_nsigma_table,
+    show_total: bool = False,
+):
+    """Show the nsigma of each and number of points of each dataset and experiment
+    of each fit, where experiment is a group of datasets according to the `experiment` key in
+    the PLOTTING info file, computed with the theory corresponding to the fit. Dataset that are not
+    included in some fit appear as `NaN`
```
scarlehoff, line 1166:
```suggestion
    """Show the nsigma of each and number of points of each dataset and experiment for each fit,
    computed with the theory corresponding to the fit.
    Datasets that are not included in one of the fit appear as `NaN`
```

There are probably a lot of `PLOTTING` references that we should remove...

validphys2/src/validphys/results.py
```diff
@@ -1098,6 +1153,53 @@ def fits_chi2_table(
...
+    """
+    lvs = fits_groups_nsigma_table.index
+    # The explicit call to list is because pandas gets confused otherwise
+    expanded_index = pd.MultiIndex.from_product((list(lvs), ["Total"]))
+    edf = fits_groups_nsigma_table.set_index(expanded_index)
+    ddf = fits_datasets_nsigma_table
+    dfs = []
+    # TODO: Better way to do the merge preserving the order?
```
scarlehoff, line 1174: Why do you want a better way?

(f there's no specific reason, I'd rather remove the `TODO`, I don't really see the problem with what you are doing. If you want leave a comment like

```
# Merge the columns of this and that into a dataframe
```
in case someone wants to have a go in the future)

I'm realizing we have many `TODO` that are really not that important. I mean, there might be a `pandas.take_by_index(edf, lv)` or whatever that together with `pandas.join_magically` does what you want... but the older I get the more I appreciate simple loops and appends that can be ported to some other language or library in the future.

(I'm rambling, I did say I was getting old)

scarlehoff (approved):
All comments are changes to docstring or comments so this is good to go from my side.

I haven't tested it myself but I trust you are currently using it so it should work :P 

Given this is just a transformation of the list of `chi2` I'm not entirely sure a test is necessary, but you might want to add one to make sure nobody breaks it in some way in the future.
I think if you do the test at the level of `dataspecs_nsigma_table` that will be touching all functions.

But if you want to merge it like this (after the changes to the docstr are applied) feel free.

scarlehoff:
If you rebase on top of master the tests should pass.

validphys2/src/validphys/results.py
```diff
@@ -1098,6 +1153,52 @@ def fits_chi2_table(
...
+    computed with the theory corresponding to the fit.
+    Datasets that are not included in one of the fit appear as `NaN`
+    """
+    lvs = fits_groups_nsigma_table.index
+    # The explicit call to list is because pandas gets confused otherwise
+    expanded_index = pd.MultiIndex.from_product((list(lvs), ["Total"]))
+    edf = fits_groups_nsigma_table.set_index(expanded_index)
+    ddf = fits_datasets_nsigma_table
+    dfs = []
+    # TODO: Better way to do the merge preserving the order?
```
scarlehoff, line 1173:
```suggestion
    # Merge the columns from groups and per-dataset preserving the order coming from `fit_groups`
```

Please check that this comment is correct (if not change it accordingly).

comane: this comment was already in `fits_chi2_table`. It's there because, I mainly just copied that function

comane: If you want I can change the comment to your suggestion or remove it completely

scarlehoff: both work

PR #2133, opened 2024-07-24
Refactor boundary condition as part of the PDF object

scarlehoff:
This PR gives the fktables the decision power of what kind of PDF is to be used.

The PDF object (which can be eventually be called, `QCDObject`? or `HadronObject`) contains both the polarized and a reference to the unpolarized PDF.

This removes the necessity of writing polarized-specific code within validphys, and, as long as the max. number of convolutions is 2, it should be completely general.
It would also work for > 2, but changes would be required in the `convolutions.py` module (which at the moment does only hadronic or dis).

## TODO
- [x] Take only the central value for the boundary condition


Radonirinaunimi (approved):
Thanks again for this! LGTM!

I've performed various checks (including re-running a proper fit) and everything is consistent with before.

I've just left tiny typographical comments below.

PR #2135, opened 2024-07-25
Disable jit compilation in tf > 2.16

scarlehoff:
This is enough to run fits in GPU with TF > 2.16. At least in my systems. I can run 120 replicas (it took me a few iterations of cuda drivers to make it work).

This disables XLA _also_ in CPU (if it is active, which at the moment I think it isn't by default)
```
--tf_xla_cpu_global_jit=false           bool    Enables global JIT compilation for CPU via SessionOptions.
```

I would keep running with 2.15 since that one we know for sure it works and I've only tested in one system for now (python 3.12, TF 2.17, RTX 3070).

I don't appreciate any degradation of performance but this GPU only has 8 GB of RAM so maybe I was already bottlenecked due to the memory before.

Note that before TF 2.16, XLA compilation was disabled by default. Funny thing is, if you enable it for TF 2.15 we also see some problems (even in CPU).
I don't know whether this is a fundamental problem with XLA (and thus nothing we can do) or whether there is a problem _in our code_ that makes XLA not work.
If someone wants to investigate, probably the best thing to do is to try TF 2.15 and set `JIT_COMPILE=True`, because from 2.16 onwards it will work but it will just crash due to the memory leak.

### DIFF
--- a/n3fit/src/n3fit/backends/keras_backend/MetaModel.py
+++ b/n3fit/src/n3fit/backends/keras_backend/MetaModel.py
@@ -25,6 +25,10 @@ elif hasattr(tf_utils, "sync_to_numpy_or_python_type"):  # from TF 2.5
 else:  # in case of disaster
     _to_numpy_or_python_type = lambda ret: {k: i.numpy() for k, i in ret.items()}
 
+# Starting with TF 2.16, a memory leak in TF https://github.com/tensorflow/tensorflow/issues/64170
+# makes jit compilation unusable in GPU.
+# Before TF 2.16 it was set to `False` by default. From 2.16 onwards, it is set to `True`
+JIT_COMPILE = False
 
 # Define in this dictionary new optimizers as well as the arguments they accept
 # (with default values if needed be)
@@ -307,7 +311,7 @@ class MetaModel(Model):
                 target_output = [target_output]
             self.target_tensors = target_output
 
-        super().compile(optimizer=opt, loss=loss)
+        super().compile(optimizer=opt, loss=loss, jit_compile=JIT_COMPILE)
 
     def set_masks_to(self, names, val=0.0):
         """Set all mask value to the selected value
--- a/n3fit/src/n3fit/stopping.py
+++ b/n3fit/src/n3fit/stopping.py
@@ -27,6 +27,7 @@
     which will tell `Validation` that no validation set was found and that the training is to
     be used instead.
 """
+
 import logging
 
 import numpy as np
@@ -345,6 +346,8 @@ class Stopping:
         self._threshold_chi2 = threshold_chi2
         self._stopping_degrees = np.zeros(self._n_replicas, dtype=int)
         self._counts = np.zeros(self._n_replicas, dtype=int)
+        # Keep track of the replicas that should not be stopped yet
+        self._dont_stop_me_now = np.ones(self._n_replicas, dtype=bool)
 
         self._dont_stop = dont_stop
         self._stop_now = False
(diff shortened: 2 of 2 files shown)


scarlehoff:
Here's a report:

https://vp.nnpdf.science/8vyjWCvZRvSzu0JGbLbErQ==/

The computer used to run this fit was
Intel(R) Core(TM) i9-10900K CPU @ 3.70GHz, 16 GB of RAM
RTX 3070 8 GB

Took about 60 minutes for 120 replicas.

RoyStegeman:
Do you know why the training length distribution looks so different?

scarlehoff:
The specific pattern looks more different than it is because by chance the fits stopped in a way that made them flatter. But the fits that arrive to the last bin (which is the only relevant one) are ~25 vs ~17 

That said, for some reason and with N=2 fits in the GPU seem to produce flatter diatributions. Don't know whether there's a reason:  https://vp.nnpdf.science/WbBCvsjfQV-6ncIQ3GhCVw==/ 

or whether it is related to the change in the positivity datasets

RoyStegeman:
I should have done a cpu fit with that runcard and master at some point, but let me quickly just redo it to be sure.

RoyStegeman:
We already discussed it, but for later reference let me put here the results of the fit with the current version of the master branch but on cpu and using the same runcard as in your fit: : https://vp.nnpdf.science/C0EQblACS2qzaMbpNLuePQ==/

As expected, the TL distribution looks different, but it would be interesting to see how the TL distribution changes on GPU if the pseudodata sampling seed is changed as in both your fits this is the same.

scarlehoff:
Good eye! 

Indeed, there was a problem with the multi-replica stopping, where the replica could become active again. Here's the same fit with this corrected:

https://vp.nnpdf.science/RWnazARaSb6TbKMtkxONGA==

It doesn't seem to have an effect on the fit (in this report even the seed for the pseudodata is also different) but good that it has been caught :)

@Cmurilochem in the same manner, it should not impact your hyperopt runs, but if you need to run new ones, better if you use this branch I guess

scarlehoff:
For good measure, a fit using the same seeds as the first one https://vp.nnpdf.science/CyjXLYyARtKfWK1iUnJrYg==/

This is ready to merge

RoyStegeman (commented):
Also with the same runcard as your first (`240726-jcm-004` wasn't uploaded so couldn't compare to that) https://vp.nnpdf.science/rr476H7nT9GjOb6WBasw7Q==/

Took 2:15h and 25 gb on a V100

Anyway, looks good to me

RoyStegeman (approved):

scarlehoff:
> Also with the same runcard as your first (240726-jcm-004 wasn't uploaded so couldn't compare to that)

they are the same

goord:
From the XLA docs:

All operations must have inferrable shapes
XLA needs to be able to infer the shapes for all of operations it compiles given the inputs to the computation. So a model function that produces a Tensor with an [unpredictable shape](https://gist.github.com/jlebar/76e73ca2626231bd588cb03ec5f74fa3) will fail with an error when run. (In this example, the shape of the output of tf.expand_dims depends on random_dim_size which cannot be inferred given x, y and z.)

Note that because XLA is a JIT compiler, the shapes can vary across runs, as long as they can be inferred given the inputs to the cluster. So [this example](https://gist.github.com/jlebar/a0ff40f378f2fad99d059a2ab33e68da) is fine.

----
Could it be that the masking layers give rise to 'unpredictable shape' tensors?

scarlehoff:
Then it should crash at compilation time (so either way it is a bug in their side). If, for fun, you want to look further into this I suggest this as a starting point: https://github.com/NNPDF/nnpdf/pull/2137
a code that is at the same time compatible with pytorch and tensorflow is the most explicit thing you can get (and actually, to make it compatible with pytorch I had to add the  output shape in a few places where tensorflow as able to infer them but pytorch wasn't).

Issue #2134, opened 2024-07-25
Move from `tensorflow.keras` to `keras`


scarlehoff:
This will make the code more general. If we can remove the tensorflow dependency as much as possible (it would be impossible right now to remove it completely) that would be great (it means we could test pytorch for instance).

At the moment (tf 2.17) `tensorflow.keras` is still provided, but it is not a given that it will continue like that and in particular the conda forge package for macos (maybe a bug on their side, I'm not sure) seems to be missing `tensorflow.keras`.

RoyStegeman:
There is no guarantee either way, some years ago this was their position: https://github.com/keras-team/keras/releases/tag/2.3.0

I'd argue that we should keep using `tensorflow.keras` for possible benefits from better integration with tensorflow. If someone is serious about providing a pytorch backend in `n3fi`, we can reconsider and in that case swapping out `tensorflow.keras` for `keras` should be a quick job.

scarlehoff:
> There is no guarantee either way, some years ago this was their position

I think a change of position in 5 years is still more stable than other programs we use

I've needed to add this already.
https://github.com/NNPDF/nnpdf/pull/2110/commits/042d9950889122b729e6e54552a1aa2087ff7c1c
and, more generally, I think it is preferable to stick with keras which is actively being developed, that makes sure that the features we use are not deprecated (as opposed to using some `tf.keras` that hasn't been removed yet when it should).

RoyStegeman:
Conda packages can always be a mess, separating keras and tensorflow may also result in problems with compatibility between versions. 

Feel free to do this if you want to change it, but I don't think that at this point we can say that separating them is clearly the better choice.

scarlehoff:
> separating keras and tensorflow may also result in problems with compatibility between versions.

No. This doesn't generate (new) problems because most of the calls are already using the separate package. If you do `from tensorflow import keras` you are already getting the separated keras package.

But we risk using features that don't actually exist in the separated `keras` package (or that have changed name)
https://github.com/tensorflow/tensorflow/tree/master/tensorflow/python/keras#stop
e.g. https://github.com/NNPDF/nnpdf/blob/4d7284455f4080de0c953350ca742f1830e2b511/n3fit/src/n3fit/backends/keras_backend/MetaModel.py#L24

PR #2137, opened 2024-07-26
Make the code compatible with other keras backends

scarlehoff:
This deals with #2134 

It exchanges tf imports to keras (with small changes when the functions signatures change). It was necessary to add a few explicit output shapes here and there so that Keras knows how to compile the graph (and then we can also use the `jit_compile` flag, at least with the tensorflow backend).

Now, for some comparisons:

1. In this report master is compared to this branch https://vp.nnpdf.science/ZGmx7ro7SNCESqFEmR2IcA==, TensorFlow backend and 70 replicas running in GPU (A6000) (53 min vs 43 min)
2. Tensorflow Vs Pytorch (both running in this branch and GPU) https://vp.nnpdf.science/zKyfSvs8TxSI1FXGR72f3w== (43 min vs 178 min)

------ 

Turns out that the code is mostly general enough to work with pytorch ootb. Very few changes are needed* (the second commit, and not even, because some of the changes are `tf.` that I missed).

However, I've run a few quick benchmarks and turns out that it is about 4 times slower than the tensorflow backend. This is not surprising since we have been using tensorflow for a long time so we might be doing thing that just happen to be better for tensorflow. Also, while jit_compile is set to auto, I think the default for pytorch is False.

### DIFF
--- a/doc/sphinx/source/n3fit/methodology.rst
+++ b/doc/sphinx/source/n3fit/methodology.rst
@@ -8,8 +8,8 @@ different in comparison to the latest NNPDF (i.e. `NNPDF3.1 <https://arxiv.org/a
 methodology.
 
 .. warning::
-   The default implementation of the concepts presented here are implemented with Keras and
-   Tensorflow. The ``n3fit`` code inherits its features, so in this document we avoid the discussion of
+   The default implementation of the concepts presented here are implemented with Keras.
+   The ``n3fit`` code inherits its features, so in this document we avoid the discussion of
    specific details which can be found in the `Keras documentation <https://keras.io/>`_.
 
 .. note::
@@ -90,7 +90,7 @@ random numbers used in training-validation, ``nnseed`` for the neural network in
 Neural network architecture
 ---------------------------
 
-The main advantage of using a modern deep learning backend such as Keras/Tensorflow consists in the
+The main advantage of using a modern deep learning backend such as Keras consists in the
 possibility to change the neural network architecture quickly as the developer is not forced to fine
 tune the code in order to achieve efficient memory management and PDF convolution performance.
 
@@ -132,41 +132,36 @@ See the `Keras documentation <https://www.tensorflow.org/api_docs/python/tf/kera
 
 .. code-block:: python
 
-   from tensorflow.keras.utils import plot_model
-   from n3fit.model_gen import pdfNN_layer_generator
-   from validphys.api import API
+   from keras.utils import plot_model
+    from n3fit.model_gen import pdfNN_layer_generator
+    from validphys.api import API
 
-   fit_info = API.fit(fit="NNPDF40_nnlo_as_01180_1000").as_input()
-   basis_info = fit_info["fitting"]["basis"]
+    fit_info = API.fit(fit="NNPDF40_nnlo_as_01180_1000").as_input()
+    basis_info = fit_info["fitting"]["basis"]
 
-   pdf_models = pdfNN_layer_generator(
-       nodes=[25, 20, 8],
-       activations=["tanh", "tanh", "linear"],
-       initializer_name="glorot_normal",
-       layer_type="dense",
-       flav_info=basis_info,
-       fitbasis="EVOL",
-       out=14,
-       seed=42,
-       dropout=0.0,
-       regularizer=None,
-       regularizer_args=None,
-       impose_sumrule="All",
-       scaler=None,
-       parallel_models=1,
-   )
+    pdf_model = pdfNN_layer_generator(
+        nodes=[25, 20, 8],
+        activations=["tanh", "tanh", "linear"],
+        initializer_name="glorot_normal",
+        layer_type="dense",
+        flav_info=basis_info,
+        fitbasis="EVOL",
+        out=14,
+        seed=42,
+        dropout=0.0,
+        regularizer=None,
+        regularizer_args=None,
+        impose_sumrule="All",
+        scaler=None,
+    )
 
-   pdf_model = pdf_models[0]
-   nn_model = pdf_model.get_layer("NN_0")
-   msr_model = pdf_model.get_layer("impose_msr")
-   models_to_plot = {
-           'plot_pdf': pdf_model,
-           'plot_nn': nn_model,
-           'plot_msr': msr_model
-           }
-
-   for name, model in models_to_plot.items():
-       plot_model(model, to_file=f"./{name}.png", show_shapes=True)
+    nn_model = pdf_model.get_layer("pdf_input")
+    msr_model = pdf_model.get_layer("impose_msr")
+    models_to_plot = {
+            'plot_pdf': pdf_model,
+            'plot_nn': nn_model,
+            'plot_msr': msr_model
+            }
 
 
 This will produce for instance the plot of the PDF model below, and can also be used to plot the
--- a/doc/sphinx/source/n3fit/methodology.rst
+++ b/doc/sphinx/source/n3fit/methodology.rst
@@ -132,7 +132,7 @@ See the `Keras documentation <https://www.tensorflow.org/api_docs/python/tf/kera
 
 .. code-block:: python
 
-   from keras.utils import plot_model
+    from keras.utils import plot_model
     from n3fit.model_gen import pdfNN_layer_generator
     from validphys.api import API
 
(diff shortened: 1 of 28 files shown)


scarlehoff:
This makes the code:

1. Keras-only (so no more tensorflow functions in the middle)
2. Compilable wtih xla (the fit is very much not the bottleneck anymore, but for reference, running in GPU it goes from 20 replicas in 23 minutes to 20 replicas in 18 minutes)
3. Compatible with `pytorch` in both cpu and gpu (_however_ running pytorch in graph mode crashed)
4. Almost compatible with jax. For jax there are 2 problems: a) the preprocessing layer b) the computation of the losses. a) is fixable but b) is too much work imho so I won't do anything.

n3fit/src/n3fit/backends/keras_backend/operations.py
```diff
@@ -216,57 +215,39 @@ def tensor_ones_like(*args, **kwargs):
...
 # modify properties of the tensor like the shape or elements it has
-@tf.function
-def flatten(x):
-    """Flatten tensor x"""
-    return tf.reshape(x, (-1,))
 
 
-@tf.function
 def reshape(x, shape):
     """reshape tensor x"""
```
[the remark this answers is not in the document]
scarlehoff: Yes, I should clean the multitude of functions that have become just a 1-to-1 wrapper as named imports...

I'll do that as part od this PR

n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -121,7 +108,7 @@ def __init__(self, input_tensors, output_tensors, scaler=None, input_values=None
         self.compute_losses_function = None
         self._scaler = scaler
 
-    @tf.autograph.experimental.do_not_convert
+    # @tf.autograph.experimental.do_not_convert
```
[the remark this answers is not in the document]
scarlehoff: I'm not  sure yet about this one. It was introduced here https://github.com/NNPDF/nnpdf/commit/2ac98976d42453fed0b6ecec5da8939ec5d862c7 and I'm not sure why (that function is also -one of- the part that is meesing with jax).

If I get 1-to-1 correspondance running in parallel with and without that I'll remove it.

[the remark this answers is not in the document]
scarlehoff: seems to make no difference in master

scarlehoff:
Ok, this is ready for review. Some comparisons in the first post.

I've heard you are completely free this weekend @RoyStegeman in case you want to have another look :P

n3fit/src/n3fit/backends/keras_backend/internal_state.py
```diff
@@ -64,8 +82,7 @@ def set_number_of_cores(max_cores=None, max_threads=None):
 
     log.info("Setting the number of cores to: %d", cores)
     try:
-        tf.config.threading.set_inter_op_parallelism_threads(threads)
-        tf.config.threading.set_intra_op_parallelism_threads(cores)
+        set_threading(threads, cores)
     except RuntimeError:
         # If pdfflow is being used, tensorflow will already be initialized by pdfflow
         # maybe it would be good to drop completely pdfflow before starting the fit? (TODO ?)
```
RoyStegeman, line 88: What causes the runtimeerror when I don't have tensorflow but use torch instead? The log.warning below seems incorrect

n3fit/src/n3fit/backends/keras_backend/callbacks.py
```diff
@@ -15,9 +15,11 @@
 import logging
 from time import time
 
+# Callbacks need tensorflow installed even if the backend is pytorch
```
RoyStegeman, line 18: Really? :| 

It seems keras defaults to the tf backend so there should either be a toggle to set the backend from the runcard and/or and explanation in the docs saying to do `KERAS_BACKEND=torch n3fit`

n3fit/src/n3fit/backends/keras_backend/internal_state.py
```diff
@@ -13,20 +14,37 @@
...
-    tf.config.run_functions_eagerly(flag)
+    def set_eager(flag=True):
+        """Pytorch is eager by default"""
+        pass
+
+    def set_threading(threads, core):
+        """Not implemented"""
+        pass
+
+elif K.backend() == "tensorflow":
```
RoyStegeman, line 34: Should there be an error for `else`, i.e. if the backend is somehow something else?

n3fit/src/n3fit/backends/keras_backend/operations.py
```diff
@@ -18,37 +16,66 @@
     layer generation:
         Instanciate a layer to be applied by the calling function
 
-    Some of these are just aliases to the backend (tensorflow or Keras) operations
+    Most of the operations in this module are just aliases to the backend
+    (Keras in this case) so that, when implementing new backends, it is clear
+    which operations may needd to be overwritten.
```
RoyStegeman, line 21:
```suggestion
    which operations may need to be overwritten.
```

n3fit/src/n3fit/model_gen.py
```diff
@@ -99,13 +99,10 @@ def _generate_experimental_layer(self, pdf):
         the input PDF is evaluated in all points that the experiment needs and needs to be split
         """
         if len(self.dataset_xsizes) > 1:
-            splitting_layer = op.as_layer(
-                op.split,
-                op_args=[self.dataset_xsizes],
-                op_kwargs={"axis": 2},
-                name=f"{self.name}_split",
+            sp_layer = op.tensor_splitter(
```
RoyStegeman, line 102: I think splitting was more informative than sp

RoyStegeman (commented):
Thanks, looks good. Main point is that it should be documented

n3fit/src/n3fit/backends/keras_backend/internal_state.py
scarlehoff: There should be no runtime error in that case (partly because with torch I'm not setting the threads, but also because there will no conflict).

The warning is written really badly but it is not incorrect I think?

pyproject.toml
```diff
@@ -73,6 +73,7 @@ reportengine = { git = "https://github.com/NNPDF/reportengine" }
 # Fit
 psutil = "*"
 tensorflow = "*"
```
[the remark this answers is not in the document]
scarlehoff: I think with pip I was able to run with 2.15, but I'll retry with an empty environment to be safe.

n3fit/src/n3fit/backends/keras_backend/internal_state.py
scarlehoff: I can put an `else` for completeness, but Keras will just fail way before arriving here.

RoyStegeman: Okay it might be worth to put a comment explaining what happens if neither condition is met, but I'll leave it up to you

n3fit/src/n3fit/backends/keras_backend/callbacks.py
scarlehoff: This is because of tensorboard. But I'll just wrap it with try-except

pyproject.toml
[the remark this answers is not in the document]
scarlehoff: Yes. It is important that Keras is `^3.1` but tensorflow can be whatever is compatible with it.

doc/sphinx/source/n3fit/methodology.rst
```diff
@@ -132,41 +132,36 @@ See the `Keras documentation <https://www.tensorflow.org/api_docs/python/tf/kera
...
-   msr_model = pdf_model.get_layer("impose_msr")
-   models_to_plot = {
-           'plot_pdf': pdf_model,
-           'plot_nn': nn_model,
-           'plot_msr': msr_model
-           }
-
-   for name, model in models_to_plot.items():
-       plot_model(model, to_file=f"./{name}.png", show_shapes=True)
+   from keras.utils import plot_model
```
RoyStegeman, line 135:
```suggestion
    from keras.utils import plot_model
```

doc/sphinx/source/tutorials/run-fit.rst
```diff
@@ -184,25 +184,26 @@ After obtaining the fit you can proceed with the fit upload and analisis by:
...
+and it is tested to run with the `Tensorflow <https://www.tensorflow.org/>`_
+and `pytorch <https://pytorch.org>`_ backends.
+This also means that anything that make any of these packages faster will also
+make ``n3fit`` faster.
+Note that at the time of writing, ``TensorFlow`` is approximately 4 times faster than ``pytorch``.
+
+The default backend for ``keras`` is ``tensorflow``.
+In order to change the backend, the environment variable ``KERAS_BACKENDD`` need to be set (e.g., ``KERAS_BACKEND=torch``).
+
+The best results are obtained with ``tensorflos[and-cuda]`` installed from pip.
```
RoyStegeman, line 197:
```suggestion
The best results are obtained with ``tensorflow[and-cuda]`` installed from pip.
```

RoyStegeman (approved):

PR #2138, opened 2024-07-31
Saving pseudodata of replicas whit multiple fits

achiefa:
When using parallel models, pseudodata are not saved for each replica, resulting in the following error when running multiple fits with n3fit.
```shell
Cannot request that multiple replicas are fitted and that pseudodata is saved. Either set `fitting::savepseudodata` to `false` or fit replicas one at a time.
```

This branch ensures that pseudodata for each replica are saved, even when `parallel_models=true` is set.

### DIFF
--- a/n3fit/src/n3fit/scripts/n3fit_exec.py
+++ b/n3fit/src/n3fit/scripts/n3fit_exec.py
@@ -140,8 +140,8 @@ class N3FitConfig(Config):
                 #    "to `false` or fit replicas one at a time."
                 #)
                 # take same namespace configuration on the pseudodata_table action.
-            training_action = namespace + "training_pseudodata"
-            validation_action = namespace + "validation_pseudodata"
+            training_action = namespace + "replicas_training_pseudodata"
+            validation_action = namespace + "replicas_validation_pseudodata"
 
             N3FIT_FIXED_CONFIG['actions_'].extend((training_action, validation_action))
 
--- a/n3fit/src/n3fit/scripts/n3fit_exec.py
+++ b/n3fit/src/n3fit/scripts/n3fit_exec.py
@@ -11,7 +11,7 @@ import shutil
 import sys
 import warnings
 
-from reportengine import colors
+from reportengine import colors, collect
 from reportengine.compat import yaml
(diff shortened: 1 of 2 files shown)


scarlehoff:
Do you actually need this? We disabled this for multiple replicas due to issues with reproducibility.

And the (main) issue is not that easy to solve, namely that in parallel replicas datasets with a single point must _all_ enter training or validation. So if you generate the data in parallel then you cannot reproduce it with the vp functions to do so.

If you _need it_ we need to find a way to tag the data as having been generated in parallel.

RoyStegeman:
Thanks for starting this. I actually meant doing something along the lines of 
`replicas_training_pseudodata = collect("training_pseudodata", ("replicas",))` in n3fit_data.py. 

I suspect this will save the pseudodata all in the same folder since somehow the output folder is probably the nnfit/replica_1 folder, using the `table_folder` from `N3FitEnvironment` but perhaps there is a way to solve this.

RoyStegeman:
> And the (main) issue is not that easy to solve, namely that in parallel replicas datasets with a single point must all enter training or validation. So if you generate the data in parallel then you cannot reproduce it with the vp functions to do so.

This is easy to solve by always including datasets with a single point in the training data. 

But indeed, we do need it because otherwise doing parameter determinations with CRM will be sketchy and with TCM will be impossible using GPU.

scarlehoff:
> This is easy to solve by always including datasets with a single point in the training data.

I would not do that. I'd rather go for the tag.

In any case, before continuing, could you check that the sequential and parallel fit (without taking into account the 1 point datasets) produce exactly the same pseudodata and trvl masks? (if that works we can think about the rest)

RoyStegeman:
> I would not do that. I'd rather go for the tag.

Why? If we're serious about using GPU as well as CPU fits, then I'd say they should be as close as possible in behaviour, no? The choice of what to do with those datasets was a bit arbitrary and has been changed over time anyway.

scarlehoff:
Then we should go for an actual solution instead of changing the behaviour every time it becomes an inconvenience.

scarlehoff:
I've moved the commits from this branch to #2276

PR #2139, opened 2024-08-01
Add NNPDFpol2.0 hyperopt card

Radonirinaunimi:
The distribution of datasets into the following folds seems homogeneous. As discussed, the JLAB datasets are always fitted and not part of the folds.

Here is a [report link](https://vp.nnpdf.science/rK01O61yQ4ep9uy2bJfHeg==/#testing-the-different-folds) containing comparisons of the baseline NNLO with fits in which datasets from each fold are removed in turn. For all the folds, the pPDFs are perfectly consistent except for $\Delta V$ and $\Delta V3$ for which there are small differences (consistent with uncertainties).

### DIFF
--- a/n3fit/runcards/examples/nnpdfpol20_hyperopt.yml
+++ b/n3fit/runcards/examples/nnpdfpol20_hyperopt.yml
@@ -1,0 +1,240 @@
+############################################################
+# Configuration file for N3FIT Polarized Fits hyperopt scan
+############################################################
+description: NNPDFpol2.0 NNLO Hyperoptimization Card
+############################################################
+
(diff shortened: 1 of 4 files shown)


Radonirinaunimi:
The only we might want to change is the training fraction. Metric- and pPDF-wise, `0.60` and `0.75` have little to no difference ([report](https://vp.nnpdf.science/rK01O61yQ4ep9uy2bJfHeg==/#testing-different-training-fraction)). Perhaps the only reason to prefer `0.60` over `0.75` is the training distribution.

scarlehoff (approved):

PR #2140, opened 2024-08-05
Only return lhapdf data paths that do exist

scarlehoff:
I haven't modified the behaviour of starting from the end.

Use case (or situation in which this is a problem): if you happen to use some shared system with lhapdf data paths you don't have access to (like in some lxplus views).

### DIFF
--- a/validphys2/src/validphys/lhaindex.py
+++ b/validphys2/src/validphys/lhaindex.py
@@ -133,6 +133,10 @@ def parse_info(name):
 def get_lha_datapath():
     """Return an existing datapath from LHAPDF, starting from the end.
     If no path is found to exist, recover the old behaviour and returns the last path.
+
+    The check for existence intends to solve problems where a previously filled `LHAPATH`
+    or `LHAPDF_DATA_PATH` environment variable is pointing to a non-existent path or shared
+    systems where LHAPDF might be compiled with hard-coded paths not available to all users.
     """
     for lhapath in lhapdf.paths()[::-1]:
         if Path(lhapath).exists():
--- a/validphys2/src/validphys/lhaindex.py
+++ b/validphys2/src/validphys/lhaindex.py
@@ -10,6 +10,7 @@ from functools import lru_cache
 import glob
 import os
 import os.path as osp
+from pathlib import Path
 import re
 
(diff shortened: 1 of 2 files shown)


validphys2/src/validphys/lhaindex.py
```diff
@@ -130,6 +131,12 @@ def parse_info(name):
 
 
 def get_lha_datapath():
+    """Return an existing datapath from LHAPDF, starting from the end.
+    If no path is found to exist, recover the old behaviour and returns the last path.
+    """
+    for lhapath in lhapdf.paths()[::-1]:
+        if Path(lhapath).exists():
```
RoyStegeman, line 138: Maybe add the usecase here, because it's not so obvious

scarlehoff: This is my particular usecase, but tbh what is the point of returning paths that don't exist?

RoyStegeman: It's not so obvious how you could have an lhapdf path that doesn't exist without it being a bigger problem (i.e. meaning the user accidentally removed the LHAPDF folder instead of your usecase).

Having sad that, I agree with your points,but then should we just raise an error instead of recovering the old behaviour?

scarlehoff: The thing is I don't know why it was done in this way and both lhapdf and this module sometimes do things I don't understand so...

When it is a problem vp will fail anyway 🤷‍♂️

RoyStegeman: Okay so in that case, could you please clarify the usecase in which you would have non-existent lhapdf paths without that being a problem the user would want to be aware of? 

I think just adding this if-statement without that would only make it more black magic

scarlehoff: I've completed the docstr with some situations in which the check is useful.
I could not find out why do we start counting from the end.

RoyStegeman: Thanks!

scarlehoff:
Is this ok to be merged?

RoyStegeman (approved):
Thanks

PR #2141, opened 2024-08-09
Skip the iterations argument if not given

scarlehoff:
At the moment when using an `EXA` theory the number of iterations is taken from the theory card but if the user gives an explicit value to the cli, that takes precedence.

The problem is that not giving any value at all was also taking precedence. This PR fixes that.


andreab1997 (approved):
Thanks for fixing this, it looks good to me

PR #2144, opened 2024-08-13
Complete the list of experiments in docs

scarlehoff:
When linking the dataset name convention to @giacomomagni the other day, I realized we left most experiment as TODO.
I don't think we need to have a wiki page for all of this, but at least a few words would be nice.

Please feel free to push directly to this branch to modify any of the descriptions or add some that are missing

cc @Radonirinaunimi @giacomomagni @cschwan 

This is just to remove the ugly TODOs.

Btw, I was unable to find what Chorus stand for.
And I didn't add the meaning of BCDMS because I don't think having a list of cities is very useful.


RoyStegeman:
> Btw, I was unable to find what Chorus stand for.

CERN Hybrid Oscillation Research Apparatus

PR #2146, opened 2024-08-16
Fix for old theories

scarlehoff:
After the changes for polarized PDFs old theories were no longer working because they are missing some of the pineappl info that we are now expecting.

Now... I'm thinking of dropping support for old theories altogether once we drop support for the old data (which will happen in about a month) because it is a lot of extra work for something nobody seems to care about, but for now let's keep it. I've added a test so that we are checking for now that the old theories do run.

### DIFF
--- a/n3fit/src/n3fit/tests/regressions/quickcard_old.yml
+++ b/n3fit/src/n3fit/tests/regressions/quickcard_old.yml
@@ -1,6 +1,6 @@
 #
 # Configuration file for n3fit regression tests
-# This runcard includes two DIS datasets, one Hadronic dataset
+# This runcard includes one DIS dataset, one Hadronic dataset
 # and two positivity datasets
 #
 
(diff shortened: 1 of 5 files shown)


validphys2/src/validphys/coredata.py
```diff
@@ -48,6 +48,11 @@ class FKTableData:
         ``xgrid`` indicating the points in ``x`` where the PDF should be
         evaluated.
 
+    convolution_types: tuple[str]
+        The type of convolution that the FkTable is expecting for each of the
+        functions to be convolved with (usually the two types of PDF from the two
+        incoming particles).
```
Radonirinaunimi, line 54:
```suggestion
        incoming hadrons).
```

n3fit/src/n3fit/tests/regressions/quickcard_old.yml
```diff
@@ -0,0 +1,72 @@
+#
+# Configuration file for n3fit regression tests
+# This runcard includes two DIS datasets, one Hadronic dataset
```
Radonirinaunimi, line 3:
```suggestion
# This runcard includes one DIS dataset, one Hadronic dataset
```

Radonirinaunimi (approved):
Good to me! Thanks.

Fixing only some typos below.

PR #2147, opened 2024-08-20
Compose variants as a tuple

### LINKED ISSUE #1964: Accept variant as a list
The reason why this is going to be necessary is that for the pheno paper many grids are being recomputed and they are not necessarily in the same structure as before (e.g., they might be normalized differently, or even be a different number of grids -like for jets-).

Therefore it will be necessary to be able to say `variant: ["legacy_uncertainties", "legacy_theory"]` so that both the uncertainties and the theory are updated.

That said, if people already (even if the theories are different and the yamldb repeated in many places) would use for the new grids (when they are replacing a new one) the same names that were used in Theory 708 / Theory 700 that will make it more likely that they work.

(and remember, we are separating Z0, WP, WM! There are still grids here and there with all three of them concatenated but they should be considered deprecated)

@achiefa @niclaurenti @andreab1997 @cschwan @t7phy @comane

scarlehoff:
This might be necessary going forward to avoid having "legacy", "legacy_with_old_theory", etc.

For now just a draft since it needs tests and if it ends up being necessary only for a bunch of datasets it might be avoidable... I'd like to drop all theories below 700 tbh.

e.g.,

```yaml
dataset_inputs:
    - {dataset: ATLAS_JETS, variant=("legacy", "old_theory")}
```
Closes #1964 

<details>

<summary> Script to update format of all datafiles </summary>

```python
from ruamel.yaml import YAML
from glob import glob

yaml = YAML()
yaml.preserve_quotes = True
yaml.width = 4096

for metadata_file in glob("*/metadata.yaml"):
    with open(metadata_file, "r") as f:
        metadata = yaml.load(f)

    with open(metadata_file, "w") as f:
        yaml.dump(metadata, f)
```

</details>

<details>

<summary> Script to separate legacy into legacy_theory and legacy_data </summary>

```python
from ruamel.yaml import YAML
from glob import glob
from pathlib import Path

yaml = YAML()
yaml.preserve_quotes = True
yaml.width = 4096

changed_datasets = []

for metadata_file in glob("*/metadata.yaml"):
    with open(metadata_file, "r") as f:
        metadata = yaml.load(f)

    for observable in metadata["implemented_observables"]:
        variants = observable.get("variants", {})
        legacy = variants.get("legacy", {})
        if "theory" in legacy:
            legacy_data = dict(legacy)
            legacy_theory = {"theory": legacy_data.pop("theory")}

            # Not all will have both legacy data and theory, in those cases skip
            if not legacy_data:
                continue
            variants["legacy_data"] = legacy_data
            variants["legacy_theory"] = legacy_theory

            changed_datasets.append(
                f"{Path(metadata_file).parent.name}_{observable['observable_name']}"
            )

    with open(metadata_file, "w") as f:
        yaml.dump(metadata, f)

print("\n".join(changed_datasets))
```

</details>

List of changed datasets:

```
ATLAS_TTBAR_13TEV_TOT_X-SEC
ATLAS_TTBAR_8TEV_TOT_X-SEC
CMS_TTBAR_7TEV_TOT_X-SEC
CMS_TTBAR_13TEV_TOT_X-SEC
CMS_TTBAR_8TEV_2L_DIF_MTTBAR-YT-NORM
ATLAS_1JET_8TEV_R06_PTY
CMS_TTBAR_8TEV_TOT_X-SEC
ATLAS_TTBAR_7TEV_TOT_X-SEC
CMS_1JET_8TEV_PTY
ATLAS_TTBAR_8TEV_LJ_DIF_YT-NORM
ATLAS_TTBAR_8TEV_LJ_DIF_YTTBAR-NORM
```


RoyStegeman:
Just to be sure I understand. To use new implementation with old fktables we'll have to go through all the metadata.yaml files and separate the `legacy` variant into `data_legacy` and `theory_legacy` variants?

scarlehoff:
I've separated the dataset including "theory" in "legacy" into "legacy_data" and "legacy_theory" and added a test that legacy_data + legacvy_theory == legacy.

The scripts that results in the first and third commit (update of all .yaml files + update of files including theory) are in the OP so if necessary they can be rerun without much trouble.

RoyStegeman:
One thing to keep into account is that some datasets use legacy_decorrelated (ATLAS_1JET_8TEV_R06_PTY) or legacy_10

scarlehoff:
In principle together with theory only `legacy_decorrelated` might exist since it is only the jets which have changed, so I'd say that one can be done by hand.

RoyStegeman:
Used in this fit: https://vp.nnpdf.science/ronNrRZ2QJ-eG8Da-sla4Q==/

RoyStegeman (approved):
Thanks!

PR #2153, opened 2024-09-16
Remove old commondata

scarlehoff:
This PR removes **all** old commondata from the repository.

All other changes are updates to tests and examples so that only the new commondata is being used. I have removed the old commondata parser from `loader.py` but I've left the routines and, crucially, the conversion of the filtered old data to new data. That I will remove after tagging so that there is a tag that can be used to convert data.

cc @RoyStegeman @enocera @Radonirinaunimi mostly for information, I don't expect a review (~6000 files changed). I think it is preferrable to have bugs in master for a few weeks rather than risking any new project to start with some old data file.

For reference, this is the last commit with the old commondata available e01e21ec40ed5b0dd65c2252ea47ab0361023bab


RoyStegeman:
Why does this commit also add 300,000+ lines?

scarlehoff:
I changed `new_commondata` -> `commondata`

RoyStegeman:
Ah of course!

scarlehoff:
A comparison with two fits, one from 2021 and one from this year https://vp.nnpdf.science/c4uR9NVDRS2SgsJqdN3S2g==/
At least the 4.0 dataset can be used without a problem with this branch. Merging now, brace for impact.

PR #2157, opened 2024-09-22
Accept more than one mapping per dataset in dataset_names.yml

scarlehoff:
At the moment only one mapping can be used per dataset in `dataset_names.yml` but, in principle, several old dataset names could refer to the same new dataset (it is strange, but it _could_ happen).

In particular, it is like this for `ATLAS_WP_JET_8TEV_PT` because that name was wrong at some point and since fits like e.g., `240701-02-rs-nnpdf40-baseline`  used it we added it to `dataset_names.yml`. This PR fixes that by letting `legacy_names` be a list.

In addition, now `legacy_names` is only used when something fails (e.g., filters not found).

Since I was at it, I also removed a bunch of `.legacy` that won't work anymore anyway because I removed all `.dat` files.

@Radonirinaunimi this would fix the comparefit you were doing for instance. Hopefully it doesn't break anything in the way.

Btw: we should add also closure tests to the extra_tests that act as regeressions because they are not fully tested...

### DIFF
--- a/validphys2/src/validphys/filters.py
+++ b/validphys2/src/validphys/filters.py
@@ -645,11 +645,14 @@ class Rule:
         # is different to the case where the rule does apply,
         # but the point was cut out by the rule.
         if (
-            dataset.setname != self.dataset
+            not (
+                dataset.setname == self.dataset
+                or
+                # for old rules, we might be using an old name of this dataset
+                (dataset.legacy_names is not None and self.dataset in dataset.legacy_names)
+            )
             and process_name != self.process_type
             and self.process_type != "DIS_ALL"
-            # for old rules, we might be using an old name of this dataset
-            and self.dataset not in dataset.legacy_names
         ):
             return None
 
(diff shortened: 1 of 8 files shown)


validphys2/src/validphys/filters.py
```diff
@@ -645,7 +645,12 @@ def __call__(self, dataset, idat):
         # is different to the case where the rule does apply,
         # but the point was cut out by the rule.
         if (
-            (dataset.setname != self.dataset and dataset.legacy_name != self.dataset)
+            not (
+                dataset.setname == self.dataset
+                or
+                # for old rules, we might be using an old name of this dataset
+                (dataset.legacy_names is not None and self.dataset in dataset.legacy_names)
+            )
```
Radonirinaunimi, line 653: Wouldn't it better to split these into multiple if statements? Having so many conditions in one single statement is a bit too much and (surely) breaks some PEP8 guides.

Radonirinaunimi (commented):
Currently still looking into it 😢

PR #2158, opened 2024-09-25
Polish polarized fit bases

giacomomagni:
This PR addresses the following:
- [X] enforce $s=\bar{s}$ by setting $V=V_8$ as agreed in Morimondo
- [X] extend polarized flavour basis and use correct labels for plotting
- [x] add more (momentum) sum rules to the report


Radonirinaunimi (approved):
All the polishing related to the bases are done, so I'd say this can now be merged.

PR #2165, opened 2024-10-04
Update feature scaling

RoyStegeman:
Some modifications to the feature scaling, in particular it makes sure that the input passed tot he photon computation is the correct one. This is because with feature scaling we pass both the scale and non-scaled input to the model, where the non-scaled input is used in the computation of sum rules and integrability.

Some reports of fits with this branch are:
- baseline: https://vp.nnpdf.science/djKZSQN-SXCJATxTw1_-MA==
- QED: https://vp.nnpdf.science/pzcyQi_TQZWMMecQtNEycQ==/
- MHOU: https://vp.nnpdf.science/vA12B_xXQvag33OIeV22QQ==/
- N3LO + MHOU + QED: https://vp.nnpdf.science/FgGpfSP8SOyqV4xIycO-yg==/


n3fit/src/n3fit/scaler.py
```diff
@@ -27,43 +27,54 @@ def generate_scaler(
...
 
     # Define an evenly spaced grid in the domain [0,1]
-    # force_set_smallest is used to make sure the smallest point included in the scaling is
-    # 1e-9, to prevent trouble when saving it to the LHAPDF grid
+    # force_set_smallest is used to make sure the smallest point included in the scaling is 1e-9, to
+    # prevent trouble when saving it to the LHAPDF grid
     force_set_smallest = input_arr.min() > 1e-9
+    include_endpoint = (
+        1.0 in input_arr
+    )  # if 1.0 is in the xgrid it should also be 1.0 in the output xgrid
```
scarlehoff, line 35: Completely irrelevant, but could you put the comment above the line instead?

n3fit/src/n3fit/scaler.py
```diff
@@ -27,43 +27,54 @@ def generate_scaler(
...
-    map_from = map_from_complete[selected_points]
-    map_from = np.log(map_from)
-    map_to = map_to_complete[selected_points]
+    selected_points += [1]  # add also this one since 1e-9 is just an outlier
 
+    # make a mask of which pints to keep
+    mask = np.zeros(len(map_from), dtype=bool)
+    mask[selected_points] = True
+
+    # apply the mask and lot the input
```
scarlehoff, line 70:
```suggestion
    # apply the mask and log the input
```

extra_tests/regression_fits/feature_scaling_81.json
```diff
@@ -51,45 +51,45 @@
   ],
   "stop_epoch": 1100,
   "best_epoch": 1099,
-  "erf_tr": 2.6529221534729004,
-  "erf_vl": 2.1242189407348633,
-  "chi2": 1.4108729362487793,
+  "erf_tr": 2.803447723388672,
+  "erf_vl": 2.2027039527893066,
+  "chi2": 1.5391007661819458,
```
scarlehoff, line 56: so the mapping has changed?

RoyStegeman: Yes slightly, I think it's because now the x=1 is usually included in the scaled xgrid where before it was not

scarlehoff: and why is it needed? Its contribution should be 0.

RoyStegeman: Because previously x=1 was not in the fktable xgrids so I could just use the largest x-grid point as the last step in the grid before x=1. With pineappl x=1 is there and it should remain at 1. 

Before this change the x=1 in pineappl would be the last step/interval before 1 after scaling.

scarlehoff (approved):

PR #2168, opened 2024-10-10
Parallelize Evolution at the level of `evolven3fit`

Radonirinaunimi:
This is not really in the direction of https://github.com/NNPDF/eko/issues/408 (specifically this [comment](https://github.com/NNPDF/eko/issues/408#issuecomment-2370691311)) but the following naive parallelization already provides a huge improvements.
- A fit with 120 replicas: ~5mn vs. ~20mn (numbers provided by @giacomomagni, not sure on how many cores)
- A fit with 1500 replicas:  ~20mn vs. ~4hrs on 32 cores

### DIFF
--- a/n3fit/src/evolven3fit/evolve.py
+++ b/n3fit/src/evolven3fit/evolve.py
@@ -5,7 +5,9 @@ import pathlib
 import sys
 
 from ekobox import apply, genpdf, info_file
+from joblib import Parallel, delayed
 import numpy as np
+import psutil
 
 import eko
 from eko import basis_rotation, runner
@@ -22,9 +24,19 @@ LOGGING_SETTINGS = {
     "level": logging.DEBUG,
 }
 
+NUM_CORES = psutil.cpu_count(logical=False)
+
 
 def evolve_fit(
-    fit_folder, q_fin, q_points, op_card_dict, theory_card_dict, force, eko_path, dump_eko=None
+    fit_folder,
+    q_fin,
+    q_points,
+    op_card_dict,
+    theory_card_dict,
+    force,
+    eko_path,
+    dump_eko=None,
+    ncores=1,
 ):
     """
     Evolves all the fitted replica in fit_folder/nnfit
@@ -117,9 +129,15 @@ def evolve_fit(
         info["NumFlavors"] = theory.heavy.num_flavs_max_pdf
         dump_info_file(usr_path, info)
 
-        for replica, pdf_data in initial_PDFs_dict.items():
-            evolved_blocks = evolve_exportgrid(pdf_data, eko_op, x_grid)
+        def _wrap_evolve(pdf, replica):
+            evolved_blocks = evolve_exportgrid(pdf, eko_op, x_grid)
             dump_evolved_replica(evolved_blocks, usr_path, int(replica.removeprefix("replica_")))
+
+        # Choose the number of cores to be the Minimal value
+        nb_cores = min(NUM_CORES, abs(ncores))
+        Parallel(n_jobs=nb_cores)(
+            delayed(_wrap_evolve)(pdf, r) for r, pdf in initial_PDFs_dict.items()
+        )
 
     # remove folder:
     # The function dump_evolved_replica dumps the replica files in a temporary folder
--- a/n3fit/src/n3fit/scripts/evolven3fit.py
+++ b/n3fit/src/n3fit/scripts/evolven3fit.py
@@ -7,10 +7,10 @@ import logging
 import pathlib
 import sys
 
-from evolven3fit import cli, eko_utils, evolve, utils
 import numpy as np
 
 from eko.runner.managed import solve
+from evolven3fit import cli, eko_utils, evolve, utils
 from n3fit.io.writer import XGRID
 from validphys.loader import FallbackLoader, Loader
 
@@ -169,6 +169,7 @@ def main():
             args.force,
             eko_path,
             None,
+            args.n_cores,
         )
     else:
         # If we are in the business of producing an eko, do some checks before starting:
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -40,6 +40,7 @@ requirements:
         - eko >=0.14.2
         - fiatlux
         - sphinx >=5.0.2
+        - joblib
         - sphinx_rtd_theme >0.5
         - sphinxcontrib-bibtex
         - ruamel.yaml <0.18
--- a/n3fit/src/evolven3fit/cli.py
+++ b/n3fit/src/evolven3fit/cli.py
@@ -2,7 +2,7 @@ from . import evolve, utils
 
 
 def cli_evolven3fit(
-    configuration_folder, q_fin, q_points, op_card_info, theory_card_info, force, load, dump
+    configuration_folder, q_fin, q_points, op_card_info, theory_card_info, force, load, dump, ncores
 ):
     """Evolves the fitted PDFs.
 
@@ -23,5 +23,13 @@ def cli_evolven3fit(
     """
     utils.check_is_a_fit(configuration_folder)
     return evolve.evolve_fit(
-        configuration_folder, q_fin, q_points, op_card_info, theory_card_info, force, load, dump
+        configuration_folder,
+        q_fin,
+        q_points,
+        op_card_info,
+        theory_card_info,
+        force,
+        load,
+        dump,
+        ncores,
     )
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,6 +73,7 @@ reportengine = { git = "https://github.com/NNPDF/reportengine" }
 psutil = "*"
 tensorflow = "*"
 eko = "^0.14.1"
+joblib = "*"
 # Hyperopt
 hyperopt = "*"
 seaborn = "*"


scarlehoff:
Please, let's fix it in EKO. Doing this just hides the problem (and potentially creates a huge problem if you happen to have many cores and not a lot of memory, this should definitely not use every core of the computer)

n3fit/src/evolven3fit/evolve.py
```diff
@@ -22,6 +24,8 @@
     "level": logging.DEBUG,
 }
 
+NUM_CORES = multiprocessing.cpu_count()
```
scarlehoff, line 27: It should not be the default and it should *not* use every available core. At the very least it should be limited by the number of cores set in the runcard and controlled by a `--parallel` flag.

But, in any case, please let fix this in EKO.

Radonirinaunimi:
> Please, let's fix it in EKO. Doing this just hides the problem (and potentially creates a huge problem if you happen to have many cores and not a lot of memory, this should definitely not use every core of the computer)

I agree that ultimately this **should** be fixed in EKO but I don't see that happening soon for many reasons but mainly:
- the next release of EKO would likely introduce a few breaking changes with all the refactoring
- it is not giving that one will not run with a similar memory problem so the the number of parallel evolution would have to be controlled anyway by the end user

The number of cores to use (depending on the memory) can always be easily set by the user by defining some environment variable or using `taskset`.

scarlehoff:
The next change of eko will probably be a disaster for pineko but I think it will be fine to just evolve a PDF I would hope.

> it is not giving that one will not run with a similar memory problem so the the number of parallel evolution would have to be controlled anyway by the end user

The memory problem is driven by the operators. Doing it in eko ensures that you do one value of Q at a time. Here yo uend up with (cores * size of the full eko) in the worst case scenario, vectorizing the convolution means that the worst case scenario is (size of the full eko).

> The number of cores to use (depending on the memory) can always be easily set by the user by defining some environment variable or using taskset.

The default should *never* be taking all cores. Programs that do that should be forbidden from running in any kind of shared system.

n3fit/src/evolven3fit/evolve.py
Radonirinaunimi: > It should not be the default and it should not use every available core. At the very least it should be limited by the number of cores set in the runcard and controlled by a --parallel flag.

For sure this can't be because the number of cores in the runcard are usually set as low as possible on the cluster (in order to run as many replicas as possible) while for the evolution one would want to run on as many cores as possible.

Anyway this was not really intended to be merged in the first place but I'd keep it around as it is handy.

Radonirinaunimi:
> The memory problem is driven by the operators. Doing it in eko ensures that you do one value of Q at a time. Here yo uend up with (cores * size of the full eko) in the worst case scenario, vectorizing the convolution means that the worst case scenario is (size of the full eko).

Actually no. The operator is only loaded once into memory and then shared by all the replicas. So it is not a function of the number of cores.

> The default should never be taking all cores. Programs that do that should be forbidden from running in any kind of shared system.

Yes, I agree with this.

felixhekhorn:
> > Please, let's fix it in EKO. Doing this just hides the problem (and potentially creates a huge problem if you happen to have many cores and not a lot of memory, this should definitely not use every core of the computer)

I agreee EKO is the correct place

> I agree that ultimately this **should** be fixed in EKO but I don't see that happening soon for many reasons but mainly:

actually, the main reason is lack of man power, so please get involved in eko if that is possible :innocent: 

> 
>     * the next release of EKO would likely introduce a few breaking changes with all the refactoring

there are a few breaking changes pending, that's why we will need some testing. My strategy so far is to delay the next tag of eko until it is actually required for this reason. However, if we now needed a tag, we make a tag.

> 
>     * it is not giving that one will not run with a similar memory problem so the the number of parallel evolution would have to be controlled anyway by the end user

the proposed solution is just to adjust the `einsum` call - i.e. if(!) there is no copying going on somewhere the memory footprint is the same as before: holding all replicas and holding the eko

> >    The default should never be taking all cores. Programs that do that should be forbidden from running in any kind of shared system.
> 
> Yes, I agree with this.

Unfortunately I have to admit that EKO is such a case :see_no_evil: (and it does create problems - see Como). We should change that. Alessandro was always arguing in favour of dropping this parallelization level and maybe, maybe with Rust we may be able to do so

scarlehoff:
> Unfortunately I have to admit that EKO is such a case

Is it? I though that EKO took only one if nothing was put in the runcard. Maybe that was the default that Andrea put in pineko/evolve then.

In any case, for the purposes of the evolution EKO certainly uses only one so it is fine.

> However, if we now needed a tag, we make a tag.

We need it only if this is implemented so that we can use it asap, but otherwise not really I believe.

Radonirinaunimi:
> actually, the main reason is lack of man power, so please get involved in eko if that is possible 😇

As I told @scarlehoff in some private exchanges, with @giacomomagni, we initially wanted to implement the changes directly in EKO but didn't do so yet because of this:

> there are a few breaking changes pending, that's why we will need some testing. My strategy so far is to delay the next tag of eko until it is actually required for this reason.

and everything it will imply. If it was just about adding this change that'd be straightforward.

So this is just a really needed dirty alternative for all the fits we need ASAP.

scarlehoff:
> So this is just a really needed dirty alternative for all the fits we need ASAP.

As already mentioned, I'm happy with merging this as long as it is not default (a --parallel flag or something) and it doesn't take all cores.

However, I would be extra happy with the proper fix :P

Radonirinaunimi:
> As already mentioned (here or somewhere else), I'm happy with merging this as long as it is not default (a --parallel flag or something) and it doesn't take all cores.
> 

Just added a `--ncores` flag to `evolve` to control this.

> However, I would be extra happy with the proper fix :P

Yes, of course!

n3fit/src/n3fit/scripts/evolven3fit.py
```diff
@@ -82,6 +82,13 @@ def construct_evolven3fit_parser(subparsers):
     parser.add_argument(
         "-d", "--dump", type=pathlib.Path, default=None, help="Path where the EKO is dumped"
     )
+    parser.add_argument(
+        "-n",
+        "--ncores",
+        type=int,
+        default=1,
+        help="Specify the number of cores to parallelize evolution",
+    )
```
scarlehoff, line 91: Careful with the naming, there _used_ to be a `-n` flag already.

n3fit/src/evolven3fit/evolve.py
```diff
@@ -125,7 +133,9 @@ def _wrap_evolve(pdf, replica):
             evolved_blocks = evolve_exportgrid(pdf, eko_op, x_grid)
             dump_evolved_replica(evolved_blocks, usr_path, int(replica.removeprefix("replica_")))
 
-        Parallel(n_jobs=NUM_CORES)(
+        # Choose the number of cores to be the Minimal value
+        nb_cores = min(NUM_CORES, ncores)
```
scarlehoff, line 137: Since you put `1` as the default I think it is fine if you just accept what the user asks for.

Radonirinaunimi: Right! I was mainly thinking in the situation in which someone is asking more than what's available 🙈

n3fit/src/n3fit/scripts/evolven3fit.py
Radonirinaunimi: But they live in different "parser", no? In the same way as `-p` for example (which appears at the top level `evolvenefit`, `produce_eko`, and `produce_eko_photon`).

scarlehoff: ah, not sure tbh, I just remember having used it.
But then can't you reuse that `-n`?

Radonirinaunimi: In principle we could. I am just not sure if there is a situation in which these two have to be different.

But actually given that this is a really temporary solution, let me do so.

scarlehoff: > I am just not sure if there is a situation in which these two have to be different.

No, the existing one is used for the generation of the eko. But the generation and the evolution are always running separately so no problem

Radonirinaunimi: Done in [a1f5abd](https://github.com/NNPDF/nnpdf/pull/2168/commits/a1f5abd0ed5a674369caa0682fa19e82402c6d4f).

n3fit/src/evolven3fit/evolve.py
Radonirinaunimi: Done in [a1f5abd](https://github.com/NNPDF/nnpdf/pull/2168/commits/a1f5abd0ed5a674369caa0682fa19e82402c6d4f). 

In the decided to keep the `min` as it doesn't make more sense to run more concurring jobs than the available `cores`.

scarlehoff:
Hi @Radonirinaunimi is it fine if I rebase, squash and merge this PR? 

That way it will be easier to rollback once #2181 can be merged

Radonirinaunimi:
> Hi @Radonirinaunimi is it fine if I rebase, squash and merge this PR?
> 
> That way it will be easier to rollback once #2181 can be merged

Yes, please do so!

PR #2169, opened 2024-10-10
Implementation of ATLAS_Z0J_8TEV PT-Y and PT-M in the new format

achiefa:


nnpdf_data/nnpdf_data/commondata/ATLAS_Z0J_8TEV/kinematics_legacy_PT-Y.yaml
```diff
@@ -0,0 +1,1441 @@
...
+    min: null
+    mid: 8000.0
+    max: null
+- k1:
+    min: null
+    mid: 2.2
+    max: null
+  k2:
+    min: null
+    mid: 302500.0
```
achiefa, line 1436: @enocera @scarlehoff I don't understand why, in the old implementation, the mid value of the bin is squared, even though the x-sec is differential in $p_T$ and not in $(p_T)^2$. For instance, the last bin in $p_T$ in hepdata is $[200,900]$ GeV, while here we have $(550)^2$. Is there a subtle reason for that? And most importantly, should I re-implement the kinematics using $p_T$ or $(p_T)^2$?

Screenshot 2024-10-10 at 18 55 06

scarlehoff: Because these k2 would get a sqrt() somewhere in plotoptions.

Just select a process option from here https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/process_options.py (or implement a new one). My vote is to make sure the process type in process options accepts pT and use pT since that's what we plot

Radonirinaunimi:
Btw @achiefa, how are you generating (re-)generating the `YAML` files? Do you have something like the `filter.py`? If so, could you also upload that one. We want to make sure that we'll be able to re-generate these data in the future.

achiefa:
@Radonirinaunimi I'm not regenerating anything at the moment. I simply changed the name of the old kinematics to `legacy`. I still need to write the `filter.py`.

nnpdf_data/nnpdf_data/commondata/ATLAS_Z0J_8TEV/metadata.yaml
```diff
@@ -102,7 +107,7 @@ implemented_observables:
         description: Variable k3
         label: k3
         units: ''
-    file: kinematics_PT-M.yaml
+    file: kinematics_legacy_PT-M.yaml
```
scarlehoff, line 110: Just for reference, we don't need to keep the legacy kinematics, you can just remove these files.
The information should be the same as in the new ones, and the new ones have all the info we need (name of the variables mainly)

achiefa: Thank you. However, I'd say that the information is not the same. The "old" kinematics used the mid value of the bin (squared), while the new one that I'm going to implement will contain min and max value of the bin (non-squared). If you say that it doesn't make any difference, and we can get rid of the old kinematics, I will remove the latter and implement the new kinematics as mentioned above.

scarlehoff: Yes, you can remove the old one. You can fill min, max and the midpoint will be filled automatically (and should be the same as before).

The reason only the midpoint was used before is that validphys could not deal with that extra information, but having the min-max is much better because then we can use these datasets to autogenerate runcards (for instance the NNLOJET runcards need this information)

achiefa: Ok, thank you so much. This is indeed what I was trying to understand. I go on.

achiefa:
Apparently, the tables in hepdata do not provide a full breakdown of the systematic uncertainties. Thus, I don't know how to reproduce the breakdown of systematic uncertainties reported in the legacy file. Should I only use the diagonal entries in the rawdata tables?

Radonirinaunimi:
> Apparently, the tables in hepdata do not provide a full breakdown of the systematic uncertainties. Thus, I don't know how to reproduce the breakdown of systematic uncertainties reported in the legacy file. Should I only use the diagonal entries in the rawdata tables?

I fear this will be a recurring theme the more datasets are re-implemented, so decisions have to taken with @enocera on dataset-by-dataset basis (?). 

I would point out however that, usually (as is the case for this particular dataset), HepData includes the raw data in which the break-down of the systematics are given if one download the full thing with the resource files. (So to say that even the distinction between HepData and non-HepData is ambiguous).

achiefa:
Ok, I think I sort of understood how to use the resource files to retrieve the old break-down of the uncertainties. So, what do you want me to do?

enocera:
@achiefa My suggestion is to implement the full breakdown of uncertainties, whenever this can be retrieved from HepData. This is the case, even if this amounts to look into the "Resources".

achiefa:
I've implemented the dataset with the full break-down of the systematic uncertainties. However, there are a few things that are worth mentioning:
- The new implementation of `PT-Y` matches the legacy one.  You can check that using the filter file, where I've implemented an automatic comparison new_vs_legacy for central data and uncertainties.
- For `PT-M`, the number of points provided in HepData is higher than the legacy implementation. Thus, I could not reproduce the legacy version from HepData.
- Within the new implementation of `PT-M`, I ran into a binning problem for table 39 in HepData. Specifically, there is a mismatch between the mid values of the pT bins in table 39 and those also reported in `ZcombPt_born_m66116_y0024/tab.dat`, which is the file containing the break-down of the sys uncertainties for the same table. It might be that the break-down was worked out with a different bin size. I attach the log of the filter script, which gives more information ([log.log](https://github.com/user-attachments/files/17400510/log.log). As you can see, the differences are really small, and only for 3 of the bins.

One last thing - Is there a `xQ2` map for `PT-M`?

nnpdf_data/nnpdf_data/commondata/ATLAS_Z0J_8TEV/filter.py
```diff
@@ -0,0 +1,61 @@
+import yaml
```
Radonirinaunimi, line 1: I'm noticing that the indentation in the Python files are not consistent. You could run a [pre-commit hooks](https://pre-commit.com/) to make sure that everything is properly formatted.

nnpdf_data/nnpdf_data/commondata/ATLAS_Z0J_8TEV/filter_utils.py
```diff
@@ -0,0 +1,284 @@
+import numpy as np
```
Radonirinaunimi, line 1: Please move the utils to [nnpdf_data/nnpdf_data/filter_utils](https://github.com/NNPDF/nnpdf/tree/deda5b8c89d31eee94eeaa450d66d756a4ec0dd8/nnpdf_data/nnpdf_data/filter_utils), especially if they are re-usable.

nnpdf_data/nnpdf_data/commondata/ATLAS_Z0J_8TEV/filter_utils.py
```diff
@@ -0,0 +1,284 @@
...
+      kin_min = bin['low']
+      kin_max = bin['high']
+      kin2_bin = self.kin2_dict[f'Table{tab_number}']
+      kin_bin = {
+        label[0]: {'min': kin2_bin[0],
+                  'mid': None,
+                  'max': kin2_bin[1]},
+        label[1]: {'min': kin_min, 
+                  'mid': None, 
+                  'max': kin_max},
```
Radonirinaunimi, line 71: You should also fill the mid-values. Actually, `mid` is the field that is used everywhere in the codes.

Radonirinaunimi (changes requested):
Thanks for this @achiefa! I will need to look in detail into the mismatch in `PT-M`, however, here are some quick comments.

Also, whenever possible (in case the datapoints and everything correspond to the old implementation, as is the case for `PT-Y`), could you make sure that the covmats are the same?

```python
from validphys.api import API
import numpy as np

inp1 = {"dataset_input": {"dataset": f"{new_implementation}"}, "theoryid": 40_000_000, "use_cuts": "internal"}
inp2 = {"dataset_input": {"dataset": f"{old_implementation}", "variant": "legacy"}, "theoryid": 40_000_000, "use_cuts": "internal"}

np.isclose(covmat1, covmat2)
``` 

This not only makes sure that the implementations are consistent but most importantly that the datasets could be loaded properly.

achiefa:
For `PT-Y` the two covmats are the same.
Screenshot 2024-10-17 at 12 30 28

nnpdf_data/nnpdf_data/commondata/ATLAS_Z0J_8TEV/filter_utils.py
scarlehoff: I think (if not, it would be a nice feature to have) that I added in the parser that if the mid is None it gets automatically filled _by the parser_

(of course, given that we technically know it at this point we might also fill it here)

Radonirinaunimi:
> For PT-M, the number of points provided in HepData is higher than the legacy implementation. Thus, I could not reproduce the legacy version from HepData.

If the number of datapoints are different, then the re-implementation in the new format becomes a new dataset. Please refer around this [comment](https://github.com/NNPDF/nnpdf/pull/2170#discussion_r1796522864) for more context.

> Within the new implementation of PT-M, I ran into a binning problem for table 39 in HepData. Specifically, there is a mismatch between the mid values of the pT bins in table 39 and those also reported in ZcombPt_born_m66116_y0024/tab.dat, which is the file containing the break-down of the sys uncertainties for the same table. It might be that the break-down was worked out with a different bin size. I attach the log of the filter script, which gives more information ([log.log](https://github.com/user-attachments/files/17400510/log.log). As you can see, the differences are really small, and only for 3 of the bins.

For this, let's simply use the rawdata, even for the binned/centra values.

scarlehoff:
> For PT-Y the two covmats are the same.

Nice! Could you also check that the t0 covmat is the same, to be completely sure?


> For PT-M, the number of points provided in HepData is higher than the legacy implementation. Thus, I could not    reproduce the legacy version from HepData.
> If the number of datapoints are different, then the re-implementation in the new format becomes a new dataset. Please refer around this https://github.com/NNPDF/nnpdf/pull/2170#discussion_r1796522864 for more context.

This is tricky indeed :___ Which points have changed? If they are points that we don't want to include maybe it makes sense to remove them directly in the filter.py file (for instance, if they are points for pT < 30 GeV we can safely discard them and then later on if we need it add a atlas_zoj_lowpt dataset)

achiefa:
> Nice! Could you also check that the t0 covmat is the same, to be completely sure?

I tried to use `t0_covmat_from_systematics` from the API, but I got this `AttributeError`
```
AttributeError: 'NoneType' object has no attribute 'is_polarized'
```

scarlehoff:
The t0 needs also a `pdf=NNPDF40_nnlo_as_01180` or `t0set=` or something like that

achiefa:
Ok, the t0 matrices are the same as well.

> This is tricky indeed :___ Which points have changed? If they are points that we don't want to include maybe it makes sense to remove them directly in the filter.py file (for instance, if they are points for pT < 30 GeV we can safely discard them and then later on if we need it add a atlas_zoj_lowpt dataset)

That I didn't check. But I have a feeling that the new implementation adds a range in the mass of the lepton pair, but using the same bins in pT. I'll check that.

achiefa:
I have investigated the discrepancy in the `PT-M` distribution between the legacy and the new versions. Just to summarise, the new implementation has six different kinematics regions for the lepton pair mass (in GeV), each of them with bins in rapidity:
```
 12.0  <  m_ll  <  20.0 
 20.0  <  m_ll  <  30.0
 30.0  <  m_ll  <  46.0
 46.0  <  m_ll  <  66.0
 66.0  <  m_ll  <  116.0
 116.0 <  m_ll  <  150.0
```
The legacy version doesn't have the second-to-last kin region in $M_{ll}$. On the other hand, the bins in rapidity and the respective central data agree between the two versions up to the fourth kin region ($46.0  <  M_{ll}  <  66.0$). The last kin region also matches in rapidity and central data. However, they have used a different binning for this latter kin region, as in the legacy version $\textrm{mid}[M_{ll}] = 138$ GeV, whereas in the new implementation $\textrm{mid}[M_{ll}] = 133$ GeV. Note that for the legacy version, I only have the mid value of the bins, but not the extremes.

I don't know if this tiny difference in the last kin region is a problem for the grids. If not, we could drop the second-to-last bin and use the same grids as in the legacy implementation.

scarlehoff:
@enocera do you know why are we avoiding that bin?

RE the binning, @achiefa if the central data is the same, it might be a mistake on the kinematic value here. Could you check whether the grid agrees with the values here or with the values in hepdata?

enocera:
Dear @achiefa (cc @scarlehoff), I am a little confused about the description of this data set. There must indeed be 6 bins in the invariant mass of the final state, as @achiefa correctly mentions above. All of these are then single-differential in pT (and not in rapidity, as @achiefa seems to suggest above). The bin corresponding to the Z-mass peak (`66.0  <  m_ll  <  116.0`) also comes double differential in `pT` and in `y`. So what we did in the legacy implementation is the following: we took all the invariant mass bins, except the one on the Z-mass peak, that are single-differential in pT and we put them into the `ATLAS_Z0J_8TEV_PT-M` data set. We then took the double differential (`pT`-`y`) Z-mass peak bin and put it in the `ATLAS_Z0J_8TEV_PT-Y` data set. This is because the more differential we are the better for PDF determination. If we had put the Z-mass peak single-differential bin in `ATLAS_Z0J_8TEV_PT-M` we would have had double counting with `ATLAS_Z0J_8TEV_PT-Y`. I hope that this clarifies the situation.

achiefa:
Hi @enocera, I apologise. I meant to say pT and not rapidity, as you pointed out.

Note that the invariant mass bins are in the kinematic region $0 < y_{ll} < 2.4$, including the one at the Z-mass peak. So I'm confused, becuase $0 < y_{ll} < 2.4$ covers all the bins in rapidity for the `PT-Y` distribution. How did you include the invariant mass bin at the Z-mass peak in the `PT-Y` distribution then?

enocera:
@achiefa as you can see from Hepdata
https://www.hepdata.net/record/ins1408516
Tables 29-34 refer to the cross section on the Z-mass peak (66 GeV<m_ll<116 GeV) which is double differential in `pT` and `y` (each table corresponds to a different `y` bin). These tables form the `ATLAS_Z0J_8TEV_PT-Y` data set. Tables 35-40 refer to the cross section on the various mass bins, including the Z-mass peak bin (Table 39) which is single-differential in `pT` (each table corresponds to a different invariant mass bin). These tables, except table 39, form the `ATLAS_Z0J_8TEV_PT-M` data set. We could have put table 39 in this data set and forgotten about the other data set (no double counting), but, as I said, the more differential we are the better. Is this clearer?

achiefa:
Ok, now I understand. Thank you so much.

achiefa:
> Could you check whether the grid agrees with the values here or with the values in hepdata?

The FK table for the last range in the invariant mass is `ATLASZPT8TEVMDIST-ATLASZPT8TEV-MLLBIN6_ptZ`. Is there a way to check that using the pineappl CLI?

scarlehoff:
Given the explanation above by @enocera everything seems consistent. But for the future, when you do `pineappl convolve` you get the binning as part of the output. Otherwise, you can obtain that information with `pineappl read`

achiefa:
Dear @scarlehoff, I'm trying to plot the x-q2 map using `_dybosonpt_xq2map`. I can do that with the `PT-M` distributions but not with the `PT-Y`distribution. The reason is that the xq2 map needs either `m_Z2` or `m_W2`, while the kinematic_coverage for `PT-Y` is `[y, pT, sqrts]`. Is it a problem to add `m_Z2` to the variables in the kinematic coverage for the `PT-Y` distribution? It would be the same value for all bins, but I don't know if the commondata is meant to have only three kinematic variables.

achiefa:
Dear @scarlehoff @Radonirinaunimi, have a look at the two reports that I posted in the description of this PR. I have three observations that I'd like to make:
1. The data-theory comparisons look unchanged for both distributions `PT-Y` and `PT-M`.
2. I added a $(x,Q^2)$ map for the distribution differential in pT and rapidity:
    ```
    def _dybosonptrap_xq2map(kin_info):
       """
       Computes x and q2 mapping for DY Z or W -> 2 leptons + jet process
       using the rapidity of the final lepton pair.
       """
       pT = kin_info[_Vars.pT]
       eta = kin_info.get_one_of(_Vars.eta, _Vars.y)
       sqrts = kin_info[_Vars.sqrts]
       ET2 = pT * pT
       x1 = np.sqrt(ET2) / sqrts * np.exp(-eta)
       x2 = np.sqrt(ET2) / sqrts * np.exp(+eta)
       x = np.concatenate((x1, x2))
      return x, np.concatenate((ET2, ET2))
    ```
   With this map, I can reproduce the kinematic plot of the legacy version for the distribution `PT-Y`. However, I'm not entirely sure about `ET2`. I think it should be something like `ET2 = m2_ll + pT * pT`, where `m2_ll` is the square of the invariant mass. Hence, the Bjorken variables should be `x = (np.sqrt(ET2) + pT) / sqrts * np.exp(+-eta)`. Of course, the kinematic plot would differ from the legacy one if I were to do so.
3. I used `_dybosonpt_xq2map` for the distribution differential in pT and invariant mass (`PT-M`) but substituted `m_Z2` with `m2_ll`. However, the kinematic plot does not match the legacy version. Should I use a fixed mass value, like `m_Z2 = 91.1876`?

Finally, I want to mention that I modified `filters.yaml` so that the cuts are applied to pT and not to pT2. Let me know if that is correct.

scarlehoff (commented):
> ET2 = m2_ll + pT * pT

Indeed, this is the right definition. It might be that the previous implementation was using muR=muF=pT as the scale, but since the theoretical predictions for ZpT usually use Et, I think Et is the right variable here (basically the previous one was "wrong", probably it doesn't shift the mapping that much though?)

> Finally, I want to mention that I modified filters.yaml so that the cuts are applied to pT and not to pT2. Let me know if that is correct.

Yes, perfect.

Do you understand why do you have that (admittely small) change in the chi2 in your reports?

(other than that, looks good!)

scarlehoff:
Yes, please, rebase on top of master so that you can use the nice bot that @Radonirinaunimi added to regenerate the data.

And ensure that you are using the yaml prettifier so that things like 6.000000000001 are rounded

scarlehoff:
If the EIC data is breaking the tests (and were not broken before) it must be somethinf that has been changed in this branch.

RE the prettifier, look e.g. at the filter.py of the jet data, basically adding a function for the yaml parser.

achiefa:
I fixed the test for the commondata, but `test_overfit_chi2` keeps failing. I don't know what is causing that, though.

scarlehoff (commented):
Try to run the test locally to find out what broke the test.

Btw, culd you also remove the jupyter notebook for the filter?

achiefa:
It seems to be `test_overfit_chi2` and in particular:
```
assert False
E        +  where False = <built-in method all of numpy.ndarray object at 0x16c07cdb0>()
E        +    where <built-in method all of numpy.ndarray object at 0x16c07cdb0> = array([0.00040202, 0.00034751, 0.00020966]) < 0.0001.all
E        +      where array([0.00040202, 0.00034751, 0.00020966]) = abs((array([1.86702616, 1.97033108, 1.84250167]) - [1.867428183555603, 1.9699835777282715, 1.842711329460144]))
```
The order of magnitude for the tolerance is fine though.

achiefa:
Ok, I figured out the problem. It is due to the round of the digits in the HepData table:

>I think that is because the central data points in hepdata are rounded to 3 decimal places, while the legacy version considered digits up to the fifth (or something similar).

Indeed, I tried to use the legacy variant for the central data, and the tests succeeded. The legacy data points can be gathered from the source files. What do you think I should do? @scarlehoff

scarlehoff:
For the chi2 that's good.

For the tests, just regenerate the regression test with the new data if that's the only difference.

Btw, could you remove from the rawdata all tables that are not being used by the filter (if any)

achiefa:
Hi @scarlehoff, the PR is ready to be reviewed now.

Note that all tests passed in the previous commit, but now they fail because of a Gateway error:
```
Error: : Failed to fetch remote theories index https://nnpdf.web.cern.ch/nnpdf/tables/theorydata.json: 504 Server Error: Gateway Time-out for url: https://nnpdf.web.cern.ch/nnpdf/tables/theorydata.json
Error: : Failed to fetch remote theories index https://nnpdf.web.cern.ch/nnpdf/tables_box/theorydata.json: 504 Server Error: Gateway Time-out for url: https://nnpdf.web.cern.ch/nnpdf/tables_box/theorydata.json
Error: : Resource not in the remote repository: Theory 41100010 not available.
Error: : Failed processing key theoryid.
Error: : Bad configuration encountered:
Could not find theory 41100010. Folder '/usr/share/miniconda/envs/test/share/NNPDF/theories/theory_41100010' not found
Instead of '41100010', did you mean one of the following?
 - 704
 - 162
 - 398
 - 399
 - 40000000
 - 708
 - 712
 ```
I suspect that is not up to me because, as I said, all tests passed last week, and the last commit just corrected a nan value in one of the uncertainties variants.

**Comment by scarlehoff** [comment]
When there is only one test failing, it might be due to an internet problem. Just resubmit the test (I just did).

**Comment by achiefa** [comment]
Ok, all tests passed!

File: validphys2/src/validphys/process_options.py [hunk 6]
```diff
@@ -359,10 +377,17 @@ def _dybosonpt_xq2map(kin_dict):
...
     "DY W or Z (2 leptons) + j boson transverse momentum",
-    accepted_variables=(_Vars.pT, _Vars.m_W2, _Vars.m_Z2, _Vars.sqrts),
+    accepted_variables=(_Vars.pT, _Vars.m_W2, _Vars.m_Z2, _Vars.sqrts, _Vars.y, _Vars.m_ll2),
     xq2map_function=_dybosonpt_xq2map,
 )
 
+DY_PT_RAP = _Process(
+    "DY_PT",
+    "DY W or Z (2 leptons) + j boson transverse momentum",
+    accepted_variables=(_Vars.pT, _Vars.m_W2, _Vars.m_Z2, _Vars.sqrts, _Vars.y, _Vars.m_ll2),
```
scarlehoff, line 387: The function accepts also eta I believe.

scarlehoff (commented):
Thanks for this. Just a final comment, please remove all the files from hepdata that are not explicitly used in the fitler, otherwise the size of the repository will grow a lot without need.

nnpdf_data/nnpdf_data/commondata/ATLAS_Z0J_8TEV/rawdata/hepdata/F2AV_README
```diff
@@ -0,0 +1,525 @@
+
```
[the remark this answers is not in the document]
achiefa: Hi @scarlehoff, thanks for the comments. I would keep it because it contains non-trivial information about the breakdown of the systematic uncertainties in the source files, which are used to build the covmat. I think it might be a good idea to keep this kind of information, also for future reference.

validphys2/src/validphys/process_options.py
achiefa: Ok, done!

nnpdf_data/nnpdf_data/commondata/ATLAS_Z0J_8TEV/rawdata/hepdata/F2AV_README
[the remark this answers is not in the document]
scarlehoff: Going forward we are removing _all_ data in the `rawdata` directory, see https://github.com/NNPDF/nnpdf/issues/2199

If there is some information that is important, summarize it and add it as a comment or module-level docstring to the `filter.py`.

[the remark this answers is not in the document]
achiefa: Ok, I'll summarise the relevant information in a module-level docstring.

achiefa:
Ok, green light from me.

scarlehoff (approved):
Great! Thanks

scarlehoff (commented):
Sorry, one last minor comment before merging. Could you remove the kinematics_legacy? since those are not needed anymore

Thanks!

achiefa:
Yes, I can. Do you want me also to remove the data legacy?

scarlehoff:
Yes, please, since it is not being loaded anymore.

RoyStegeman:
Should these commits be squashed into a single one? In particular I'm looking at https://github.com/NNPDF/nnpdf/pull/2169/commits/f9bcdc9d17fccce508e0670e6940aaec437ad908 and it would be nice not to have in the history in master.

achiefa:
Good to go!

scarlehoff:
I've recommited the squash to remove the merge commits, let's wait for the tests then merge. Thanks for this!

PR #2170, opened 2024-10-10
Check Fixed target DIS 

giacomomagni:
This PR is to implement the fixed target dis datasets from hepdata. 

## Summary

| Experiment | Reproducibility | Note | TODO |
|------------|--------|---------------|------|
| BCDMS | ✅   | Two new observables, which superseede the old ones. |  |
| CHORUS | ✅ | Kinematics and central data match. Last point added manually. Minor change in Q2 | |  
| EMC |  ✅   | Kinematics and central data match. Branching ratio of 0.82 wrt to Hepdata. Add syst of 15%  |  |
| NMC  | 🟡    | The ratio dataset matches (except for uncertainties). ⚠️ $\sigma$ normalisation not clear for the P dataset. Two new observables, which are alternative to the old ones.  |
| NUTEV  | 🟡  | Implementation matches with Mason theses. Br and acceptance corrections where provided directly by the author. Minor change in Q2 | acc_err not used ? where is the normalization error coming from ? |
| SLAC  | 🔴  |  Numbers are consistent with Phys.Lett.B 282 (1992) 475, but the source is unavailable. See [comment below](https://github.com/NNPDF/nnpdf/pull/2170#issuecomment-2431966128). |

### Common questions
* ⚠️ Old implementation used the breakdown of systematics when available. HepData use the combined one. 
How do we want to proceed? NO, only in the legacy data.
* ⚠️ Do we want to store the information about the old systematics somewhere? NO, only in the legacy data
* ⚠️ Do we want to store the old rawdata for SLAC ? NO, only in the legacy data

## New variants

| Dataset                                 |  hepdata | arxiv | Action | Motivation | Uncertainties | Variants |
|--------------------------- |----------|------|-------|------------|--------------|---------|
| BCDMS_NC_NOTFIXED_P | [link](https://www.hepdata.net/record/ins276661) | [link](https://cds.cern.ch/record/194232) |  no action taken, replaced by new observables |  The old implementation (data and kinematic) coincides with the R=0, not averaged on $\sqrt{s}$ values. However the averaged and not averaged tables do not have the same kinematic. |  Breakdown of syst. not given in HepData  |   |
| BCDMS_NC_NOTFIXED_D | [link](https://www.hepdata.net/record/ins285497) | [link](https://cds.cern.ch/record/203765) |  no action taken, replaced by new observables |  The old implementation (data and kinematic) coincides with the R=0, not averaged on $\sqrt{s}$ values. However the averaged and not averaged tables do not have the same kinematic. |  Breakdown of syst. not given in HepData  |  |
| CHORUS_CC_NOTFIXED_PB | [link](https://www.hepdata.net/record/ins699123) | [link](https://cds.cern.ch/record/906644) |  new variant  |  Kinematics and central value match | Breakdown of syst. not given in HepData |  `hepdata` |
| EMC_NC_250GEV_DW | [link](https://www.hepdata.net/record/ins180921) | [link](https://cds.cern.ch/record/140481/files/198211222.pdf) |  new variant  |  Kinematics and central match value |  Old implementation has a 0.15 % syst not given in HepData. + nuclear u.   |  `rzero` |
| NMC_NC_NOTFIXED | [link](https://www.hepdata.net/record/ins426595) | [link](https://arxiv.org/abs/hep-ex/9611022) |  new variant  |  Kinematics and central value match  | Breakdown of syst. not given in HepData |  `hepdata` |
| NMC_NC_NOTFIXED_P | [link](https://www.hepdata.net/record/ins424154) | [link](https://arxiv.org/pdf/hep-ph/9610231) |  no action taken, see new observables  |  Kinematics match, $\sigma$ was computed in the old buildmaster from R and F2  |  Full breakdown of systematics included in legacy. | |
| NUTEV_CC_NOTFIXED_PB |  | [link](https://inspirehep.net/literature/712244) |  new variant  |  Kinematics and central value match | copied from old buildmaster |  `hepdata` with the updated BR and no nuclear uncertanties |
| SLAC_NC_NOTFIXED_P |  | | no action taken. metadata amended.  |  |  |  |  |  |
| SLAC_NC_NOTFIXED_D |  | | no action taken.  metadata amended. |  |  |  |  |  |

## New HEPDATA Observables 

| Observable                           |  hepdata | arxiv | Motivation |  Variants |
|--------------------------- |----------|------|------------|---------|
| BCDMS_NC_NOTFIXED_P_EM-F2-HEPDATA | [link](https://www.hepdata.net/record/ins276661) | [link](https://cds.cern.ch/record/194232) |  Averaged values on $\sqrt{s}$ has different kinematics.  |  `rzero`, `rqcd` |
| BCDMS_NC_NOTFIXED_D_EM-F2-HEPDATA | [link](https://www.hepdata.net/record/ins285497) | [link](https://cds.cern.ch/record/203765) |  Averaged values on $\sqrt{s}$ has different kinematics.  |  `rzero`, `rqcd` |
| NMC_NC_NOTFIXED_P_EM-F2-HEPDATA | [link](https://www.hepdata.net/record/ins424154) | [link](https://arxiv.org/pdf/hep-ph/9610231) |  Old implementation was $\sigma$, not $F_2$, breakdown of syst. not given |   |
| NMC_NC_NOTFIXED_D_EM-F2-HEPDATA | [link](https://www.hepdata.net/record/ins424154) | [link](https://arxiv.org/pdf/hep-ph/9610231) |  Not available before (possible double counting), breakdown of syst. not given  |   |


giacomomagni:
Okay here we have the first interesting case: 
central data and and kinematics (even if I swap sqrts with y) do not match exactly. 
What do we want to do? New dataset ? New observable ?

enocera:
My opinion is as follows:
- implement a new data set, consistent with the information on Hepdata, given that the legacy data se is there, we could study later the impact of the "new" dataset w.r.t. the old;
- we want to download and store the tables.

scarlehoff:
Should we do a complete new dataset or just have a legacy variant?

(this is a new situation due to the kinematics change so a decision has to be made!! I'm personally partial to have either a variant or an observable even if the legacy didn't come from the same paper)

giacomomagni:
So the legacy and the hepdata version should come from the same reference. 
I'd be in favor of creating a variant (with different kinematic, unc and data) so this way it will be clear that it's completely exclusive wrt to the legacy one.

scarlehoff:
> But I'm not sure if kinematic variants are easy to support (or already working)

It might be working and if not it's very easy to support so no problem

giacomomagni:
> It might be working and if not it's very easy to support so no problem

But are you sure, for instance it will requires also different FKtables, ecc...

scarlehoff:
Theory and data/uncertainties are supported by variants (at the beginning it was only for uncertainties but then we realized data and theory might need to change for these legacy cases).

We decided to create a new dataset if e.g. the number of points changed.

validphys2/src/validphys/commondataparser.py
```diff
@@ -362,6 +348,26 @@ def apply_label(self, var, value):
...
+    A variant can overwrite a number of keys, as defined by this dataclass
+    """
+
+    data_uncertainties: Optional[list[ValidPath]] = None
+    theory: Optional[TheoryMeta] = None
+    data_central: Optional[ValidPath] = None
+    kinematic_coverage: Optional[list] = None
+    kinematics: Optional[ValidKinematics] = None
+    tables: Optional[list] = None
+    ndata: Optional[int] = None
```
scarlehoff, line 364: Hi @giacomomagni let's not do this one. If the number of datapoints actually changes, that's a completely different dataset.

Actually, I would say the same is true for `kinematic_coverage` and `plotting`. Those are enough to say they are a different dataset (and, tbh, being the same dataset it could be recasted to the same kinematic variables, so only the numerical values should change)

giacomomagni: Okay, it was just an attempt, i'll revert it.
So then we go for new datasets, we can include `_HEPDATA_` in the name to distinguish it.

scarlehoff: Let me invoke the master @enocera 

I'm not sure whether maybe we want to change the _old ones_ to be called `whatever....-legacy` so that the new ones have the "easy name"?

giacomomagni: So renaming old dataset will be not backward compatible but it's also possible, I mean in the long terms you'd like to get rid of them.

scarlehoff: By putting the name of on the `dataset_names.yml` all fits using the (very) old name will be reproducible just the same, and the new ones (the ones that @RoyStegeman is collecting in the wiki) will now refer to the hepdata dataset which is good I'd say (_if_ we consider that the hepdata dataset is the final truth, of course)

giacomomagni: I see, okay than we can add ``_LEGACY_`` to the old datsets and proceed with the new one. 
I'll do it asap.

enocera: I confirm we should go for a new data set (as I mentioned at the beginning of the thread). I validate @scarlehoff's proposal.

giacomomagni: To support full backward compatibility  with the current studies using the new commondata, 
it's better to change the name of the new dataset.

scarlehoff:
https://www.slac.stanford.edu/pubs/slacpubs/5250/slac-pub-5442.pdf

giacomomagni:
This is a (hopefully) conclusive report about the investigations regarding the source of the **SLAC data**. 
Thanks to @enocera  for the precious collaboration. 

1. Legacy data are in agreement with Figs 2 (proton) e 4 (deuteron) di [Phys.Lett.B 282 (1992) 475](https://inspirehep.net/literature/319089), as now mentioned in the `metadata.yaml` file. These plots match with Figs 5.9 and Fig 5.10 of [Whitlow thesis](https://inspirehep.net/files/546b978b7c6224876d39b3867cb1bb2c).
2. Data are the re-analysis of all the SLAC experiments:  E49a, E49b, E61, E87, E89a, E89b, E139, E140 and rebinned with the same `x` kinematics of the BCDMS experiment: 
```
Data from all eight experiments (E140 included) are fused into the x-spectra shown in Figures 5.9 and 5.10. Each datum represents the weighted average of as many as twelve F2 extractions from as many as five experiments. The x-spectra presented in these figures match the x-binning of BCDMS. Numerical values for these spectra are tabled in Appendix E
```
3. No HepData source is available. In particular the [tables linked to Whitlow thesis](https://www.hepdata.net/record/ins295113) do not match with the BCDMS binning, so are not able to reproduce Fig 5.9 of the thesis, see also the different kinematic coverage in the plot below. It is unclear whether they have better accuracy than the already implemented data.
4. CT does not include SLAC data in their fits. MSHT might be using our very same implementation.

**Action taken**

1. Metadata files of the legacy versions have been amended regarding the source of the data.
2. There **will not be and Hepdata implementation** of these data. The idea is to exclude SLAC measurements in the long term.

**Additional files**
Kinematic coverage of the HepData tables for proton target vs our legacy implementation: [kinplot.pdf](https://github.com/user-attachments/files/17491141/Markers0_plot_xq2.pdf)

scarlehoff:
I honestly don't understand what's happening with the tests in this branch. The failure is a systematics overlap between NMC and LHCB and the overlap is an uncertainty called LHCB something which is obviously not in NMC.

Then the test of new data fails because `ModuleNotFoundError: No module named 'nnpdf_data.commondata.NMC_NC_NOTFIXED_P_HEPDATA.filter'` but... it is there afaiks?

I am _very_ confused.

Radonirinaunimi:
> Then the test of new data fails because `ModuleNotFoundError: No module named 'nnpdf_data.commondata.NMC_NC_NOTFIXED_P_HEPDATA.filter'` but... it is there afaiks?
> 
Python cannot resolve it because there is no `__init__.py`. We should not import submodules/functions from the commondata folder. If it has to be imported across multiple files then `filter_utils` is a better place.

> I honestly don't understand what's happening with the tests in this branch. The failure is a systematics overlap between NMC and LHCB and the overlap is an uncertainty called LHCB something which is obviously not in NMC.
> 
The tests in other places (e.g. https://github.com/NNPDF/nnpdf/pull/2204) are also failing for mysterious reasons....

giacomomagni:
Thanks for spotting it @Radonirinaunimi !

scarlehoff:
Ok, this was a funny bug. Apart from the test being not very useful and the `print_systype_overlap` being unnecesarily complicated, it seems that the problem was the `lru` part of the cache (I just changed `lru_cache` to `cache` but it worked just the same by increasing the size of the cache).

At some point (I couldn't figure out why) the `load` function for a NMC dataset returns a previously loaded LHCb:

```
ipdb> a1.name
'NMC_NC_NOTFIXED_EM-F2'
ipdb> b1.name
'LHCB_WPWM_7TEV_MUON_Y'
ipdb> a1.load()
CommonData(setname='LHCB_WPWM_7TEV_MUON_Y', ndata=16, commondataproc='EWK_RAP', nkin=3, nsys=35, legacy=False, legacy_names=['LHCBWMU7TEV'], kin_variables=['y', 'M2', 'sqrts'])
ipdb> b1.load()
CommonData(setname='LHCB_WPWM_7TEV_MUON_Y', ndata=16, commondataproc='EWK_RAP', nkin=3, nsys=35, legacy=False, legacy_names=['LHCBWMU7TEV'], kin_variables=['y', 'M2', 'sqrts'])
```

I don't really have time to figure out the why, but the reason (a hash collision? but I checked and it doesn't seem to be the case?) is possibly intrincate enough to warrant some fun investigation in the future.
So I leave here the commit and the minimal group of test files that need to be run in order to trigger the error.

https://github.com/NNPDF/nnpdf/pull/2170/commits/a5ef4c2833b677e34b5eb5a0fe08cf6108031148

```
 test_datafiles.py  test_filter_rules.py  test_fitdata.py  test_metaexps.py
```

giacomomagni:
> I don't really have time to figure out the why, but the reason (a hash collision? but I checked and it doesn't seem to be the case?) is possibly intrincate enough to warrant some fun investigation in the future. So I leave here the commit and the minimal group of test files that need to be run in order to trigger the error.

but why was this happening only here?? 
Shall I merge `master` into this branch ?

scarlehoff:
I don't think it is worth it.

> but why was this happening only here??

I'm not sure. As I said it is a rabbit hole that I don't have time to enter right now, if I have to guess, by adding a few extra datasets you went over the n~128 which is the default limit of `lru_cache` (at which point starts to removing the old information).

giacomomagni:
As agrred with @enocera  we have added some modifications to the EMC and CHORUS data w.r.t the plain `hepdata` implementation. 

* For Chorus. We have decided to retain in the hepdata variant also  `CHORUSISOTARGCOR` and `CHORUSQEDRADCOR`, computed exactly as in the `legacy` variant. The first uncertainty is due the isoscalar target correction interpreted as uncertainty, while the secondo one takes into account possible QED effects (QED radiation correction interpreted as uncertainty).

* For EMC. We added to the `hepdata` implementation an addtional 15% "general" systematical uncertainty, which was taken as from buidmaster. We also take into account a 1.2% error from the Branching ration. 

This should conclude the revision of these datasets.  So the PR is ready for rewiev.

scarlehoff:
Thanks @giacomomagni should we merge #2192 here before doing any checks?

giacomomagni:
> Thanks @giacomomagni should we merge #2192 here before doing any checks?

sure, we can do it!

giacomomagni:
> A quick look seems ok (I have to check the failing fitbot, it seems vp crashed?) but I have two questions:
> 
> 1. Why is there no default variant? I would say that we want to use the dataset without having to specify `variant: xx`. Just select the variant that you consider the "most correct" and promote it to be the default.
> 
> 1.b Shouldn't we have a `variant: dw` for some of the nuclear uncertainties or in those cases we just need to use the legacy?

which dataset are you referring exactly? 

> 
> 2. Wouldn't it make more sense (for organizational purposes) to have the `_HEPDATA` datasets as a separate obaervable within the same folder as the others? I guess this is somewhat irrelevant, but given that the others are just a port I think they can live safely in the same folder.
> 

This is not poissbile because the kinematic is not the same, or the number of points was different, see the tables at the top of the PR... 

> RE the other PR, actually it might be possible to merge it directly to master so on second thoughts better finish it and merge the other by itself since it is basically harmless.

okay

scarlehoff:
> which dataset are you referring exactly?

All of them, you left `dataset_uncertainties` empty for all of them (including the ones that are purely new)

> This is not poissbile because the kinematic is not the same, or the number of points was different, see the tables at the top of the PR...

You can have two different observables with different kinematics. I'm not talking about variants. So you have `BCDMS_NC_NOTFIXED_D_EM-F2` and `BCDMS_NC_NOTFIXED_D_EM-F2-HEPDATA`, both under `BCDMS_NC_NOTFIXED_D`.

scarlehoff:
> Okay sorry, let me fix it and rerun the fitbot.

I'm not sure what the problem of the fitbot is

> But here we're going in loop:

No, we just meant different things by dataset*, I think you meant a "set of data" while I meant really a dataset name that you would put in a runcard.
What I'm suggesting is having two datasets: `BCDMS_NC_NOTFIXED_D_EM-F2` and `BCDMS_NC_NOTFIXED_D_EM-F2-HEPDATA` both part of the set `BCDMS_NC_NOTFIXED_D`.

*I know, the terminology has become impossible, probably the per-folder organization was not a good idea but now it's too late.

nnpdf_data/nnpdf_data/commondata/BCDMS_NC_NOTFIXED_D/metadata.yaml
```diff
@@ -77,3 +77,58 @@ implemented_observables:
...
+    units: ''
+  process_type: DIS_NC
+  tables: [2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23]
+  npoints: []
+  ndata: 159
+  data_uncertainties: []
+  data_central: data_rqcd.yaml
+  kinematic_coverage:
+    - x
+    - Q2
```
scarlehoff, line 93: Up to now we were using `Q`, this is one of the magic things that plotoptions was doing, the `k2` (which was indeed `Q2`) was being converted to `Q`.

So you have two options, if you leave it like this all new plots for DIS will have `Q2` instead of `Q`, i.e., 

image

will become

image

or you need to change all `Q2` to `Q` and do `sqrt` to the `kinematics` file.

I think the second option is better since that way plots don't change and are directly comparable, but while in other cases I have a strong opinion about it, DIS is the one case where I'd be fine with using Q2 in the plots...

Choose your poison.

giacomomagni: If the plots in Q2 are still readable (as the one you posted) I guess that's the quickest solution and then, if needed, one can convert them in Q in a subsequent PR (together with the collider DIS data, which should have the same problem, right?)

scarlehoff: Yes, but the collider DIS is still to be finished so 🤷‍♂️

But I'm ok with keeping it Q^2.

Btw, the fktables for the `hepdata` observables exist in some theory or are still to be done?

giacomomagni: To be done...

scarlehoff: Then I'll run a few more checks and merge. Thanks.

scarlehoff (approved):

PR #2176, opened 2024-10-16
Fixed target DIS DW datasets as variant

giacomomagni:
Move the DW fixed target DIS to be a variant.


giacomomagni:
I noticed that:

* BCDMS DW had different FKtables splitted by Q2, is that because we forgot to update 
the standard BCDMS sets? 

These 2 parameters are also different in the DW data and standard one: 
* `nnpdf_metadata.experiment`
* `observable_name.process_type`
can we add them to the variant option?

RoyStegeman:
I'm not sure if this is what you mean, but BCDMS has a different fktable for each beam energy because there is a degeneracy in Q2,x between them so pineappl cannot distinguish the measurements of a given kinematic point at different energies.

nnpdf_data/nnpdf_data/commondata/dataset_names.yml
```diff
@@ -213,62 +213,92 @@ CHORUSNB:
   dataset: CHORUS_CC_NOTFIXED_PB_NB-SIGMARED
   variant: legacy
 CHORUSNBPb_dw_ite:
-  dataset: CHORUS_CC_NOTFIXED_PB_DW_NB-SIGMARED
-  variant: legacy
+  dataset: CHORUS_CC_NOTFIXED_PB_NB-SIGMARED
+  variant: legacy_dw
+CHORUS_CC_NOTFIXED_PB_DW_NB-SIGMARED:
```
giacomomagni, line 218: This trick is not going to be enough here...

giacomomagni:
> I'm not sure if this is what you mean, but BCDMS has a different fktable for each beam energy because there is a degeneracy in Q2,x between them so pineappl cannot distinguish the measurements of a given kinematic point at different energies.

I've noticed that this has not been updated, because we usually use the DW version:

https://github.com/NNPDF/nnpdf/blob/deda5b8c89d31eee94eeaa450d66d756a4ec0dd8/nnpdf_data/nnpdf_data/commondata/BCDMS_NC_NOTFIXED_D/metadata.yaml#L65

RoyStegeman:
Ah that's your point. Yes you're right. I think you can just update the FKtables in the metadata? Because the difference is just in the uncertainties yaml. I would assume that in new theories the old fktables don't even exist

scarlehoff:
> nnpdf_metadata.experiment

Nop... this became a problem quite quickly @RoyStegeman :___

I'd say use the nuclear/deuteron type which should be more inclusive in what correlations are allowed (right @enocera ?)

> observable_name.process_type

Here even if they _used_ to be different, they can probably be made into the same one because the data is the same.

nnpdf_data/nnpdf_data/commondata/dataset_names.yml
scarlehoff: why?

giacomomagni: No ok it works! I'll try to fix the test asap and then we should be good

giacomomagni:
The 2 remaining tests: `test_pseudodata` and `test_overfit_metric` are failing because of the change in  `nnpdf_metadata.experiment`, I suspect (ie reverting the metadata makes them passing)

nnpdf_data/nnpdf_data/commondata/dataset_names.yml
giacomomagni: There were indeed som additional minimal modifications needed: https://github.com/NNPDF/nnpdf/pull/2176/commits/6750cf986324500c76dd45d82e3f3367f10b24f8

giacomomagni: yes the second one was needed to read new format filters of datasets (with `_DW_`) which are now not existing (ie are legacy).

scarlehoff: Ok, so it would've been needed sooner or later. Could you add a comment about that (I think eventually the "old path" might be removed but what you added I think should remain)

scarlehoff:
> are failing because of the change in nnpdf_metadata.experiment

Because that's making the random numbers change? (the seed takes the experiment name)

If so... I don't really see a way of fixing it without having two separate datasets, but if it is only the random numbers we can accept that it changed (this tag will not be random-seed-equivalent with the previous one anyway) although maybe we will need to redo some of the fits in the tests.

giacomomagni:
So I gave a try to add `experiment` in the variant options, 
somehow `PlottingOptions` still need a deepcopy somewhere, which I don't know how to fix.  
If a7294bdd31563caa961788c59d50bdc78e300a21 it's not going in the correct direction then I suspect we have to give up on this PR.

scarlehoff:
> somehow PlottingOptions still need a deepcopy somewhere, which I'm sure know how to fix

Why do you care about PlottingOptions? 

However... I'm wondering whether it wouldn't be better just to have them as separate datasets. The fact that two "variants" of the same _dataset_ can come from a different _experiment_ is... silly. We are just trying to get around a mistake of times past.

However, send me the runcard and I'll give it a go .

nnpdf_data/nnpdf_data/__init__.py
```diff
@@ -62,6 +62,10 @@ def new_to_legacy_map(dataset_name, variant_used):
 
         if new_name == dataset_name:
             matches.append(old_name)
+            # if it's a nuclear DIS data promote legacy to be legacy_dy
```
scarlehoff, line 65:
```suggestion
            # if it's a nuclear DIS data promote legacy to be legacy_dw
```

scarlehoff:
I think the runcards with _DW_ are not understood correctly by vp :/

giacomomagni:
> I think the runcards with _DW_ are not understood correctly by vp :/

PDFs are identical, but yes the chi2 tables are weird 😅

scarlehoff:
No, the problem is here `https://github.com/NNPDF/nnpdf/blob/87daeb1b25cc913fe45a6d7d8be779f80c7ebffc/validphys2/src/validphys/config.py#L465` since we already had a variant. Not sure how to get around this without adding too much `_DW_` specific code

The easy fix is to add:
```
# legacy_dw trumps everything
if variant is None or map_variant == "legacy_dw"
```

Btw, you need to add to this PR also the fixed target DY which also have DW variants.

giacomomagni:
> ```
> # legacy_dw trumps everything
> if variant is None or map_variant == "legacy_dw"
> ```
>
I see, Shall I go for it?

> Btw, you need to add to this PR also the fixed target DY which also have DW variants.

Yes, but these are trivial as they only have a DW legacy (as EMC), so it's just renaming but no duplication.

scarlehoff:
Yes, let's try...

scarlehoff:
Ok, this works now in terms of vp, however looking through the datafiles, they are actually not always the same, so you need to copy them as well for some of the datasets.

For instance, these two in master are not equal:

https://github.com/NNPDF/nnpdf/blob/master/nnpdf_data/nnpdf_data/commondata/NUTEV_CC_NOTFIXED_FE/data_legacy_NB-SIGMARED.yaml
https://github.com/NNPDF/nnpdf/blob/master/nnpdf_data/nnpdf_data/commondata/NUTEV_CC_NOTFIXED_FE_DW/data_legacy_NB-SIGMARED.yaml

The differences left in the fitbot are coming from there, but the fitbot does not include all datasets, so you should check. These can be added to the variant so it should just be a question of copying them over.

giacomomagni:
> Ok, this works now in terms of vp, however looking through the datafiles, they are actually not always the same, so you need to copy them as well for some of the datasets.
> 
 okay let me double check this.

giacomomagni:
Okay so somehow only Nutev had different central values. 
Here you have a small script to check:

```python
import yaml
import pathlib
import rich
import numpy as np

COMMONDATA_PATH = pathlib.Path(__file__).parent / "nnpdf_data/nnpdf_data/commondata"

def check_central(dataset, data_file):
    with open(COMMONDATA_PATH / f"{dataset}_DW" / data_file, encoding="utf-8") as f:
        central_dw = yaml.safe_load(f)['data_central']
    with open(COMMONDATA_PATH / dataset / data_file, encoding="utf-8") as f:
        central = yaml.safe_load(f)['data_central']

    try:
        np.testing.assert_allclose(central_dw, central)
    except AssertionError:
        rich.print(f"[red] {dataset}_DW has a different central value")
    return

if __name__ == "__main__":
    DATA_LIST = [
        "BCDMS_NC_NOTFIXED_D",
        "BCDMS_NC_NOTFIXED_P",
        "CHORUS_CC_NOTFIXED_PB",
        "NMC_NC_NOTFIXED",
        "NUTEV_CC_NOTFIXED_FE",
        "SLAC_NC_NOTFIXED_D",
        "SLAC_NC_NOTFIXED_P",
    ]
    for dataset in DATA_LIST:
        try:
            data_file = "data_legacy_EM-F2.yaml"
            check_central(dataset, data_file)
        except FileNotFoundError:
            data_file = "data_legacy_NB-SIGMARED.yaml"
            check_central(dataset, data_file)
            data_file = "data_legacy_NU-SIGMARED.yaml"
            check_central(dataset, data_file)

```

And the outcome of the test you suggested with the nutev data fixed:

Fit on master:
[datacuts_theory_fitting_pseudodata_table.csv](https://github.com/user-attachments/files/17488136/datacuts_theory_fitting_pseudodata_table.csv)

Fit on this branch:
[datacuts_theory_fitting_pseudodata_table.csv](https://github.com/user-attachments/files/17488178/datacuts_theory_fitting_pseudodata_table.csv)

PR #2177, opened 2024-10-16
Add t0theoryid for the correlated replicas method (CRM)

RoyStegeman:
This is needed for the correlated replicas method. 

Specifically, in the construction of the t0 covmat (which has to be the same for all values of alphas) we not only need a t0pdfset but also a t0theoryid to be the same for all values of alphas.

If we don't define a t0theoryid it uses the regular theoryid instead, which differs for different values of alphas.


validphys2/src/validphys/config.py
```diff
@@ -175,6 +175,22 @@ def parse_theoryid(self, theoryID: (str, int)):
                 str(e), theoryID, self.loader.available_theories, display_alternatives="all"
             )
 
+    def parse_t0theoryid(self, theoryID: (str, int)):
+        """A number corresponding to the database theory ID where the
+        corresponding theory folder is installed in te data directory.
```
scarlehoff, line 180:
```suggestion
        corresponding theory folder is installed in the data directory.
```

validphys2/src/validphys/config.py
```diff
@@ -661,14 +676,38 @@ def produce_dataset(
...
+        fit=None,
+        check_plotting: bool = False,
+    ):
+        """
+        Same as produce_dataset, but if a t0theoryid has been defined in the
+        runcard then those corresponding fktables will be linked.
+        """
+        ds = self.produce_dataset(
+            dataset_input=dataset_input,
+            theoryid=t0theoryid if t0theoryid else theoryid,
```
scarlehoff, line 703: Is it possible instead of doing the if-else here, for the parse rule to ensure that t0theoryid is equal to theoryid if not present?

If so I would prefer that.

validphys2/src/validphys/config.py
```diff
@@ -1628,13 +1667,12 @@ def produce_group_dataset_inputs_by_experiment(self, data_input):
...
+    def produce_scale_variation_theories(self, point_prescription, theoryid, t0theoryid=None):
         """Produces a list of theoryids given a theoryid at central scales and a point
-        prescription. The options for the latter are '3 point', '5 point', '5bar point', '7 point'
-        and '9 point'. Note that these are defined in arXiv:1906.10698. This hard codes the
-        theories needed for each prescription to avoid user error."""
+        prescription. The options for the latter are defined in pointprescriptions.yaml.
+        This hard codes the theories needed for each prescription to avoid user error."""
         pp = point_prescription
-        th = theoryid.id
+        th = t0theoryid.id if t0theoryid else theoryid.id
```
scarlehoff, line 1675: Same as above

validphys2/src/validphys/tests/test_pythonmakereplica.py
```diff
@@ -49,7 +49,6 @@ def test_commondata_unchanged(data_config, dataset_inputs, use_cuts):
     pre_mkrep_cvs = [deepcopy(cd.central_values) for cd in ld_cds]
     pre_mkrep_sys_tabs = [deepcopy(cd.systematics_table) for cd in ld_cds]
     pre_mkrep_cd_tabs = [deepcopy(cd.commondata_table) for cd in ld_cds]
-    API.make_replica(**config)
```
scarlehoff, line 52: why?

scarlehoff (commented):
Looks good to me. If it is possible to have the t0id = tid directly in the parse rule that would be great, as it avoid having an if-else every time one of the two should appear.

validphys2/src/validphys/config.py
achiefa: However, @scarlehoff's suggestion would make reading this part of the code less transparent. I honestly prefer Roy's solution, even if it is not "good-looking".

achiefa (commented):
All good by me, just a few comments following up from those of @scarlehoff.

scarlehoff:
>  honestly prefer Roy's solution, even if it is not "good-looking".

It's not because of the good looking. My suggestion goes in the spirit of separating completely what uses tid and what uses t0id.

For instance, right now `produce_t0dataset` needs a theory id even with a t0 theory is present, despite the theory id not being used in this second case.

achiefa:
I see, thanks for the clarification. I may agree, although I would state it clearly in the docstring if we do that at the parse rule level. Otherwise, a newcomer would not understand.

scarlehoff:
Agreed, if can be done (I'm not sure, produce-parse rules are something I still struggle with) it should be clear that when the t0 theoryid is not given, the normal theory id is used.

validphys2/src/validphys/config.py
RoyStegeman: This is what I wanted to do at first, but produce_ requires introducing another variable name - so it's t0theoryid in the runcard and as input to the production rule, but than the new variable should be .e.g produce_t0_theoryid. Which is done for example with t0pdfset and t0set, but I don't think is much better. 

parse_t0theoryid already requires the t0theoryid to exist, which is exactly not always the case

Since it only appears in two cases i went with the current solution, but I can do the first if you prefer

validphys2/src/validphys/tests/test_pythonmakereplica.py
RoyStegeman: Ah yes, I wasn't paying attention to context

validphys2/src/validphys/config.py
scarlehoff: Since the other case in which it appears is also t0, I think it makes sense to follow the same convention.

But :(
I think a nice addition to reportengine (cc @comane if you ever want to have a go at doing it) would be to level of parse rules so that a pase can depend on a previous parse. Maybe with a decorator `@priority(4)` or something like that.

validphys2/src/validphys/config.py
```diff
@@ -661,14 +660,37 @@ def produce_dataset(
...
+        *,
+        dataset_input,
+        t0id,
+        cuts,
+        use_fitcommondata=False,
+        fit=None,
+        check_plotting: bool = False,
+    ):
+        """
+        Same as produce_dataset, but if a t0id has been defined in the
```
scarlehoff, line 681:
```suggestion
        Same as produce_dataset, but if a ``t0theoryid`` has been defined in the
```

validphys2/src/validphys/config.py
```diff
@@ -1628,13 +1671,12 @@ def produce_group_dataset_inputs_by_experiment(self, data_input):
...
         return self.produce_group_dataset_inputs_by_metadata(data_input, "nnpdf31_process")
 
-    def produce_scale_variation_theories(self, theoryid, point_prescription):
+    def produce_scale_variation_theories(self, point_prescription, t0id):
         """Produces a list of theoryids given a theoryid at central scales and a point
-        prescription. The options for the latter are '3 point', '5 point', '5bar point', '7 point'
-        and '9 point'. Note that these are defined in arXiv:1906.10698. This hard codes the
-        theories needed for each prescription to avoid user error."""
+        prescription. The options for the latter are defined in pointprescriptions.yaml.
+        This hard codes the theories needed for each prescription to avoid user error."""
```
scarlehoff, line 1677: Actually I didn't pick this up before, why the scale variation use the t0 id and not the normal id?

RoyStegeman: Both can work since the theoryids used for the scale variations depend on what is in pointprescriptions.yaml and scalevariationstheoryids.yaml. In scalevariationstheoryids the same lists of variations are repeated many times for a different `theoryid`. But whether this `theoryid` is the t0id or the regular theoryid is an arbitrary choice that's up to us (and I think both options are fine), but the benefit of using the t0id, is that we don't have to copy the list for all theoryids correspdonding to the alphas variations

scarlehoff: Could you add a comment in the docstr with "note that this is using the t0theoryid, which by default is equal to theoryid but can be set separately" or something like this.

scarlehoff (approved):

PR #2179, opened 2024-10-17
removed unused function filter_closure_data

comane:
This PR simply removes one function that is unused from the filter.py module


scarlehoff (approved):
If the tests pass, I'm always happy with removing things. If someone needs it it can be added back.

PR #2181, opened 2024-10-17
Use eko v0.15

scarlehoff:
This allows the usage of this PR from eko https://github.com/NNPDF/eko/pull/415

Which makes (in my computer) a 100 replica fit evolve in ~2 minutes (instead of ~20m) (which would make #2168 unnecessary) but requires also https://github.com/NNPDF/eko/pull/417

So I'm leaving this PR here and depending on how things work out in eko we will merge this directly or #2168 first and then revert it when we merge this.

### DIFF
--- a/n3fit/src/evolven3fit/evolve.py
+++ b/n3fit/src/evolven3fit/evolve.py
@@ -8,6 +8,7 @@ from ekobox import apply, genpdf, info_file
 import numpy as np
 
 import eko
+from eko import basis_rotation, runner
 from eko.interpolation import XGrid
 from eko.io import manipulate
 from eko import basis_rotation, runner
@@ -26,14 +27,7 @@ LOGGING_SETTINGS = {
 
 
 def evolve_fit(
-    fit_folder,
-    q_fin,
-    q_points,
-    op_card_dict,
-    theory_card_dict,
-    force,
-    eko_path,
-    dump_eko=None,
+    fit_folder, q_fin, q_points, op_card_dict, theory_card_dict, force, eko_path, dump_eko=None
 ):
     """
     Evolves all the fitted replica in fit_folder/nnfit
@@ -128,12 +122,12 @@ def evolve_fit(
         for rep_idx in range(1, n_replicas + 1):
             # swap photon postion to match eko.basis_roation.flavor_basis_pids
             pdfgrid = np.array(initial_PDFs_dict[f"replica_{rep_idx}"]["pdfgrid"])
-            pdfgrid = np.append(pdfgrid[:,-1].reshape(x_grid.size,1), pdfgrid[:,:-1], axis=1)
+            pdfgrid = np.append(pdfgrid[:, -1].reshape(x_grid.size, 1), pdfgrid[:, :-1], axis=1)
             # and divide by x
-            all_replicas.append(pdfgrid.T / x_grid )
+            all_replicas.append(pdfgrid.T / x_grid)
 
         # reshape the xgrid eko if necessary
-        if not XGrid(x_grid) == eko_op.xgrid:
+        if XGrid(x_grid) != eko_op.xgrid:
             for _, elem in eko_op.items():
                 elem = manipulate.xgrid_reshape(
                     elem,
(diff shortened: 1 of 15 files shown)


scarlehoff:
Test to do:

- [ ] Evolve fit with 0.13 ekos
- [ ] Evolve fit with 0.14 ekos

scarlehoff:
In principle we could now continue with this using the master branch from eko, right @felixhekhorn ?

If @giacomomagni is fine with it, I'll rebase on top of master and resolve the conflicts so that this is ready to merge when the release of eko is out.

giacomomagni:
> In principle we could now continue with this using the master branch from eko, right @felixhekhorn ?
> 
yes that should work now! 

> If @giacomomagni is fine with it, I'll rebase on top of master and resolve the conflicts so that this is ready to merge when the release of eko is out.

Okay!

giacomomagni:
> Do we have an actual "official" theory with a 0.15 eko already? (otherwise I'll compute one with the latest commit of eko)

Not yet, mine were unofficial for sure!!

felixhekhorn:
> Do we have an actual "official" theory with a 0.15 eko already? (otherwise I'll compute one with the latest commit of eko)

if you want an official one, we should tag eko - which I believe we can do

and then we will break in v0.16 again (e.g. https://github.com/NNPDF/eko/issues/441 is an "open" breaking change and would a bugfix, e.g. https://github.com/NNPDF/eko/issues/447 , also be one?)

scarlehoff:
I think both the changes you linked would be very easy to work around. By official I meant "used in a paper" anyway.

scarlehoff:
afaics I get the exact same PDF with 0.14 and 0.15 ekos

I should test it also with a QED eko I guess, but I'll wait for the release to do that test...

n3fit/src/evolven3fit/evolve.py
```diff
@@ -107,10 +97,6 @@ def evolve_fit(
             raise ValueError(f"dump_eko not provided and {eko_path=} not found")
 
     with eko.EKO.edit(eko_path) as eko_op:
```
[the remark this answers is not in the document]
scarlehoff: But I guess it cannot be known until it is loaded, right?

scarlehoff:
Looking at the error in the CI, it seems to be due to the "low precision ekos", which were ekos for which the internal interpolation grid has only 15 points. This turns out to be a spurious error (the xgrid reshape should not trigger) because of a combination of https://github.com/NNPDF/eko/blob/c69a7b964a68342cfac728627be1a2c1084287ed/src/eko/io/v1.py#L26 and the eko for theory_399 being 0.13.4, for which `xgrid` was the internal interpolation instead of the externally-accessible grids. I'll create new "low precision theories" with newer ekos since 399/398 are outdated anyway, and for the time being I'll just work around that issue.

Besides that, (cc @felixhekhorn) I think it would be good for EKO to provide a `xgrid_reshape` which acts on the entire operator struct (and updates the metadata accordingly). The new interface is a bit clunky I feel and it is easy to forget steps (like updating the items in place or updating the xgrid in the operator, which were forgotten here indeed!)

n3fit/src/evolven3fit/evolve.py
```diff
@@ -106,18 +97,51 @@ def evolve_fit(
...
 
+        eko_original_xgrid = eko_op.xgrid
+        if XGrid(x_grid) != eko_original_xgrid:
+            # If the xgrid of the eko is not directly usable, construct a copy in memory
+            # by replacing the internal inventory of operators in a readonly copy
+            new_xgrid = XGrid(x_grid)
+            new_metadata = dataclasses.replace(eko_op.metadata, xgrid=new_xgrid)
+
+            new_operators = {}
+            for target_key in eko_op.operators:
```
felixhekhorn, line 117: what is `target_key` actually? the "canonical" way is to directly iterate `eko_op` (I would say), which should yield `ep -> elem`

n3fit/src/evolven3fit/evolve.py
```diff
@@ -106,18 +97,51 @@ def evolve_fit(
...
+                elem = eko_op[target_key.ep]
+
+                if eko_op.metadata.version == "0.13.4":
```
felixhekhorn, line 120: Do we need to support 0.13.4? EKO itself currently only supports `>=0.13.5` cc @evagroenendijk

felixhekhorn (commented):
If this works, I think it is fine for the moment - there is room for future improvements left

n3fit/src/evolven3fit/evolve.py
scarlehoff: I was debating myself whether to include this or not, but in the end I decided to be nice. There are some theories people use that have 0.13.4, and this will let them use it for some time (unless you are now forbidding these for being read at all, when I tested this PR they were read but they were not 100% correct).

n3fit/src/evolven3fit/evolve.py
scarlehoff: This is needed to construct a rotated copy of the original grid in memory, since I need to replace the `cache` variable of the operator.

felixhekhorn: not sure I understand, but if it works - fine. As said maybe something for later improvement ...

n3fit/src/evolven3fit/q2grids.py
```diff
@@ -1,12 +1,13 @@
...
-    Definition of default Q2 grids
+Definition of default Q2 grids
 
-    This file includes:
+This file includes:
 
-        ``Q2GRID_DEFAULT``: default NNPDF Q2 grid for evolution (55 points, starts at Q=1GeV)
-        ``Q2GRID_NNPDF40``: q2 grid used in the fits for the NNPDF4.0 release (49 points, starts at Q=1.65 GeV)
-        ``Q2GRID_Nf03``: q2 grid used in the perturvative charm fits for the NNPDF4.0 release (48 points, starts at Q=1GeV)
+    ``Q2GRID_DEFAULT``: default NNPDF Q2 grid for evolution (55 points, starts at Q=1GeV)
```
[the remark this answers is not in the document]
scarlehoff: boh, do I?

I think the evolven3fit module is not being parsed by the docs anyway :__

n3fit/src/evolven3fit/evolve.py
evagroenendijk: I think indeed they are read but a while ago when I tested versions <0.13.5 did not exactly reproduce the benchmark

n3fit/src/evolven3fit/q2grids.py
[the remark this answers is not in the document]
scarlehoff: I see (I didn't write them though, they just got caught by one of the pre-commit hooks).

I'll update the docs so that evolven3fit is also parsed so I can catch these as well.

n3fit/src/evolven3fit/evolve.py
felixhekhorn: I see.... but this way you loose the memory management of eko, as you are holding again all EKOs in memory, which for n3fit might still be a significant amount (do you know how many?). Again, if it works let's postpone improvements

scarlehoff: Take into account that this is almost never used.

scarlehoff:
@felixhekhorn should I understand your last msg as an approval?

felixhekhorn (approved):

PR #2182, opened 2024-10-18
add NNLOxQED scalevar to scalevariationtheoryids.yaml

RoyStegeman:
This is the NNLO counterpart of #2167

I also removed some outdated theoryids in an attempt to keep the file a bit clean


scarlehoff:
Wait, don't remove 600 _just yet_ they are an ugly mix of fktables coming from different theories but I think @t7phy is using them.

they will be overwritten by a new set of theories coming from the grids in https://github.com/NNPDF/theories_slim/pull/42

scarlehoff (approved):

PR #2183, opened 2024-10-18
add theory cards for alphas variations of QED theories (NNLO and N3LO)

RoyStegeman:
theories 40_003_xxx are based on theory 702 and theories 40_004_xxx are based on theory 824. In both cases only the alphas values are changed (and the `Comments`).


nnpdf_data/nnpdf_data/theory_cards/40004000.yaml
giacomomagni, line 1: Hi @RoyStegeman, since the EKO default choice of N3LO anomalous dimension will change in `v0.15` from 
`use_fhmruvv=False` to `use_fhmruvv=True`, maybe you want to add here that parameter, just to be on the safer side. (It has not been switched yet inside `nnpdf_data`, but you might want to be aware of this).

RoyStegeman: Thanks for pointing that out. It's already put in the `nnpdf_data` interface through which pineko has to interact with these theories. I say "has to" because these NNPDF theories are missing keys that eko/pineko expect so if these are used directly without `nnpdf_data` then pineko will raise an error.

This line sets the default: https://github.com/NNPDF/nnpdf/blob/87daeb1b25cc913fe45a6d7d8be779f80c7ebffc/nnpdf_data/nnpdf_data/theory.py#L84

I think that's enough? Though it does means that we will have to explicitly put `use_fhmruvv=True` in the NNPDF theorycards in which we do want to use their parametrisation.

giacomomagni: For the moment this is enough, but not in the future if we'll change the default here... 
My idea, when I changed the eko default, is that at some point we will drop `use_fhmruvv = False` 
and continue supporting only the version of `use_fhmruvv = True`.

RoyStegeman: Yes I imagine 4.1 will be done with FHMRUVV. Do you mean you plan to remove the implementation for your own parametrization from eko entirely? In that case these theorycards will be broken anyway (or at least they can't be used to produce the corresponding theory on the server), regardless if we put `use_fhmruvv` in the theory or not. 

I suppose we could put it in all the current N3LO theories for our own future reference as a reminder that these are done with a (at some point in the near future) not backward-compatible version of EKO.

giacomomagni: > I suppose we could put it in all the current N3LO theories for our own future reference as a reminder that these are done with a (at some point in the near future) not backward-compatible version of EKO.

yes that would be the best, thanks !!

RoyStegeman: Please check if you're happy with this

nnpdf_data/nnpdf_data/theory_cards/1000.yaml
```diff
@@ -1,6 +1,7 @@
 ID: 1000
 PTO: 3
 FNS: FONLL-E
+use_fhmruvv: False
```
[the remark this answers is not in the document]
RoyStegeman: ad variations are in the `Comments` perhaps this was from a time before we had it in the theorycards?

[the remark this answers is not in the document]
giacomomagni: yes indeed.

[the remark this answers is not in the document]
RoyStegeman: Okay I updated it, please check again if you're happy.

nnpdf_data/nnpdf_data/commondata/NNPDF_FIATLUX_NOTFIXED/kinematics.yaml
RoyStegeman, line 1: why is this so short?

scarlehoff: It's needed because the commondataparser requires a kinematics file but for this file is irrelevant since this is never read as a dataset

nnpdf_data/nnpdf_data/theory_cards/1023.yaml
```diff
@@ -44,5 +45,6 @@ SIN2TW: 0.23126
 TMC: 1
 MP: 0.938
 Comments: NNPDF4.0 N3LO ad variation (0, 0, 0, 0, 0, 0, 0), FHMV splitting
+n3lo_ad_variation: [0, 0, 0, 0, 0, 0, 0]
```
giacomomagni, line 48:
```suggestion
n3lo_ad_variation: [0, 0, 0, 0, 0, 0, 0]
n3lo_cf_variation: -1
```

nnpdf_data/nnpdf_data/theory_cards/1024.yaml
```diff
@@ -44,5 +45,6 @@ SIN2TW: 0.23126
 TMC: 1
 MP: 0.938
 Comments: NNPDF4.0 N3LO ad variation (0, 0, 0, 0, 0, 0, 0), FHMV splitting
+n3lo_ad_variation: [0, 0, 0, 0, 0, 0, 0]
```
giacomomagni, line 48:
```suggestion
n3lo_ad_variation: [0, 0, 0, 0, 0, 0, 0]
n3lo_cf_variation: 1
```

PR #2184, opened 2024-10-20
Apple Silicon GPU compatibility for Tensorflow

comane:
This pull request includes updates to the `doc/sphinx/source/n3fit/runcard_detailed.rst` file to clarify instructions for running parallel models and using GPUs on M1/M2 Macs.

Updates to parallel model instructions:

* Added a note that `savepseudodata` must be set to `false` in the `fitting` section of the runcard to run with parallel models. (`doc/sphinx/source/n3fit/runcard_detailed.rst`)

Updates for GPU usage on M1/M2 Macs:

* Added instructions to install specific packages (`tensorflow-deps`, `tensorflow-macos`, `tensorflow-metal`, and `wandb`) to run replicas in parallel using GPUs on M1/M2 Macs. (`doc/sphinx/source/n3fit/runcard_detailed.rst`)

### DIFF
--- a/doc/sphinx/source/n3fit/runcard_detailed.rst
+++ b/doc/sphinx/source/n3fit/runcard_detailed.rst
@@ -332,9 +332,9 @@ top-level option:
 
   parallel_models: true
 
-
-And then run ``n3fit`` with a replica range to be parallelized
-(in this case from replica 1 to replica 4).
+Note that currently, in order to run with parallel models, one has to set ``savepseudodata: false``
+in the ``fitting`` section of the runcard. Once this is done, the user can run ``n3fit`` with a 
+replica range to be parallelized (in this case from replica 1 to replica 4).
 
 .. code-block:: bash
 
--- a/doc/sphinx/source/n3fit/runcard_detailed.rst
+++ b/doc/sphinx/source/n3fit/runcard_detailed.rst
@@ -346,7 +346,7 @@ should run by setting the environment variable ``CUDA_VISIBLE_DEVICES``
 to the right index (usually ``0, 1, 2``) or leaving it explicitly empty
 to avoid running on GPU: ``export CUDA_VISIBLE_DEVICES=""``
 
-Note that in order to run the replicas in parallel using the GPUs of a M1/M2 Mac, it is necessary to also install 
+Note that in order to run the replicas in parallel using the GPUs of an Apple Silicon computer (like M1 Mac), it is necessary to also install 
 the following packages:
 
 .. code-block:: bash
@@ -355,6 +355,7 @@ the following packages:
    pip install tensorflow-macos==2.13.0 tensorflow-metal wandb==0.15.9
 
 
+See also the following issue for more information: `protobuf issue <https://github.com/wandb/wandb/issues/5935>`_.
 
 .. _otheroptions-label:
 
--- a/doc/sphinx/source/n3fit/runcard_detailed.rst
+++ b/doc/sphinx/source/n3fit/runcard_detailed.rst
@@ -346,6 +346,15 @@ should run by setting the environment variable ``CUDA_VISIBLE_DEVICES``
 to the right index (usually ``0, 1, 2``) or leaving it explicitly empty
 to avoid running on GPU: ``export CUDA_VISIBLE_DEVICES=""``
 
+Note that in order to run the replicas in parallel using the GPUs of a M1/M2 Mac, it is necessary to also install 
+the following packages:
+
+.. code-block:: bash
+
+   conda install -c apple tensorflow-deps
+   pip install tensorflow-macos==2.13.0 tensorflow-metal wandb==0.15.9
+
+
 
 .. _otheroptions-label:
 


doc/sphinx/source/n3fit/runcard_detailed.rst
```diff
@@ -346,6 +346,15 @@ should run by setting the environment variable ``CUDA_VISIBLE_DEVICES``
 to the right index (usually ``0, 1, 2``) or leaving it explicitly empty
 to avoid running on GPU: ``export CUDA_VISIBLE_DEVICES=""``
 
+Note that in order to run the replicas in parallel using the GPUs of a M1/M2 Mac, it is necessary to also install 
+the following packages:
+
+.. code-block:: bash
+
+   conda install -c apple tensorflow-deps
+   pip install tensorflow-macos==2.13.0 tensorflow-metal wandb==0.15.9
```
Radonirinaunimi, line 355: Why is it necessary to really pin this version?

Radonirinaunimi (commented):
Thanks @comane for this! Good to see that this is indeed what's needed to make it run.

Out of curiosity, how is the performance (how many replicas could you run, etc.)?

doc/sphinx/source/n3fit/runcard_detailed.rst
comane: It is related to this issue here: https://github.com/wandb/wandb/issues/5935

I was not able to make it run on MaC M2 GPUs with other versions.

comane:
> Out of curiosity, how is the performance (how many replicas could you run, etc.)?

The performance is not really good, at least on my laptop, it takes longer than using cpus. But this might be different for someone else with a more powerful mac

scarlehoff:
Out of curiosity, it doesn't work on M3 at all or you only had M1 and M2 to test?

comane:
> Out of curiosity, it doesn't work on M3 at all or you only had M1 and M2 to test?

I only tested for M2, but the above mentioned issue is for M1.
Maybe @ecole41, if time allows, can test it with her M3?

But I assume it works for M3 as well.

Radonirinaunimi:
> > Out of curiosity, how is the performance (how many replicas could you run, etc.)?
> 
> The performance is not really good, at least on my laptop, it takes longer than using cpus. But this might be different for someone else with a more powerful mac

When you say longer, how much is it? With how many replicas? (Maybe you are hitting memory bottleneck?)

But in any case, if it is not 4/5 times slower I'd say that's still good because you get all the replicas at then same time.

scarlehoff:
If @ecole41 can test it that would be great.

I'd suggest anyway changing from M1/M2 to something along the lines of "Apple Sillicon". 

> But in any case, if it is not 4/5 times slower I'd say that's still good because you get all the replicas at then same time.

I would say even 4/5 is still good. In my case, I can run entire fits in ~3 hours in my desktop's GPU, while a single replica takes about 40 minutes. It's about 5 times more but when the cluster is busy is the difference between having the fits ready in the same morning or one day later.

Radonirinaunimi:
> I would say even 4/5 is still good. In my case, I can run entire fits in ~3 hours in my desktop's GPU, while a single replica takes about 40 minutes. It's about 5 times more but when the cluster is busy is the difference between having the fits ready in the same morning or one day later.

That's absolutely true! My threshold was really pessimistic 😅

comane:
> When you say longer, how much is it? With how many replicas? (Maybe you are hitting memory bottleneck?)

Running with 10 replicas only on GPUs takes 15 minutes to get to epoch 4400 / 17000. If I run on cpu the same thing (still on my laptop) it takes 2 min 45 sec.
So, I think that at least on my computer it's more convenient to run things on CPU. 

> I would say even 4/5 is still good. In my case, I can run entire fits in ~3 hours in my desktop's GPU, while a single replica takes about 40 minutes. It's about 5 times more but when the cluster is busy is the difference between having the fits ready in the same morning or one day later.

@scarlehoff when you say on your desktop do you mean a MaC Os?

An interesting warning that I am getting is the following:
Pasted Graphic

scarlehoff:
> @scarlehoff when you say on your desktop do you mean a MaC Os?

Nop, a linux desktop with an nvidia gpu (at some point I tried it as well with an AMD one and it worked fwiw)

doc/sphinx/source/n3fit/runcard_detailed.rst
```diff
@@ -346,6 +346,15 @@ should run by setting the environment variable ``CUDA_VISIBLE_DEVICES``
 to the right index (usually ``0, 1, 2``) or leaving it explicitly empty
 to avoid running on GPU: ``export CUDA_VISIBLE_DEVICES=""``
 
+Note that in order to run the replicas in parallel using the GPUs of a M1/M2 Mac, it is necessary to also install
```
scarlehoff, line 349:
```suggestion
Note that in order to run the replicas in parallel on an Apple Silicon computer (like a M1 Mac) it is necessary to also install 
```

doc/sphinx/source/n3fit/runcard_detailed.rst
scarlehoff: Could you also add a reference to that issue in the docs?

scarlehoff (approved):
Thanks for testing this (and adding it to the docs!!!)

ecole41:
Hello @scarlehoff @comane @Radonirinaunimi , I tested this on M3 GPUs and it worked, the performance on GPUs was much slower than on CPUs on my Mac.
For 200 epochs and 100 replicas for the nnpdf40-like runcard:
- GPUs : 3800s
- CPUs: 2000s

ecole41:
Both GPU and CPU for 100 replicas

RoyStegeman:
Is that the timing only for the 200 replicas or does it include overhead?

ecole41:
I'm not sure how to check this, let me know if this helps:

This is the time is get for the GPU:
GPU

This is for the CPU:
CPU

scarlehoff:
I think it includes overhead. But in any case, it seems that running a fit on a Mac is not really going to be doable just yet :(

Maybe there's some low hanging fruit to improve it but not sure the effort is worth it

PR #2185, opened 2024-10-21
Tevatron inclusive DY

jacoterh:
This PR implements filter functions for the Tevatron inclusive DY datasets:
- CDF_Z0_1P96TEV_ZRAP: agreement with legacy implementation
- D0_WPWM_1P96TEV: agreement with legacy, except for possible rounding errors in the old implementation of the kinematics, see specifically the last two bins [here](https://github.com/NNPDF/nnpdf/compare/tevatron_inc_DY?expand=1#diff-3707afb74bc71486f8753643eea441ee88b3e8cba56677edd781367505cdd9e0) 
- D0_Z0_1P96TEV: agreement with legacy. One difference: systematic in 3rd bin is symmetric already on hepdata, while the legacy implementation seems to have symmetrised plus and minus uncertainties in the past here. This also affects the central value. 

General comment: small overall differences are due to formatting. For example: 1.2999999999999998 instead of 1.3. Is there any function that was used in the past to uniform the way floats were formatted?


Radonirinaunimi:
Thanks for this @jacoterh!
 
> Small overall differences are due to formatting. For example: 1.2999999999999998 instead of 1.3. Is there any function that was used in the past to uniform the way floats were formatted?

Now, you can use this function:
https://github.com/NNPDF/nnpdf/blob/bf8c89043987790bef41b9349b26369d65937951/nnpdf_data/nnpdf_data/filter_utils/utils.py#L402-L423

and call it as follows in the filter:
https://github.com/NNPDF/nnpdf/blob/bf8c89043987790bef41b9349b26369d65937951/nnpdf_data/nnpdf_data/commondata/ATLAS_1JET_8TEV_R06/filter.py#L7

Also, same as in the other PRs, could you check that the datasets are loaded properly and that the (t0) covmats are the same?

```python
from validphys.api import API
import numpy as np

inp1 = {"dataset_input": {"dataset": f"{new_implementation}"}, "theoryid": 40_000_000, "use_cuts": "internal", "t0pdfset": "NNPDF40_nnlo_as_01180", "use_t0": True}
inp2 = {"dataset_input": {"dataset": f"{old_implementation}", "variant": "legacy"}, "theoryid": 40_000_000, "use_cuts": "internal", "t0pdfset": "NNPDF40_nnlo_as_01180", "use_t0": True}

np.all(np.isclose(covmat1, covmat2))
np.all(np.isclose(t0_covmat1, t0_covmat2))
```

nnpdf_data/nnpdf_data/commondata/CDF_Z0_1P96TEV/metadata.yaml
```diff
@@ -5,21 +5,21 @@ nnpdf_metadata:
...
-  version: -1
+  url: https://www.hepdata.net/record/ins856131?version=1&table=Table%202
+  version: 1
 implemented_observables:
 - observable_name: ZRAP
   observable:
     description: Drell-Yan Rapidity Distribution
     label: CDF $Z$ rapidity (new)
     units: ''
   process_type: EWK_RAP
```
scarlehoff, line 21: Hi @jacoterh 

Remember to put as `process_type:` one of the processes that you will find in this file: https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/process_options.py (at the bottom, either DY_Z_Y or DY_W_ETA)

and also put `kinematic_override: identity`.

This is to avoid going through some old code that we are trying to remove.

When you do that you will see that that file defines which kinematic variables are allowed, sometimes it is necessary to change from Q -> Q^2 or pT -> pT^2 (the kinematic info in the end is the same,  but we are trying to standardize it a bit more)

jacoterh:
Thanks @Radonirinaunimi! The numbers have been formatted correctly now. I have also performed the check you proposed using the snippet you provided, see for instance [here](https://github.com/NNPDF/nnpdf/pull/2185/files#diff-582891e8e5178b770593c8381b3ac60c7c701f9f77fab622d50eabf2575024f0) . If you run this script, you'll find that the 3rd bin in D0_Z0_1P96TEV differs from the legacy implementation for the reason I mentioned in the beginning of this PR.

Some questions:
1. Why do the tests suddenly fail? Running `pytest tests/test_datafiles.py` locally works fine. 
2. I explicitly checked whether the comparison against the legacy implementation failed when I randomly changed some numbers. It did for all datasets, except for the 1st bin in D0_WPWM_1P96TEV. The test does neem to care about whatever I write here. This doesn't happen for the D0_Z0_1P96TEV and CDF_Z0_1P96TEV datasets.
3. In the metadata,  which hepdata table do I link for the url in case multiple entered the implementation?

nnpdf_data/nnpdf_data/commondata/CDF_Z0_1P96TEV/metadata.yaml
jacoterh: Ok, done!

scarlehoff:
The current failure is because it is `plotting::kinematics_override` (the probability that I get a variable name or word in general correct without looking in the first try is close to 0)

The previous one probably related to some change in digits looking at the logs.

jacoterh:
validphys reports:
- [legacy](https://vp.nnpdf.science/FlwD_DN4RmKw7cB3U89y_Q==/) 
- [new commondata ](https://vp.nnpdf.science/1RsjFWR7RimNLYEdbqWkcQ==/)

scarlehoff:
Thanks Jaco. One last thing, could you check that also the experimental covmat for the 3 datasets together doesn't change for non-diagonal entries? 

(you can also e.g. just check that the experimental chi2 for all 3 considered together is unchanged)

(this will be important for the singletop data as well, to check that possible experimental correlations across datasets are also captured)

jacoterh:
Sure, I made a plot of the ratio against the legacy implementation. There are no cross correlations between datasets here though (which I find odd as we have two datasets from D0), so the covmat is block diagonal and the datasets could be cross checked separately. I will also do this check for single top, thanks.
ratio_exp_covmat_tevatron_inc_dy

scarlehoff (commented):
Hi @jacoterh 

Thanks for this. All seems fine, the only thing, could you bump the `version` and change the `version_comment` in the `metadata.yml` files?

jacoterh:
@scarlehoff I have tried to follow your steps but I am a bit lost. This is what I did.

```
vp-get fit pseudodata_test_fit_n3fit_240916 
vp-fitrename pseudodata_test_fit_n3fit_240916 pseudodata_test_fit_n3fit_241115
cp pseudodata_test_fit_n3fit_241115/filter.yml pseudodata_test_fit_n3fit_241115.yml  
``` 

and launched the fit with 
```
n3fit pseudodata_test_fit_n3fit_241115.yml 10
```

This gives me a fitted replica, but then what do I commit to make sure the tests don't fail anymore?

jacoterh:
Thanks, I ran the fit on 20 replicas, evolved them, but now postfit filters out 90% of them: 
`[INFO]: 2 replicas pass IntegNumber_3`.
Should I just fit more replicas, i.e. is this expected? 90% seems a bit harsh.

scarlehoff:
Yes, you don't even need to finish the fit. That's why if you use fit rename you get the previous one and just take need the new data csvs

scarlehoff:
Let me know whether I can merge this.

nnpdf_data/nnpdf_data/commondata/CDF_Z0_1P96TEV/metadata.yaml
```diff
@@ -53,9 +53,5 @@ implemented_observables:
     - - CDFZRAP_NEW-CDF_ZRAP_MCgrid_500M_last_two_bins_combined
   data_uncertainties:
     - uncertainties_ZRAP.yaml
-  variants:
-    legacy:
-      data_uncertainties:
-      - uncertainties_legacy_ZRAP.yaml
```
scarlehoff, line 58: You need to leave the variant here for now ^^U
but just make it point to the same `- uncertainties_ZRAP.yaml`

Once all datasets are done we can remove the variant (but many of them will need to remain... pointing to old fktables which might have different names)

nnpdf_data/nnpdf_data/commondata/D0_WPWM_1P96TEV/metadata.yaml
```diff
@@ -9,6 +9,7 @@ arXiv:
   journal: Phys. Rev. D91 (2015) 032007
 iNSPIRE:
   url: https://inspirehep.net/literature/1253555
+  version: 1
```
scarlehoff, line 12: I mean the `version_comment` in the top of the file, but I could not leave the comment there.

nnpdf_data/nnpdf_data/commondata/D0_Z0_1P96TEV/metadata.yaml
```diff
@@ -9,6 +9,7 @@ arXiv:
   journal: Phys. Rev. D76 (2007) 012003
 iNSPIRE:
   url: https://inspirehep.net/literature/744624
+  version: 1
```
scarlehoff, line 12: same

jacoterh:
For future reference, this is the (hopefully) final validphys comparison: https://vp.nnpdf.science/PceSQWJaSLCrCp7wHzrsWg==/

If the tests pass, this branch should be ready for merging

scarlehoff (approved):

PR #2187, opened 2024-10-27
Add uncertainty variants for select datasets

t7phy:
Adds uncertainty variants for some gluon-PDF sensitive processes where luminosity uncert is excluded, for reproducibility of pheno paper


nnpdf_data/nnpdf_data/commondata/ATLAS_2JET_13TEV_DIF/filter.py
```diff
@@ -91,3 +93,29 @@ def processData():
...
+
+
+def lumiless(source_file, destination_file):
+    shutil.copy(source_file, destination_file)
+    with open(destination_file, 'r') as file:
+        content = file.read()
+    content = content.replace(search_block, replace_block)
+    with open(destination_file, 'w') as file:
+        file.write(content)
+
```
scarlehoff, line 119: Please, thes efunctions which are all equal or very similar, lift them to the utils module.

(Also, instead of running isort and black separately, you can use pre-commit before doing commits!)

scarlehoff:
hi @t7phy for the datasets for which the uncertainty is well separated you can create two files

- uncertainties_wo_lumi (which you already have)
- lumi_uncertainty 

And then the data_uncertainties entry can be a list and they will be concatenated. That way the information doesn't need to be repeatead in the file with and without uncertainties (and the size of these files will be reduced conaiderably)

t7phy:
@scarlehoff but we need to have them as separate variants otherwise how do you compute the z-score where the lumi is skipped?

nnpdf_data/nnpdf_data/commondata/ATLAS_1JET_13TEV_DIF/uncertainties_wo-lumi.yaml
```diff
@@ -1346,7 +1346,7 @@ definitions:
   syst_lumi:
     description: ''
     treatment: MULT
-    type: CORR
+    type: SKIP
```
scarlehoff, line 1349: In the spirit of making the files a bit smaller, could you also remove the lines with `syst_lumi` from this file? (that way we don't need to even have the type skip here)

nnpdf_data/nnpdf_data/filter_utils/utils.py
```diff
@@ -421,3 +424,50 @@ def prettify_float(dumper, value):
...
+    if 'definitions' in content:
+        definitions = content['definitions']
+        if uncert_name in definitions and 'type' in definitions[uncert_name]:
+            content_uncert['definitions'] = {uncert_name: copy.deepcopy(definitions[uncert_name])}
+            content_uncert['bins'] = {}
+            definitions[uncert_name]['type'] = 'SKIP'
+            bins = []
+            for i in range(len(content['bins'])):
+                bins.append({uncert_name: content['bins'][i][uncert_name]})
+            content_uncert['bins'] = bins
```
scarlehoff, line 467: Basically dropping the lumi bin from the content here.

scarlehoff (commented):
Thanks for this. Just a small comment.

However, wouldn't it be better to add this during the filter or is there a reason why this cannot be done in this way?

So, while running the filter, instead of creating the full file and afterwards separating the lumi; can't it be separated from the beginning?

Issue #2162, opened 2024-10-02
Update matplotlib


scarlehoff:
We have a dependency pin with matplotlib because there was _some plot_ that didn't work with a recent update. Is this still the case?

Adding you @Radonirinaunimi since I think you already looked into this?

Radonirinaunimi:
I will look back into this ASAP.

PR #2191, opened 2024-10-30
Update `matplotlib` version

Radonirinaunimi:
A simple change to address https://github.com/NNPDF/nnpdf/issues/2162 and the issue reported in https://github.com/NNPDF/nnpdf/pull/1809 related to the axis transformation for `matplotlib >= 3.8`. This also seems to fix the non-deterministic failing tests in https://github.com/NNPDF/nnpdf/pull/2099 (at least for me locally).

PS: Need to be tested further.


scarlehoff:
Nice! I would be happy with merging it and crossing our fingers since it seems to work also in the ci

Radonirinaunimi:
> Nice! I would be happy with merging it and crossing our fingers since it seems to work also in the ci

I still need to perform further tests but indeed it seems to work. After understanding the issues, I think it happened that magically it worked before `v3.8`.

scarlehoff:
I'd be happy merging this even if it doesn't fix the random failure. If it does, so much the better.

What @RoyStegeman said about the DW variants worries me because those are relatively recent, if it fails again I'll have a deeper look.

Thanks!

Radonirinaunimi:
I had to re-run the fitbot because the first one was started before the change in the `conda-recipe/meta.yaml`. So the last reports are the correct ones.

So , yes, this basically solves the issues with the offset plotting with a still minor issue (which should be easy to solve) that for some very few ratio plots there are points that are cut ways: for e.g. [reference](https://vp.nnpdf.science/upbwXyHTSNK-7FjKEeb2zw==/figures/matched_datasets_from_dataspecs7_dataset_report_Datanorm_plot_fancy_dataspecs_0.pdf) vs [now](https://vp.nnpdf.science/jbCBEMzMQAS9btI69QF3nQ==/figures/matched_datasets_from_dataspecs7_dataset_report_Datanorm_plot_fancy_dataspecs_0.pdf).

scarlehoff:
i see, do you have an idea of how to solve that? Re-axing (if that's a thing?)

Radonirinaunimi:
> i see, do you have an idea of how to solve that? Re-axing (if that's a thing?)

Not yet xd but I will have a look in a while. It shouldn't be difficult.

RoyStegeman:
Something I didn't realise before: I can reproduce this error locally (while yesterday I said I couldn't), but while it occurs if I run all tests it doesn't if I try to run only the failing tests

Radonirinaunimi:
So I haven't found a fancy way to construct the ranges so the $y$ limits are computed manually [f7c6595](https://github.com/NNPDF/nnpdf/pull/2191/commits/f7c659521f934cc45bd47366ceefb2901b792c8f).

> Something I didn't realise before: I can reproduce this error locally (while yesterday I said I couldn't), but while it occurs if I run all tests it doesn't if I try to run only the failing tests

Just to confirm that locally, sometimes, I am still experiencing this issue. So updating dependency versions definitely does not _magically_ solve it.

Radonirinaunimi:
> Do you experience the issue also if you only run the failing tests?

I've never experienced it when running separately the failing tests. But also it's not every time that tests are failing when I run everything all at once. It seems to occur randomly when running all the tests altogether...

RoyStegeman:
Yes same here

Radonirinaunimi:
> Yes same here, I tried two more times running all locally but they just passed

Hmhmh, this issue deserves a closer look. If we are happy with the last fitbot, we can merge this and investigate the issue in another PR.

RoyStegeman:
I think the problem might be the caching of `parse_new_metadata`. I'll have a closer look tomorrow

scarlehoff:
The latest fitbot actually picked up a problem!

RE the failure, I agree, let's investigate it somewhere else.

Radonirinaunimi:
> The latest fitbot actually picked up a problem!
> 
Yes, the error is that for some instances all the values are all `NaN`s, so it will be sufficient to simply skip these instances when setting `ax.get_ylim`. I missed this before because I haven't checked all the datasets that are run/fitted by the bot.

validphys2/src/validphys/tests/test_plots.py
```diff
@@ -41,6 +41,21 @@ def test_dataspecschi2():
...
+@pytest.mark.linux
+@pytest.mark.mpl_image_compare(tolerance=TOLERANCE_VALUE)
+def test_plotfancy():
+    return next(
+        iter(
+            API.plot_fancy(
+                dataset_input={'dataset': 'NMC_NC_NOTFIXED_P_EM-SIGMARED', 'variant': 'legacy'},
+                theoryid=THEORYID,
+                pdfs=[PDF],
+                use_cuts='internal',
```
scarlehoff, line 53: Could you use one of the data configs from comftest.py? (perhaps with minimal modifications)

Radonirinaunimi: Yes, I can do that. This was first a test to check if what I generated locally could match the CI....and as always I seem to have really different architecture/configurations from the CI...

scarlehoff: I think we are even skipping the matplotlib tests in mac because of how hard it was to make them compatible with linux...

Radonirinaunimi: But the weird thing is that it is passing on macOS while failing on ubuntu 😅 I will investigate later.

scarlehoff:
This is still not working completely right :/ https://vp.nnpdf.science/ju590ltLStii5BG9mkW-Hw==/matched_datasets_from_dataspecs24_dataset_report_report.html#normalized

Radonirinaunimi:
> I've pushed my solution (+ rebase on top of master). Basically I use the existing ymax/ymin instead of recalculating it, seems to work locally... let's see whether the bot thinks the same thing and we are not just playing whack-a-mole where fixing one plot breaks the following one.
>
This is what I first tried before but didn't work **locally** because (back then if I remember correctly the issue was) the values of the `mins/maxs` were computed based on the data `cv+/-err` and therefore might cut away the predictions. But looking at the reports, everything seems to be perfect now. So, good for it.

The only thing remaining here is updating the regression plot.

scarlehoff:
I'll try to update the plot later. The change is basically a shift, maybe there's a way of making the test more stable.

scarlehoff:
Since the python installation is already checking the plot, I put a less restrictive check for the conda installation (I'll increase it until the tests pass).

scarlehoff (approved):
LGTM! If you are happy with my change to the tests please merge ^^

I've checked all plots in the bot's report and I think they are ok, but at some point the points were dancing around my eyes... I hope I haven't missed anything.

Radonirinaunimi:
> I've checked all plots in the bot's report and I think they are ok, but at some point the points were dancing around my eyes... I hope I haven't missed anything.

I've also gone through all of them before, so let's hope that the parts where the points started dancing in our case did not overlap xD

I'm merging now.

PR #2192, opened 2024-10-30
Exterminate k1,k2,k3 from Fixed target data

giacomomagni:
This PR is to remove the `k1,k2,k3` variables from the Fixed target legacy dataset.

I split the commit here, in case there are further changes needed in the plotting options.


scarlehoff:
This looks good, thank you!

The error is weird, and I cannot reproduce it... I wonder whether it is related to the other error we see ramdomly...

scarlehoff:
@giacomomagni can we try to rebase both this and the parent PR on master to see whether the problems dissapear?

PR #2196, opened 2024-11-01
dynamically set year for the docs

RoyStegeman:
For some reason `%Y` as suggested in the [sphinx documentation](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-project_copyright) doesn't seem to work


scarlehoff (approved):
although the previous one at least gave an idea of when was the last time the documentation was up-to-date

RoyStegeman:
🥲

PR #2198, opened 2024-11-05
Sum covmats from various point prescriptions

RoyStegeman:
We now have various sources of uncertainty as the default option (IHOU DIS, IHOU DGLAP, MHOU 7pt) and various other sources of uncertainty that we regularly use. This number will only grow as we use the theory covariance method for the determination/implementation of higher twist, alphas, mtop, etc. 

To avoid an exponentially growing `pointprescriptions.yaml` as a result of having to implement a new point prescription for each combination of uncertainties that we sum, this PR enables to put a list of `point_prescriptions` in the n3fit runcard instead of a single `point_prescription`. These point prescriptions each have their own associated covmat where then the sum of them is stored. 

We also store the individual covmats because the theory covariance method involves recomputing the theory covmat of the relevant parameter, so it's a good consistency check to see if the same is reproduced e.g. in case a different commit is used between fit and post-fit analysis using the theory covariance method. 

As a proof of concept here is a N3LO QED+MHOU fit report comparing the current master branch to this branch: https://vp.nnpdf.science/B8uVm8CQQPK1LDNJoKdvJg==/

To see that the covmats are the same for both fits, you can use the following script
<details>
  <summary>Code snippet</summary>

  ```python
import numpy as np
from validphys.api import API
import pandas as pd

one_covmat_fit = API.fit(fit="241104-02-rs-n3lo-mhou-qed") # done with current master
multiple_covmat_fit = API.fit(fit="241104-01-rs-n3lo-mhou-qed-sumcovmat") # done with update_th_covmat branch

one_covmat = pd.read_csv(
    one_covmat_fit.path / "tables/datacuts_theory_theorycovmatconfig_theory_covmat_custom.csv",
    index_col=[0, 1, 2],
    header=[0, 1, 2],
    sep="\t|,",
    engine="python",
).fillna(0)

summed_covmat = pd.read_csv(
    multiple_covmat_fit.path / "tables/datacuts_theory_theorycovmatconfig_theory_covmat_custom.csv",
    index_col=[0, 1, 2],
    header=[0, 1, 2],
    sep="\t|,",
    engine="python",
).fillna(0)

print(np.allclose(one_covmat, summed_covmat, rtol=1e-2))
```
</details>

It does break backward compatibility and old runcards can not be used to run n3fit, though old fits can still be analyzed in the same way since the name of the theory covmat has not changed. 

In particular the following changes impact how to write an n3fit runcard:
1) the `point_prescription` key is no longer supported and only `point_prescriptions` is, which takes a list of point prescriptions. Of course, that list can contain only a single point prescription. 
2) theoryids are always read from `scalevariationtheoryids.yaml` so there is no need to explicitly declare it each time, i.e. no more
```  
theoryids:
    from_: scale_variation_theories
``` 
3) some point prescriptions that are sums of various sources of uncertainty have been removed and are now only available as the sum of other point prescriptions, for a translation table from previously implemented point prescriptions in terms of the new see the bottom of `pointprescriptions.yaml`.

### CODE REVIEW

File: validphys2/examples/data_theory_comparison_w_sv.yaml [hunk 1]
```diff
@@ -7,8 +7,6 @@ use_cuts: "internal"
 
 theoryid: 717 # define the central theory
 point_prescription: "3 point"
-theoryids:
```
scarlehoff, line 10: Does this runcard work even if you don't have a list?

RoyStegeman: Yes, all that changed is that now `theory_covmat_custom` (the function that creates the covmat for `vp-setupfit`) is collected over point_prescriptions, but most other actions still take a regular point_prescription as input so their functionality hasn't changed.

scarlehoff: So the reason the old runcards cannot run is just because the `from_` stuff is gone?

RoyStegeman: Yes it would clash with the produce_theoryids that this PR introduces

validphys2/src/validphys/scalevariations/pointprescriptions.yaml
```diff
@@ -1,29 +1,32 @@
...
 # Polarized missing: 3 point renormalization scales for JETS and DIJETS
 '3pt missing': ['(1, 1)', '(1, 0.5 missing)', '(1, 2 missing)']
+
+# n3lo ad covmat            -> n3lo ad ihou
+# n3lo ihou                 -> n3lo ad ihou + n3lo dis ihou
+# n3lo full thcovmat        -> n3lo ad ihou + n3lo dis ihou + n3lo 7 point
+# n3lo 3pt missing          -> n3lo ad ihou + n3lo dis ihou + n3lo 3pt missing
+# n3lo 3pt hadronic         -> n3lo ad ihou + n3lo dis ihou + n3lo 3pt hadronic
+# n3lo fhmv full thcovmat   -> n3lo fhmv ihou + n3lo dis ihou + n3lo 7 point
+# n3lo fhmv 3pt hadronic    -> n3lo fhmv ihou + n3lo 3pt hadronic + n3lo 3pt hadronic fhmv + n3lo dis ihou fhmv
```
giacomomagni, line 32:
```suggestion
# n3lo fhmv 3pt hadronic    -> n3lo fhmv ihou + n3lo dis ihou + n3lo 3pt hadronic 
```

giacomomagni (commented):
Thanks @RoyStegeman, it looks much nicer now!

This prescription could also be removed as is the same as `covmat_3pt`, no? 

https://github.com/NNPDF/nnpdf/blob/15d26d414fcfdac10ca6a4192d6b527fb0d2b4b6/validphys2/src/validphys/theorycovariance/construction.py#L97-L105

validphys2/src/validphys/config.py
```diff
@@ -1696,44 +1701,54 @@ def produce_scale_variation_theories(self, point_prescription, t0id):
...
+                for scale in scales:
+                    if scale not in available_scales:
+                        missing_scales.append(scale)
+                missing_scales_string = ", ".join(missing_scales)
+                raise ConfigError(
+                    "For this central theoryid, the requested point prescription is not currently "
+                    + "available. To use this point prescription for this central theoryid, theoryids "
+                    + "that correspond to the following scale choices must be created and added to "
+                    + "validphys2/src/validphys/scalevariations/scalevariationtheoryids.yaml: "
+                    + f"(k_F, k_R) = {missing_scales_string}."
```
giacomomagni, line 1741:
```suggestion
                    + f"missing variations = {missing_scales_string}."
```

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -516,13 +470,23 @@ def total_theory_covmat(theory_covmat_custom, user_covmat):
...
+    )
     return df
 
 
+theory_covmats_fitting = collect(theory_covmat_custom_per_prescription, ("point_prescriptions",))
+
+
+@table
+def theory_covmat_custom(theory_covmats_fitting):
+    return sum(theory_covmats_fitting)
```
giacomomagni, line 487:
```suggestion
    """Sum all the theory covmat listed in `point_prescriptions` ."""
    return sum(theory_covmats_fitting)
```

validphys2/src/validphys/scalevariations/pointprescriptions.yaml
```diff
@@ -1,29 +1,29 @@
-# Mapping between a point prescription and the scale combinations that form it
-# IMPORTANT: scale combinations must be listed according to (k_F, k_R) in the following order:
-# (1, 1), (2, 1), (0.5, 1), (1, 2), (1, 0.5), (2, 2), (0.5, 0.5), (2, 0.5), (0.5, 2)
+# Shifts are always computed with respec to the first item/theory in the list
```
scarlehoff, line 1:
```suggestion
# Shifts are always computed with respect to the first item/theory in the list
```

validphys2/src/validphys/config.py
```diff
@@ -210,13 +210,21 @@ def produce_replicas(self, nreplica: int):
...
     def produce_inclusive_use_scalevar_uncertainties(
-        self, use_scalevar_uncertainties: bool = False, point_prescription: (str, None) = None
+        self,
+        use_scalevar_uncertainties: bool = False,
+        point_prescriptions: (list, None) = None,
+        point_prescription: (str, None) = None,
     ):
         """Whether to use a scale variation uncertainty theory covmat.
         Checks whether a point prescription is included in the runcard and if so
         assumes scale uncertainties are to be used."""
```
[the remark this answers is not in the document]
RoyStegeman: `point_prescription` shouldn't be here anymore. I probably left it originally because I was considering supporting both

validphys2/src/validphys/config.py
```diff
@@ -210,13 +210,21 @@ def produce_replicas(self, nreplica: int):
...
-        if (not use_scalevar_uncertainties) and (point_prescription is not None):
+        if (not use_scalevar_uncertainties) and (
+            point_prescriptions is not None or point_prescription is not None
+        ):
```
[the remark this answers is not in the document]
RoyStegeman: yes I think the use_scalevar uncertainties is an unnecessary complexity

validphys2/src/validphys/config.py
```diff
@@ -1671,11 +1675,12 @@ def produce_group_dataset_inputs_by_experiment(self, data_input):
     def produce_group_dataset_inputs_by_process(self, data_input):
         return self.produce_group_dataset_inputs_by_metadata(data_input, "nnpdf31_process")
 
-    def produce_scale_variation_theories(self, point_prescription, t0id):
+    def produce_point_prescriptions_theoryids(
+        self, t0id, point_prescription, point_prescriptions=None
+    ):
```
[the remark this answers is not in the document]
RoyStegeman: `point_prescriptions` is basically only used in theory_covmat_custum (which has been redefined to be the sum of covmats generated by looping over point_prescriptions). Everywhere else in vp point_prescription is used and functionality doesn't change

validphys2/src/validphys/config.py
```diff
@@ -1696,44 +1701,54 @@ def produce_scale_variation_theories(self, point_prescription, t0id):
...
+                    "Scale variations are not currently defined for this point prescription. This "
+                    + "configuration only works when 'point_prescription' is equal to one of the "
+                    + f"following: {valid_pps}. Please use one of these instead if you wish to "
+                    + "include theory uncertainties here."
+                )
+
+            # Get dictionary containing theoryid and variations for central theory from runcard
+            for scalevarsfor_dict in scalevarsfor_list:
+                if scalevarsfor_dict["theoryid"] == int(th):
+                    theoryid_variations = scalevarsfor_dict
```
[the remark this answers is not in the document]
RoyStegeman: I removed the loop so this problem is no longer there

validphys2/src/validphys/config.py
```diff
@@ -210,13 +210,16 @@ def produce_replicas(self, nreplica: int):
...
+
     def produce_inclusive_use_scalevar_uncertainties(
-        self, use_scalevar_uncertainties: bool = False, point_prescription: (str, None) = None
+        self, use_scalevar_uncertainties: bool = False, point_prescriptions: (list, None) = None
     ):
         """Whether to use a scale variation uncertainty theory covmat.
         Checks whether a point prescription is included in the runcard and if so
         assumes scale uncertainties are to be used."""
-        if (not use_scalevar_uncertainties) and (point_prescription is not None):
+        if not use_scalevar_uncertainties and point_prescriptions is not None:
```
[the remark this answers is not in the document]
RoyStegeman: I will still address that, but I think that will involve cleaning up how the n3fit config is written since I think that can be done in a less convoluted way, but I'll have to check

scarlehoff:
Other than the tests this looks good. I'm submitting the bot just in case there is an expected change in behaviour somewhere.

> It does break backward compatibility and old runcards can not be used to run n3fit, though old fits can still be analyzed in the same way since the name of the theory covmat has not changed.

This is important, would it be possible to catch early the error in the old fit and print a message saying what to change? Since some students may want to reproduce eg `NNPDF40_nnlo_as_mhou` and will notice it doesn't work and stop there. We don't need to catch every possible variation, but it should be able to tell you what to do for the few fits we have released.

RoyStegeman:
> This is important, would it be possible to catch early the error in the old fit and print a message saying what to change? 

Yes I wasn't sure it would be necessary, but it's also quick to add so I'll do that

validphys2/src/validphys/scalevariations/pointprescriptions.yaml
```diff
@@ -1,29 +1,29 @@
...
+# IHOU DIS massive coefficient functions
+'dis ihou': ['central DIS','lower bound DIS','upper bound DIS']
+# 3 point renormalization scale variations for subset of data
+'3pt missing': ['(1, 1)','(1, 0.5 missing)', '(1, 2 missing)']
+# 3 point scale renormalization variations for hadronic datasets
+'3pt hadronic': ['(1, 1)','(1, 0.5 hadronic)', '(1, 2 hadronic)']
+# FHMV splitting function IHOUs
+'fhmv ihou': ['(0, 0, 0, 0, 0, 0, 0)','(1, 0, 0, 0, 0, 0, 0)','(2, 0, 0, 0, 0, 0, 0)','(0, 1, 0, 0, 0, 0, 0)','(0, 2, 0, 0, 0, 0, 0)', '(0, 0, 1, 0, 0, 0, 0)','(0, 0, 2, 0, 0, 0, 0)','(0, 0, 0, 1, 0, 0, 0)','(0, 0, 0, 2, 0, 0, 0)','(0, 0, 0, 0, 1, 0, 0)','(0, 0, 0, 0, 2, 0, 0)','(0, 0, 0, 0, 0, 1, 0)','(0, 0, 0, 0, 0, 2, 0)','(0, 0, 0, 0, 0, 0, 1)','(0, 0, 0, 0, 0, 0, 2)']
```
giacomomagni, line 21: Since we are updating names in any case...

```suggestion
'fhmruvv ihou': ['(0, 0, 0, 0, 0, 0, 0)','(1, 0, 0, 0, 0, 0, 0)','(2, 0, 0, 0, 0, 0, 0)','(0, 1, 0, 0, 0, 0, 0)','(0, 2, 0, 0, 0, 0, 0)', '(0, 0, 1, 0, 0, 0, 0)','(0, 0, 2, 0, 0, 0, 0)','(0, 0, 0, 1, 0, 0, 0)','(0, 0, 0, 2, 0, 0, 0)','(0, 0, 0, 0, 1, 0, 0)','(0, 0, 0, 0, 2, 0, 0)','(0, 0, 0, 0, 0, 1, 0)','(0, 0, 0, 0, 0, 2, 0)','(0, 0, 0, 0, 0, 0, 1)','(0, 0, 0, 0, 0, 0, 2)']
```

RoyStegeman:
@scarlehoff I added some tests and the error if people use an old runcard, if you're happy I think this can be merged

n3fit/src/n3fit/scripts/vp_setupfit.py
```diff
@@ -153,7 +153,14 @@ def from_yaml(cls, o, *args, **kwargs):
             filter_action = 'datacuts::theory::fitting filter'
             check_n3fit_action = 'datacuts::theory::fitting n3fit_checks_action'
         SETUPFIT_FIXED_CONFIG['actions_'] += [check_n3fit_action, filter_action]
-        if file_content.get('theorycovmatconfig') is not None:
+        if (thconfig := file_content.get('theorycovmatconfig')) is not None:
+            if thconfig.get('point_prescription') is not None:
+                raise ConfigError(
+                    "`point_prescription` is deprecated in favor of a list of "
```
scarlehoff, line 159:
```suggestion
                    "`point_prescription` has been removed in favor of a list of "
```

PR #2203, opened 2024-11-08
Old commondata thcovamat compatibility

giacomomagni:
This is needed in order to read the old commondata theory covmat.


RoyStegeman (approved):
Tested and works, thanks a lot!

PR #2209, opened 2024-11-11
Re-implementation of ATLAS prompt photon production

jacoterh:
This PR reimplements ATLAS prompt photon production at 13 Tev. The validphys report is linked [here]( https://vp.nnpdf.science/OqNZQGGdTsi06BU8qgzm7g==/).

```
observables = ["XSEC"]

for obs in observables:

    new_implementation = f"ATLAS_PH_13TEV_{obs}"
    old_implementation = f"ATLAS_PH_13TEV_{obs}"

    inp1 = {
        "dataset_input": {"dataset": f"{new_implementation}"},
        "theoryid": 200,
        "use_cuts": "internal",
        "t0pdfset": "NNPDF40_nnlo_as_01180",
        "use_t0": True,
    }
    inp2 = {
        "dataset_input": {"dataset": f"{old_implementation}", "variant": "legacy"},
        "theoryid": 200,
        "use_cuts": "internal",
        "t0pdfset": "NNPDF40_nnlo_as_01180",
        "use_t0": True,
    }

    print(f"Comparison for {new_implementation}")
    print(np.all(np.isclose(covmat1, covmat2)))
    print(np.all(np.isclose(t0_covmat1, t0_covmat2)))
```

gives

```
Comparison for ATLAS_PH_13TEV_XSEC
False
False
```

The ratio of the eigenvalues of the legacy and the new covmat is close to unity though. A summary of what changed:
- k1 (the photon pseudo rapidity) in kinematics_XSEC was incorrect. It did not correspond to the midpoint of each bin, but to half the bin width.
- All systematics reported on HEPdata are now taken into account. Before there were only two, now we have 91. This obviously affects the central value (through shifts due to asymmetric uncertainties) and the covmat itself.


validphys2/src/validphys/process_options.py
```diff
@@ -162,6 +163,16 @@ def _shp_xq2map(kin_info):
     return np.clip(x, a_min=None, a_max=1, out=x), np.concatenate((q2, q2))
 
 
+def _pht_xq2map(kin_info):
+    # Then compute x, Q2
```
scarlehoff, line 167:
```suggestion
```

scarlehoff (commented):
Instead of ET^2, I think it makes more sense to have ET in the kinematics (and change the `process_options.py` accordingly.

I'm also not sure the value of `x` is computed correctly, since we should have x1 and x2. 

Also, since you have the rapidity of the photon I think it makes more sense to use that for to access the x1/x2 kinematics? (because Et/s at best is giving you something like that x1*x2)

jacoterh:
Hi @scarlehoff , I should have fixed the x Q2 mapping, please check. The updated vp report can be found [here](https://vp.nnpdf.science/Yc5gmOILTrqwyzh-B21OBw==/). Regarding the pseudo rapidity on master,  it did not correspond to the midpoint of each bin, but to half the bin width. The new vp report looks fine to me. Let me know whether you agree. 

By the way, I don't find any old implementation of ATLAS_PH_8TEV_XSEC even though this dataset is listed on the wiki. What's up with that?

jacoterh:
I see, then let's try to just merge this and work on the 8TeV one later. It's ready for review in any case!

jacoterh:
Done, I simply forgot, thanks for pointing it out

scarlehoff (approved):

PR #2213, opened 2024-11-14
New theories for mt variation ``40_009_xxx``

jacoterh:
This PR is for new theories where we vary the top mass for a simultaneous alphas and mt determination. Not much has been added yet, but I just wanted to claim theoryid  ``40_009_xxx`` if still available.


RoyStegeman:
Thanks for opening the PR. 
 
Our theories will be variations of 40_000_000: one exact copy (the theorycard will be the same but in the theory we will have fully NNLO FKtables for the top production dataset) and then variants where we change `mt`. You can just copy that theorycard to have one that is up-to-date.

RoyStegeman:
It's only for the top datasets since we don't want to present the first fit with NNLO grids - or probably we can but I imagine that won't be appreciated by ppl who computed them.

PR #2216, opened 2024-11-18
remove use_theorycovmat key

RoyStegeman:
I was adding this to #2198 but when I tried to commit I saw that had been merged - I'd still like to add it though 

This key was only used in a check, where it is forced to be set to true, though an error is raised if this is not the case, so it serves little purpose. Probably a remnant of a time where it could be used to toggle the use of the theory covmat.


validphys2/src/validphys/results.py
```diff
@@ -538,14 +538,11 @@ def procs_corrmat(procs_covmat):
 def results(dataset: DataSetSpec, pdf: PDF, covariance_matrix, sqrt_covmat):
     """Tuple of data and theory results for a single pdf. The data will have an associated
     covariance matrix, which can include a contribution from the theory covariance matrix which
-    is constructed from scale variation. The inclusion of this covariance matrix by default is used
-    where available, however this behaviour can be modified with the flag `use_theorycovmat`.
+    is constructed from scale variation.
```
scarlehoff, line 541: So this docstring was not true?

But the part of `The inclusion of this covariance matrix by default is used` is still true isn't it?

scarlehoff (commented):
Sorry, I thought you wanted to merge the other PR already!

Please remove the key also from the examples where it is used (data_theory comparison and the pineappl notebook)

RoyStegeman:
I did want to merge that, so thanks for that - I had just forgotten I also wanted to do this. But doesn't really matter that it has it's own PR

validphys2/src/validphys/results.py
RoyStegeman: I didn't trace back how `covariance_matrix` and `sqrt_covmat` are produced and whether they include the theory covmat or not, but at least the comment about `use_theorycovmat` was wrong

scarlehoff (approved):

PR #2217, opened 2024-11-18
Update PineAPPL version to `v1`

Radonirinaunimi:
For the time being, this is primarily needed because `pineko` (https://github.com/NNPDF/pineko/pull/206) depends on it. This does not introduce physic's changes yet, ie still does not allow for an arbitrary number of convolutions.


scarlehoff:
sorry, wouldn't it be better to instead of having a branch here just for the pineko dev branch, to remove for now nnpdf from the pineko dev branch?

Or do you need nnpdf there?

validphys2/src/validphys/coredata.py
```diff
@@ -68,7 +68,7 @@ class FKTableData:
     ndata: int
     xgrid: np.ndarray
     sigma: pd.DataFrame
-    convolution_types: tuple[str] = None
+    convolution_types: list | None = None
```
scarlehoff, line 71: Tuple would be best so it can be hashed if needed (not 100% important since in principle one given fktable will have only one entry for this I guess)

validphys2/src/validphys/pineparser.py
```diff
@@ -155,26 +153,18 @@ def pineappl_reader(fkspec):
...
-        # Sanity check (in case at some point we start fitting things that are not protons)
-        if parton1 != "2212":
-            raise ValueError("vp can only read hadronic fktables with 2 protons!")
+    # Get the convolution types for this FK table
+    convolutions = pine_rep.convolutions
+
+    # Is it hadronic? For the time being, hadronic is defined with `len(convolutions) == 2`.
+    # Hadronic could also involve 3 convolutions in processes such as `pp->H` (with FFs).
+    # DIS FK table now only contains ONE single convolutions.
+    hadronic = len(convolutions) == 2
```
scarlehoff, line 162: The pid type is important to check to avoid problems like unknowingly fittingna p pbar fktable for instance

validphys2/src/validphys/pineparser.py
```diff
@@ -272,7 +265,7 @@ def pineappl_reader(fkspec):
         sigma=sigma,
         ndata=ndata,
         Q0=Q0,
-        convolution_types=tuple(conv_types),
+        convolution_types=convolutions,
         metadata=fkspec.metadata,
```
scarlehoff, line 269: The comment about tupling

scarlehoff (commented):
ok then! I've left some small comments, although it is WIP I thought better to mention these small points sooner rather than later.

Just one question, in principle old fktables we will always be able to read, right?

Radonirinaunimi:
> ok then! I've left some small comments, although it is WIP I thought better to mention these small points sooner rather than later.
> 
Thanks for the comments @scarlehoff!

> Just one question, in principle old fktables we will always be able to read, right?

Yes, old FK tables will still be read as before perfectly. The only difference is that now when PineAPPL reads these FK tables it'll add onto them the new attributes (for e.g. `convolutions`) .

validphys2/src/validphys/coredata.py
Radonirinaunimi: I agree!

On a side note, having `list[pineappl.convolutions.Conv]` or `tuple[pineappl.convolutions.Conv]` might not work for hyperopt (?) because the `Conv` object is not pickleable.

validphys2/src/validphys/pineparser.py
Radonirinaunimi: Good point! I will modify this accordingly.

validphys2/src/validphys/coredata.py
scarlehoff: I don't think the fktables go down the pickle with hyperopt, do they?

Because if they do we just found the memory leak :__

Radonirinaunimi: They shouldn't! But somehow the hyperopt is failing because it cannot pickle `Conv`:
https://github.com/NNPDF/nnpdf/actions/runs/11896970990/job/33150122459?pr=2217#step:6:2168

Not really sure what is happening here.

scarlehoff: It should certainly not. Good we found that ^^

scarlehoff: I have looked a bit at this. We definitely should prune what goes into the pickle there _however_ I think the parser should take the pineappl object and convert them to strings. We don't want to have pineappl objects during the training in the fit since we want to be able to act exclusively with keras + numpy objects, sobe  the `FKTableData` should definitely keep the `convolution_types: tuple[str] = None`

Radonirinaunimi: Thanks for looking into this!

> We don't want to have pineappl objects during the training in the fit since we want to be able to act exclusively with keras + numpy objects. The FKTableData should definitely keep the convolution_types: tuple[str] = None

If the state of the `FKTableData` object is still passed onto the fit, then I agree that `convolution_types` should be `tuple[str]`. But to be honest, I didn't expect this to be the case 👀

So, in the meantime, yes I will do as so.

scarlehoff: Well, either at the parser or later we need to parser `convolution_types` because we need to know which tensorflow function to use. So, since it needs to be parsed anyway, it should be done as soon as possible so that we are always using the parsed values instead of having parts of the code that use the pineappl value and part of the code that use the strings.

scarlehoff:
As of `1.0.0a4` there's a problem with fktables that include bins that are 0. It is possible to still fits simply by removing these 0 points as they are also ignored by the nnpdf code:

```
#!/usr/bin/bash

for i in *.lz4
do
    if ! python -c "import pineappl ; pineappl.fk_table.FkTable.read(\"${i}\").table()" 2>/dev/null
    then
        echo "Removing extra points from" $i
        bins=$(pineappl read ${i} --bins | awk 'END{print $1}')
        pineappl write ${i} --delete-bins 1-${bins} tmp.lz4
        mv tmp.lz4 ${i}
    fi
done
```

scarlehoff:
This needs to be rebased to ensure that everything works.

I would say let's ignore here the fact that pineappl>=1.0 is not compatible with eko>=0.15 since they never need to talk for this repo (it is only a problem for pineko, but pineko depends only on nnpdf-data not in the full nnpdf)

scarlehoff:
It was mentioned in the last code meeting, but I don't know the details.

Radonirinaunimi:
> Probably that's off-topic, but I just wanted to mention that with pineappl v1, we need to update the API extension notebook too. I was thinking that this branch might be the most appropriate place to update it. I can take care of that if you don't mind.

Yes, sure.

PS: There isn't a conda version of v1 yet so for the time being please just ignore the failing tests.

scarlehoff:
If you could update this as well https://github.com/NNPDF/nnpdf/blob/master/validphys2/examples/API_extension_Pineappl.ipynb before merging that would be great. Otherwise I can have a go tomorrow or the day after and merge it then.

PR #2222, opened 2024-11-21
Remove the multi dense layer

scarlehoff:
We are not using this layer and, if we want to do something similar, we will need to rewrite it from scratch because the current formulation doesn't work anyway.

And also, since we want to fit several architectures at once going forward, this doesn't really help there...


RoyStegeman (approved):
😢

PR #2224, opened 2024-11-22
add modsv key if xif not 1

RoyStegeman:
This adds the `ModSV: expanded` key in case `XIF` is not equal to 1 as discussed in the telegram channel. 

Please check if for the polarized the update is also correct or if it should only be applied to a subset of the theories.

I used the script below (apologies about the stylistic changes, they're coming from ruamel)

```python
import os
from ruamel.yaml import YAML
from nnpdf_data import theory_cards as theory_cards_folder

yaml = YAML()
yaml.preserve_quotes = True

for filename in os.listdir(theory_cards_folder):
    if filename.endswith(".yaml") or filename.endswith(".yml"):
        filepath = os.path.join(theory_cards_folder, filename)

        with open(filepath, 'r') as file:
            data = yaml.load(file)

        if data.get("XIF") != 1:
            data["ModSV"] = "expanded"

        if data.get("XIF") == 1:
            data.pop("ModSV", None)

        with open(filepath, 'w') as file:
            yaml.dump(data, file)
```

### DIFF
--- a/nnpdf_data/nnpdf_data/theory.py
+++ b/nnpdf_data/nnpdf_data/theory.py
@@ -107,6 +107,17 @@ class TheoryCard:
                     f"Trying to use {self.ID} with {self.Qedref} != {self.Qref}. This is not supported!"
                 )
 
+        if self.XIF == 1 and self.ModSV is not None:
+            raise TheoryCardError(
+                f"Theory: {self.ID}, error: XIF is {self.XIF} while ModSV is {self.ModSV}. "
+                "If XIF is equal to 1.0, ModSV should not be defined."
+            )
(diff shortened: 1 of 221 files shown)


scarlehoff:
Please, also add to https://github.com/NNPDF/nnpdf/blob/6edf250cebbf47d72d32fb48baa3e3e4ac4bdfce/nnpdf_data/nnpdf_data/theory.py#L101

```python
        if self.XIF != 1 and self. ModSV:
             raise WRONG
        elif self.XIF == 1 and self.ModSV is not None:
             raise ALSO WRONG
```

scarlehoff (approved):
Thanks!

PR #2225, opened 2024-11-23
Separated nnpdf data package

scarlehoff:
<s>This will remain a separate branch until the commondatas are all merged since merging this will create approx. one thousand conflicts. </s>

Ready for review. Keeping this on top of master has not been my favourite thing to do, but I think I didn't break anything.

The package is available in pypi https://pypi.org/project/nnpdf-data/


scarlehoff:
The code for this is now finished. Version 0.0.3 in pypi includes all data currently in master so it should be enough to reproduce 4.0', the alpha_s theories etc.

`nnpdf_data` is enough to read and inspect all commondata files (coredata.CommonData has been moved to `nnpdf_data`). It depends on both pandas and numpy but I think that's ok.

For the time being I will continue updating it by hand, but the goal would be for the CI to upload to pypi `nnpdf` (containing `validphys` and `n3fit`) and `nnpdf_data` (containing only the data) for every nnpdf tag, so that `nnpdf_data` and `nnpdf` are always synchronized for their versions.

RoyStegeman (approved):
:)

PR #2226, opened 2024-11-25
unpin ruamel.yaml version

RoyStegeman:
Along with https://github.com/NNPDF/reportengine/pull/67 this PR addresses the dependence on the ruamel.yaml version which is pinned to `<0.18` due to changes in the API. 

Hopefully this also allows us to resolve the problem with `vp-nextfitruncard` removing newlines and comments from the runcard 🤞

### DIFF
--- a/nnpdf_data/nnpdf_data/utils.py
+++ b/nnpdf_data/nnpdf_data/utils.py
@@ -2,9 +2,13 @@ import pathlib
 
 from ruamel.yaml import YAML
 from validobj import ValidationError, parse_input
+import yaml
 
 yaml_rt = YAML(typ="rt")
-yaml_fast = YAML(typ="safe", pure=False)
+try:
+    Loader = yaml.CLoader
+except AttributeError:
+    Loader = yaml.Loader
 
 
 def parse_yaml_inp(input_yaml, spec):
@@ -15,7 +19,7 @@ def parse_yaml_inp(input_yaml, spec):
     https://validobj.readthedocs.io/en/latest/examples.html#yaml-line-numbers
     """
     input_yaml = pathlib.Path(input_yaml)
-    inp = yaml_fast.load(input_yaml.read_text(encoding="utf-8"))
+    inp = yaml.load(input_yaml.read_text(encoding="utf-8"), Loader=Loader)
     try:
         return parse_input(inp, spec)
     except ValidationError as e:
--- a/nnpdf_data/nnpdf_data/utils.py
+++ b/nnpdf_data/nnpdf_data/utils.py
@@ -1,7 +1,10 @@
 import pathlib
 
+from ruamel.yaml import YAML
 from validobj import ValidationError, parse_input
-import yaml
+
+yaml_rt = YAML(typ="rt")
+yaml_fast = YAML(typ="safe", pure=False)
 
 
 def parse_yaml_inp(input_yaml, spec):
@@ -12,14 +15,14 @@ def parse_yaml_inp(input_yaml, spec):
     https://validobj.readthedocs.io/en/latest/examples.html#yaml-line-numbers
     """
     input_yaml = pathlib.Path(input_yaml)
-    inp = yaml.load(input_yaml.read_text(encoding="utf-8"), yaml.Loader)
+    inp = yaml_fast.load(input_yaml.read_text(encoding="utf-8"))
     try:
         return parse_input(inp, spec)
     except ValidationError as e:
         current_exc = e
-        # In order to provide a more complete error information, use round_trip_load
+        # In order to provide a more complete error information, use round trip
         # to read the .yaml file again (insetad of using the CLoader)
-        current_inp = yaml.load(input_yaml.open("r", encoding="utf-8"), yaml.Loader)
+        current_inp = yaml_rt.load(input_yaml.open("r", encoding="utf-8"))
         error_text_lines = []
         while current_exc:
             if hasattr(current_exc, 'wrong_field'):
--- a/nnpdf_data/nnpdf_data/utils.py
+++ b/nnpdf_data/nnpdf_data/utils.py
@@ -27,7 +27,7 @@ def parse_yaml_inp(input_yaml, spec):
         while current_exc:
             if hasattr(current_exc, 'wrong_field'):
                 wrong_field = current_exc.wrong_field
-                # Mappings compping from ``round_trip_load`` have an
+                # Mappings compping from yaml_rt have an
                 # ``lc`` attribute that gives a tuple of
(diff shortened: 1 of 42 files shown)


nnpdf_data/nnpdf_data/utils.py
```diff
@@ -1,12 +1,9 @@
...
 
-import ruamel.yaml as yaml
+from ruamel.yaml import YAML
 from validobj import ValidationError, parse_input
 
-try:
-    Loader = yaml.CLoader
-except AttributeError:
-    Loader = yaml.Loader
+yaml = YAML(typ='safe')
```
scarlehoff, line 6: Using libyaml (CLoader) when available makes a huge difference in time when loading large (or many) files.

scarlehoff (commented):
Many thanks for this, you are doing god's work here.

One thing, wouldn't it make sense to have the import ruamel.yaml ; yaml = YAML() in one single place (nnpdf_data for instance) and import it from there?

Would make your life easier I think debugging and ensuring that everything is doing the same thing.

nnpdf_data/nnpdf_data/utils.py
RoyStegeman: CLoader is PyYaml, do we want to keep using both yaml packages?

RoyStegeman:
> One thing, wouldn't it make sense to have the import ruamel.yaml ; yaml = YAML() in one single place (nnpdf_data for instance) and import it from there?

I'm still trying to understand some of the errors, it seems to depend on what settings are used during import (which is a bit concerning that it actually affects the way the yaml is parsed rather than just the formatting upon dumping). If this is the case then we may need a different `YAML` per module, though in general you're right and I'll probably clean it up once I understand the issues. 

If not then you may be right

nnpdf_data/nnpdf_data/utils.py
scarlehoff: I could not find an explicit mention in the docs but it might be that as long as you use `safe` (and purre is not True) it will use CLoader under the hood https://sourceforge.net/p/ruamel-yaml/tickets/159/#8f94 

but the comment is from 2017... we should check. The difference in performance was enough to think about dropping ruamel if we can no longer use CLoader

validphys2/src/validphys/commondataparser.py
```diff
@@ -41,36 +41,21 @@
...
-    Loader = yaml.CLoader
-except AttributeError:
-    # fallback to the slow loader
-    Loader = yaml.Loader
-
-
-def _quick_yaml_load(filepath):
-    return yaml.load(filepath.read_text(encoding="utf-8"), Loader=Loader)
-
+yaml = YAML(typ='rt')
```
scarlehoff, line 58:
```suggestion
yaml = YAML(typ='safe')
```

Just benchmarked it, with this change is (almost) as fast as master. Not as fast though, it takes 1.3x more. 
With `rt` it takes 10x more.

As I said, I think the best thing at this point is to have in utils `safe_load` and `rt_load` or whatever with a comment explaining why. That will make it easier if we need to change it again.

RoyStegeman:
I've added a single initialization in`vp.utils` for the different options we may want to use in vp and n3fit. In `nnpdf_data` mostly `pyyaml` was used and `ruamel` only in some places where formatting doesn't matter, so I made `nnpdf_data` `pyyaml`-only.

scarlehoff:
commondataparser will be moved to nnpdf_data, can it use pyyaml as well?

nnpdf_data/nnpdf_data/utils.py
```diff
@@ -17,14 +12,14 @@ def parse_yaml_inp(input_yaml, spec):
...
-    inp = yaml.load(input_yaml.read_text(encoding="utf-8"), Loader=Loader)
+    inp = yaml.load(input_yaml.read_text(encoding="utf-8"), yaml.Loader)
     try:
         return parse_input(inp, spec)
     except ValidationError as e:
         current_exc = e
         # In order to provide a more complete error information, use round_trip_load
         # to read the .yaml file again (insetad of using the CLoader)
-        current_inp = yaml.round_trip_load(input_yaml.open("r", encoding="utf-8"))
+        current_inp = yaml.load(input_yaml.open("r", encoding="utf-8"), yaml.Loader)
```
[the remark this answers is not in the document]
RoyStegeman: Turns out ruamel is built on top of pyyaml, so CLoader is the same thing in both. I took the passing ci as indication that this was fine, but I'll revert it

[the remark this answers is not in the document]
scarlehoff: Then you need to use CLoader for the other one to speed things up (and you need the try-except since libyaml might not be installed everywhere).
I think all in all it is better to use ruamel also here, pyyaml doesn't really simplify things.

> I took the passing ci as indication that this was fine, but I'll revert it

We would need a test that not only checks for errors but looks also at the message... and in this case is a `ValidationError` from reportengine. I'll put it in my "probably never happening but who knows" to do list

RoyStegeman:
I wasn't too sure from the ruamel code, but it seems that yaml_fast does the same as CLoader:
```
yaml_rt: 0.002312 seconds per load
yaml_fast: 0.000502 seconds per load
CLoader: 0.000502 seconds per load
```

Here is the script for reference:

```python
import pathlib
from timeit import timeit

from ruamel.yaml import YAML
import yaml

import nnpdf_data

yaml_rt = YAML(typ="rt")
yaml_fast = YAML(typ="safe", pure=False)

file_path = nnpdf_data.theory_cards / "410.yaml"


def load_with_yaml_rt():
    input_yaml = pathlib.Path(file_path)
    return yaml_rt.load(input_yaml.read_text(encoding="utf-8"))


def load_with_yaml_fast():
    input_yaml = pathlib.Path(file_path)
    return yaml_fast.load(input_yaml.read_text(encoding="utf-8"))


def load_with_CLoader():
    input_yaml = pathlib.Path(file_path)
    return yaml.load(input_yaml.read_text(encoding="utf-8"), Loader=yaml.CLoader)


num_iterations = 1000
time_yaml_rt = timeit(load_with_yaml_rt, number=num_iterations)
time_yaml_fast = timeit(load_with_yaml_fast, number=num_iterations)
time_yaml_CLoader = timeit(load_with_CLoader, number=num_iterations)

print(f"yaml_rt: {time_yaml_rt / num_iterations:.6f} seconds per load")
print(f"yaml_fast: {time_yaml_fast / num_iterations:.6f} seconds per load")
print(f"CLoader: {time_yaml_fast / num_iterations:.6f} seconds per load")
```

scarlehoff (approved):

PR #2227, opened 2024-11-25
add test_unique_theoryid_variations

RoyStegeman:
Since the list is getting quite long I foresee this causing problems at some point if we don't check


scarlehoff (approved):
lgtm

PR #2235, opened 2024-11-29
Unify the way theories are getting read

scarlehoff:
Thanks to @jacoterh's PR  #2213 I realized the tests to the theories were not being applied correctly because `fetch_all` sort of bypassed "`fetch_one`".
This just ensures that theories are always loaded through the same functions.


RoyStegeman (approved):
LGTM

PR #2238, opened 2024-12-05
Revision of CMS inclusive DY data

achiefa:
This PR addresses the four datasets listed below:
- CMS_WPWM_7TEV_ELECTRON_ASY
- CMS_WPWM_7TEV_MUON_ASY
- CMS_Z0_7TEV_DIMUON_2D
- CMS_WPWM_8TEV_MUON_Y

These datasets were already re-implemented in the new format by TR. I think the things to do here are:
1. Check the filter file is working.
2. Update the `process_type` with the processes in `process_options`.
3. Update the name of the kinematics (and re-run the filter)

See the reports for these datasets before I apply any change: [report](https://vp.nnpdf.science/dGqpwGfpQYi-3jNQmsOVpQ==)


nnpdf_data/nnpdf_data/commondata/CMS_WPWM_8TEV_MUON/uncertainties.yaml
```diff
@@ -218,8 +218,8 @@ bins:
   corr_stat_10: 4.59724656e+02
   corr_stat_11: -1.19020891e+03
   corr_stat_12: -7.78086020e+02
-  corr_stat_13: -3.25115482e+02
-  corr_stat_14: -1.74120972e+01
+  corr_stat_13: 3.25115482e+02
+  corr_stat_14: 1.74120972e+01
   corr_stat_15: 1.85441751e+02
```
achiefa, line 223: One thing to note here is that after re-running the filter without any modification, the signs of these two sources flip. I don't know the reason, but as long as the same sing-flip is propagated for all bins and consistently always with the same sources, then it should not be a problem.

nnpdf_data/nnpdf_data/commondata/CMS_Z0_7TEV_DIMUON/uncertainties.yaml
achiefa, line 1: Same comment as the previous commit.

scarlehoff: You can apply the label `regenerate-label` to this PR and then the data will be autogenerated by the bot.

nnpdf_data/nnpdf_data/commondata/CMS_WPWM_7TEV_MUON/metadata.yaml
```diff
@@ -40,8 +40,8 @@ implemented_observables:
         label: $\eta$
         units: ''
       M2:
-        description: W boson Mass
-        label: $M^2$
+        description: W boson mass squared
```
scarlehoff, line 43: You forgot the `M2` of the name of the dict in this file.

achiefa: Thanks for that. Updated!

scarlehoff: Btw, I think you will need to update also filter rules to use the right `m_w` etc

achiefa: Yes, indeed.

achiefa:
And this is the report after these changes: [report](https://vp.nnpdf.science/YkBtvBFbQS-OiknEvYs7Sg==)

achiefa:
@scarlehoff I don't understand that error message that I get from the test
```
Run echo "Changed files: $CHANGED_FILES"
  echo "Changed files: $CHANGED_FILES"
  exit 1
  shell: /usr/bin/bash -e {0}
  env:
    CHANGED_FILES: nnpdf_data/nnpdf_data/commondata/CMS_WPWM_8TEV_MUON/uncertainties.yaml
Changed files: nnpdf_data/nnpdf_data/commondata/CMS_WPWM_8TEV_MUON/uncertainties.yaml
Error: Process completed with exit code 1.
```
Can you help me?

scarlehoff:
It means that you need to use the label again. Basically it means that the bot would get results different than what you commited (@Radonirinaunimi and I were unable to make the -1 signs agree between computers so we decided to take the bot as the ultimate truth)

achiefa:
>(@Radonirinaunimi and I were unable to make the -1 signs agree between computers so we decided to take the bot as the ultimate truth)

How is that possible 🫤?

achiefa:
Ok, I think now this is solved. Wait for the tests, and then I think it can be merged.

nnpdf_data/nnpdf_data/commondata/CMS_Z0_7TEV_DIMUON/metadata.yaml
```diff
@@ -31,13 +31,13 @@ implemented_observables:
     dataset_label: "CMS Drell-Yan 2D 7 TeV 2011"
```
scarlehoff, line 31: The only thing missing is that the `kinematics_override` needs to be set to `identity` in all of the files, because otherwise validphys will `sqrt` the masses (see e.g, the kinematic coverage plot, the scales are too low)

image

PR #2239, opened 2024-12-06
add no-interactive flag to vp-nextfitruncard

RoyStegeman:


scarlehoff (approved):
Perhaps recover also the 3.10 in the tests, we'll fix that later...

PR #2241, opened 2024-12-06
Re-implementation of CMS_Z0J_8TEV in the new format

achiefa:
This PR implements CMS_Z0J_8TEV in the new format.

Legacy: [[default](https://vp.nnpdf.science/gusD-5PZSjSK2B1pqyOqbw==)], [[sys_10](https://vp.nnpdf.science/TZplVwchTH2q5gdMCZXotA==)]
New: [[default](https://vp.nnpdf.science/HIYCpdW-QnqhZJLUjpx43w==)], [[sys_10](https://vp.nnpdf.science/ZtjcWJGvTVKw3nb1dFHkTA==)]


achiefa:
Hi @scarlehoff & @enocera, I see that this dataset was implemented with three different variants:
- `sys_10_NNPDF31`
- `sys_NNPDF31` 
- I suspect they were used to account for K-factors, so they should not be necessary. Can you

scarlehoff:
As usual, @enocera should confirm. If `_10` is also a scale uncertainty I don't think it should've gone into 4.0?

In any case, if you are able to reproduce the default from hepdata (i.e., without the theory uncertainty, whatever it is), then you can always create the `_10` version (taking it from the legacy file) by concatenating that uncertainty with the hepdata ones, so that we can reproduce previous result.

achiefa:
The two variants match between the new implementation and the legacy.

I've also uploaded the data-theory comparisons in the description of this PR. The chi2's are unchanged. There is a difference in the (x,Q2) map, but I think this is expected as we move to the new format (i.e. different functions).

I'm now regenerating the common data using the bot because of the same issue encountered in #2238. After that, all tests should pass successfully.

Screenshot 2024-12-09 at 12 17 41
Screenshot 2024-12-09 at 12 17 58

achiefa:
Hi @Radonirinaunimi, I need your help here. I had to regenerate the tests after the implementation in the new format. However, the three tests in `test_plots` below keep failing even if they succeed locally on my laptop. Do you know what's happening?
```
=================================== FAILURES ===================================
________________________________ test_plotfancy ________________________________
Error: Image files did not match.
  RMS Value: 41.03842988928521
  Expected:  
    /tmp/tmp1vpabeyq/validphys.tests.test_plots.test_plotfancy/baseline.png
  Actual:    
    /tmp/tmp1vpabeyq/validphys.tests.test_plots.test_plotfancy/result.png
  Difference:
    /tmp/tmp1vpabeyq/validphys.tests.test_plots.test_plotfancy/result-failed-diff.png
  Tolerance: 
    22
________________________________ test_plot_xq2 _________________________________
Error: Image files did not match.
  RMS Value: 33.0809017673408
  Expected:  
    /tmp/tmp1vpabeyq/validphys.tests.test_plots.test_plot_xq2/baseline.png
  Actual:    
    /tmp/tmp1vpabeyq/validphys.tests.test_plots.test_plot_xq2/result.png
  Difference:
    /tmp/tmp1vpabeyq/validphys.tests.test_plots.test_plot_xq2/result-failed-diff.png
  Tolerance: 
    22
_____________________________ test_plot_xq2_custom _____________________________
Error: Image files did not match.
  RMS Value: 32.44088131829135
  Expected:  
    /tmp/tmp1vpabeyq/validphys.tests.test_plots.test_plot_xq2_custom/baseline.png
  Actual:    
    /tmp/tmp1vpabeyq/validphys.tests.test_plots.test_plot_xq2_custom/result.png
  Difference:
    /tmp/tmp1vpabeyq/validphys.tests.test_plots.test_plot_xq2_custom/result-failed-diff.png
  Tolerance: 
    22
```

RoyStegeman:
Thanks. For me they do fail (are you using the `--mpl` flag and removing the decorator `@pytest.mark.linux`?). Tests can be quite sensitive to the mpl version, and I imagine OS as well, so you could try generating the figures on a linux machine while making sure the mpl version is the same as in the CI. 

I'm leaving the office now, but can have a closer look tomorrow.

achiefa:
Hi @RoyStegeman, thanks for that. Indeed I was only removing the `@pytest.mark.linux` decorator. I'll generate the new figures tomorrow.

Radonirinaunimi:
(Re-)Generating reference plots are always tricky, not only they are highly dependent on the package versions, but also on the platform on which they are run. In #2191 I had asked @scarlehoff to generate the plots.

achiefa:
Hi @Radonirinaunimi, thanks for that. So, what do you think is the best way to go forward, given the circumstances? Is it worth trying what @RoyStegeman suggested?

RoyStegeman:
Yes please do try

achiefa:
I've regenerated the files using a linux machine. I'm not sure this will work out, but let's wait for the test.

Incidentally, I was thinking that it may be worth implementing a bot that re-generates the test plots when needed, similar to the bot that regenerates data. That would produce figures consistently, right?

RoyStegeman:
> Incidentally, I was thinking that it may be worth implementing a bot that re-generates the test plots when needed, similar to the bot that regenerates data. That would produce figures consistently, right?

In principle yes,  but with the repo-size in mind I'm not sure if we want to regenerate all figures if the tests fail because of a subset. A workaround for that would be to have the bot share the plots via a comment

RoyStegeman:
I increased the tolerance, but I don't know how strict a tolerance of 24 is. Maybe it's too lenient? 2 is the default setting.

validphys2/src/validphys/tests/baseline/test_plotfancy.png
RoyStegeman, line 1: Is there a reason the variable on the x-axis has been changed?

achiefa:
Ok, so just to wrap up, this is ready to go now, right?

RoyStegeman:
No I still need to complete my review will try to do so by tomorrow

achiefa:
Yes, sorry, I meant ready for review.

nnpdf_data/nnpdf_data/commondata/CMS_Z0J_8TEV/metadata.yaml
```diff
@@ -1,75 +1,78 @@
...
     legacy_NNPDF31:
       data_uncertainties:
-      - uncertainties_PT-Y_sys_NNPDF31.yaml
+      - uncertainties_legacy_PT-Y_sys_NNPDF31.yaml
     legacy_10_NNPDF31:
       data_uncertainties:
-      - uncertainties_PT-Y_sys_10_NNPDF31.yaml
+      - uncertainties_legacy_PT-Y_sys_10_NNPDF31.yaml
     legacy_10:
       data_uncertainties:
```
RoyStegeman, line 67: Do we need to keep all these? For backward compatibility I guess we will at least want to keep `legacy_10` since that's what has been used in fits so far, but I don't think any of the other have?

nnpdf_data/nnpdf_data/commondata/CMS_Z0J_8TEV/filter.py
```diff
@@ -0,0 +1,40 @@
...
+from filter_utils import Extractor
+import numpy as np
+import yaml
+
+logging.basicConfig(level=logging.INFO, format='[%(levelname)s] %(message)s')
+
+
+def check_dat_with_legacy(observable, rtol=1e-03):
+    """
+    Same as `check_unc_with_legacy`, but for central data points.
```
RoyStegeman, line 16: `check_unc_with_legacy` lives in a different directory at the same level so I don't think it makes much sense to refer to it here

nnpdf_data/nnpdf_data/commondata/CMS_Z0J_8TEV/filter_utils.py
```diff
@@ -0,0 +1,305 @@
...
+                'treatment': 'MULT',
+                'type': 'UNCORR',
+            }
+        elif variant != 'default':
+
+            raise ValueError(f'The variant {variant} is not implemented yet.')
+
+        return unc_definitions
+
+    def generate_data(self, variant='default', save_to_yaml=False, path='./'):
```
RoyStegeman, line 233: Please put a docstring

nnpdf_data/nnpdf_data/commondata/CMS_Z0J_8TEV/filter.py
```diff
@@ -0,0 +1,40 @@
...
+        try:
+            assert np.allclose(new_val, legacy_val, rtol=rtol)
+        except AssertionError as e:
+            logging.warning(f"Inconsistency, {new_val} != {legacy_val} in bin: {i+1}")
+
+
+if __name__ == "__main__":
+    CMS_Z0J_8TEV = Extractor("./metadata.yaml", "PT-Y", mult_factor=1000)
+    _, _, _ = CMS_Z0J_8TEV.generate_data(variant='default', save_to_yaml=True)
+    _, _, _ = CMS_Z0J_8TEV.generate_data(variant='sys_10', save_to_yaml=True)
```
RoyStegeman, line 40:
```suggestion
    CMS_Z0J_8TEV.generate_data(variant='default', save_to_yaml=True)
    CMS_Z0J_8TEV.generate_data(variant='sys_10', save_to_yaml=True)
```

validphys2/src/validphys/tests/baseline/test_plotfancy.png
RoyStegeman: If it's unique for this dataset then I think the class should be trimmed to only the stuff used for this dataset

nnpdf_data/nnpdf_data/commondata/CMS_Z0J_8TEV/metadata.yaml
achiefa: I agree that only `legacy_10` was used for the fits However, since the new implementation does not match with the old one, I would keep both just in case. Do you think this is a problem?

nnpdf_data/nnpdf_data/commondata/CMS_Z0J_8TEV/filter.py
achiefa: I kept it because I thought it could have been helpful, but it is unnecessary here. I'll delete it.

achiefa:
@RoyStegeman  Thanks for the comments. 

>I.e. no getters, always save to yaml, tab number, maybe other things..
I can remove the getters (I used them for development reasons), but I don't understand why I should remove `save_to_yaml`. Also, do you want me to hard-code the tab numbers? I believe it is much easier if these are read from the metadata file directly.

RoyStegeman:
In some sense I'm being overly picky, and I am aware Juan accepted these things before, so feel free to make your own final decision on this. But to explain my reasoning: at them moment there is a fair bit of functionality that is unused, either because, as you say, it was used during development or in other cases you copy from another dataset but can't have a shared class because they're too different for that. If we keep everything that was used during development our code would be much larger than it is now. 

The point is that seeing what is going on for another person (or yourself later on) takes more time, the more code there is. So there really is value in trimming the unused functionality. 

 > I don't understand why I should remove save_to_yaml

because it's never `False` so there is no need for the option. 

> do you want me to hard-code the tab numbers?

I meant that `__extract_kinematics` has an unused variable `tab_number`

I also commented more on style since I'm not so familiar with how you (and others) have been doing the data implementation so more technical comments will follow later (or maybe they won't be needed).

achiefa:
Ho @RoyStegeman, thanks for the clarification. I understand your point, and I will remove the unused functionalities in the filter file. I'll also clean the other data implementations in the open PRs, so that they will be more uniform. Please let me know if there is anything else you'd like to change.

RoyStegeman (approved):
I went through it more rigorously mainly because I still had to understand better how the data implementation is being done. Most changes that I made are probably subjective though there are some things which could be improved, e.g. `functools.cache` exists so there is no need to implement a caching for the tables yourself. 

Anyway, once you're back, have a look at the changes to see if you agree. The final results are unchanged so if you're happy, this can be merged.

achiefa:
Hi @RoyStegeman , thanks for these suggestions. I agree with you that many of the things you removed were unnecessary or controversial. Now, the filter file is cleaner and more readable. Thanks also for the suggestion concerning `functools.cache`.

For me this is fine, and can be merged.

I'm now cleaning up my other PR's following your guidelines. We can discuss them when we are both back.

PR #2244, opened 2024-12-09
New implementation of CMS_WCHARM_13TEV_WPWM-TOT-UNNORM

achiefa:
This PR implements CMS_WCHARM_13TEV_WPWM-TOT-UNNORM in the new format.

General comments
==============
This dataset delivers the differential distribution in function of the absolute rapidity of the lepton pair. Each data point is accompanied by a (symmetric) statistical uncertainty and a (asymmetric) systematical uncertainty. The latter is the sum in quadrature of the different sources of uncertainty. The breakdown of these systematic sources is not delivered in the HepData format, but it is given in Table 1 of the paper.

The legacy version has the variant `sys_10`, which should not be implemented because it was meant to account for the 3pt prescription.

Legacy: [[default](https://vp.nnpdf.science/NZMg5ENvQOi7RlaewChB0Q==)],
New: [[default w/ shifts](https://vp.nnpdf.science/pHYGj9lnSw6f6vMCUc_Kzg==)], [[default w/o shifts](https://vp.nnpdf.science/H98HYsu5RZ2C04voYU85Iw==)]


achiefa:
Hi @RoyStegeman , I think I'm done here. Please, see the following comments:
- In this dataset, uncertainties must be symmetrised. This means that the central data must be shifted accordingly. However, I suspect that the old implementation didn't have the data shifted. I produced two comparisons (see description), w/ and w/o shift, respectively. The one that does *not* implement the shift gives the same chi2 as the legacy version. You can also check by naked eye that the non-shifted data are closer to the legacy data.
- I also believe that the old implementation rounded some data, and indeed the central data in HepData do not match with the legacy implementation.
- I can't reproduce the legacy covmat, but it might be that the cause is one or both of the differences above. Below, you can find the two matrices.
output

Honestly, I can't judge whether these differences are relevant or not. The difference in chi2 is not negligible if one accounts for the shifts. On the other hand, the difference in the t0 matrices does not really worry me as I was able to reproduce the chi2 of the legacy implementation provided shifts were removed. @RoyStegeman, what do you think? Maybe it is worth asking @enocera.

achiefa:
> Do you know why the fktables of this dataset only exist in theories 704 (0.5,0.5) and 705 (0.5,1)?

No, maybe @enocera does.

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/filter.py
```diff
@@ -0,0 +1,16 @@
+'''
+Filter script for CMS_WCHARM_13TEV
+'''
+
+import logging
+
+from filter_utils import Extractor
+import numpy as np
+import yaml
```
RoyStegeman, line 9:
```suggestion
```

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/filter_utils.py
```diff
@@ -0,0 +1,304 @@
+import logging
+import os
+
+import numpy as np
+import pandas as pd
```
RoyStegeman, line 5:
```suggestion
```

RoyStegeman, line 5:
```suggestion
```

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/filter_utils.py
```diff
@@ -0,0 +1,304 @@
...
+UNCORR_SYS_UNC = ['UnfoldMCstat', 'UnfoldOtherGen', 'UnfoldReweight']
+ART_LABEL = 'art_corr_unc'
+STAT_LABEL = 'stat_uncorr_unc'
+TABLE = ''
+
+# From Table 1 of the paper
+SYS_UNC_by_bin = [{}]
+
+
+class Extractor:
```
RoyStegeman, line 28: Some of this stuff seems pretty universal. Would it be worth defining a base class that is shared between datasets?

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/filter_utils.py
```diff
@@ -0,0 +1,304 @@
+import logging
+import os
```
RoyStegeman, line 2:
```suggestion
```

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/metadata.yaml
```diff
@@ -1,65 +1,61 @@
...
     dataset_label: CMS $W+c$ total 13 TeV
     y_label: $d\sigma(W+c)/d|\eta^\mu|$ (fb)
     x_label: $|\eta^\mu|$
-    plot_x: k1
-  kinematic_coverage:
-  - k1
-  - k2
-  - k3
+    plot_x: abs_eta
+  kinematic_coverage: [abs_eta, m_W2]
```
RoyStegeman, line 34: Why do the legacy kinematics have three variables and the new implementation just two? And does the legacy dataset not use this metadatafile and thus cause issues if the kinematics don't match?

From the value it seems the removed variable was just supposed to indicate the 13TeV beam energy, which I don't think would be used anywhere. But this question is more about how the code deals with it.

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/sys_uncertainties.py
```diff
@@ -0,0 +1,143 @@
+import numpy as np
```
RoyStegeman, line 1:
```suggestion
```

RoyStegeman (commented):
I'm not so familiar with dataset implementations so this is going to take me some time to figure out... 

For now I just have a question regarding the `Extractor` class. There are also a lot of unused imports, are you using an lsp?

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/metadata.yaml
achiefa: The third variable in the old implementation was the beam energy. However, this is optional in the new common data parser as the $\sqrt{s}$ is automatically read from the name of the dataset. In other words, using sqrts here would be redundant.

The legacy dataset does use this metadata file. However, since the kinematics is the same between the two versions, the legacy file can be removed (and I will).

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/filter_utils.py
achiefa: I've been thinking about making this class more universal. However, I gave up because there are many differences between datasets, even within the same experiment. If we want to use a more universal `extractor`, then we should all agree on standard common specifics amongst datasets. For now, the extractor class is rather specialised to the datasets that I implemented.

achiefa:
> There are also a lot of unused imports, are you using an lsp?
These were copy and paste. My bad. BTW, what is a ``lsp``?

RoyStegeman:
language server protocol. It's the software that highlights tokens based on their role in the python syntax. Including unused imports

achiefa:
Oh, then I am. But I just forgot to delete the unused imports.

enocera:
> > Do you know why the fktables of this dataset only exist in theories 704 (0.5,0.5) and 705 (0.5,1)?
> 
> No, maybe @enocera does.

I think that the reason is as follows: W+c data were not included in NNDPF4.0 because, at that time, NNLO corrections to the matrix elements were not known. When the MHOU, QED, and aN3LO determinations were produced, the leitmotiv was to put them on the same grounds as NNPDF4.0. Therefore W+c did not go into them. At some point  I raised the question whether we should include it (in the same way as we do, e.g. for LHC data in the N3LO fit). Initially their answer was yes, but then they retracted. So I suspect that Andrea started to compute the FK tables, but then stopped.

achiefa:
According to what ERN said in the last code meeting, this one is also ready for review.

RoyStegeman (approved):
Thanks! The implementation looks great and was much easier to go through. I checked the data-th plots and chi2 and the changes between the [old](https://vp.nnpdf.science/DKy-LzGSSo2o6CD46Sb-3A==/) and [new](https://vp.nnpdf.science/Ix_XwQwiQgSw4oQlsXCugw==/) implementations seem acceptable

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/metadata.yaml
```diff
@@ -1,65 +1,61 @@
...
-  data_uncertainties: []
   variants:
     legacy:
       data_uncertainties:
       - uncertainties_legacy_WPWM-TOT-UNNORM.yaml
     legacy_10:
       data_uncertainties:
-      - uncertainties_WPWM-TOT-UNNORM_sys_10.yaml
-  data_central: data_legacy_WPWM-TOT-UNNORM.yaml
+      - uncertainties_legacy_WPWM-TOT-UNNORM_sys_10.yaml
```
RoyStegeman, line 60: Actually, one question: should the `legacy` variant not also include the old central values?

RoyStegeman (changes requested):

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/metadata.yaml
```diff
@@ -1,65 +1,61 @@
...
 implemented_observables:
 - observable_name: WPWM-TOT-UNNORM
   observable:
     description: Jet Rapidity Distribution
     label: CMS $W+c$ total 13 TeV
-    units: ''
-  process_type: EWJ_RAP
-  tables: []
-  npoints: []
+    units: '[fb]'
```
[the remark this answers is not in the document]
achiefa: I'm pretty sure the existing fk tables are in fb. You can check that by comparing table 1 in HepData with the legacy central data. For the HepData table, x-secs are given in pb and their order of magnitude is ~1.e2. If you look at the legacy central data, the order of magnitude is three times higher, thus fb. That's why in the filter file there is the multiplicative factor 1e3.

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/filter_utils.py
```diff
@@ -0,0 +1,248 @@
...
+        The function generates the kinematics by reading and processing it from
+        the referenced table. Kinematics is processed in the format of a list of
+        dictionaries. The keys in each dictionaries specify the label (i.e. name)
+        for the kinematic variables. For this dataset, they are 'abs_eta' and 'm_W2'.
+        The labels are taken from the matadata file. The corresponding values are
+        'min', 'mid', and 'max'.
+
+        For this dataset, 'm_W2' is used in the computation of the (x,Q2)-map and
+        does not have any active role in the fit. For that reason, every bin has the
+        same value. Moreover, only the mid value is used.
```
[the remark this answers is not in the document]
achiefa: What I mean here is that MW2 is used only in the computation $(x,Q^2)$-plane for the kinematic coverage. Contrary to the case of sqrts, MW2 is not deduced by the name of the dataset. So you're point is right: despite MW2 is a fixed parameter, the code still needs it.

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/filter_utils.py
```diff
@@ -0,0 +1,248 @@
...
+        if not self.metadata['ndata'] == ndata:
+            raise ValueError(
+                f"Mismatch in 'ndata': expected {self.metadata['ndata']}, but got {ndata}"
+            )
+        return kinematics
+
+    def _generate_data_and_unc(self):
+        """
+        Return a list with central data points and two additional lists with the corresponding
+        statistical uncertainties. For this dataset, statistical uncertainties
```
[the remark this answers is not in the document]
achiefa: Yes, my fault. It is only one list.

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/metadata.yaml
achiefa: Good point. Yes, it should, given that it has changed. I'll restore it and add it to the metadata.

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/metadata.yaml
achiefa: Yes

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_13TEV/metadata.yaml
achiefa: I've included the legacy central data in the variants. I've also updated the report in the description for the default legacy implementation.

RoyStegeman (approved):
Thanks, once the tests pass this can be merged

PR #2245, opened 2024-12-10
implementation of CMS_WCHARM_7TEV in the new format

achiefa:
Comments
========
This dataset delivers two different observables:
- Total x-sec
- Ratio of x-secs

The variant `sys_10` accounts for the 3pt prescription and thus it should be deprecated in the new format.

Legacy: [[default](https://vp.nnpdf.science/ODh8VYk-QTKhFHrp6o-9MA==)]

### DIFF
--- a/nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_7TEV/metadata.yaml
+++ b/nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_7TEV/metadata.yaml
@@ -1,17 +1,21 @@
 setname: CMS_WCHARM_7TEV
-version: 1
-version_comment: Port of old commondata
+
 nnpdf_metadata:
   nnpdf31_process: DY CC
   experiment: CMS
+
 arXiv:
   url: https://arxiv.org/abs/1310.1138
   journal: JHEP 02 (2014) 013
 iNSPIRE:
-  url: ''
+  url: https://inspirehep.net/literature/1256938
 hepdata:
-  url: ''
-  version: -1
+  url: https://www.hepdata.net/record/ins1256938
+  version: 1
+
+version: 2
+version_comment: Implementation in the new format
+
 implemented_observables:
 - observable_name: WPWM-RATIO
   observable:
@@ -62,6 +66,7 @@ implemented_observables:
       - uncertainties_WPWM-RATIO_sys_10.yaml
   data_central: data_legacy_WPWM-RATIO.yaml
   ported_from: CMSWCHARMRAT
+
 - observable_name: WPWM-TOT
   observable:
     description: Jet Rapidity Distribution


achiefa:
This PR is still WIP. Here you can find a report of the new implementation, which is preliminary:  [report](https://vp.nnpdf.science/0ojlsi0fRRC_h-NTTUu_-Q==)
I have some comments:
- The RATIO xsec in HepData provides statistical and systematic uncertainties. However, the legacy version implemented only the statistical uncertainty (I'm not considering the two additional sys uncertainties for the scale variations). Is there something here that I don't understand?
- For the TOT xsec, I can't reconstruct the covmat. Note that, for this observable, we have the correlation matrix.

achiefa:
I have finalised the implementation of this dataset. Besides the new report posted in the description of this PR, I have the following comments:
- For the `TOT` observable, the covariance matrix is applied to the systematic uncertainties and not to the statistical uncertainties, as I mistakenly did before.
- For the `TOT` observable, I didn't manage to reproduce the t0 matrix of the legacy implementation. I suspect the data in HepData have changed since the first implementation of this dataset. This is indeed the case for the `RATIO` observable, for which the statistical uncertainties in HepData differ from the statistical uncertainties reported in the legacy file.
- For the `RATIO` observable, I didn't manage to reproduce the legacy t0 matrix either. The reason is the one explained above. Moreover, the HepData provides an additional systematic uncertainty that the legacy implementation does not have.
- In terms of data-theory comparison, the plots look really similar between the new and the legacy implementations. The major difference is in the uncertainty bars, as expected for the reasons above. The chi2's for the `RATIO` is slightly better: 2.45 (new) VS 2.49 (legacy). For the `TOT` observable, the chi2 worsens in the new implementation: 1.72 (new) VS 0.96 (legacy). However, this is expected, given the discussion above.

Please, @RoyStegeman have a look at the implementation and let me know if you find something unclear or wrong.

RoyStegeman (commented):
for both WPWM-RATIO and WPWM-TOT:
- The `legacy` variant gives exactly the same chi2 between master and this branch. 
- WPWM-RATIO: Difference between [legacy](https://vp.nnpdf.science/0BNw8-KMQlSdjaKme8By9A==/) and [new](https://vp.nnpdf.science/TgM3bI27Ryq5P9lYKcev_Q==/) implementation is minimal
- WPWM-TOT: Difference between [legacy](https://vp.nnpdf.science/18Dl5RGWSuCRJLC4IGq17g==/) and [new](https://vp.nnpdf.science/gB6xqJ7USCm2eKfZBIj3_w==/) implementation is minimal

I still want to go over the code but if it's similar to the previous one I'm sure that will be fine

nnpdf_data/nnpdf_data/commondata/CMS_WCHARM_7TEV/rawdata/Table9.yaml
RoyStegeman, line 1: This table is not used

RoyStegeman (approved):
Thanks! Looks good

PR #2205, opened 2024-11-09
Reimplement Atlas Z0 8 TeV lowmass

comane:
Possible bugs were found in the old implementation:

1. The overall multiplicative factor of the luminosity uncertainty given on hep-data (see https://www.hepdata.net/record/117854) is 1.8 %. In the old implementation 1.9 % was taken

-> I have incorporated this modification as it makes not a big difference 

(this branch) https://vp.nnpdf.science/XpC1u8GqSyapV3e-ffvpwg==

(master) https://vp.nnpdf.science/7KddpIE6S9u2aBAfuTybkQ==

2. In the old implementation Lumi uncertainty is assumed to be MULT UNCORR. I have noticed that setting the Lumi uncertainty to CORR the chi2 grows by an order of magnitude. I am not sure I understand this. Perhaps in this case Lumi is supposed to be UNCORR?

PR #2246, opened 2024-12-10
Implementing ATLAS_PH-8TEV_XSEC

jacoterh:
This PR implements ATLAS prompt photon production at 8 TeV, similar to #2209 at 13 TeV. No legacy implementation exists for this dataset as it wasn't ported from buildmaster. 

I tried to produce a validphys report, but it complains it cannot use an old theory with a purely new dataset:

`[ERROR]: Bad configuration encountered:
Cannot use an old theory with a purely new dataset (ATLAS_PH_8TEV_XSEC)
`

I used this [card](https://github.com/user-attachments/files/18083779/runcard.txt).


jacoterh:
Hi @RoyStegeman , I forgot how we agreed to proceed here at the last code meeting. Do you remember it? Sorry!

RoyStegeman:
It was suggested to transform the APPLgrid to pineappl grid and add it to the new theory (708)

nnpdf_data/nnpdf_data/commondata/ATLAS_PH_8TEV/filter.py
```diff
@@ -0,0 +1,259 @@
...
+        uncertainties_all = pd.concat([uncertainties_all, uncertainties])
+        central_values_all = np.concatenate([central_values_all, central_values])
+        kinematics_all += kinematics
+
+    uncertainties_all.index = [f"bin {i}" for i in range(uncertainties_all.shape[0])]
+
+    n_sources = uncertainties_all.shape[1]
+    sys_types = {
+        "treatment": ["ADD"] + ["MULT"] * (n_sources - 1),
+        "type": ["UNCORR"] * (n_sources - 1) + ["ATLASLUMI15"],
```
scarlehoff, line 246:
```suggestion
        "type": ["UNCORR"] * (n_sources - 1) + ["ATLASLUMI12"],
```

This is part of the 2012 data https://github.com/NNPDF/nnpdf/pull/2205#issuecomment-2605365260

jacoterh:
Hi @enocera, I noticed that the old buildmaster implementation of this dataset doesn't provide the applgrid for the rapidity region 1.81<|eta(gamma)| < 2.37. It isn't listed at https://github.com/NNPDF/applgrids/tree/master/ATLASPHT12, while the 13 TeV counterpart does provide the applgrid in this region. Do you recall the reason for this? The hepdata entry is https://www.hepdata.net/record/ins1457605?version=1&table=Table%204 for reference.

enocera:
Dear @jacoterh as far as I can recollect, the reason is historical. Prompt photon data (8 TeV and 13 teV) were originally studied in https://arxiv.org/pdf/1802.03021. The NNPDF4.0 implemnetation closely follows the results in that paper. On page 4, one can read: "the most forward rapidity bin, 1.81 ≤ |ηγ | < 2.37, is excluded from the ﬁt due to the tensions between the experimental data and the theoretical predictions." I guess that, for this reason, the PineAPPL grids were not generated at all. At 13 TeV, the paper does not report any restriction.

enocera:
For now, I suggest to assume that this statement (and the NNPDF4.0 selection) hold true.

scarlehoff:
@jacoterh note that it is possible to shift a bin for a number of points like in this case: https://github.com/NNPDF/nnpdf/blob/d7bfbd02415582e65c380f5e805f29716c119cfa/nnpdf_data/nnpdf_data/commondata/ATLAS_Z0J_8TEV/metadata.yaml#L107

In this example `ATLASZPT8TEVMDIST-ATLASZPT8TEV-MLLBIN4_ptZ` and `ATLASZPT8TEVMDIST-ATLASZPT8TEV-MLLBIN6_ptZ` both get shifted by 10 to account for the gap between the data and the theory.

jacoterh:
FK tables have been added to theory 708. See the below link for the data theory comparison with the new commondata implementation: https://vp.nnpdf.science/RyQDSR8VRSONW9yLre6EzA==/

The grids excluded the first 4 ET bins in each rapidity bin, this shift has been accounted for in the filter as well.

nnpdf_data/nnpdf_data/commondata/ATLAS_PH_8TEV/metadata.yaml
```diff
@@ -0,0 +1,59 @@
...
+  kinematic_coverage:
+  - eta
+  - ET
+  - sqrts
+  kinematics:
+    variables:
+      eta:
+        description: Absolute photon pseudorapidity
+        label: $|\eta|$
+        units: $GeV$
```
scarlehoff, line 40:
```suggestion
        units: ''
```

nnpdf_data/nnpdf_data/commondata/ATLAS_PH_8TEV/filter.py
```diff
@@ -0,0 +1,263 @@
...
+import numpy as np
+import pandas as pd
+import yaml
+
+from nnpdf_data.filter_utils.utils import prettify_float
+
+yaml.add_representer(float, prettify_float)
+
+MT_VALUE = 172.5
+SQRT_S = 8_000.0
```
scarlehoff, line 12: I haven't paying much attention to the filter files themselves, but won't many of the functions be similar/equal to those from ATLAS_PH_13TEV?

scarlehoff (commented):
Given the shape of the ET distribution, @enocera, do we want to cut a few more bins in ET? The 13 TeV distribution looks a bit better but it might be because of EWK k-factors? (although they are quite small...)

nnpdf_data/nnpdf_data/commondata/ATLAS_PH_8TEV/filter.py
jacoterh: Yes, the filter is equivalent to ATLAS_PH_13TEV, that's right

scarlehoff: Will you merge both or do you want to keep it explicitly separted?

PR #2228, opened 2024-11-27
Remove HepData raw tables from the repository

Radonirinaunimi:
Addresses https://github.com/NNPDF/nnpdf/issues/2199 by removing the HepData raw tables from the repository. This adds a module that downloads the raw tables from the **server** and from the **HepData** when the commondata need to be regenerated.

**TO DO:**
- [ ] download the raw tables from the NNPDF server
- [ ] propagate across all the datasets
  - [ ] fixed-target DIS
  - [ ] collider DIS
  - [ ] fixed-target inclusive DY
  - [ ] Tevatron inclusive DY
  - [ ] ATLAS inclusive DY
  - [ ] ATLAS DY+jet
  - [ ] ATLAS single-inclusive jet and dijet production
  - [ ] ATLAS top pair production
  - [ ] ATLAS single top production
  - [ ] ATLAS pormpt photon production
  - [ ] CMS inclusive DY
  - [ ] CMS DY+jet
  - [ ] CMS single-inclusive jet and dijet production
  - [ ] CMS top pair production
  - [ ] CMS single top production
  - [ ] LHCb inclusive DY
- [ ] fail commondata tests **only** when `kinematics.yaml`, `data.yaml`, and `uncertainties*.yaml` are modified

PR #2248, opened 2024-12-11
Reimplement DY

peterkrack:
Overall, this is close to completion except for `DYE605_Z0_38P8GEV_DW_PXSEC` in which the legacy implementation contains one extra-source of systematic uncertainties (cc @enocera).

| Dataset  | Status  | Check CovMat  | Check $t0$ CovMat  | Comments  |
|---|---|---|---|---|
| DYE605_Z0_38P8GEV_DW_PXSEC | ❌ | ❌ | ❌  | The old implementation contains one extra-source of uncertainties  |
| DYE866_Z0_800GEV_PXSEC  | ✅  | ❌  | ❌  |  Slight numerical differences due to rawdata source (with maxdiff of ~2%)  |
| DYE866_Z0_800GEV_DW_RATIO_PDXSECRATIO | ✅  | ✅  | ✅  | -  |
| DYE906_Z0_120GEV_DW_PDXSECRATIO |  ✅  | ✅  | ✅  | -  |

These being said, none of the remaining differences are really visible at the data vs theory comparisons [report](https://vp.nnpdf.science/o_-sou46Qf-V7iOp96CjDw==/).

### DIFF
--- a/nnpdf_data/nnpdf_data/commondata/DYE866_Z0_800GEV/metadata.yaml
+++ b/nnpdf_data/nnpdf_data/commondata/DYE866_Z0_800GEV/metadata.yaml
@@ -429,6 +429,7 @@ implemented_observables:
   data_uncertainties: [uncertainties.yaml]
   variants:
     legacy:
+      data_central: data_legacy.yaml
       data_uncertainties:
       - uncertainties_legacy_PXSEC.yaml
   data_central: data.yaml
--- a/nnpdf_data/nnpdf_data/commondata/DYE906_Z0_120GEV_DW/filter.py
+++ b/nnpdf_data/nnpdf_data/commondata/DYE906_Z0_120GEV_DW/filter.py
@@ -45,8 +45,8 @@ class E906_DW_RATIO_commondata(commondata):
         MP = 0.938
         s = 2 * MP * (Ebeam + MP) * y / y
 
-        kin = pd.concat([y, data["M"] ** 2, s], axis=1)
-        kin = kin.set_axis(["y", "M2", "s"], axis=1)
+        kin = pd.concat([y, data["M"] ** 2, np.sqrt(s)], axis=1)
+        kin = kin.set_axis(["y", "M2", "sqrts"], axis=1)
         self.kinematics = kin.astype(float).to_numpy()
         self.kinematic_quantities = ["y", "M2", "sqrts"]
 
--- a/nnpdf_data/nnpdf_data/commondata/DYE906_Z0_120GEV_DW/kinematics.yaml
+++ b/nnpdf_data/nnpdf_data/commondata/DYE906_Z0_120GEV_DW/kinematics.yaml
@@ -9,7 +9,7 @@ bins:
     max: null
   sqrts:
     min: null
-    mid: 2.26879688e+02
+    mid: 1.50625260e+01
     max: null
 - y:
     min: null
@@ -21,7 +21,7 @@ bins:
     max: null
   sqrts:
     min: null
-    mid: 2.26879688e+02
+    mid: 1.50625260e+01
     max: null
 - y:
     min: null
@@ -33,7 +33,7 @@ bins:
     max: null
   sqrts:
     min: null
-    mid: 2.26879688e+02
+    mid: 1.50625260e+01
     max: null
 - y:
     min: null
@@ -45,7 +45,7 @@ bins:
     max: null
   sqrts:
     min: null
-    mid: 2.26879688e+02
+    mid: 1.50625260e+01
     max: null
 - y:
     min: null
@@ -57,7 +57,7 @@ bins:
     max: null
   sqrts:
     min: null
-    mid: 2.26879688e+02
+    mid: 1.50625260e+01
     max: null
 - y:
     min: null
@@ -69,5 +69,5 @@ bins:
     max: null
   sqrts:
     min: null
-    mid: 2.26879688e+02
+    mid: 1.50625260e+01
     max: null


Radonirinaunimi (changes requested):
Hi @peterkrack! Happy new year and thanks for this!

Here are some preliminary comments:
- Could you please fix the metedata regarding the arXiv, inSpire, and HepData urls? These will be crucial when moving into https://github.com/NNPDF/nnpdf/pull/2228
- Could you make sure that the filters run properly. Currently, this is not the case as these filters contain some calls to non-existing functions.
- Add the following to the filters to properly format the floats: https://github.com/NNPDF/nnpdf/blob/3f590241452ae68d4c577a8b98ed27b23eb14739/nnpdf_data/nnpdf_data/commondata/ATHENA_NC_45GEV_EP/filter.py#L12-L14
- Use pre-commit hooks https://pre-commit.com/ to make sure that all the files are properly formatted before each commit.

Could you please take of these first before I can move further into the details of the implementation?

Radonirinaunimi:
Hi @peterkrack, please let me know if there is anything I can help with in the meantime.

peterkrack:
> I was wondering if there are some updates regarding the pending issues above and whether some helps are needed?

For DYE605_Z0_38P8GEV_DW is suspect it is a bug in the old buildmaster. The other two I have to figure out what exactly is causing the discrepancy between legacy and reimplementation.

Radonirinaunimi:
> > I was wondering if there are some updates regarding the pending issues above and whether some helps are needed?
> 
> For DYE605_Z0_38P8GEV_DW is suspect it is a bug in the old buildmaster. The other two I have to figure out what exactly is causing the discrepancy between legacy and reimplementation.

I can also have a look in details, but would you be able to take care of https://github.com/NNPDF/nnpdf/pull/2248#pullrequestreview-2536510268 above?

peterkrack:
> Overall, this is close to completion except for `DYE605_Z0_38P8GEV_DW_PXSEC` in which the legacy implementation contains one extra-source of systematic uncertainties (cc @enocera).

Concerning the extra uncertainty:
rep00001 to rep00999
https://raw.githubusercontent.com/NNPDF/nnpdf/refs/tags/4.0.6/buildmaster/rawdata/DYE605/nuclear/output/tables/group_result_table.csv

Then in the old buildmaster
nrep is set to 1000
https://github.com/NNPDF/nnpdf/blob/f6c49ae23f38b277e109d891488b1a254ac52624/buildmaster/filters/FTDY.cc#L151

then later on the loop runs from irep=0 to irep=999; one uncertainty too much 
https://github.com/NNPDF/nnpdf/blob/f6c49ae23f38b277e109d891488b1a254ac52624/buildmaster/filters/FTDY.cc#L206

enocera:
Dear @peterkrack @Radonirinaunimi, let me try to clarify the uncertainties for the DY E605 data set.
- In the "old" commondata implementation, there were one statistical (uncorrelated) uncertainty and 1002 sources of systematic uncertainties. Of these 1002 sources, the first was a 10% uncorrelated (additive) uncertainty and the second a 15% correlated (multiplicative) normalisation uncertainty. The other 1000 uncertainties were "nuclear uncertainties", estimated as the difference between predictions obtained with proton and nuclear PDFs, taking the proton PDF fixed to the NNPDF4.0 central value and varying the nuclear predictions for each of the 1000 replicas in nNNPDF3.0.
- It seems to me that this implementation is correctly propagated into the legacy data set. Indeed, if I look at `uncertainties_legacy_PXSEC.yaml`, there are one statistical uncertainty, the two aforementioned systematic uncertainties, and 1000 nuclear uncertainties. In `uncertainties_reimplemented_PXSEC.yaml`, I'd say that one nuclear uncertainty is missing. This is consistent with what @peterkrack noticed above.
- Now, you're right in saying that there are only 999 nuclear uncertainties in the input file `rawdata/nuclear/output/tables/group_result_table.csv`. A replica must have gone missed in the generation process.

I see two ways of proceeding forward.
1) I re-generate the `rawdata/nuclear/output/tables/group_result_table.csv` with the missing nuclear uncertainty.
2) We live with 999 nuclear uncertainties for the time being and we forget about the 1000th. Nuclear uncertainties will have to be revisited anyways for NNPDF4.1.

enocera:
I have a preference for option 2.

Radonirinaunimi:
> I see two ways of proceeding forward.
> 
>     1. I re-generate the `rawdata/nuclear/output/tables/group_result_table.csv` with the missing nuclear uncertainty.
> 
>     2. We live with 999 nuclear uncertainties for the time being and we forget about the 1000th. Nuclear uncertainties will have to be revisited anyways for NNPDF4.1.

Thanks for your reply @enocera! I agree with all of your points - and the conclusions. As you see in the report, this differences is negligible so I tend to lean to the 2nd option (on top of the reason you said).

scarlehoff (changes requested):
The `kinematics_override` needs to be set to the identity. The `result_transform` we can live with for the time being.

Probably the process type needs to be changed as well, we shoudn't have a process type per dataset...

Radonirinaunimi:
> Probably the process type needs to be changed as well, we shoudn't have a process type per dataset...

This is probably very minor, but how would you call the process types then?

And are you happy with how the variants are called?

scarlehoff:
The variables are ok. For the process type... use DYP, otherwise you will need to dive into validphys to change a few things.

Radonirinaunimi:
> The variables are ok. For the process type... use DYP, otherwise you will need to dive into validphys to change a few things...
> 

Yep, this I know. But I was wondering if you want something specific after the `_`, ie `DYP_XX`. 

> although FTDY would be better imho but 🤷‍♂️

Accounting for this, I went for `DYP_FT`. I hope it is not too bad 😅

Radonirinaunimi:
> yes, that sounds ok, as long as the first three letters are DYP it will go through vp ok. It needs to be added to process options though.

When modifying the process options, I went for the easiest solution which is to simply add the variable `M2` into the `_dyboson_xq2map` instead of modifying the variables here to be `m_Z2` (although that should be the proper variable name). Doing the later turns out to be very messy as it also involves adding modifications to the filter rules.

[Final Report](https://vp.nnpdf.science/fkEf1n6bSH-eK8zShUcSrQ==/)

nnpdf_data/nnpdf_data/commondata/DYE605_Z0_38P8GEV_DW/metadata.yaml
```diff
@@ -59,5 +59,8 @@ implemented_observables:
     legacy:
       data_uncertainties:
       - uncertainties_legacy_PXSEC.yaml
-  data_central: data_legacy_PXSEC.yaml
+    reimplemented:
+      data_uncertainties:
+      - uncertainties_reimplemented_PXSEC.yaml
+  data_central: data_reimplemented_PXSEC.yaml
```
scarlehoff, line 65: Some final comments. This should not be `reimplemented`, this should just be the normal `data_uncertainties`, not a variant.

Same for all the others.

scarlehoff: we want to keep the old data under `legacy` for that dataset, and the new implementation be the default for all (so not keep `reimplemented` for any)

Radonirinaunimi: Sorry for the delay in this, but now I think everything is cleaned up.

scarlehoff: Thanks. You kept the `reimplemented` variant though (although you removed it from the names). Maybe you didn't commit that part?

scarlehoff (approved):
Looks good. I think we should change M2 to M (left a note in https://github.com/NNPDF/nnpdf/issues/2264)

Another thing to note is that for E866 one extra point is now cut by the internal cuts. This seems to be due to a small (% level) differences on `y`, but it looks like it is coming from hepdata so not much to do there. The chi2 just changes from 1.59 to 1.57 so I don't think we have to worry about it.

PR #2175, opened 2024-10-16
Reimplement hera datasets

peterkrack:
Added rawdata and filter.py script for HERA dataset.
Checks of reimplemented data are still missing.

PR #2255, opened 2025-01-10
Recover the special type and cuts for heavy quark dis

scarlehoff:
In #2175 all processes where changed to DIS_NC/CC losing the heavy quark information, and so the internal cuts for charm were not applied.

This might fix the problem you are seeing @RoyStegeman


RoyStegeman (approved):
Thanks!

PR #2257, opened 2025-01-15
Recover 3.9 test

scarlehoff:
Let's recover the 3.9 test. Perhaps we need to skip the regression tests for 3.9 but that's fine.


scarlehoff:
@RoyStegeman if you don't have anything against this, I'll merge?

RoyStegeman (approved):
Fine for me. I think we never tested 3.9 in the python installation, only in the conda one (but that wasn't really testing 3.9 as we discovered)

PR #2258, opened 2025-01-17
True Positive and Negative rates in multiclosure weighted fits

comane:
Module for computation of True Positive and Negative rates for flagging a dataset as inconsistent.

When the fit is weighted TPR and TNR are computed taking into account whether adding the weight has deteriorated the overall fit quality.

- [x] Add arXiv reference once paper is out
- [x] notation of various sets should be consistent with the one used in the paper (currently S2 <-> S3)
- [x] polish plotting functions


validphys2/src/validphys/closuretest/multiclosure_nsigma.py
```diff
@@ -0,0 +1,308 @@
...
+        else:
+            columns_bools = np.any((df_weight - df_ref).values > z_alpha, axis=0)
+
+        set3_alpha[z_alpha] = set(df_weight.columns[columns_bools])
+
+    return set3_alpha
+
+
+def set_3(dataspecs_multiclosurefits_nsigma: list, weighted_dataset: str) -> dict:
+    """
```
scarlehoff, line 270:
```suggestion
    r"""
```

validphys2/src/validphys/closuretest/multiclosure_nsigma_helpers.py
```diff
@@ -0,0 +1,170 @@
...
+
+@dataclasses.dataclass
+class CentralChi2Data:
+    value: float
+    ndata: int
+    dataset: DataSetSpec
+
+    @property
+    def reduced(self):
+        return self.value / self.ndata
```
scarlehoff, line 29: There's already `Chi2Data` in results... I wonder whether it'd be better to extend that one? Or maybe use a different name here?

The difference is that this one holds the dataset only while the other holds a whole result object.

(in any case docstr needed)

validphys2/src/validphys/closuretest/multiclosure_nsigma_helpers.py
```diff
@@ -0,0 +1,170 @@
...
+    ----------
+    dataset: validphys.core.DataSetSpec
+    pdf: validphys.core.PDF
+
+    Returns
+    -------
+    pd.DataFrame
+        index is datapoints, column is the central prediction.
+    """
+    return convolution.central_predictions(dataset, pdf)
```
scarlehoff, line 46:
```suggestion
from convolution import central_predictions
```

would be the same. You can add the type hints there if you need them.

validphys2/src/validphys/closuretest/multiclosure_nsigma_helpers.py
```diff
@@ -0,0 +1,170 @@
...
+    CentralChi2Data
+    """
+    diff = (
+        central_predictions.values.ravel()
+        - loaded_commondata_with_cuts.central_values.values.ravel()
+    )
+
+    value = calc_chi2(sqrt_covmat, diff)
+    ndata = len(central_predictions)
+    return CentralChi2Data(value=value, ndata=ndata, dataset=dataset)
```
scarlehoff, line 79: Test it because I'm not 100% sure (maybe you are doing something in the multiclosure that will break this) but _in principle_ this function could just depend on `abs_chi2_data` which would automatically get the predictions, the data and the compute the chi2.

validphys2/src/validphys/closuretest/multiclosure_nsigma_helpers.py
```diff
@@ -0,0 +1,170 @@
...
+    or there is more than one weighted dataset, an error is raised.
+
+    Parameters
+    ----------
+    fits_data: list
+        List of data for each fit.
+
+    Returns
+    -------
+    str or None
```
scarlehoff, line 132: It returns True/False

validphys2/examples/true_pos_neg_closuretests.yaml
```diff
@@ -0,0 +1,114 @@
+meta:
+  title: Inconsistency classification probabilities
+  author: Lazy Person
+  keywords: [nsigma, chi2, multiclosure test, inconsistent]
+
```
scarlehoff, line 5: Add perhaps a comment saying that the commented dataset are those to reproduce the paper.

scarlehoff (commented):
I think it is mostly ok. The only thing is that, if possible, I'd try to trim down the helpers since some of the functionality is already in `results.py` (or, the one that is not, would be better suited as enhancements there).

validphys2/src/validphys/closuretest/__init__.py
```diff
@@ -13,3 +13,6 @@
 from validphys.closuretest.multiclosure_preprocessing import *
 from validphys.closuretest.multiclosure_pseudodata import *
 from validphys.closuretest.inconsistent_closuretest.multiclosure_inconsistent_output import *
+from validphys.closuretest.multiclosure_nsigma_helpers import *
+from validphys.closuretest.multiclosure_nsigma import *
+from validphys.closuretest.multiclosure_nsigma_output import *
```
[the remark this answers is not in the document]
comane: I agree with you but I also think that it is very useful to have them there as vp-providers.
So any time someone adds a table or figure function in the _output module it does not have to also add it to the __init__ file but can already use it in the reports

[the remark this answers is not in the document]
comane: Is there another way you can think of perhaps that avoids having to hand pick the needed functions?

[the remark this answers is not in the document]
scarlehoff: You either need to handpick or play around with `importlib` and the namespaces.

But I think handpicking is good because that way one can open `__init__` and know inmediately what is available (also when using autocompletion you don't get 200 entries in the list...).

It also avoids someone writting a function in `multiclosure_nsigma_output` with the same name as in `multiclosure_pseudodata` and breaking everything as a result.

> So any time someone adds a table or figure function in the _output module it does not have to also add it to the init file but can already use it in the reports

Ideally the person would also write documentation and tests for that function, so adding it to the `__init__` is a very small part of their time.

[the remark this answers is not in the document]
comane: But it comes with some complications too, because one of the problems, and very unintuitive things, is that I also need to import function from module x if they are only used in the same module x as providers.
An example is `multiclosurefits_nsigma` which is used in a `collect` in `multiclosure_nsigma.py`.

This is quite unintuitive at first I think.
But perhaps fine since it is how vp works?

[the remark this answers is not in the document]
scarlehoff: If you want to use it as a provider and the validphys provider is pointing to `closuretest/__init__` (as it is at the moment, my personal take is that it was a mistake but alas) yes, you need to import both.

If you didn't actually want `multiclosurefits_nsigma` to be a provider instead of using it inside a `collect` it should've just been a function which takes as input something that happens to be a collection.

perhaps there is a `collection_module` reportengine function which would look inside the module automagically to do the equivalent to

```
def some_fun(collected_item):
    return map(_my_module_private_function, collected_items)
```
?

validphys2/src/validphys/closuretest/multiclosure_nsigma_helpers.py
comane: What name would you go for? Would `DatasetChi2` work for you?

scarlehoff: Yes. But also, I think it should be in results?

Btw, this is a case (and the one below) where there a clashes with other parts of the code generated by the `import *`.

comane: The one below yes, but `CentralChi2Data` is only defined in multiclosure to my knowledge

validphys2/src/validphys/closuretest/__init__.py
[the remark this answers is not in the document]
comane: ok, I opted for having the single imports.

comane:
Hi @scarlehoff, thank you very much for your review comments.

I should have addressed most of them now.

If you prefer I can also move the `CentralChi2Data` function to `results.py`

validphys2/src/validphys/closuretest/__init__.py
```diff
@@ -13,3 +13,32 @@
 from validphys.closuretest.multiclosure_preprocessing import *
 from validphys.closuretest.multiclosure_pseudodata import *
 from validphys.closuretest.inconsistent_closuretest.multiclosure_inconsistent_output import *
+from validphys.closuretest.multiclosure_nsigma_helpers import (
```
scarlehoff, line 16:
```suggestion
# TODO: 
# select the functions to be exposed from the previous modules
# and make sure they are flagged smh as closure_test providers
from validphys.closuretest.multiclosure_nsigma_helpers import (
```

scarlehoff (approved):

PR #2266, opened 2025-01-28
Remove completely kinematics_override

### LINKED ISSUE #2252: Remove `kinematics_override`.
`kinematics_override` will be removed and so all plots will be produced with whatever is in the metadata.

For datasets with `kinematics_override: identity` there will be no problems, for the rest as the problems surface they'll be updated.

scarlehoff:
This PR closes #2252, which has required datasets that either have only legacy version or for which it was just not done during the translation.

For HERA data in particular the situation was a bit dire since the plots reported Q2 but the numbers were actually Q. Now it should be Q2 everywhere, which might not be what everybody wants, but at least is correct.
The same is true for the NMC (with a horrible result, but the previous situation was just [silly](https://github.com/NNPDF/nnpdf/blob/8327d16719c638ef0fd2eb6af788d9a7836146d3/validphys2/src/validphys/plotoptions/kintransforms.py#L365) so I've modified the kinematics for that) and NUTEV.

(personally I'd move all the squared quantities to just... not the square?)

Edit: the fecking NMC legacy dataset with kinematics done with a dice had of course to be part of every single regression test 🤷 I hope we can move to the hepdata-binned version.


validphys2/src/validphys/process_options.py
```diff
@@ -229,8 +237,11 @@ def _hqp_mqq_xq2map(kin_info):
 
 def _inc_xq2map(kin_info):
     # Compute x, Q2
-    # k2 necessary to take the mass for DY inclusive cross sections still not migrated
-    mass2 = kin_info.get_one_of(_Vars.m_W2, _Vars.m_Z2, _Vars.m_t2, "k2")
+    try:
+        mass = kin_info.get_one_of(_Vars.m_W, _Vars.m_Z, _Vars.m_t, _Vars.m_ll)
```
Radonirinaunimi, line 241: Why are the first three needed? AFAICT, there are no datasets (and there shouldn't) that are expressed in these (?).

validphys2/src/validphys/plotoptions/core.py
```diff
@@ -311,10 +311,7 @@ def get_xq2map(kintable, info):
     """
     try:
         return info.process_type.xq2map(kintable, info.ds_metadata)
```
Radonirinaunimi, line 313: Unrelated but stumbled upon it during CR: the return within `try` statements is not recommended by [PEP 8](https://peps.python.org/pep-0008/#programming-recommendations).

Radonirinaunimi (commented):
Thank you very much for this (especially for fixing the remaining quirks of the old re-implemented sets)!!

So when would the `k_` be completely stripped out of the codes? And do you perhaps happen to have a comparison of the kinematic plots?

validphys2/src/validphys/process_options.py
scarlehoff: As I put in the OP, I think we should not use the squares (since they look much uglier in plots), since I was changing these files I thought I might as well prepare it.

scarlehoff:
yes, the `k` will (in a subsequent PR) all be removed. I cannot do it here because internally validphys uses the f**** kn everywhere so it requires some care.

Regarding the plots, I do have comparisons for the entire set of data. imho when they have changed they look better but I don't want to clutter the server too much, as a surrogate you can look at this comparisons between this comparefit: https://vp.nnpdf.science/a0kifkAWSlSTtrwUa7KqbQ==

and this older one: https://vp.nnpdf.science/eB5jnBi4Qn-OEc2KCq_i1g==

(if you want to look at the whole set of changes I can upload them to the server)

validphys2/src/validphys/plotoptions/core.py
scarlehoff: I can modify it here if you want. While my hope is to remove this whole module, as long as it doesn't really mess up with other parts of the code (the kinematics override was messing up) it will take me a long time to actually get to it.

validphys2/src/validphys/kinematics.py
```diff
@@ -25,12 +26,10 @@ def describe_kinematics(commondata, titlelevel: int = 1):
 
     titlelevel can be used to control the header level of the title.
     """
-    import inspect
-
     cd = commondata
     info = plotoptions_core.get_info(cd)
     proc = cd.load_commondata().commondataproc
-    src = inspect.getsource(info.kinematics_override.xq2map)
+    src = ""
```
giacomomagni, line 32: can't you drop this `src` at all?

giacomomagni (commented):
Looks cleaner!!

validphys2/src/validphys/plotoptions/core.py
scarlehoff: On second thoughts, this function is no longer needed so I'll remove it.

Radonirinaunimi: Perfect! Thanks.

Radonirinaunimi:
> yes, the `k` will (in a subsequent PR) all be removed. I cannot do it here because internally validphys uses the f**** kn everywhere so it requires some care.
> 
Okok, good!

> Regarding the plots, I do have comparisons for the entire set of data. imho when they have changed they look better but I don't want to clutter the server too much, as a surrogate you can look at this comparisons between this comparefit: https://vp.nnpdf.science/a0kifkAWSlSTtrwUa7KqbQ==
> 
> and this older one: https://vp.nnpdf.science/eB5jnBi4Qn-OEc2KCq_i1g==
> 
As far as I can tell, everything looks good.

> (if you want to look at the whole set of changes I can upload them to the server)

Maybe  even just for the record, this is good to have on the server in any case.

scarlehoff:
> Maybe even just for the record, this is good to have on the server in any case.

I put them in `/home/nnpdf2/nnpdf/PR_tests/PR_2266_removal_of_kinematics_override` (this is a second disk we got since the first one was full, I've started moving things around but still lots to do). The folders are separated by `<experiment>_master` which is run with master and `<experiment>` which is run with this branch. To first approximation I ran them commit by commit (so the LHCB folder refers to the LHCB commits and so on).

Radonirinaunimi (approved):

PR #2268, opened 2025-02-03
Modify the VP api so it can take its own objects.

scarlehoff:
Right now the validphys api is limited to strings for PDFs. This small change allows to take either already loaded PDFs or other objects (such as the PDFs coming directly from n3fit).

Note: while `from __future__ import annotations` would allow using `Union[str, PDF]` in python 3.9, this in practice changes PDF to "PDF" (which was the old way) and this is not allowed/understood by reportengine. So for now the type annotation for the pdf name must be removed for `parse_pdf` to allow both strings and classes.

### DIFF
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -7,6 +7,7 @@ import inspect
 import logging
 import numbers
 import pathlib
+from typing import Union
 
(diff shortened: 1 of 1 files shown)


tgiani (approved):
Looks good to me

PR #2269, opened 2025-02-03
correct observable: CMS_TTBAR_13TEV_LJ_2016_DIF_YTTBAR -> YT

RoyStegeman:
The `observable_name` of `CMS_TTBAR_13TEV_LJ_2016_DIF` was incorrectly defined as `YTTBAR` while it should have been `YT`.


nnpdf_data/nnpdf_data/commondata/dataset_names.yml
```diff
@@ -330,7 +330,7 @@ CMS_SINGLETOP_TCH_TOT_7TEV:
   dataset: CMS_SINGLETOP_7TEV_TCHANNEL-XSEC
   variant: legacy
 CMS_TTB_DIFF_13TEV_2016_LJ_TRAP:
-  dataset: CMS_TTBAR_13TEV_LJ_2016_DIF_YTTBAR
+  dataset: CMS_TTBAR_13TEV_LJ_2016_DIF_YT
```
scarlehoff, line 333: Could you also add a line (perhaps with a comment saying that it is a bug correction) for YTTBAR -> YT so that runcards that have been run with the wrong name still run.

scarlehoff (commented):
Could you also bump the version of the dataset?

nnpdf_data/nnpdf_data/commondata/dataset_names.yml
RoyStegeman: I think that's a bad idea since YTTBAR is also a valid process so it can lead to confusion if people think the distribution is YTTBAR (I already made a mistake based on it). I think I'll locally add a mapping in dataset_names (and other can perhaps do the same) but I don't think that should be done in master

Issue #2267, opened 2025-01-31
Chek the ATLAS Z0 8 TeV low mass data set


enocera:
In an attempt to understand what may be wrong with the ATLAS Z0 8 TeV low mass data set, I suggest to implement a variant of the data set in which we implement the piece of information of Table 6 in [1710.05167](https://arxiv.org/pdf/1710.05167). As you can see, there are just four sources of uncertainty there: the uncorrelated statistical uncertainty, the uncorrelated systematic uncertainty, the 100% correlated systematic uncertainty and the 100% 1.8% luminosity uncertainty. This is way simpler than dealing with the 200+ systematic uncertainties available on Hepdata that are currently implemented. In my opinion this variant can be useful to understand to check if the chi2 is really crazily high when correlating the lumi uncertainty. We can then go back and compare the two variants in an attempt to check the correlation of each of the 200+ sys uncs at a later stage. @scarlehoff can you assign the implementation of the variant to some of us?

PR #2270, opened 2025-02-03
Chek the ATLAS Z0 8 TeV low mass data set

achiefa:
This PR addresses #2267.


achiefa:
Dear @enocera,

As mentioned in issue #2267, I've implemented the "light" variant of the dataset. In this variant, statistical and systematic uncertainties are taken from Table 6 in [1710.05167](https://arxiv.org/pdf/1710.05167), on top of which I included the luminosity uncertainty (1.9%). However, I took the liberty to play with the luminosity uncertainty and I combined it with different variants, which are listed below:
1.  HepData v3 (__ATLASLUMI12__): Here  I took version 3 of the HepData table and I correlated the luminosity uncertainty using the key __ATLASLUMI12__. This corresponds to the new default implementation.
2. HepData v3 (__CORR LUMI__): As (1), but now the luminosity is correlated _within_ the same experiment (key __CORR__).
3. HepData v3 (__UNCORR LUMI__): As (1), but now the luminosity is _not_ correlated at all (key __UNCORR__).
4. HepData v1 (__ATLASLUMI12__): As (1), but with version 1 in HepData
5. HepData v1 (__CORR LUMI__): As (2), but with version 1 in HepData
6. HepData v1 (__UNCORR LUMI__): As (3), but with version 1 in HepData
7. Syst. light (__ATLASLUMI12__): As (1), but using the light version of systematic uncertainties.
8. Syst. light (__CORR LUMI__): As (2), but using the light version of systematic uncertainties.
9. Syst. light (__UNCORR LUMI__): As (3), but using the light version of systematic uncertainties.

For each of these variants, I computed the $\chi^2$ to `ATLAS_Z0_8TEV_LOWMASS` using the same pdf set (`250122-jth-01-data-check`, the one used by RS in his report). You can find the report [here](https://vp.nnpdf.science/ry_P0D3VTuenwXhSqCH2mA==).

I think the message is rather clear -- the $\chi^2$ improves when the luminosity uncertainty is not correlated. This is true for version 3 on HepData and the light version (__3__ and __9__ in the list above). On the other hand, the $\chi^2$ for version 1 is always high (~50) regardless of the type of correlation assigned to the luminosity.

The list above should provide a comprehensive set of combinations useful for this investigation. Please, let me know if something is missing. Also, I've pushed the uncertainty data files that I used to produce the variants, so that you can have a look if something does not convince you.

P.S. In the original re-implementation, the luminosity uncertainty was set to 1.8%, although the paper claims 1.9%. Apart from version 1, which already includes (God knows what) luminosity uncertainty in the HepData table, all the other variants use 1.9%.

scarlehoff:
Some results from today's code meeting, using directly the NNLO grids (which are slightly different) to compute the chi2 and `NNPDF40_nnlo_as_01180` as PDF.

legacy: 0.6911
default: 8.743
default with `CT18Z`: 5.649

achiefa:
Values of the $\chi^2$ per bin:
1. $[46, 66]$:  [link](https://vp.nnpdf.science/ALifN94vSlaIYwiXx2ATrA==)
2. $[66, 80]$: [link](https://vp.nnpdf.science/v0SJcynIQIiYZFujVP1YRQ==)
3. $[80, 91]$: [link](https://vp.nnpdf.science/ISELZgruRgeGc_oEz7qjOg==)
4. $[91, 102]$: [link](https://vp.nnpdf.science/zx7nhJk2RW2Os-F83hTEdA==)
5. $[102, 116]$: [link](https://vp.nnpdf.science/fjhdKZ-UQNO_Tm0-Pmw6FA==)
6. $[116, 150]$: [link](https://vp.nnpdf.science/RSE0pIOjRO-_ojMmwdUMJQ==)
7. $[150, 200]$: [link](https://vp.nnpdf.science/UtLF0vb-TAi27yVCPLKiJQ==)

Comments soon...

scarlehoff:
Thank you very much for this @achiefa 
Is this the same dataset that was problematic in the pheno paper? I thought we had checked that the old version was ok even for the peak or is this just another dataset which happens to also have problems in the peak?

achiefa:
> edit: silly me, the old version was always legacy so luminosity was always uncorrelated

Indeed, I was just about to say that.

enocera:
> edit: But maybe that can explain why we could not describe well the new ATLAS analysis at Z -peak. That's just me thinking out loudly, but maybe the new ATLAS analysis, which is at the Z -peak, fixes the problematic correlation of the luminosity. If that is true, then the old dataset (ATLAS_Z0_8_TEV_LOWMASS @ Z -peak) and the new one are simply incompatible.

Very well thought - this is the explanation I'm leaning towards.

scarlehoff:
The new one is incompatible also with other datasets though, since we were not able to fit it by itself without mhou...

enocera:
@achiefa Question: all these chi2 are w/o MHOUs, right? So that we can compare with the old NNPDF4.0 fit?

achiefa:
Yes, all those chi2's are w/o MHOUs.

enocera:
Dear @achiefa I looked into the MSHT paper, see in particular Fig. 7 of https://arxiv.org/pdf/2012.04684. They have cuts on the ATLAS 8 TeV 2D low mass data, in particular they exclude, for each invariant mass bin, the two most forward rapidity bins. They also remove the lowest invariant mass bin altogether. Could you please try to recompute the chi2 with these cuts? Thanks.

achiefa:
Hi @enocera,

I will compute the chi2. On top of the cuts that you mentioned, should I relax the cuts that remove the two last bins in the invariant mass?

enocera:
In our framework, we want to always remove the last two invariant mass bins, because these are incorporated in the high-mass data set. So I would NOT relax the cuts and continue to exclude the last two invariant mass bins.

achiefa:
## __Follow-up__

I computed the $\chi^2 as requested by @enocera, namely applying the same cuts as MSHT. You can find the report [here](https://vp.nnpdf.science/UGQb8kC3QzylI8D2E7HVFg==). As far as I can understand, I don't see any improvement in the $\chi^2$ - actually, the value worsens in any dataset variant.

I also computed the $\chi^2$ for two other configurations:
- computing the $\chi^2$ for all the bins but the one at the Z peak using the same set as in the previous comments (`250122-jth-01-data-check`). The last two invariant mass bins are still excluded. The report is [here](https://vp.nnpdf.science/ZSbqcyrFQnahroGPIXSe_Q==).
- computing the $\chi^2$ using `241003-01-rs-mhou` which includes MHOUs. The report is [here](https://vp.nnpdf.science/8a6l-lRITLuzyVx4xdtH1Q==).

I'll compute the $\chi^2$ for other configurations, for instance as done in the pheno paper where the calculation of the $\chi^2$ includes the covmat for MHOUs.

enocera:
Thanks, Amedeo. So should we conclude that, as soon as we put all the invariant mass bins together (but those on the Z peak), we get a horrible chi2?

scarlehoff:
I've been talking a bit with Tom about this problem and trying to compare our data to theirs... statistical uncertainties and data seem to be the same, and the 200+ systematics are impossible to compare well but one thing I noticed is that their luminosity uncertainty is 1.9% while we have 1.8% why is that?
Looking at the paper that we reference for this it says 1.9% https://arxiv.org/pdf/1710.05167

enocera:
This is done on purpose. The luminosity uncertainty is 100% correlated across different ATLAS experiments, therefore, even if the paper says 1.9%, I set it up to 1.8%, which is the value (also mentioned in ATLAS papers) corresponding to that specific 8 TeV run. You can try to have a variant in which this is set to 1.9% and check the difference. I bet that this is not going to solve the problem.

enocera:
Maybe the MSHT crowds have decorrelated some specific uncertainties? Or perhaps they are not correlating (certain) uncertainties across invariant mass bins?

enocera:
@scarlehoff Thanks for checking with Tom.

PR #2271, opened 2025-02-04
Force the usage of new names when running a fit

scarlehoff:
This PR adds a flag `allow_legacy_names` which defaults to  `False` in n3fit (and, for now, to `True` for other actions). As the flag indicates, when it is `False` there is no translation and people is forced to use new names.

With this, new fits will always be done with new names but older fits will still be usable for comparisons or `from_fit` actions.

(most of the changes are to the error messages, trying to be very explicit about what went wrong)

In order to avoid rogue individuals to add `allow_legacy_names: True` to the runcard, if we see it happening we can just make `vp-upload` fail in such cases.


n3fit/src/n3fit/scripts/n3fit_exec.py
```diff
@@ -159,6 +159,8 @@ def from_yaml(cls, o, *args, **kwargs):
             N3FIT_FIXED_CONFIG['use_thcovmat_in_sampling'] = thconfig.get(
                 'use_thcovmat_in_sampling', True
             )
+        # TODO: this whole function is upside down?
+        N3FIT_FIXED_CONFIG["allow_legacy_names"] = file_content.get("allow_legacy_names", False)
```
RoyStegeman, line 163:
```suggestion
```

Is changing the value from the runcard not specifically what you want to avoid?

RoyStegeman (approved):
Thanks! Perhaps it's worth adding a deprecation warning in `legacy_to_new_map`?

Let's wait with merging until @enocera has had a look at #2269

n3fit/src/n3fit/scripts/n3fit_exec.py
scarlehoff: The value in the runcard should take precedence imho.
But we can remove the option to use it (so if someone wants to be evil they need to really hack the code)

PR #2274, opened 2025-02-06
correctly load different types of theory covmats

RoyStegeman:
Since the `point_prescriptions` and `user_covmat_path` were not added to the n3fit config dict they would not be known by the actions called when running n3fit. In particular this affects the `generic_path` that defines which covmat to look for inside the tables/ folder.


scarlehoff (commented):
So this `theorycovmatconfig` namespace is ignored by everyone in vp, right?

This function really needs to be refactored very strongly :__

scarlehoff (approved):

PR #2275, opened 2025-02-11
don't resample L2 pseudodata

RoyStegeman:
I don't see why we don't make this the default. It was said that at the level of the PDFs this doesn't have an impact (which is largely true) but in cases where it does, such as alphas extractions, this is the wrong thing to do. In general, changing the distribution of the data doesn't just cut the unphysical tail but also shifts the central value so I don't think this can be justified. 

At the Amsterdam meeting it was agreed to try three options for the sampling (old resampling, no resampling and symmetric) and see the impact at the PDF level.


scarlehoff:
Let's have a last tag where we try to be compatible* with the runs that have been used up to the last batch of papers and from then onwards start introducing "breaking" changes that might staty, not stay, etc.

I see this in approximately the same "breaking" footing as using the new names for the seeding of the replicas (right now we are still using the old data).

*sadly this has been broken for the FTDY experiment since one datapoint has dissapeared 😭

scarlehoff:
Do we want this to go together with the diagonal default? I'd prefer it since then I only have to redo all tests once (and 4.1.p will have this, t0 sampling and diagonalization as breaking changes)

scarlehoff (approved):

PR #2276, opened 2025-02-12
Treat 1-point datasets equally in sequential and parallel fits

scarlehoff:
Due to the shape-changing nature of boolean masks we decided to just put single-point datasets in training when running in GPU. This PR removes that limitation by just accepting the point in both training and validation and setting to 0 the row and column in the inverse covmat (so the masking happens at the level of the loss).

If this works ok, #2138 comes for free.

@RoyStegeman @achiefa I want to run a few tests first before considering this good:

- [x] That Etr and Evl are ok for multireplica fits (they are ok for single replicas given that the regression test pass)
- [x] Running a fit with only single-point datasets (this is actually broken in master)
- [x] Experiments with one single dataset which is one-single-point.
- [x] Check whether the written down replicas are _really_ the same in sequential and parallel 
    - [x] Add a test for exactly this situation
    - [x] Separate the `.csv` files per replica

(any help running the checks would be appreciated ofc, the more eyes the better)

This is needed for tree-saving reasons.


scarlehoff:
This is ready for review.

@achiefa please have a look if you have time to see whether this works for your use case. In principle it should store the replica pseudodata in each folder in GPU just like it does when you run sequentially.

@comane to achieve this I've created a decorator which modifies reportengine's decorator. Not sure whether what I've done is legal or whether there's an easier more official way to achieve this?

n3fit/src/n3fit/model_gen.py
```diff
@@ -26,7 +26,7 @@
     base_layer_selector,
 )
 from n3fit.backends import operations as op
-from n3fit.backends import regularizer_selector
+from n3fit.backends import regularizer_selector as reg_sec
```
RoyStegeman, line 29: Why this? Presumably you called it regularizer_selector because it's more descriptive, which is now lost

validphys2/src/validphys/n3fit_data.py
```diff
@@ -274,18 +332,38 @@ def fitting_data_dict(
...
+        covmat_vl[data_zero_vl, data_zero_vl] = 1.0
+
+        invcovmat_tr = np.linalg.inv(covmat_tr)
         invcovmat_vl = np.linalg.inv(covmat_vl)
 
-    ndata_tr = np.count_nonzero(tr_mask)
-    expdata_tr = expdata[tr_mask].reshape(1, ndata_tr)
+        # Set to 0 the points in the diagonal that were left as 1
+        invcovmat_tr[np.ix_(data_zero_tr, data_zero_tr)] = 0.0
+        invcovmat_vl[np.ix_(data_zero_vl, data_zero_vl)] = 0.0
```
RoyStegeman, line 356: I don't think this is mathematically sound. You take the inverse of the entire matrix, not element-by-element, so doing the inverse and then setting to 0 the points you don't want is different from slicing and then taking the inverse. The normalisation (determinant) will account for those points you effectively remove in these last two lines

RoyStegeman: Ah, right, yes that should be fine indeed

n3fit/src/n3fit/model_gen.py
```diff
@@ -170,6 +171,10 @@ def observable_generator(
         boundary_condition: dict
             dictionary containing the instance of the a PDF set to be used as a
             Boundary Condition.
+        mask_array: np.ndarray
```
RoyStegeman, line 174:
```suggestion
        training_mask_array: np.ndarray
```

I think this would help a lot with readability

n3fit/src/n3fit/model_gen.py
```diff
@@ -170,6 +171,10 @@ def observable_generator(
...
+            training mask per replica
+        validation_mask_array: np.ndarray
+            validation mask per replica, when not given ¬mask_array will be used
```
RoyStegeman, line 177: Maybe it's worth clarifying (or citing where info can be found) in which scenario the validation mask can be different from ~training_mask

RoyStegeman (approved):

validphys2/src/validphys/n3fit_data.py
```diff
@@ -274,18 +332,38 @@ def fitting_data_dict(
...
-        invcovmat_tr = np.linalg.inv(covmat_tr)
-
         covmat_vl = covmat[vl_mask].T[vl_mask]
+
+        # Remove possible correlations for 1-point datasets
+        # that should've been masked out
+        covmat_tr[data_zero_tr, :] = covmat_tr[:, data_zero_tr] = 0.0
+        covmat_vl[data_zero_vl, :] = covmat_vl[:, data_zero_vl] = 0.0
+        # Avoid infinities
+        covmat_tr[data_zero_tr, data_zero_tr] = 1.0
```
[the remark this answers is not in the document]
RoyStegeman: Because then the determinant would be 0 and thus the inverse would be infinite

n3fit/src/n3fit/model_trainer.py
```diff
@@ -151,7 +151,6 @@ def __init__(
         self.exp_info = list(exp_info)
```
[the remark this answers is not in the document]
scarlehoff: tbf the input to this function has outgrown its intended purposes. I'll add docs for each of the arguments to make sure it says what they are now.

at some point that function input should be refactored...

scarlehoff:
> I'm okay with that. Some things were not clear to me. But I think that is because I don't have as strong a knowledge of the codebase as you have.

This is why it is important to have less experienced people doing the review, so they can become experienced! Thanks. I'll add the requested comments / docs.

validphys2/src/validphys/n3fit_data.py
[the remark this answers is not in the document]
achiefa: Thanks

PR #2280, opened 2025-02-14
sample thcovmat only for L2 data, not L1

RoyStegeman:
To do a closure test of alphas determinations with the TCM we want to sample the theory covmat in the generation of L2 data, but not in the generation of L1 data. So there should be a separate key to activate sampling in the latter rather than using `use_thcovmat_in_sampling` to decide wether the covmat should be used both at L1 and L2. 

Having said that, there are plans to also include MHOUs in a closure test which would complicate things since then we want to use the MHOU covmat in the L1 sampling but not the alphas covmat. Not sure how this can be done somewhat elegantly...

### DIFF
--- a/n3fit/src/evolven3fit/evolve.py
+++ b/n3fit/src/evolven3fit/evolve.py
@@ -12,6 +12,9 @@ import psutil
 import eko
 from eko import basis_rotation, runner
 from validphys.utils import yaml_safe
+import tempfile
+import shutil
+import os
 
 from . import eko_utils, utils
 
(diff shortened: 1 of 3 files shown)


scarlehoff:
Do we want to keep these changes?

The point about EKO I've also run into a few times and I was thinking it would be a good idea to separate eko and theory in the profile (and the folders), so one can point the eko to the node easily if needed.

scarlehoff:
> Copying the EKO to the node didn't resolve the issue

What was the issue for you? In my case it was lack of storage in $HOME so I was downloading directly to the node.

RoyStegeman:
For me the issue is that EKO cannot be loaded by multiple processes at the same time

scarlehoff:
Ah, I see. I wonder why since shutil should just... copy it? It might be fixed as well if they get directly downloaded to the node but quite annoying...

In any case, I'll close this PR.

RoyStegeman:
I mean that multiple processes cannot copy the .tar at exactly the same time, which happens sometimes if you run O(100) fits. I don't remember the exact error message, but in some cases it messes up the eko and I have to redownload it before I can evolve the fits

RoyStegeman:
> One option is to send the EKO as part of the submission so the copy is managed by the queuing system but it might make things worse?

That's what I do right now, but I only have one original copy of the EKO. So the problem occurs when different submissions try to access the same EKO at the same time. I'm not sure there is a fix other than improved performance so we don't have to precompute the PDF EKOs anymore

scarlehoff:
If the copy is part of the submission (and not a cp in the script, which should also be fine anyway) then I'd say the problem is in the condor of nikhef...

PR #2282, opened 2025-02-20
Make pymongo optional

scarlehoff:
So that we don't need to have `pymongo` (or `mongodb`) installed just to run a fit or the tests. Having to have a very specific version is becoming a pain.


Radonirinaunimi (approved):
Thanks! LGTM

PR #2285, opened 2025-02-26
Positivity fktables at Q2 = 10

comane:
pinecards: https://github.com/NNPDF/pinecards/pull/183

Report
-------
Comparison between MS Bar POS @ Q2 = 5 vs Q2 = 10

https://vp.nnpdf.science/WOLIggtvSxCX36RQv3KloA==/


scarlehoff:
Please, remove the fonll cards. At the moment when you generate them with pineko they would be created in whatever folder the original card is anyway. Personally I'd prefer if we remove completely the need for this intermediate theory cards... cc @felixhekhorn @giacomomagni

RoyStegeman:
The whole point is that nFONLL works at observable level so you need an fktable for each nf. We can indeed choose not to store the theory cards anywhere, and this is probably a reasonable thing to do (like arguably we shouldn't store the operatorcards), especially in the nnpdf_data repo, but we can't entirely avoid intermediate theory cards (and ekos, and fktables).

felixhekhorn:
> The whole point is that nFONLL works at observable level so you need an fktable for each nf.

Indeed. As I already argued in https://github.com/NNPDF/pineko/issues/202 having different theory (and operator) cards disentangles the programs: eko <-> pineko <-> pinefarm

If and where to store them (while being generated objects) is a separate issue ...

nnpdf_data/nnpdf_data/commondata/NNPDF_POS_3P16GEV/kinematics_F2U.yaml
```diff
@@ -0,0 +1,161 @@
+bins:
+- x:
+    min: null
+    mid: 5.0e-07
+    max: null
+  Q2:
+    min: null
+    mid: 5.0
```
felixhekhorn, line 8:
```suggestion
    mid: 10.0
```
?

scarlehoff:
It requires a different grid per nf but a single fktable.

Given a theory (meaning, the main theory) the theory cards for all 6 grids are fixed so there's no need (and we should not) keep all these 6 extra theories.

felixhekhorn:
> It requires a different grid per nf

mmh, almost no: there are 7 theory cards and 7 different grids. There is one eko per nf.

> but a single fktable.

for you as a PDF fitter it may look like this - but as soon as you undo the tiny final step (for getting this 1 FK table) the answer is: no, there are 7 FK tables, which are plainly summed over. And we need these 7 FK tables and they are (more or less) fully reasonable FK tables.

> Given a theory (meaning, the main theory) the theory cards for all 6 grids are fixed so there's no need (and we should not) keep all these 6 extra theories.

this is doable by not committing them here (so you don't see them) - but locally we need them.

> edit: clearly all solutions until now have been confusing for everyone who has been dealing with it.

fine, we can/should improve the documentation, but https://pineko.readthedocs.io/en/latest/theory/fonll.html gives more or less clear instructions on how to proceed. We can leave FONLL to pineko (i.e. NNPDF does not need to care), but we need the details there. And we should leave it _only_ to pineko, i.e. we will not implement the theory expansion procedure (1 -> 7) into pinefarm: those two communicate by the already established interface: theory cards.

Since @comane just went through this: do you have a suggestion on how to improve https://pineko.readthedocs.io/en/latest/theory/fonll.html ?

nnpdf_data/nnpdf_data/commondata/NNPDF_POS_3P16GEV/kinematics_F2U.yaml
RoyStegeman: In your own paper the recommended scale is 5GeV2

felixhekhorn: yes, yes but look at the title of the PR

comane: > ?

Yes thanks, I had missed the kinematics, pinecard should be fine though

felixhekhorn:
> pinecard should be fine though

Let's link the two PRs for easier navigation https://github.com/NNPDF/pinecards/pull/183

scarlehoff:
> for you as a PDF fitter it may look like this - but as soon as you undo the tiny final step (for getting this 1 FK table) the answer is: no, there are 7 FK tables, which are plainly summed over. And we need these 7 FK tables and they are (more or less) fully reasonable FK tables.

But this *is* the PDF fitting repository.

> this is doable by not committing them here (so you don't see them) - but locally we need them.

They don't need to be committed anywhere if they can be always regenerated 

> We can leave FONLL to pineko (i.e. NNPDF does not need to care),

This is already the case.

> , i.e. we will not implement the theory expansion procedure (1 -> 7) into pinefarm: those two communicate by the already established interface: theory cards.

The theory cards have never been truly compatible between the two however. 

But fine, no dependency and no code shared between pinefarm and pineko.
Then, imho, we should just add the FONLL part to pinefarm "How to do FONLL grid, go to pineko, create the theories and copy them here, done".

This doesn't create any dependency, it just tells people how to do things in the way that everybody has been doing up to now.

felixhekhorn:
> They don't need to be committed anywhere if they can be always regenerated

I believe we all agree on that, so that - so that should close the discussion

>  add the FONLL part to pinefarm

https://github.com/NNPDF/pinefarm/issues/96 - @comane could you please take care?

nnpdf_data/nnpdf_data/theory_cards/4000000000.yaml
felixhekhorn, line 1: Also this one should not be there: the "base theory" is 40_000_000 - instead 40_000_000_00 is also a generated file (in fact you are just making new (pseudo) observables - there is no need for a new theory)

comane:
@felixhekhorn , @scarlehoff can this be merged?

scarlehoff (approved):
Can you also link here the report you showed in Amsterdam?

nnpdf_data/nnpdf_data/commondata/NNPDF_POS_3P16GEV/metadata.yaml
```diff
@@ -0,0 +1,83 @@
+setname: NNPDF_POS_3P16GEV
+version: 1
+version_comment: POS at Q2 = 10 GeV^2
+nnpdf_metadata:
+  nnpdf31_process: POSITIVITY
+  experiment: NNPDF
+arXiv:
+  url: https://arxiv.org/abs/
```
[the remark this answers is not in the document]
felixhekhorn: I would still do that - or is there a reason to not?

nnpdf_data/nnpdf_data/commondata/NNPDF_POS_3P16GEV/metadata.yaml
```diff
@@ -0,0 +1,245 @@
...
+  observable:
+    description: Deep Inelastic Scattering
+    label: 'positivity dataset: $xu$ PDF $Q^2=10$ GeV$^2$'
+    units: ''
+  process_type: POS_XPDF
+  tables: []
+  npoints: []
+  ndata: 20
+  plotting:
+    theory_reference: Bertone:2013vaa
```
felixhekhorn, line 25: actually why do we have such a key? and can we please change it to
```bibtex
@article{Candido:2024rkr,
    author = "Candido, Alessandro and Hekhorn, Felix and Magni, Giacomo and Rabemananjara, Tanjona R. and Stegeman, Roy",
    title = "{Yadism: yet another deep-inelastic scattering module}",
    eprint = "2401.15187",
    archivePrefix = "arXiv",
    primaryClass = "hep-ph",
    reportNumber = "CERN-TH-2024-015, Edinburgh 2024/4, Nikhef-2024-002",
    doi = "10.1140/epjc/s10052-024-12972-7",
    journal = "Eur. Phys. J. C",
    volume = "84",
    number = "7",
    pages = "698",
    year = "2024"
}
```

felixhekhorn: > they can be removed

I think that would be better - because data can not know which "theory" was used (nor which generator nor which settings etc.)

comane: So remove`theory_reference` key within plotting?

scarlehoff: Yes. If it is fine I'll add a commit removing it from everywhere. It's only left around because of the ported data.

PR #2290, opened 2025-03-03
Implement double gauge boson production (2VBP) datasets

t7phy:


scarlehoff:
I think it makes sense to separate `WW` and `WZ` as two different processes.

PR #2296, opened 2025-03-07
Update docs, citations

scarlehoff:
I've also recovered the autodeploy for the documentation since we are still updating it.

The references look like this now

Screenshot 2025-03-07 at 14 19 56

I've also moved the tutorials to the top, since I think it is the first thing people actually look at.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -41,3 +41,15 @@ jobs:
       shell: bash -l {0}
       run: |
         conda build -q conda-recipe --package-format=1
+    - name: Build and upload sphinx documentation to NNPDF server
+      if: startsWith(matrix.os, 'ubuntu') && github.ref == 'refs/heads/master'
+      shell: bash -l {0}
+      run: |
+        KEY=$( mktemp )
+        echo "$NNPDF_SSH_KEY" | base64 --decode > "$KEY"
+        conda install nnpdf --yes
+        cd doc/sphinx
+        make html
+        scp -r -i "$KEY" -o StrictHostKeyChecking=no\
+          build/html/* \
+          dummy@packages.nnpdf.science:~/sphinx-docs/
--- a/doc/sphinx/source/releases.rst
+++ b/doc/sphinx/source/releases.rst
@@ -20,9 +20,12 @@ developments and to mark versions used to produce main results. The
 significant releases since the code was made public are:
 
 `Version 4.0.10 <https://github.com/NNPDF/nnpdf/releases/tag/4.0.10>`_
-    Last release fully compatible with the 4.0 era runcards.
-    The new commondata and theory formats have been fully adopted.
+    Last release of the 4.0 era. Runcards for 4.0 work in this release will work
+    but are no longer used the old tools, commondata or theories.
+    The new commondata and theory pipeline has been fully adopted.
     EKO is now the only evolution code supported. C++ and cmake have been removed.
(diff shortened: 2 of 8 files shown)


doc/sphinx/source/releases.rst
```diff
@@ -19,11 +19,15 @@ The code is tagged to contextualize the versioning, mark significant
 developments and to mark versions used to produce main results. The
 significant releases since the code was made public are:
 
+`Version 4.0.10 <https://github.com/NNPDF/nnpdf/releases/tag/4.0.10>`_
+    Last release fully compatible with the 4.0 era runcards.
```
felixhekhorn, line 23: I don't know if you want to change the 4.0.9 description concerning "Last" - or we could do "Final-Final" :innocent:

scarlehoff: The 4.0.9 is compatible with the entire pipeline, 4.0.10 only with the runcards. As in, the runcards work but now they use eko instead of apfel, etc.

Clearly it was not clear, I'll reword it a bit.

.github/workflows/tests.yml
```diff
@@ -41,3 +41,15 @@ jobs:
       shell: bash -l {0}
       run: |
         conda build -q conda-recipe --package-format=1
+    - name: Build and upload sphinx documentation to NNPDF server
```
felixhekhorn, line 44: Since we are moving towards a tagged release schedule, we should also provide tagged documentation.

On RTD this is almost trivial as I just found out: https://docs.readthedocs.com/platform/stable/guides/automation-rules.html . And to be fair we haven't done that for, e.g., eko so far - but now we do: https://eko.readthedocs.io/en/v0.14.6/code/Operators.html . Now this version (v0.14.6) I added manually as I think the rules are only from now on. Until now for eko it was only possible to see either the version of `master` or `v0.8` from 3 years ago, which is clearly not ideal - since we know master is not stable and v0.8 is ancient ... also it turned out that I could not add the v0.13.5 version since in that commit RTD is broken :melting_face:

scarlehoff: Yes, I was thinking doing so, but we are not using RTD but our server because we are too cool for school...

It should be easy to do in the server just the same of course, but right now all tagged docs would be outdated anyway so better to keep updating them until we catch up :__)

RoyStegeman:
I hadn't even realised the documentation was not being auto-deployed but I see it was only 5 days ago https://github.com/NNPDF/nnpdf/pull/2293

doc/sphinx/source/releases.rst
```diff
@@ -19,11 +19,19 @@ The code is tagged to contextualize the versioning, mark significant
 developments and to mark versions used to produce main results. The
 significant releases since the code was made public are:
 
+`Version 4.0.10 <https://github.com/NNPDF/nnpdf/releases/tag/4.0.10>`_
+    Last release of the 4.0 era. Runcards for 4.0 work in this release will work
+    but are no longer used the old tools, commondata or theories.
```
RoyStegeman, line 24: Runcards for 4.0 work in this release will work but are no longer used the old tools, commondata or theories.

This is not an proper sentence

RoyStegeman (approved):

doc/sphinx/source/releases.rst
scarlehoff, line 24:
```suggestion
    Last release of the 4.0 era. Runcards for 4.0 will still run using new tools, data and theory.
```

PR #2298, opened 2025-03-10
Merge all pytest workflows

scarlehoff:
This should be more gentle in the email clients when failures happen (which usually happen for all tests anyway).

The commondata tests are separated because since they are mostly independent from the rest.


.github/actions/prepare_environment/action.yml
```diff
@@ -0,0 +1,24 @@
+name: Prepare environment
```
RoyStegeman, line 1:
```suggestion
name: Prepare NNPDF conda environment
```

RoyStegeman (approved):
I would appreciate it a lot if (many) more inline comments were placed to explain why keys are given the values that they have. While 
```
        run: |
          pytest extra_tests/regression_checks.py
  ```
  
  is pretty straightforward, I can only guess what `using: "composite"` does

PR #2301, opened 2025-03-17
Diagonal covmat

jacoterh:
This PR performs the training validation split in the diagonal basis of the t0 covariance matrix.

Without any split, so pure training, the loss must be the same regardless of whether one diagonalises or not. Indeed, at float64 precision, with this [test runcard](https://github.com/user-attachments/files/19346423/diagonal_covmat_test.txt) I find,

```
float64, non diagonal

Epoch 100/17000: loss: 5.3918877
Epoch 200/17000: loss: 5.3818284
Epoch 300/17000: loss: 5.3818996
Epoch 400/17000: loss: 5.0928206
Epoch 500/17000: loss: 5.2855544
Epoch 600/17000: loss: 5.0460824
Epoch 700/17000: loss: 5.1262326
Epoch 800/17000: loss: 5.0686013
Epoch 900/17000: loss: 4.9429147
Epoch 1000/17000: loss: 5.1852478
Epoch 2000/17000: loss: 4.7091641
Epoch 4000/17000: loss: 4.3651678


float64, diagonal

Epoch 100/17000: loss: 5.3918877
Epoch 200/17000: loss: 5.3818284
Epoch 300/17000: loss: 5.3818996
Epoch 400/17000: loss: 5.0928206
Epoch 500/17000: loss: 5.2850301
Epoch 600/17000: loss: 5.4309732
Epoch 700/17000: loss: 5.1401602
Epoch 800/17000: loss: 5.1322600
Epoch 900/17000: loss: 5.1281818
Epoch 1000/17000: loss: 5.1255253
Epoch 1500/17000: loss: 4.9968909
Epoch 2000/17000: loss: 4.5648582
Epoch 2500/17000: loss: 4.4773234
Epoch 3000/17000: loss: 4.4334937
Epoch 4000/17000: loss: 4.3631501
Epoch 5000/17000: loss: 4.1540497
```

Up to the precision printed here, the losses remain in sync up to around epoch 500, while after that they start showing some small differences. But this we can explain: a tiny deviation in the loss due to precision in the beginning also leads to another update of the NNs and hence a different path through the loss landscape. What matters is that after training is done, the difference at the pdf level is of the same order as changing the seed of the NNs. 

Indeed, in a DIS only fit at float64 precision, I find https://vp.nnpdf.science/UobKaEuXQf-BwMeMN_Z2Mw==/ where the distance plots show that the difference between non-diagonal and diagonal are well within one sigma.

The diagonal basis can be activated in the runcard by adding `diagonal_basis: True` at the top level of the yaml. With this, the implementation is ready, but we should perform the following checks with a non-zero training/validation split:
- [x]  using closure pseudodata where we remove any possible hidden correlations: https://vp.nnpdf.science/xBMmCMHjQ5yB5_l0uVK_Aw==/  
- [x] Non-diagonal basis reproduces master: https://vp.nnpdf.science/obaVuxjRSv-4fyLpg80jhw==/
- [x] Using real data: https://vp.nnpdf.science/fTfKodChTtKShpHWpiRdCg==/
- [ ]  using inconsistent closure pseudodata


scarlehoff:
I wonder where the difference is coming from?

How are the cuts applied differently (the training/validation cuts?) what would happen if you do the closure fit without tr/vl split like in your test? 

Perhaps we should do this in two steps: 1) First apply tr/vl cuts, validate that resutls are the same 2) Then change cuts from databaset-by-dataset to full-dataset
Alternatively do a 1000-replica fit where hopefully the difference due to how we cut away the data is washed away... I guess doing the cut in a dataset-per-dataset basis has some merit as it ensures that for every replica for every process there's a contribution to both training and validation.

scarlehoff:
> could you be more specific?

At the beginning you did a test with trvl: 1.0 and you found small numerical differences. My suggestion to check whether these grow enough to be noticeable are:

- A fit with trvl: 1.0 (there diagonal or not, the result should be equivalent)
- A fit in which trvl: 0.75 but you first apply the cuts, then diagonalize training and validation (perhaps with genrep = false to make it easier).

These two tests will ensure that the differences that we see in your report are coming from the cuts (entire dataset vs dataset-by-dataset) and then we'll need to think about the implications of that*


*e.g., in the diagonal case it doesn't make sense to have frac-per-dataset or to concatenate the masks

jacoterh:
> * A fit with trvl: 1.0 (diagonal or not, the result should be equivalent)

Results are equivalent, as you can see in this DIS only fit without any split: https://vp.nnpdf.science/UobKaEuXQf-BwMeMN_Z2Mw==/

jacoterh:
> * A fit in which trvl: 0.75 but you first apply the cuts, then diagonalize training and validation (perhaps with genrep = false to make it easier).

That agrees with what I've been doing so far. I apply the training/validation split at the individual dataset level before rotating to the diagonal basis. I am not sure it matters when we apply the mask. We can either concatenate all the individual masks or compute one big mask only at the end. As long as datapoints get randomly assigned and the split is 75/25 in both, these should be equivalent.

scarlehoff:
I thought you were applying the mask _after_ diagonalizing since you apply it to `eig_vals`. If so the data that is cut away is not the same and it will actually be a quite different set (since the order gets all messed up).

>  I am not sure it matters when we apply the mask. We can either concatenate all the individual masks or compute one big mask at the very end after combining all datasets. As long as datapoints get randomly assigned and the split is 75/25 in both, these should be equivalent, no?

The results might be equivalent in the limit of many replicas. Also, by concatenating the individual masks you are not covering some situations (e.g. you will never get the top-25% of eigenvalues being masked) so it is not completely random. 

Btw, I don't know how deterministic `eigh` is either, which means we should probably order them in order to be able to reproduce fits.

jacoterh:
Very good, I see. Just to clarify that I am _applying_ the mask after rotation, I just ment that I _construct_ the mask before the rotation. So indeed, I am applying it to the eigenvalues.

jacoterh:
Hi @scarlehoff , I have performed the check you proposed. That is, comparing
1. First splitting, then diagonalise training and validation separately
2. Same split, don't diagonalise 

I put the relevant code I needed to perform this check on a separate branch as it would interfere with what we actually want (diagonalising before splitting) https://github.com/NNPDF/nnpdf/tree/diagonal_covmat_exc_juan

The loss values agree between 1 and 2 (up to precession effects)
[log_splitting_diag_trvl_sep.txt](https://github.com/user-attachments/files/19688139/log_splitting_diag_trvl_sep.txt)

scarlehoff:
Nice, thanks!

Then the next step I'd say is to do a large fit (many replicas) to see how the choice of diagonal vs diagonal might affect the results.

Btw, perhaps instead of doing the splitting in the dictionary creation it makes sense to do it in the function `tr_masks` above. If the diagonal is on instead of going through the loop of datasets you exit early with what you are doing now but outputing the same kind of object. This *might* help with some of the errors.

Some others are related to the fact that we try to write training and validation data down during the fit and I'm not sure whether we want to do that in the diagonal basis...

jacoterh:
Just for the record and to help us to keep track, these are the reports with the negative modes removed in the diagonal basis:
- real data: https://vp.nnpdf.science/V-SA51LcS_Kx7wEx8p6vFg==
- fake data: https://vp.nnpdf.science/rceHcI8MTdaWpZlKkb05vw==

The split is done at the level of the full covariance matrix, not per dataset separately.

scarlehoff:
> The split and change of basis is done at the level of the full covariance matrix, not per dataset separately.

This I don't understand. I had a look yesterday after the meeting and indeed the function in which the diagonalization occurs is called experiment by experiment (not by dataset, perhaps I said dataset when I meant experiment, apologies).

In any case, all the data in the runcard is only taken together when MHOUs are enabled. So it would be interesting to see whether the issue we discussed actually disappears in that case.

jacoterh:
Indeed, I was also just checking this and you're right, it's called by experiment! This wasn't obvious to me before as I had tested it only on datasets within the same experiment and I noticed all datasets got combined.

jacoterh:
@scarlehoff In case of MHOU, do you have an idea where in validphys the various experiments get collected such that they are treated as one? I'm seeing it works this way, but I cannot find what part in the code is responsible for this.

jacoterh:
It seems the grouping case of MHOU is handled by this statement: https://github.com/NNPDF/nnpdf/blob/3d10bbc0a6ba93d33d7951a573fa6bbb8d91456a/validphys2/src/validphys/config.py#L1654

jacoterh:
Hi @scarlehoff, I am working on the tests, and I'm seeing that `test_pseudodata.py` results in 

```
FAILED validphys2/src/validphys/tests/test_pseudodata.py::test_recreate_fit_pseudodata - AttributeError: 'NoneType' object has no attribute 'is_polarized'
FAILED validphys2/src/validphys/tests/test_pseudodata.py::test_recreate_pdf_pseudodata - AttributeError: 'NoneType' object has no attribute 'is_polarized'
FAILED validphys2/src/validphys/tests/test_pseudodata.py::test_read_matches_recreate - AttributeError: 'NoneType' object has no attribute 'is_polarized'
```

This is triggered by `coredata.py` in `determine_pdfs` where `pdf` is none such that `pdf.is_polarized` returns the `AttributeError ` error above. Do you have any idea how this could be related to the implementation of the diagonal basis, I don't see how. Thanks!

jacoterh:
When I deliberately remove `dataset_inputs_t0_predictions` validphys tells me it's needed to process 


```
E  reportengine.configparser.InputNotFoundError: A parameter is required: dataset_inputs_t0_predictions.
E           This is needed to process:
E            - recreate_fit_pseudodata
E           through:
E            - fit_masks
E           through:
E            - replica_mask_table
E           through:
E            - replica_mask
E           through:
E            - exps_masks
E           through:
E            - masks
E           through:
E            - dataset_inputs_fitting_covmat
```

jacoterh:
Hi @scarlehoff , regarding the last failing test, is it enough to rerun the fit bot to update the jsons in `extra_tests/regression_fits` to fix this? At the moment, I am getting 

```
FAILED extra_tests/regression_checks.py::test_regression_fit[False-normal_fit-72] - AssertionError: 
FAILED extra_tests/regression_checks.py::test_regression_fit[False-diagonal-45] - AssertionError: 
FAILED extra_tests/regression_checks.py::test_regression_fit[False-no_lagrange-27] - AssertionError: 
FAILED extra_tests/regression_checks.py::test_regression_fit[False-polarized_evol-34] - AssertionError: 
FAILED extra_tests/regression_checks.py::test_regression_fit[False-t0theoryid-100] - AssertionError: 
```

scarlehoff:
Yes. I think it is only the diagonal one failing in the CI, all the others fail because of differences between your computer and the CI.

scarlehoff:
Oh, the t0 sampling. I think it would be better if you skip the last commit by the bot and rebase [475bbe7](https://github.com/NNPDF/nnpdf/pull/2301/commits/475bbe73c4a9a3c4f246b6d04ab4abcf37ef9572) on top of master, then run the bot.

validphys2/src/validphys/pseudodata.py
```diff
@@ -421,20 +420,19 @@ def make_level1_data(data, level0_commondata_wc, filterseed, data_index, sep_mul
     return level1_commondata_instances_wc
 
 
-_group_recreate_pseudodata = collect(
-    'indexed_make_replica', ('group_dataset_inputs_by_experiment',)
-)
+# TODO: by by_metadata or by_experiment?
```
scarlehoff, line 423: What does this todo mean?

validphys2/src/validphys/results.py
```diff
@@ -242,7 +242,7 @@ def data_index(data):
 procs_data = collect("data", ("group_dataset_inputs_by_process",))
 
 
-def groups_index(groups_data):
+def groups_index(groups_data, diagonal_basis=False):
```
scarlehoff, line 245: Please add the effect of diagonal basis to the docstr.

Also, make sure to run pre-commit for the modified files.

validphys2/src/validphys/n3fit_data.py
```diff
@@ -102,23 +102,68 @@ def replica_luxseed(replica, luxseed):
...
     ``group_name`` and ``seed`` define the ``masks``.
+
+    In case of the diagonal basis, the ``rotation`` is a tuple of the eigenvalues and
+    eigenvectors of the fitting covariance matrix.
     """
 
-    def __init__(self, group_name, seed, masks=None):
-        self.masks = masks
+    def __init__(
+        self, group_name, seed, tr_masks, vl_masks, diagonal_basis=False, eig_vals=None, u=None
```
scarlehoff, line 117: I think u is not self-explanatory enough, I think it should be added to the docstr or perhaps call it diagonal_rotation or something here in the class (then later on u = mask.diagonal rotation is self explanatory)

validphys2/src/validphys/n3fit_data.py
```diff
@@ -102,23 +102,68 @@ def replica_luxseed(replica, luxseed):
...
         super().__init__(group_name, seed)
 
-    def __iter__(self):
-        yield from self.masks
 
+def _diagonal_masks(
+    data,
+    replica_trvlseed,
+    dataset_inputs_fitting_covmat,
+    diagonal_basis=False,
```
scarlehoff, line 133: Does diagonal basis have an effect here? (at the end of the call it is set to true) I guess this is a remnant of the previous version?

validphys2/src/validphys/n3fit_data.py
```diff
@@ -347,19 +394,40 @@ def fitting_data_dict(
...
+        )
+        invcovmat_vl = np.einsum(
+            "i, ij, j -> ij",
+            diag_inv_sqrt_covmat_vl,
+            np.linalg.inv(cormat_vl),
+            diag_inv_sqrt_covmat_vl,
+        )
+
+        # invcovmat_tr = np.linalg.inv(covmat_tr)
+        # invcovmat_vl = np.linalg.inv(covmat_vl)
```
scarlehoff, line 419:
```suggestion
```

validphys2/src/validphys/n3fit_data.py
```diff
@@ -425,6 +493,7 @@ def replica_nnseed_fitting_data_dict(replica, exps_fitting_data_dict, replica_nn
 groups_replicas_indexed_make_replica = collect(
     "indexed_make_replica", ("replicas", "group_dataset_inputs_by_experiment")
 )
+# TODO: by_experiment or by_metadata?
```
scarlehoff, line 496:
```suggestion
```

I guess this todo doesn't apply anymore?

validphys2/src/validphys/n3fit_data.py
```diff
@@ -394,14 +463,13 @@ def fitting_data_dict(
...
-        "data_transformation_vl": dt_trans_vl,
-        "data_transformation": dt_trans,
+        "data_transformation": u if diagonal_basis else None,
     }
     return dict_out
 
 
-exps_fitting_data_dict = collect("fitting_data_dict", ("group_dataset_inputs_by_metadata",))
+# TODO: by_experiment or by_metadata?
+exps_fitting_data_dict = collect("fitting_data_dict", ("group_dataset_inputs_by_experiment",))
```
jacoterh, line 472: @scarlehoff The TODOs here are a reminder to myself to take a look at this. When we do `group_dataset_inputs_by_metadata` the correlation matrix that is diagonalised is the total cormat collected over all experiments, while `group_dataset_inputs_by_experiment` results in a separate correlation matrix for each experiment. This line needs to be conditioned on `diagonal_basis` , which is why I left this comment.

jacoterh: I convinced myself that one can always use `group_dataset_inputs_by_metadata` as it defaults to `experiment`, while it collects all experiments when the diagonal_basis is turned on

validphys2/src/validphys/results.py
jacoterh: I have added some documentation to the doctoring. Which modified files do you mean, the ones in this PR? Because I did already run the pre-commit on those if I'm not mistaken.

scarlehoff: Yes, just these. Thanks!

scarlehoff (approved):
I know there were a few more checks requested during the PC, let's wait for those before merging but (expecting that nothing will break there) I think this is good to go.

jacoterh:
Thanks @scarlehoff , what's going on with the tests? I had fixed them all a couple of days ago (up to the regression_tests), but now it seems like new ones have appeared...

scarlehoff:
Not sure, but in the last few changelogs from numpy they seem to be working on the thread-safety of some functions so that can change the order of operations. And diagonalizing a huge not-super-well-behaved matrix is very sensitive to  that (for the data the change happened for CMS 1 Jet).

jacoterh:
For reference, links to reports
- comparing two fits with and without diagonal basis in the presence of MHOUs and t0 sampling: https://vp.nnpdf.science/1a6KBrruQdyDUTw_lCyaRQ==/
- comparing the no-diagonal basis implementation on master and the current branch: https://vp.nnpdf.science/3CKAWh5LQP2eI0dJVnJAqQ==/. Difference is probably due to the new way we take the inverse of the covariance, also in the non-diagonal basis - we take the inverse of the correlation and convert this to the inverse covariance later.  Before this change we got perfect agreement, see: https://vp.nnpdf.science/obaVuxjRSv-4fyLpg80jhw==/. To be checked again.

Radonirinaunimi:
Could we merge this please and then address the other points in another PR? We are relying on this with the hyperopt PR and it is getting a bit of a pain to merge locally while at the same time having changes.

PR #2302, opened 2025-03-24
some things used in alphas studies

RoyStegeman:
- don't iterate filterseed (why should we? The experimental data doesn't change either)
- add `reuse_luxset_replica`. Instead of only using each replica of the input PDF set in the computation of the photon PDF once, we can reuse them 
- removes the error raised when trying to sample a theory covmat in a closure test
- adds some alphas covmat point prescriptions


scarlehoff:
> don't iterate filterseed (why should we? The experimental data doesn't change either)

Why shouldn't we iterate it? 

> add reuse_luxset_replica. Instead of only using each replica of the input PDF set in the computation of the photon PDF once, we can reuse them

What does this mean? If it is "use always central replica" then it should be `use_always_central_replica`

RoyStegeman:
> Why shouldn't we iterate it?

because it generates the L1 data and we also don't iterate the experimental sampling in a real data fit 

> What does this mean? If it is "use always central replica" then it should be use_always_central_replica

It means taking MOD(nrep) instead of nrep

scarlehoff:
> because it generates the L1 data and we also don't iterate the experimental sampling in a real data fit

`mcseed` is iterated in a real data fit, don't we have a seed for L2 (which I guess would be te equivalent)?

> It means taking inputrep = nrep MOD totrep instead of nrep

Ah, I see. I wonder whether this should actually be the behaviour always. This is used just to generate the errors and if we have less members in the original photon so be it...?

RoyStegeman:
> mcseed is iterated in a real data fit, don't we have a seed for L2 (which I guess would be te equivalent)?

I think that's still `mcseed`

> Ah, I see. I wonder whether this should actually be the behaviour always. This is used just to generate the errors and if we have less members in the original photon so be it...?

Given the small photon uncertainty I don't think it all matters very much. Practically I definitely think this is the sensible thing to do

RoyStegeman:
> Everyone in the collaboration who has run a QED fit is in this conversation anyway.

🥲

scarlehoff (approved):

PR #2303, opened 2025-03-25
Stop using old theories or dataset names in examples or tests

### LINKED ISSUE #2151: Update tests and examples so that they are no longer using "low precision theories"
Before 4.0.9 theories were very big and fitting or loading them was very slow, and so "low precision" theories were used for the tests.

This is no longer needed, all tests should use standard theories, i.e., all NNLO tests should use `40_000_000`, all NLO and QED tests the corresponding versions etc.

This task is not particularly complicated and will make the tests much easier to maintain. In addition, no old theory should be used any longer.

1. Create a version of all fits used in the tests (https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/tests/conftest.py#L75) with the corresponding theory (most of them 40_000_000)
2. Remove `THEORYID=162` and `THEORYID_NEW=399` from that same file. Update all tests to use `THEORYID=40_000_000`
3. Run the test and fix any remaining problems  (e.g., the regression tests will surely need to be updated, cfactors might be different, etc)
4. Do the same with the examples in the validphys test folder https://github.com/NNPDF/nnpdf/tree/master/validphys2/examples make sure they all still run!
5. Update the documentation to no longer talk about using low precision theories.

scarlehoff:
Closes #2151

Before merging I'll redo every test fit and check that every example runs, but first I'd like #2300 to be merged.


scarlehoff:
I believe I updated all tests and examples. Now we are using new datasets (and without `legacy` whenever possible) and the theories 40_000_000 and 40_000_100 which are NNLO QCD and +QED respectively.

We are no longer using low-precision theories as the code seems to be efficient enough now, so we are actually testing what it is used. This is not pointing to master but instead to #2275 since the regression tests are updated anyway.

n3fit/src/n3fit/tests/regressions/filter.yml
```diff
@@ -25,10 +25,10 @@ datacuts:
 
 ############################################################
 theory:
-  theoryid: 162        # database id
+  theoryid: 40_000_000     # database id
 
 ############################################################
-genrep: True    # True = generate MC replicas, False = use real data
+genrep: True    # on = generate MC replicas, False = use real data
```
RoyStegeman, line 31:
```suggestion
genrep: True    # True = generate MC replicas, False = use real data
```

nnpdf_data/nnpdf_data/theory_cards/40000001.yaml
RoyStegeman, line 1: This and the other theorycard contains outdated keys

RoyStegeman (approved):
This is a lot to go through, but if the tests pass and the theorycard issue is addressed I'm happy

PR #2304, opened 2025-03-25
Remove all old theories

scarlehoff:
As we discussed. For now I've left the low precision theories and test theories (will remove after #2303) and 200/201/etc.

I've changed also the script that generates the tables in the docs so that now we have two tables, one with all theories, and a summary that prints out only the "000" theories (which are usually the central one).

For this to be useful it requires a bit of hygiene when creating new theories which we can try to enforce during the reviews (e..g, we should avoid writting just 'Same as theory X' which I'm also guilty of).

Screenshot 2025-03-25 at 11 06 37


scarlehoff:
@enocera are you fine with these old theories being removed?

(they will be available online of course, I won't remove them from the server)

enocera:
@scarlehoff You've got my blessing to proceed with the removal of the old theories. Perhaps we should save one entry for 3.1prime which went into PDF4LHC21?

scarlehoff:
I can keep that. Which number is it?

Issue #780, opened 2020-05-15
Build packages from branches


Zaharid:
We should try to have packages from non-master branches that can be trivially used to e.g. run fits.

At the moment we can upload the packages that get built, but those are indistinguishable from nnpdf-master so that is not very good. Instead we should have:

  - Different package names.
  - Packages that are mutually incompatible (you can only have one NNPDF package from any branch installed).

Conda used to make that very complicated but now there is a feature that may be just right:
https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/packages.html#mutex-metapackages

The idea would be that you can just do `conda install nnpdf-flbasis` and run that out of the box.

scarlehoff:
since we can install now directly with pip from the repository, running from a branch in e.g., a cluster is trivial

scarlehoff:
But actually, this is now closed by #2305 :)

PR #2305, opened 2025-03-26
Miscellaneous improvements to docs and tests

scarlehoff:
These are some suggestions that were made by the revieweres from the HSF, to clarify certain points.

Since I've updated the message in `postfit` (which tells you to upload to a server that most people won't have access to), I've updated the "positivity veto" to "convergence veto" to avoid the common misconception regarding positivity.

Since postfit had not been touched in ages, `pre-commit` had a field day with it.

### DIFF
--- a/.github/workflows/all_tests_nnpdf.yml
+++ b/.github/workflows/all_tests_nnpdf.yml
@@ -5,6 +5,9 @@ name: All tests NNPDF
 # the custom actions in .github/actions/
 
 on: [push]
+concurrency:
+  group: ${{ github.head_ref }}-${{ github.ref }}
+  cancel-in-progress: true
 
 env:
   # https://keras.io/getting_started/faq/#how-can-i-obtain-reproducible-results-using-keras-during-development
@@ -79,6 +82,7 @@ jobs:
         if: ${{ !cancelled() }}
         uses: actions/upload-artifact@v4
         with:
+          name: nnpdf-package-${{ github.job }}-${{ strategy.job-index }}.tar.bz2
           path: ${{ env.CONDA_PREFIX }}/conda-bld/noarch/*.tar.bz2
 
   run_pytorch:
--- a/.github/workflows/check_newcd.yml
+++ b/.github/workflows/check_newcd.yml
@@ -1,5 +1,5 @@
 # A CI script with github workflow to test the new commondata
-name: Test new commondata
+name: CommonData
 
 on:
   push:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,3 +22,13 @@ repos:
     hooks:
     -   id: isort
         args: ['--settings-path=./pyproject.toml']
+
+-   repo: https://github.com/pre-commit/pygrep-hooks
+    rev: v1.10.0  # Use the ref you want to point at
+    hooks:
+    -   id: python-check-blanket-noqa
+    -   id: python-check-blanket-type-ignore
+#     -   id: rst-backticks # too many false positive but many true positives as well
+    -   id: rst-directive-colons
+    -   id: rst-inline-touching-normal
+    -   id: text-unicode-replacement-char
--- a/doc/sphinx/source/get-started/installation.rst
+++ b/doc/sphinx/source/get-started/installation.rst
@@ -3,11 +3,16 @@ Installing the code on Linux or macOS
 
 Installing the NNPDF code requires a system with a recent version of python and
 either a Linux-based operating system or macOS.
+At least 4 GB of storage are needed in order to download all data and theory ingredients
+necessary for a fit or most analyses.
+Some specific features such as theory uncertainties or QED fits might require more space.
+
 The code can be installed either with :ref:`conda <condainstall>` or with :ref:`pip <pip>`,
 the former offering a "batteries included" approach where non-python software
 such as LHAPDF and pandoc will be installed automatically.
 We also made available an automatic installation of conda, see :ref:`bootstrap-installation`
 to install conda as well as set up relevant channels.
+
 
 If you plan to contribute to the development of NNPDF, please see :ref:`Source`.
 
@@ -33,6 +38,7 @@ You are now ready to use the NNPDF code! Check out the :ref:`Tutorials`.
 .. note::
 
    Make sure to be using a recent version of `conda <https://docs.anaconda.com/miniconda/install/>`_. These instructions have been tested with conda 25.
+
 
 
 .. _pip:
@@ -84,7 +90,7 @@ a helper script exists to aid the configuration using conda. To obtain it use:
    each time you want to use the environment). On remote machines, the
    addition to .bashrc should read as follows
 
-   ::
+::
 
         if shopt -q login_shell; then
             . <path-to-conda>/etc/profile.d/conda.sh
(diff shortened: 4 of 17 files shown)


.github/workflows/check_newcd.yml
```diff
@@ -1,5 +1,5 @@
 # A CI script with github workflow to test the new commondata
-name: Test new commondata
+name: CommonData
```
comane, line 3:
```suggestion
name: Test CommonData

```

.github/workflows/check_newcd.yml
```diff
@@ -1,5 +1,5 @@
 # A CI script with github workflow to test the new commondata
```
comane, line 1:
```suggestion
# A CI script with github workflow to test the commondata
```

doc/sphinx/source/get-started/installation.rst
```diff
@@ -147,3 +153,36 @@ This will open a bash shell with the ``nnpdf`` environment already activated, wi
...
+-----------------------------
+
+The NNPDF framework would not be possible with a number of dependencies.
+These are automatically installed when using conda and the full list can be consulted in the
+`conda recipe <https://github.com/NNPDF/nnpdf/blob/master/conda-recipe/meta.yaml#L20>`_ available in the repository.
+
+Below we list some of the most relevant external libraries than underpin the different aspects of this framework.
+
+::
+
```
comane, line 170: reportengine

validphys2/src/validphys/fitveto.py
```diff
@@ -75,9 +75,9 @@ def determine_vetoes(
         )
 
     # Positivity veto
-    posmask = np.array([replica.is_positive for replica in fitinfos], dtype=bool)
-    vetoes = {"Positivity": posmask}
-    total_mask = posmask.copy()
```
comane, line 80: Does this mean that no positivity veto is applied anymore? In such case would probably also remove the comment at line 77

scarlehoff: Ups, forgot the comment.

The veto hasn't changed, but since we decided to be strict on the positivity, the fit cannot converge if the positivity is not ok. And equally so, if the fit doesn't converge the positivity is not triggered. So calling it "positivity veto" was a misnomer.

doc/sphinx/source/get-started/installation.rst
scarlehoff: That's not really external so it is also part of the suggested citations. I'm also skipping EKO and pineappl for the same reasons.

validphys2/src/validphys/fitveto.py
```diff
@@ -75,9 +75,9 @@ def determine_vetoes(
         )
 
     # Positivity veto
```
scarlehoff, line 77:
```suggestion
    # Veto on convergence & positivity
```

scarlehoff:
We no longer have to maintain a conda server :) 

https://anaconda.org/conda-forge/nnpdf

I'll update the docs accordingly (and we'll maintain the server for the older versions, but now we don't depend on it anymore)

scarlehoff:
I haven't added it to the doc because I think it might confuse people (since it makes interacting with pip a bit more difficult) but for those who want to enjoy new experiencies, installing with pixi is much faster https://pixi.sh/latest/

scarlehoff:
@comane @RoyStegeman are you fine with merging this or would you like to have a second look?

.github/workflows/all_tests_nnpdf.yml
```diff
@@ -5,6 +5,9 @@ name: All tests NNPDF
 # the custom actions in .github/actions/
 
 on: [push]
+concurrency:
+  group: ${{ github.head_ref }}-${{ github.ref }}
+  cancel-in-progress: true
```
RoyStegeman, line 10:
```suggestion
# Define a group based on on branch name (github.head_ref) and .... (github.ref) 
# use this group to identify which processes to cancel due to cancel-in-progress
concurrency:
  group: ${{ github.head_ref }}-${{ github.ref }}
  cancel-in-progress: true
```

Some explanation of what this does would be usueful. Also, do we need both head_ref and ref. Is one not sufficient?

.github/workflows/all_tests_nnpdf.yml
```diff
@@ -79,6 +82,7 @@ jobs:
         if: ${{ !cancelled() }}
         uses: actions/upload-artifact@v4
         with:
+          name: nnpdf-package-${{ github.job }}-${{ strategy.job-index }}.tar.bz2
```
RoyStegeman, line 85: Why this and what is the purpose of keeping it as an artefact?

.pre-commit-config.yaml
```diff
@@ -22,3 +22,13 @@ repos:
     hooks:
     -   id: isort
         args: ['--settings-path=./pyproject.toml']
+
+-   repo: https://github.com/pre-commit/pygrep-hooks
+    rev: v1.10.0  # Use the ref you want to point at
```
RoyStegeman, line 27:
```suggestion
    rev: v1.10.0
```

.pre-commit-config.yaml
```diff
@@ -22,3 +22,13 @@ repos:
...
+    hooks:
+    -   id: python-check-blanket-noqa
+    -   id: python-check-blanket-type-ignore
+#     -   id: rst-backticks # too many false positive but many true positives as well
```
RoyStegeman, line 31:
```suggestion
```

pyproject.toml
```diff
@@ -165,3 +168,18 @@ cipaths = [ # In order to combine files from both the ubuntu and macos runners
     "/Users/runner/work/nnpdf/nnpdf",
     "/home/runner/work/nnpdf/nnpdf"
 ]
+
+[tool.repo-review]
+select = ["PY", "GH", "PP", "PC"]
+
+[tool.repo-review.ignore]
+"PY004" = "Using doc/ as folder"
```
RoyStegeman, line 176: Can we not just change the folder name?

RoyStegeman (approved):

.github/workflows/all_tests_nnpdf.yml
scarlehoff: Just one might either resolve to `master` or be empty https://gist.github.com/toshimaru/f14386ab69394e6ffb54aa799b9b124b#file-github-actions-ref-vs-head_ref-md 

but I haven't studied it in more detail, maybe there's a key for "only the workflows in this branch in this PR"

pyproject.toml
scarlehoff: We should but didn't want to deal with the possible conflicts with active PRs for just an `s`.

.github/workflows/all_tests_nnpdf.yml
scarlehoff: You can download exactly the conda package for a given branch and use it locally.

And closes https://github.com/NNPDF/nnpdf/issues/780 :)

RoyStegeman: So where is this uploaded, I thought env.CONDA_PREFIX would just point to the conda folder in the CI. An inline comment giving some details would be good

scarlehoff: the path is of the object to be uploaded, which then remains with the workflow...

but the environment is not being populated... https://github.com/NNPDF/nnpdf/actions/runs/14476915646/job/40604671855?pr=2305#step:5:15
I tested it giving it the exact path before, but within this action the environment doesn't activate it seems.

I'll fix that then merge.

PR #2308, opened 2025-04-04
Introduce frag scale to theory card

t7phy:
@felixhekhorn what do we want to do about ModSV, in particular, do we introduce a new one for frag scale or do we make the current one work for fac and frag scales?

### DIFF
--- a/nnpdf_data/nnpdf_data/theory.py
+++ b/nnpdf_data/nnpdf_data/theory.py
@@ -1,5 +1,5 @@
 """
-This module provides an unique source and definition for all the possible parameters
+This module provides a unique source and definition for all the possible parameters
 that a theory card can contain.
 """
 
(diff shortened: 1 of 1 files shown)


nnpdf_data/nnpdf_data/theory.py
```diff
@@ -54,6 +54,7 @@ class TheoryCard:
     alphas: float  # Value of alpha_s at the scale Qref
     alphaqed: float  # Values of alpha QED at the scale Qref
     Qref: float  # [GeV] Reference scale for alphas and alphaqed
+    XIA: float = None  # Fragmentation scale over the hard scattering scale ratio
```
felixhekhorn, line 57:
```suggestion
    XIA: float = 1.0  # Fragmentation scale over the hard scattering scale ratio
```
?

felixhekhorn:
> @felixhekhorn what do we want to do about ModSV, in particular, do we introduce a new one for frag scale or do we make the current one work for fac and frag scales?

ModSV is providing different ways to estimate MHOU, so it does not matter that much, but to be consistent I would just use one setting - i.e. we should reuse it.

scarlehoff (approved):

PR #2311, opened 2025-04-16
Hotfix for macos 15.4 , rsync -> openrsync

scarlehoff:
I think the problem with the server is due to this recent change https://derflounder.wordpress.com/2025/04/06/rsync-replaced-with-openrsync-on-macos-sequoia/

Server-side I'm manually keeping the permissions of the parent folders, but probably uploading reports in new versions of mac with old versions of the code will lead to issues.

@Radonirinaunimi I've updated my mac in order to test this, I've put you as a reviewer since I guess you are the most likely to have an updated mac? But if not, testing it works in the _previous_ version would also be useful.

### DIFF
--- a/validphys2/src/validphys/uploadutils.py
+++ b/validphys2/src/validphys/uploadutils.py
@@ -60,6 +60,7 @@ class Uploader:
 
     upload_host = _profile_key('upload_host')
     _profile = Loader().nnprofile
+    _macos_compatibility = False
 
     def get_relative_path(self, output_path):
         """Return the relative path to the ``target_dir``."""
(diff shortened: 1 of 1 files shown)


scarlehoff:
> In principle, the user could also choose between the rsync and openrsync implementation by setting the CHOSEN_RSYNC environment, and therefore no changes are needed here. Or wouldn't that work/be an option?

Not anymore. They both changed the defaullt _and_ removed the previous version in one go :__

However, even if they recover that option in a point update, it requires people to change their environments (and if they don't, the server is broken for everyone) so I think it is safer doing it here.
If someone was hiding secret information in their validphys reports (this makes them readable for everyone) that was a bad idea anyway.

Radonirinaunimi:
Ok, all good then!

PR #2313, opened 2025-04-17
Make parallel_models=true the default

scarlehoff:
The only reason it wasn't at this point was that the pseudodata was not written down. This is not a concern anymore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,9 @@ compile_commands.json
 CTestTestfile.cmake
 build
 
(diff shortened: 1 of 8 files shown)


scarlehoff:
Since I don't think anyone is doing multireplicas without parallelizing anymore (and there's no longer a reason) I'll merge this.

Radonirinaunimi (approved):
Sorry for having taken so long to look into this.

PR #2314, opened 2025-04-24
Multiarchitecture models

scarlehoff:
This PR will implement the multiarchitecture parlallel training. The way I'm planning to do it is by first refactoring everything that happens below `generate_pdf_model`, which is the only function that gets called from the outside, without changing anything from the interface (other than perhaps making it more strict on the input). 

The refactoring in this case means separating very clearly the options that can be modified in a per-replica basis and the options that must be generic to the entire model (e.g., the number of layer of each replica can be different, but the output shape must be the same).

I will try my best to do things such that it also prepares the future implementation of FF and nuclear PDFs.

After that I will add a very minimal and proof-of-concept sampling upon which we will build the one for 4.1

I wanted to open the draft early so that people can pitch in, in case you have some idea already about how to structure this. @RoyStegeman @Radonirinaunimi @tgiani

### DIFF
--- a/n3fit/src/n3fit/model_gen.py
+++ b/n3fit/src/n3fit/model_gen.py
@@ -324,7 +324,7 @@ def observable_generator(
 
 
 @dataclass
-class _ReplicaSettings:
+class ReplicaSettings:
     """Dataclass which hold all necessary replica-dependent information of a PDF.
 
     Parameters
@@ -351,8 +351,8 @@ class _ReplicaSettings:
     seed: int
     nodes: list[int]
     activations: list[str]
-    architecture: str
-    initializer: str
+    architecture: str = "dense"
+    initializer: str = "glorot_normal"
     dropout_rate: float = 0.0
     regularizer: str = None
     regularizer_args: dict = field(default_factory=dict)
@@ -379,24 +379,19 @@ class _ReplicaSettings:
             )
 
 
-# TODO:
-#   1. Decide whether the sampling occurs in this function or before entering here
-#      at the moment, it occurs here.
+def sample_pdf_models():
+    """
+    Given a set of parameters, sample a set of PDF models
+    """
+
+
 def generate_pdf_model(
-    nodes: list[int] = None,
-    activations: list[str] = None,
-    initializer_name: str = "glorot_normal",
-    layer_type: str = "dense",
+    replicas_settings: list[ReplicaSettings],
     flav_info: dict = None,
     fitbasis: str = "NN31IC",
     out: int = 14,
-    seed_list: list[int] = None,
-    dropout: float = 0.0,
-    regularizer: str = None,
-    regularizer_args: dict = None,
     impose_sumrule: str = None,
     scaler: Callable = None,
-    num_replicas: int = 1,
     photons: Photon = None,
 ):
     """
@@ -439,31 +434,33 @@ def generate_pdf_model(
 
     Parameters:
     -----------
-        nodes: list(int)
-            list of the number of nodes per layer of the PDF NN
-        activation: list
-            list of activation functions to apply to each layer
-        initializer_name: str
-            selects the initializer of the weights of the NN. Default: glorot_normal
-        layer_type: str
-            selects the type of architecture of the NN. Default: dense
-        flav_info: dict
-            dictionary containing the information about each PDF (basis dictionary in the runcard)
-            to be used by Preprocessing
-        fitbasis: str
-            fitbasis used during the fit. Default: NN31IC
+        replica_settings: list[ReplicaSettings]
+            list of ReplicaSettings objects which must contain the following information
+                nodes: list(int)
+                    list of the number of nodes per layer of the PDF NN
+                activation: list
+                    list of activation functions to apply to each layer
+                initializer_name: str
+                    selects the initializer of the weights of the NN. Default: glorot_normal
+                layer_type: str
+                    selects the type of architecture of the NN. Default: dense
+                seed: int
+                    the initialization seed for the NN
+                dropout: float
+                    rate of dropout layer by layer
+                regularizer: str
+                    name of the regularizer to use for the NN
+                regularizer_args: dict
+                    options to pass down to the regularizer (if any)
+                flav_info: dict
+                    dictionary containing the information about each PDF (basis dictionary in the runcard)
+                    to be used by Preprocessing
+                fitbasis: str
+                    fitbasis used during the fit. Default: NN31IC
         out: int
             number of output flavours of the model (default 14)
-        seed_list: list(int)
-            seed for the initialization of the Neural Network
         impose_sumrule: str
             whether to impose sumrules on the output pdf and which one to impose (All, MSR, VSR, TSR)
-        dropout: float
-            rate of dropout layer by layer
-        regularizer: str
-            name of the regularizer to use for the NN
-        regularizer_args: dict
-            options to pass down to the regularizer (if any)
         scaler: callable
             Function to apply to the input. If given the input to the model
             will be a (1, None, 2) tensor where dim [:,:,0] is scaled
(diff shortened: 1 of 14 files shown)


Radonirinaunimi:
> Ok, I'll try to finish this asap

Thanks! Will do the same for the review.

scarlehoff:
I haven't added a way of using it with the runcards but added a test for it so you can see how it works.

Adding it to the runcard is somewhat trivial but it requires some discussion (perhaps a `sample_architecture` dictionary?, do we want to read directly from some hyperopt file?) and I think it is better to find problems on the implementation itself in case tehre's something wrong.

n3fit/src/n3fit/model_gen.py
```diff
@@ -323,84 +323,202 @@ def observable_generator(
...
+            self.regularizer_args = dict()
+
+        # Checks
+        if len(self.nodes) != len(self.activations):
+            raise ValueError(
+                f"nodes and activations do not match ({self.nodes} vs {self.activations}"
+            )
+        if self.regularizer_args and self.regularizer is None:
+            raise ValueError(
+                "Regularizer arguments have been provided but not regularizer is selected"
```
Radonirinaunimi, line 378:
```suggestion
                "Regularizer arguments have been provided but no regularizer is selected"
```

n3fit/src/n3fit/model_gen.py
```diff
@@ -323,84 +323,202 @@ def observable_generator(
     return layer_info
 
 
+@dataclass
+class ReplicaSettings:
+    """Dataclass which hold all necessary replica-dependent information of a PDF.
```
Radonirinaunimi, line 328:
```suggestion
    """Dataclass which holds all necessary replica-dependent information of a PDF.
```

n3fit/src/n3fit/model_gen.py
```diff
@@ -323,84 +323,202 @@ def observable_generator(
...
+        if self.regularizer_args and self.regularizer is None:
+            raise ValueError(
+                "Regularizer arguments have been provided but not regularizer is selected"
+            )
+
+
+def sample_pdf_models():
+    """
+    Given a set of parameters, sample a set of PDF models
+    """
```
Radonirinaunimi, line 385: What exactly do you intend to address here? If this is the sampling of the models from the hyperopt scan, I think it is better have it in the `hyper_optimization` submodule. The idea would be that such a function will sample the sets of hyperpameters, dump them in a run card which could be run later.

n3fit/src/n3fit/model_gen.py
```diff
@@ -323,84 +323,202 @@ def observable_generator(
...
-def pdfNN_layer_generator(
-    nodes: list[int] = None,
-    activations: list[str] = None,
-    initializer_name: str = "glorot_normal",
-    layer_type: str = "dense",
+def _pdfNN_layer_generator(
+    replicas_settings: list[ReplicaSettings],
     flav_info: dict = None,
     fitbasis: str = "NN31IC",
     out: int = 14,
```
Radonirinaunimi, line 519:
```suggestion
    output_size: int = 14,
```

n3fit/src/n3fit/model_gen.py
```diff
@@ -323,84 +323,202 @@ def observable_generator(
...
-    seed: int = None,
-    dropout: float = 0.0,
-    regularizer: str = None,
-    regularizer_args: dict = None,
     impose_sumrule: str = None,
     scaler: Callable = None,
-    num_replicas: int = 1,
     photons: Photon = None,
```
Radonirinaunimi, line 395: Just to say that in the (near) future, the arguments of this function will contain `nuclear_info` which will contain the information about the atomic mass number $A$ of the hadron. Then the input of the NN will become `(x, log(x)/scaler(x), A)`.

n3fit/src/n3fit/model_gen.py
```diff
@@ -572,50 +659,39 @@ def pdfNN_layer_generator(
...
+    list_of_nn_pdfs = []
+    for i, replica_settings in enumerate(replicas_settings):
+        rep_pdf = _generate_nn(x_input, i, **asdict(replica_settings))
+        # And build them all with the same input layer
+        list_of_nn_pdfs.append(rep_pdf(x_input))
 
-    # Evolution layer
-    layer_evln = FkRotation(output_dim=out, name="pdf_FK_basis")
+    # Stack all replicas together as one single object
+    nn_pdfs = Lambda(lambda nns: op.stack(nns, axis=1), name=f"stack_replicas")(list_of_nn_pdfs)
```
Radonirinaunimi, line 677:
```suggestion
    nn_pdfs = Lambda(lambda nns: op.stack(nns, axis=1), name="stack_replicas")(list_of_nn_pdfs)
```

n3fit/src/n3fit/model_gen.py
```diff
@@ -671,153 +773,119 @@ def compute_unnormalized_pdf(x):
...
+        layer_photon = AddPhoton(photons=photons, name="add_photon")
         PDFs = layer_photon(PDFs)
 
+    # Return a PDF without a replica axis, to extract single replicas from an ensemble
     if not replica_axis:
         PDFs = Lambda(lambda pdfs: pdfs[:, 0], name="remove_replica_axis")(PDFs)
 
-    pdf_model = MetaModel(model_input, PDFs, name=f"PDFs", scaler=scaler)
-    return pdf_model
+    return MetaModel(model_input, PDFs, name=f"PDFs", scaler=scaler)
```
Radonirinaunimi, line 790:
```suggestion
    return MetaModel(model_input, PDFs, name="PDFs", scaler=scaler)
```

n3fit/src/n3fit/model_gen.py
```diff
@@ -671,153 +773,119 @@ def compute_unnormalized_pdf(x):
...
-    activations: list[str],
-    initializer_name: str,
-    replica_seeds: list[int],
-    dropout: float,
-    regularizer: str,
-    regularizer_args: dict,
-    last_layer_nodes: int,
+# TODO: is there a way of keeping sincronized the input of this function and ReplicaSettings
+# beyond a test of it? In principle we might want to have the arguments explicitly here...
+def _generate_nn(
```
Radonirinaunimi, line 795: Why doesn't this simply take `ReplicaSetting` as part of the argument? This would simplify things and prevents doing something like this for example:
```Python
_generate_nn(x_input, i, **asdict(replica_settings))
```

n3fit/src/n3fit/model_gen.py
```diff
@@ -323,84 +323,202 @@ def observable_generator(
...
-    seed: int = None,
-    dropout: float = 0.0,
-    regularizer: str = None,
-    regularizer_args: dict = None,
     impose_sumrule: str = None,
     scaler: Callable = None,
-    num_replicas: int = 1,
     photons: Photon = None,
     replica_axis: bool = True,
 ):  # pylint: disable=too-many-locals
```
Radonirinaunimi, line 524:
```suggestion
):
```
This should no longer be needed.

n3fit/src/n3fit/model_gen.py
```diff
@@ -459,41 +577,29 @@ def pdfNN_layer_generator(
 
     >>> import numpy as np
     >>> from n3fit.vpinterface import N3PDF
-    >>> from n3fit.model_gen import pdfNN_layer_generator
+    >>> from n3fit.model_gen import _pdfNN_layer_generator
     >>> from validphys.pdfgrids import xplotting_grid
     >>> fake_fl = [{'fl' : i, 'largex' : [0,1], 'smallx': [1,2]} for i in ['u', 'ubar', 'd', 'dbar', 'c', 'cbar', 's', 'sbar']]
     >>> fake_x = np.linspace(1e-3,0.8,3)
-    >>> pdf_model = pdfNN_layer_generator(nodes=[8], activations=['linear'], seed=[2,3], flav_info=fake_fl, num_replicas=2)
+    >>> pdf_model = _pdfNN_layer_generator(nodes=[8], activations=['linear'], seed=[2,3], flav_info=fake_fl)
```
Radonirinaunimi, line 584:
```suggestion
    >>> pdf_model = _pdfNN_layer_generator(nodes=[8], activations=['linear'], flav_info=fake_fl)
```

Radonirinaunimi (commented):
Thank you very much for this 🔥! This will not only be crucial for the hyperopt studies but also for future extensions which will include nPDFs and FFs. For the former, I guess one thing missing is propagating the fact that each replica can have different NN architecture in the runcard?

Here are some comments after a first pass.

n3fit/src/n3fit/model_gen.py
scarlehoff: You are right. I'll remove it.

n3fit/src/n3fit/model_gen.py
scarlehoff: I think that can be at the same level as photon. I was thinking (but let's leave it to another PR) to add a `nuclear` and `fragmentation` namespaces which will arrive here together with photon and include their own replica settings (which I don't think we want to sample as the precision is not there yet)

But I still don't know whether it is the best option

PR #2315, opened 2025-05-06
fix sorting of postfit symlinks

comane:
Currently postfit symlinks are sorted only based on the first number of the replica folder eg

1, 10, 11, 12, 13, ..., 2, 20, ...

This PR fixes this st the order is sorted

### DIFF
--- a/validphys2/src/validphys/scripts/postfit.py
+++ b/validphys2/src/validphys/scripts/postfit.py
@@ -62,6 +62,16 @@ def set_lhapdf_info(info_path, nrep):
         f.truncate()
 
 
+def idx(path):
+    """
+    Get the index of the replica from the path.
+    """
+    # get the final folder name, e.g. "replica_12"
+    name = os.path.basename(os.path.normpath(path))
+    # extract the number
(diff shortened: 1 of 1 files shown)


RoyStegeman:
Note that postfit writes a `FromMCReplica`, which allows you to know which fitted replica a given PDF replica corresponds to. Of course if you don't like the sorting it can be changed, but practically there is no problem with the current ordering.

comane:
> Note that postfit writes a `FromMCReplica`, which allows you to know which fitted replica a given PDF replica corresponds to. Of course if you don't like the sorting it can be changed, but practically there is no problem with the current ordering.

Hi @RoyStegeman, I agree that practically there is no problem with that. In fact I did not label it as a bug.

However, there is a `sorted` operation done in the code and it is confusing since it does not sort properly.

scarlehoff:
The sorting is proper in the sense that it is the same as you get with `ls` and kept constant regardless of the system.

Is there an advantage? The replicas and members are not expected to be aligned anyway.

PR #2317, opened 2025-05-16
new hyperopt figure of merit

tgiani:
Implementation of hyperopt figure of merit taking into account pdf error and penalty term

### DIFF
--- a/n3fit/src/n3fit/vpinterface.py
+++ b/n3fit/src/n3fit/vpinterface.py
@@ -32,6 +32,7 @@ from validphys.covmats import covmat_from_systematics, sqrt_covmat
 from validphys.lhapdfset import LHAPDFSet
 from validphys.pdfbases import ALL_FLAVOURS, check_basis
 from validphys.results import abs_chi2_data, phi_data, results
+from validphys.calcutils import calc_chi2
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


Radonirinaunimi:
> @Radonirinaunimi the only commit missing from the other PR is [d51b3b6](https://github.com/NNPDF/nnpdf/commit/d51b3b65a28e7f6e7f01f72a1299704b691d6073) can you cherry-pick it? Then I'll close the other. Otherwise I can do it.
> 
> It was only a small refactoring and comments.

Done ✅!

Radonirinaunimi:
> I updated last week the pypi package for fiatlux, I think it should be ok now. I've rerun the tests.

I think the problem is that there is no pre-built wheel for ARM macOS and therefore it has to compile `fiatlux`. But now the problem is that macOS images no longer ship a Fortran compiler, hence `CMAKE_Fortran_COMPILER` not found.

scarlehoff:
I only created it for python 3.13 because it was the only one missing https://pypi.org/project/fiatlux/0.1.5/#files for python 3.9 to 3.12 it was already there https://pypi.org/project/fiatlux/0.1.4/#files and it has been working until now.

Radonirinaunimi (approved):
Now that the checks are passing, this could be merged.

PR #2327, opened 2025-06-02
update scalevariationtheoryids for alphas fits

RoyStegeman:
For the CRM and TCM fits we used only central theoryids for the covmat (and the t0theoryid in case of CRM) around 0.118-0.120, so that's why none of those are there yet. However, it was agreed to also provide fits where we assume the input value of alphas are all values from 0.114 to 0.125 without fixing the t0covmat. 

In principle we could reuse many of these lists since they are all the same (e.g. the MHOU and IHOU covmats are always constructed with theories at alphas=0.118), but that would involve introducing something like a `theorycovmattheoryid`, such that we can point to the theoryid in scalevariationtheoryids for which we want to use the list. I think that would overcomplicate things and is not worth it.


scarlehoff:
I think you repeated a few.

Being `yaml` you can do `variations: &nnloqed` and then in all the others that are the same `varations: *nnloqed` so that the file is shorter, so nothing needs to be done in the side of the code.

RoyStegeman:
Yes I was lazy and decided to just have the ci tell me which are already there

Good point about yaml. I completely didn't think about that!

PR #2328, opened 2025-06-05
loop over replica_ folders rather than replica numbers starting from 0

RoyStegeman:
If for some reason the folders are not a consecutive list from 1 to N, that is a problem for evolven3fit


scarlehoff (approved):

RoyStegeman:
I made another change since you approved, it's in the last commit.

scarlehoff:
> I made another change since you approved, it's in the last commit.

Are the result of the dictionary ordered? Because that assumes they are (but also, without that if you have a gap it would fail as well)

PR #2329, opened 2025-06-09
Remove accidental 's' from commondata 

RoyStegeman:
This fixes the ci failing. I think it might be due to the updated pandas version, but didn't check


scarlehoff (approved):
Do you understand what generated the error? I was looking at the pandas 2.3.0 changes and couldn't see why

PR #2330, opened 2025-06-13
Inclusive xsecs

enocera:
This PR incorporates the commondata implementation for the inclusive cross sections of the following papers:
- 2403.12902
- 2308.09529
- 2207.01354
- 1812.10505
- 2303.10680


enocera:
Dear @jacoterh @t7phy this PR contains some new total ttbar cross sections and a total W+-Z cross section. I think that, for completeness, the total ttbar cross sections can be useful for your mt study.
@t7phy Can you please check the implementation?
@jacoterh For most of these cross sections, we can re-use the grids that you have already produced (the c.m.energies are the same as for the NNPDF4.0 data, so I imagine that you already have these). There is an exception. There are two data sets from Run III at 13.6 TeV, for which we don't have grids yet. Can you consider to run these with MATRIX? Thanks.

enocera:
@t7phy Where are your suggestions?

jacoterh (approved):
Thanks @enocera, looks good! I took the freedom to implement @t7phy's suggestions. On the theory side we have all the corresponding predictions except for W+W- and Z production. I checked and saw these can also be computed with MATRIX. I'll open a dedicated PR for this when I have them. The ttbar predictions of this PR I've added to https://github.com/NNPDF/theories_slim/pull/43

nnpdf_data/nnpdf_data/commondata/ATLAS_TTBAR_13P6TEV_TOT/kinematics.yaml
```diff
@@ -0,0 +1,13 @@
...
+    mid: 0
+    max: null
+  m_t2:
+    min: null
+    mid: 29756.25
+    max: null
+  sqrts:
+    min: null
+    mid: 13600
+    max: null
```
scarlehoff, line 13: I guess too late now, but for the future, remember that both `zero` and `sqrts` are not necessary! (even `mid` and `max` are not)

t7phy: yes I thought of removing it, but it was a single bin so I didn't bother xD

PR #2333, opened 2025-06-16
fix photon replica treatment if replica is equal to the number of replicas in the luxset

RoyStegeman:
`photonreplica = replica % luxset_members` returned 0 if `replica==luxset_members` while it should return `luxset_members`. 

I also moved the logic inside the Photon class. This allows us to only use the `photonreplica` when extracting the replica from the luxset, but anywhere else (basically just the luxseed determination) we use the replica number for which we call `n3fit`, for whatever that's worth.


scarlehoff:
> I also moved the logic inside the Photon class. This allows us to only use the photonreplica when extracting the replica from the luxset, but anywhere else (basically just the luxseed determination) we use the replica number for which we call n3fit, for whatever that's worth.

What is the practical effect of this?

n3fit/src/n3fit/model_trainer.py
```diff
@@ -947,12 +947,8 @@ def hyperparametrizable(self, params):
 
         # Initialize all photon classes for the different replicas:
         if self.lux_params:
-            luxset_members = self.lux_params["luxset"].get_members() - 1  # -1 is for replica 0
-            # we take the MOD of the luxset_members to avoid failing due to limited number of
-            # replicas in the luxset
-            photonreplicas = tuple(r % luxset_members for r in self.replicas)
             photons = Photon(
-                theoryid=self.theoryid, lux_params=self.lux_params, replicas=photonreplicas
+                theoryid=self.theoryid, lux_params=self.lux_params, replicas=self.replicas
```
scarlehoff, line 951: Let me try to answer myself: You save in the photon class the replicas from the n3fit call, and then, internally, only when you are going to call the luxpdfset, you take the mod. So practically nothing changes. Is that correct?

RoyStegeman:
To be clear, the bug is resolved by the `or luxset_members`, so for a 100 replica fit instead of  taking replicas 0-99 we now take 1-100, which is what we want since the central rep0 is also included. `or luxset_members` makes sure the 0 becomes a 100.

Issue #2332, opened 2025-06-16
Make the diagonal fit the default


scarlehoff:
It has been decided that for 4.1 we will want the diagonal covmat to be the default way in which fits are run, see this comment:  
https://github.com/NNPDF/nnpdf/pull/2301#issuecomment-2922178907

(it would be best if this is done as one of the first PR after 4.1.0 is released, so by looking at the version of the code one can redo previous fits with or without diagonal)

In addition to make it default, we want to make sure that by using the diagonal basis we reduce the cost on memory of the fits, so that PR would be the perfect place to benchmark that and improve the performance of the fits.

scarlehoff:
One point that I'm not sure we looked at is what to do in multireplica fits. I think you will have to move the diagonalization logic outside of the `fitting_data_dict`, as this function is called once per replica, so `dataset_inputs_fitting_covmat` should already be diagonal when you get there.

PR #2338, opened 2025-06-23
Separate covmat so that the inversion can be hashed

scarlehoff:
This is necessary for https://github.com/NNPDF/nnpdf/issues/2332

I've basically moved the covmat inversion out of the generic function and added a "Hashed array" class. The advantage (right now) is that you can do a very large number of replicas in a GPU with a single covariance matrix (if the trvl faction is set to 1.0).

There are probably more pieces that can be taken out but that's left for #2332

### DIFF
--- a/validphys2/src/validphys/n3fit_data.py
+++ b/validphys2/src/validphys/n3fit_data.py
@@ -22,6 +22,14 @@ from validphys.n3fit_data_utils import validphys_group_extractor
 log = logging.getLogger(__name__)
 
 
+class Hashrray(TupleComp):
+    """Wrapper class to hash a numpy array so it can be cached."""
+
+    def __init__(self, array):
+        self.array = array
+        super().__init__(hash(self.array.tobytes()))
+
+
 def _per_replica(f):
     """Decorator to be used on top of reportengine's decorators.
     It replaces the preparation step of the decorator with a custom function,
@@ -153,7 +161,7 @@ class _Masks(TupleComp):
         super().__init__(group_name, seed)
 
 
-def _diagonal_masks(
+def diagonal_masks(
     data, replica_trvlseed, dataset_inputs_fitting_covmat, diagonal_frac=1.0, threshold_eigvals=0
 ):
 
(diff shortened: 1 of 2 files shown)


validphys2/src/validphys/n3fit_data.py
```diff
@@ -224,6 +235,9 @@ def _standard_masks(data, replica_trvlseed):
         vl_mask = ~tr_mask
         trmask_partial.append(tr_mask)
         vlmask_partial.append(vl_mask)
+    # if we are not masking, remove the seed from the object
+    if nomasking:
+        replica_trvlseed = None
```
jacoterh, line 240: What happens now if `frac = 1.0` for only one one dataset and not for the rest? In that case `nomasking` evaluates to `False`, but some datasets still have `frac` unequal to 1, in which case a mask will still have to be applied. It's a corner case, I know, but I'd like to understand the logic you added here :)

scarlehoff: Indeed, as you said, as soon as one dataset has`frac != 1.0`, `nomasking` will evaluate to `False` so you still get a Mask (which for the subset of datasets with `frac == 1.0` will be `True` for all points).

> It's a corner case, I know, but I'd like to understand the logic you added here :)

If it doesn't work as intended is a bug, corner or not!

jacoterh:
Thanks for this, it looks good, but I'm a bit confused about the use case, because as you said, we only have a single covmat when we don't do any split at all, and I'm not sure when we'd ever need such a case.

scarlehoff:
In the non-diagonal case this is useful only in the "no splitting corner case", however, the moment we turn on the diagonalization, we only have (in practice) one covmat, since the mask happens after diagionalization (and the inversion is trivial in that case).

However, right now, even when the diagonal basis is being used, a separate covariance matrix is generated for each replica. With the added problem that it is the "single-exp" covmat, which is bigger than the usual experiment-by-experiment one. 

This is just the scaffolding: the `Hashrray` container allows you to use the `@functools.cache`, and the separation of cached_covmat -> cached_inv makes sures that reportengine don't recompute it many times.
The no-splitting corner case is just an example (useful, since otherwise running thcovmat fits or diagonal-basis fits in the GPU takes way too much memory). Making it work (in an useful way) with the diagonal basis will still require extra work ^^U

jacoterh:
All clear, thanks!

jacoterh (approved):

PR #2339, opened 2025-06-23
start dataset implementation

vschutze-alt:


scarlehoff:
> Regarding this dataset, should I change figure_by from 'eta' to 'radius' in the metadata file? Given that the two subfigures I'm implementing are differentiated by radius.

For the kinematic variables, `sqrts` is not necessary as this is taken directly from the dataset (13 TeV).

Then, regarding the rest. I'd say these are either two different observables or two variants (@enocera should decide), but the radius is not part of the kinematic variables. `eta` and the transverse energy instead are (as we might want to apply cuts on them for instance).

enocera:
@vschutze-alt (cc @scarlehoff ) Concering the radius, I think that here the situation is similar to that we have with single-inclusive jet or di-jet production, in which indeed we can have two versions of the same data set that differ for the jet radius. In that case we have implemented two variants of the same data set, therefore I would do the same here. To have an idea of how a variant works in this case, you can have a look at `CMS_1JET_13TEV_DIF`.
@scarlehoff Yes, this measurement will make the companion 13 TeV measurement already in NNPDF4.0 outdated. But we want to have grids for both. We may end up finding issues with this new data set and decide to stick to the old one. Also I imagine that Stefano, Richard an Juan, at some, point, would like to study both data sets with NNLO grids. But of course you decide how to proceed with the grid computation. Are, by any chance, the binning and the cuts the same for the two data sets, so that we can re-use the same grids?
@vschutze-alt Concerning the rapidity binning, it seems to me that the bins displayed in the plot above are consistent with those for which there are Hepdata tables. In other words: I don't see Hepdata tables for the 1.37<eta<1.56 bin (but I'm in an airport lounge as I write, so I may be wrong). My feeling is that the experimental setup is such that there is a "missing" rapidity bin. Possibly some additional piece of information is contained in the paper?

enocera (approved):
Dear @VLTSML26 I have revised this PR. I've made the following alterations.
- `metadata.yaml`: I have corrected a typo in the journal name; I have changed the names of the observables; I have added the c.m. energy (even if not required) in the kinematic specifications; I have added the (provisional) theory block.
- `data*`, `kinematics*` and `uncertainteis*` files: the names are modified according to the new names of the observables; note that, after a careful look at the paper, I came to a slight different conclusion on the uncorrelated systematic uncertainties. Note that I set all the uncorrelated uncertainties to the `ADD` treatment. They indeed look of statistical origin, therefore it seems more appropriate for them to be additive. I have left a comment about that in filter.py.
- `filter.py`: I have changed a little the way in which the data is parsed, and I have implemented the new names for the observables.

Note that tests are now passing (I suspect that there was a kinematic mismatch).

nnpdf_data/nnpdf_data/commondata/ATLAS_PH_13TEV_139fb/metadata.yaml
```diff
@@ -0,0 +1,114 @@
...
+hepdata:
+  url: https://www.hepdata.net/record/ins2628741
+  version: 1
+
+implemented_observables:
+
+- observable_name: ET-ETA-R02
+  observable:
+    description: "ATLAS direct photon production 13 TeV at 139 fb^-1 R=0.2"
+    label: r"dsigma/dE_T^\gamma"
```
scarlehoff, line 25: Since this is differential in both I think the label should also have d\eta

nnpdf_data/nnpdf_data/commondata/ATLAS_PH_13TEV_139fb/metadata.yaml
```diff
@@ -0,0 +1,114 @@
...
+    plot_x: ET
+  kinematic_coverage:
+  - eta
+  - ET
+  - sqrts
+  kinematics:
+    variables:
+      eta:
+        description: "Absolute photon pseudorapidity"
+        label: "$|\eta|$"
```
scarlehoff:
```suggestion
        label: r"$|\eta|$"
```

Or use `\\e` to escape the `e`

scarlehoff (commented):
I only have the NLO for now, the NNLO is still running, it might be ready by Amsterdam but in any case:

Somehow the chi2 is great even if there's a 10% difference in the data-th comparison still:

```
Exp chi2: 1.274
```
Image

This is for the R=0.4. I'll run the R=0.2 afterwards. But I think given this NLO this can be merged (upon fixing a bit the labels, the title didn't work because of a combination of unescaped `\` and latex).

PR #2342, opened 2025-06-30
Fixing Jax Backend

ecole41:
This PR aims to fix the jax backend within` n3fit`. So far, the input to the model has been converted from a tracked dictionary into a standard Python dictionary and early stopping has been removed.

When setting the `KERAS_BACKEND=jax` and running n3fit, this model stops training before the first epoch.

### DIFF
--- a/.github/workflows/all_tests_nnpdf.yml
+++ b/.github/workflows/all_tests_nnpdf.yml
@@ -108,6 +108,30 @@ jobs:
       run: |
         cd n3fit/runcards/examples
         n3fit Basic_runcard.yml 4
+        cat Basic_runcard/nnfit/*/Basic_runcard.json
+
+  run_jax:
+    runs-on: ubuntu-latest
+    env:
+      KERAS_BACKEND: jax
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
+      with:
+        python-version: "3.12"
+    - name: Install nnpdf without LHAPDF
+      shell: bash -l {0}
+      run: |
+        pip install .[nolha,jax]
+        # Since there is no LHAPDF in the system, initialize the folder and download pdfsets.index
+        lhapdf-management update --init
+    - name: Test we can run one runcard
+      shell: bash -l {0}
+      run: |
+        cd n3fit/runcards/examples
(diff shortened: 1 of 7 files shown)


scarlehoff:
This last commit seems to fix several problems. The only one I am not sure yet why is it happening is the "Issue with Preprocessing". It seems that calling the validation model's `compute_losses` fail (while calling the training one doesn't).

But the interesting thing is that the layer is the same, so one can call the training, and once it is built, call the validation. This is obviously and morally wrong but seems to work... there _must_ be something that can be done to the validation model in order to call it directly without the extra step.

ecole41:
@scarlehoff I'm finding that with this version, I'm getting an indexing error when running with `n3fit/runcards/examples/Basic_runcard.yml`:

```bash
  File "/Users/ellacole/codes/nnpdf/nnpdfgit/nnpdf/n3fit/src/n3fit/backends/keras_backend/MetaModel.py", line 175, in perform_fit
    history = super().fit(
              ^^^^^^^^^^^^
  File "/Users/ellacole/miniconda3/envs/nnpdf_dev/lib/python3.12/site-packages/keras/src/utils/traceback_utils.py", line 122, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "/Users/ellacole/miniconda3/envs/nnpdf_dev/lib/python3.12/site-packages/optree/ops.py", line 747, in tree_map
    return treespec.unflatten(map(func, *flat_args))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
IndexError: index 15 is out of bounds for axis 0 with size 1
```
The shapes of the inputs look correct so I'm not sure why this is happening. :

```bash
DEBUG: x_params shapes:
  pdf_input: (100, 152, 1)
  xgrid_integration: (100, 2000, 1)
DEBUG: y shapes:
  y[0]: (100, 1)
  y[1]: (100, 1)
  y[2]: (100, 1)
  y[3]: (100, 1)
  y[4]: (100, 1)
  y[5]: (100, 1)
DEBUG: steps_per_epoch: 100
DEBUG: epochs: 900
```
Do you maybe have a run card which works with this

PR #2344, opened 2025-07-15
New evolution grid

evagroenendijk:
Since we changed the values of the quark masses, we have to account for these new values in the Q grids. Otherwise the matching scheme does not work correctly.


scarlehoff:
It needs to be done dynamically (either obly for those two values or for the entire grid) otherwise the 40* theories cannot be reproduced.

evagroenendijk:
> It needs to be done dynamically (either only for those two values or for the entire grid) otherwise the 40* theories cannot be reproduced.

Alright, of course! So we should define a new QGRID_NNPDF41 and QGRID_IC_41 then, right?

evagroenendijk:
@felixhekhorn @scarlehoff Does one of you perhaps know a quick way to make a test eko? I adjusted evolven3fit and would like to test if "evolven3fit produce_eko" now makes the eko correctly, but I would prefer to test quickly.

My test now is to make an eko only for 3 points in the q_grid (1.502, 1.502 and 1.65), but this already takes very long. Thanks!

n3fit/src/evolven3fit/eko_utils.py
```diff
@@ -74,6 +92,7 @@ def construct_eko_cards(
         },
         theory["nf0"],
         legacy40=legacy40,
+        theory_41=False,
```
scarlehoff, line 95: Basically, I don't think we should have this if condition, but rather modify the `generate_q2grid` so that it is generated instead of taking a fixed value.

Or, at worst, the `generate_q2grid` below should, in the situations in  which it would get `Q2GRID_DEFAULT`,  say "ok, and now this defaullt should have instead of 1.65 -> `thresholds["c"]`, instead of `mb` -> `thresholds["b"]`

evagroenendijk: Okay, also I agree! I wanted to at least quickly fix it for now. I checked that my fix works, but I get that the permanent solution should be a bit more elegant!

n3fit/src/evolven3fit/utils.py
```diff
@@ -29,16 +29,39 @@ def generate_q2grid(Q0, Qfin, Q_points, match_dict, nf0=None, legacy40=False, th
...
-            return Q2GRID_DEFAULT
-        elif theory_41:
-            return Q2GRID_DEFAULT_41
-        elif nf0 is None:
-            raise ValueError("In order to use a default grid, a value of nf0 must be provided")
-        else:
-            raise NotImplementedError(f"No default grid in Q available for {nf0=}")
+        else: 
+            grids = []
+            Q_low = 1
```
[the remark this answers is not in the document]
evagroenendijk: So my idea was to fully remove all default grids except for the NNPDF4.0 legacy grid. So what I did was:

1. create the grid dynamically, unless legacy40 is specified. The grid is filled in between the quark thresholds and the starting scale, because those have to be in there anyway. I now did it analogously to what was already done when Qini and Qfin are specified (maybe I can merge those together to make the code shorter). If you don't agree with the number of points or the separation of points I put, let me know because that can be easily changed. I noticed that I got an error that my q grid is not exactly the same as ones that were used before, but I thought maybe that is ok
2. I did it s.t. the grid always starts at Q=1, because to be honest I did not really see why we should not just use the whole Q range from 1 to 1e5 always
3. Also I did not really see why you would have to get a q grid error if you don't specify nf0, because now we would just get the same Q grid for all nf0's

Please let me know if you have different thoughts. To me this seemed like the simplest way to define the q grid, but I don't know if there were specific reasons for having the nf0 error, all the different default grids, etc.

n3fit/src/evolven3fit/utils.py
```diff
@@ -19,45 +19,84 @@ def get_theoryID_from_runcard(usr_path):
...
+    
+    # If flag --legacy40 is set return handmade legacy grid
+    if legacy40:
+        return Q2GRID_NNPDF40
+    # Otherwise dynamically create the grid from Q2_min --> Q2_max
+    else: 
+        if Qfin is None and Q_points is None:
+            Q2_min = 1.0**2
+            Q2_max = 1e5**2
+            Q0 = 1.65
```
scarlehoff:
```suggestion
```

I meant given by the input, instead of hardcoded

n3fit/src/evolven3fit/utils.py
```diff
@@ -19,45 +19,84 @@ def get_theoryID_from_runcard(usr_path):
...
+        return Q2GRID_NNPDF40
+    # Otherwise dynamically create the grid from Q2_min --> Q2_max
+    else: 
+        if Qfin is None and Q_points is None:
+            Q2_min = 1.0**2
+            Q2_max = 1e5**2
+            Q0 = 1.65                    
+            Lambda2 = 0.0625                  
+            total_points = 50
+            total_points_ic = 6
```
felixhekhorn: as @scarlehoff suggested: make all of those parameters an input to this function - except `Lambda2`, which I suggest to make a module constant instead (since it is unlikely to change) (i.e. also rename it to `LAMBDA2`)

n3fit/src/evolven3fit/utils.py
```diff
@@ -19,45 +19,84 @@ def get_theoryID_from_runcard(usr_path):
...
+        # Collect all node Q2's from Q0^2 --> Q2_max
+        q0_2 = Q0**2
+        node_Q2 = [q0_2, (match_dict["mb"]*match_dict["kbThr"])**2, Q2_max]
+
+        # Define map
+        def Q2_to_t(q2: float) -> float:
+            return np.log(np.log(q2 / Lambda2))
+
+        def t_to_Q2(t: float) -> float:
+            return Lambda2 * np.exp(np.exp(t))
```
felixhekhorn: move these on module level

n3fit/src/evolven3fit/utils.py
```diff
@@ -19,45 +19,84 @@ def get_theoryID_from_runcard(usr_path):
...
+        for q2 in q2_vals:
+            while subgridindex < n_intervals - 1 and q2 >= node_Q2[subgridindex + 1]:
+                subgridindex += 1
+            nQpoints[subgridindex] += 1
+        
+        # Make t grid from Q2_min --> Q0^2
+        t_vals_ic = np.linspace(Q2_to_t(Q2_min), Q2_to_t((match_dict["mc"]*match_dict["kcThr"])**2), total_points_ic)
+        q2_vals_ic = t_to_Q2(t_vals_ic)
+        
+        # Now build each subgrid to contain the points we want
```
felixhekhorn: naively it looks to me this function can be split at least in to three parts:
1. some initialization
2. some preparation
3. the actual thing

if that is true, please split the code accordingly, e.g. 2) and 3) can be independent functions taking a smaller set of arguments

felixhekhorn (changes requested):

n3fit/src/evolven3fit/utils.py
```diff
@@ -19,45 +19,84 @@ def get_theoryID_from_runcard(usr_path):
     return my_runcard["theory"]["theoryid"]
 
 
-def generate_q2grid(Q0, Qfin, Q_points, match_dict, nf0=None, legacy40=False):
-    """Generate the q2grid used in the final evolved pdfs or use the default grid if Qfin or Q_points is
-    not provided.
+def generate_q2grid(Q0, Qfin, Q_points, match_dict, legacy40=False):
```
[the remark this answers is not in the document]
scarlehoff: No. We want to keep the `legacy40` to return the grid of 4.0. Everything else can go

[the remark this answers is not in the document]
felixhekhorn: then let's do that: keep `legacy40=False` (for legacy reason), keep `Q0` and `match_dict` (because they are needed), drop `Qfin` and `Q_points` (which don't serve anything any longer), and add the new ones mentioned below.

felixhekhorn:
Please "resolve conversation" when you think you have addressed the issue(s) - I'll unresolve them when I disagree or open a new one with the partial problem (else I can not know when to do that)

n3fit/src/evolven3fit/utils.py
evagroenendijk: Then I did not understand what you meant. I split the function up into multiple lines now, I thought that was what you were saying. What do you mean then?

n3fit/src/evolven3fit/eko_utils.py
```diff
@@ -62,19 +62,21 @@ def construct_eko_cards(
...
-        q_points,
-        {
-            theory["mc"]: thresholds["c"],
-            theory["mb"]: thresholds["b"],
-            theory["mt"]: thresholds["t"],
+        Q0=mu0,
+        Qmin=1.0,
+        Qmax=1E5,
+        total_points=50,
+        total_points_ic=6,
```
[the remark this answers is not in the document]
scarlehoff: not at all

n3fit/src/n3fit/tests/test_evolven3fit.py
```diff
@@ -73,27 +73,36 @@ def check_lhapdf_dat(dat_path, info):
...
-    t2 = 100.0 * 1.0
+    np.testing.assert_allclose((1.0) ** 2, matched_grid[0])
+    np.testing.assert_allclose((100.0) ** 2, matched_grid[-1])
 
-    np.testing.assert_allclose((1.65) ** 2, matched_grid[0])
-    np.testing.assert_allclose((1.0e5) ** 2, matched_grid[-1])
-    assert t1**2 in matched_grid
-    assert t2**2 in matched_grid
+    # Test the legacy40 grid
+    legacy40 = utils.generate_q2grid(1, 1, 1, {}, 1, 1, legacy40=True)
```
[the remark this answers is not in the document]
evagroenendijk: So in my opinion it's not such a problem, because it will take the legacy40 grid whatever the other parameters are. And I think you can just set the flag --legacy40 in the command line without specifying the other parameters. But if you have another suggestion I can look at it!

PR #2346, opened 2025-07-21
Implementation of CMS TTBAR 2L 2024 data

t7phy:


nnpdf_data/nnpdf_data/commondata/CMS_TTBAR_13TEV_2L_138FB-1_DIF/kinematics_dSig_dmttBar.yaml
```diff
@@ -0,0 +1,57 @@
+bins:
```
jacoterh, line 1: Are we sure we don't need 

```
- sqrts:
    min: null
    mid: 13000.0
    max: null
```

anymore?

t7phy: It's no longer needed as the code pulls the energy from the dataset name now

PR #2349, opened 2025-07-24
add sample theorycard for alphas mtop variations at n3lo

RoyStegeman:
@jacoterh I took 40_005_000 as a starting point and changed ModEv from TRN to EXA. I'll add the complete set later, but what is the alphas grid for which we want to produce these? Do we just want the 6 points you listed earlier, or should we do all mtop for all values of alphas? Or something in between? 

For reference of what you requested:
(0.116, 172.5), (0.118, 172.5), (0.120, 172.5).
(0.118, 170.0), (0.118, 172.5), (0.118, 175.0)


nnpdf_data/nnpdf_data/theory_cards/40013000.yaml
```diff
@@ -0,0 +1,48 @@
...
+QED: 0
+Qmb: 4.92
+Qmc: 1.51
+Qmt: 172.5
+Qref: 91.2
+SIN2TW: 0.23126
+TMC: 1
+XIF: 1.0
+XIR: 1.0
+alphaqed: 0.007496252
```
jacoterh, line 36: This value of `alphaqed` corresponds to scale `Qedref=1.777`. However, if  `Qedref` is not specified like in this theory card, the scale assumed for `alphaqed` is taken from `Qref`, which coincides with the Z pole. But at the Z pole we have `alphaqed: 0.0077553`, so we should change `0.007496252` to `0.0077553`  (or specify `Qedref=1.777`, but I'm not sure whether this key is supported still)

nnpdf_data/nnpdf_data/theory_cards/40013000.yaml
```diff
@@ -0,0 +1,48 @@
...
+GF: 1.1663787e-05
+HQ: POLE
+IC: 1
+MP: 0.938
+MW: 80.398
+MZ: 91.1876
+MaxNfPdf: 5
+ModEv: EXA
+ModSV: null
+NfFF: 4
```
jacoterh, line 24: What is the meaning of the keys `NfFF`, `ModSV`, `nfref` (nr of flavours at the reference scale Q0?), `nf0`, `n3lo_ad_variation`, `n3lo_cf_variation`  and  `use_fhmruvv`. I couldn't find these  at https://docs.nnpdf.science/theory/theoryparamsdefinitions.html. I can update the website if you tell me :)

jacoterh:
Thanks for this, to answer your question, let's add also the variations at alphas=0.115, 0.117, 0.119, 0.121, 0.122 and 0.123. Probably 0.115 won't be necessary, but better safe than sorry while we're at it

nnpdf_data/nnpdf_data/theory_cards/40013000.yaml
felixhekhorn: Sorry for jumping in, but I believe these are all question which I can answer. You have to specify $\alpha_{em}$ and $\alpha_s$ at the same scale if you are doing a joint evolution, i.e. `Qedref` is no longer supported as it would be inconsistent. However, since you also have `QED=0` you are doing QCD only and thus `alphaqed` is completely ignored I believe. Specifically, it is not used in eko nor yadism nor mg5 (https://github.com/NNPDF/pinefarm/issues/101) nor Matrix (most likely - although that interface is not publicly available :innocent: )

nnpdf_data/nnpdf_data/theory_cards/40013000.yaml
```diff
@@ -0,0 +1,48 @@
...
+- 0.999152
+DAMP: 0
+FNS: FONLL-E
+GF: 1.1663787e-05
+HQ: POLE
+IC: 1
+MP: 0.938
+MW: 80.398
+MZ: 91.1876
+MaxNfPdf: 5
```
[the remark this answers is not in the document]
felixhekhorn: this is no longer needed https://github.com/NNPDF/eko/pull/295 (i.e. dropping is the right thing)

nnpdf_data/nnpdf_data/theory_cards/40013000.yaml
felixhekhorn: If you could update the documentation that would be great :see_no_evil: 
- `NfFF` basically ignored (it can be used in yadism for a FFNS computation and would specify the number of light flavours there, but in nnpdf we never do FFNS (but FONLL))
- `ModSV` scale variation scheme - since we always use "scheme B" (see [old MHOU paper, sec. 3.3](https://inspirehep.net/literature/1741422)), it should be `expanded`
- `nfref` number of flavours at the couplings reference scale. As `Qref` is typically $M_Z$ it should be `5`
- `nf0` number of flavours at the fitting scale. As `Q0` is typically 1.65 GeV it should be `4`
- `n3lo_ad_variation` IHOU variation of the N3LO anomalous dimensions. Order is [(gg, gq, qg, qq, nsp, nsm, nsv)](https://github.com/NNPDF/eko/blob/f8b9910417f979d179b67a9366e2df1224c87f49/src/ekore/anomalous_dimensions/unpolarized/space_like/__init__.py#L42) Values are 0 or 1 for FHMRUVV and 0..n for in-house (n we would need to dig out somewhere)
- `n3lo_cf_variation` IHOU variations of the N3LO massive DIS coefficient functions. Values are -1, 0, 1 (see [here](https://github.com/NNPDF/yadism/blob/072fc9ecd5f2496fbb5b21c143673fe7232f3ff7/src/yadism/coefficient_functions/heavy/n3lo/__init__.py#L15))
- `use_fhmruvv` use N3LO anomalous dimensions by FHMRUVV group instead of in-house? should default to `true` (after decision at some PC)

scarlehoff: > `NfFF`

This then should be updated in the definition of the Theory, which I see was added not only as an important one but even mandatory! (although from the comment it shouldn't be) https://github.com/NNPDF/nnpdf/blob/aa7cd57bdd75382be194dd47b874ef6b8921eba4/nnpdf_data/nnpdf_data/theory.py#L32

> `use_fhmruvv`

The decision was not propagated to the repository though:
https://github.com/NNPDF/nnpdf/blob/aa7cd57bdd75382be194dd47b874ef6b8921eba4/nnpdf_data/nnpdf_data/theory.py#L82

> `nfref`, `nf0`

These are only needed when you actually want to do something inconsistent. imho it is better to leave them out and let the code decide automatically depending on Q and Q0.

scarlehoff: > which code?

The file I linked! A bit below https://github.com/NNPDF/nnpdf/blob/aa7cd57bdd75382be194dd47b874ef6b8921eba4/nnpdf_data/nnpdf_data/theory.py#L137

nnpdf_data/nnpdf_data/theory_cards/40013000.yaml
jacoterh: Thanks Felix for helping out! We will also do aN3LO + QED at some point, so it's good to know alphaQED and alphas need to be defined at the same scale

jacoterh:
@RoyStegeman I've added the alphas variations theory cards, please check whether you agree :)

RoyStegeman:
> Question: what do we do if our prior on alphas is not centered around the alphas value 0.118 we use in mtop? 

We should be able to convince ourselves (and other people..) that this effect is either small and under control and we attribute some uncertainty to it, or completely negligible. If it is not small, I don't think we can get around creating a finer spacing in alphas. Although of course the same applies to mtop. 

The point is that when the prediction does not agree with the prior, we are relying on the linear interpolation encoded in the theory covmat. At some point the uncertainty associated with the interpolation will become too large compared to the precision we're aiming for. When exactly that point is, I don't know. 

Having said that, I don't think it's too expensive to do a more finely spaced grid of 0.001 stepsize between 0.114 and 0.125. We need four sets of theories (NNLO, NNLO+QED, aN3LO, aN3LO+QED) and for the mtop variations we only have to do the top datasets and the rest can be re-used.

jacoterh:
@felixhekhorn small question: when I generate the FK tables with `ModSV` set to `expanded`, I'm getting

`XIF is 1.0 while ModSV is expanded. If XIF is equal to 1.0, ModSV should not be defined.`

Should I set it to `none` instead?

PR #2350, opened 2025-07-25
Remove spurious lhapdf imports

scarlehoff:
This PR finalizes hiding the `lhapdf` imports under the `lhapdf_compatibiltiy` module, which at the moment was still imported explicitly in various places meaning one could not cleanly use a different backend.

Among other things, this means that now every management or fitting task can be performed without LHAPDF: `n3fit`, `postfit`, `vp-get *`, etc.
And management tasks even without `pdfflow` (or eventually `neopdf`).

In principle reports could also be done without LHAPDF, but `pdfflow` takes way too much memory for many members and `neopdf` is not implemented yet.


Radonirinaunimi (approved):
This looks good to me!

PR #2351, opened 2025-07-28
Modify commondata theory for jets

evagroenendijk:


scarlehoff:
Thanks for this. Please add a `legacy_theory` variant as well with the old namea so that old results can be reproduced.

nnpdf_data/nnpdf_data/commondata/CMS_1JET_8TEV/metadata.yaml
```diff
@@ -68,7 +68,13 @@ implemented_observables:
       theory: *id002
   theory:
     FK_tables:
-    - - CMS_1JET_8TEV-CMS_1JET_8TEV
```
scarlehoff, line 71: I think in this one we need (like in the one above) the separated `legacy` `legacy_theory`, etc.

nnpdf_data/nnpdf_data/commondata/CMS_2JET_7TEV/metadata.yaml
```diff
@@ -53,10 +53,18 @@ implemented_observables:
   data_central: data.yaml
   data_uncertainties:
   - uncertainties.yaml
-
+  variants:
+    legacy:
```
scarlehoff, line 57:
```suggestion
    legacy_theory:
```

I think this makes a bit clear that we only have a legacy difference for the theory.

scarlehoff (commented):
The only thing missing is the `legacy_theory` in one of the datasets, for the rest I think this can be merged.

PR #2352, opened 2025-07-29
CMS_Z0_13TEV

evagroenendijk:
Implementation of the Drell-Yan forward-backward asymmetry at high dilepton masses in proton-proton collisions at  13 TeV.

### DIFF
--- a/validphys2/src/validphys/cuts/filters.yaml
+++ b/validphys2/src/validphys/cuts/filters.yaml
@@ -1,4 +1,4 @@
-- dataset: CMS_Z0_13TEV-AFB
+- dataset: CMS_Z0_13TEV_AFB
   reason: |
    We remove the Z-mass invariant mass bin where Ew corrections are
    expected to be large


enocera:
Thanks @evagroenendijk . We have the possibility of defining two different data sets, as you propose and did. Or also to define a single data set with two different distributions (one for electrons and one for muons). We don't have strong reasons to prefer one than the other, given that these are essentially uncorrelated. So I suggest that you proceed with the implementation that you've already started.

scarlehoff:
Two questions/comments

1. Is this pure charged current, neutral current or both? If both we need to separate W and Z (and if only one of the two, it should be either Z0 or WPWM). This is necessary for the theory covmat.

2. If the paper is the same, the rawdata is the same and the definition of the bins is the same, I think it is better to have them in one single folder with two observables like in LHCB Z0 13 TeV https://github.com/NNPDF/nnpdf/tree/master/nnpdf_data/nnpdf_data/commondata/LHCB_Z0_13TEV

enocera:
@scarlehoff This is the Z forward backward asymmetry, so NC only.

enocera:
@evagroenendijk Incidentally, I would implement only AFB (electrons, muons), not Delta AFB.

evagroenendijk:
> @evagroenendijk Incidentally, I would implement only AFB (electrons, muons), not Delta AFB.

Ok! I thought because the uncertainties are different (the pdf uncertainty cancels out mostly for \Delta A_FB), I thought it would maybe be useful? But I can take only the A_FB!

enocera:
The thing is that, in the SM, at parton level, Delta AFB should be zero. So this observable is better suited to study new physics than constraints on PDFs.

evagroenendijk:
> The thing is that, in the SM, at parton level, Delta AFB should be zero. So this observable is better suited to study new physics than constraints on PDFs.

Alright, clear! Thanks

And shall I put them into one dataset as Juan suggested then?

evagroenendijk:
> Dear @evagroenendijk in view of the upcoming Morimondo meeting, I have reviewed this PR and implemented the following alterations.
> 
> * I have removed the separate electron and muon channels, and I have retained only the combined measurement, the reason being that the breakdown of systematic uncertainties is available only for this.
> * I have fixed some inaccuracies in `metadata.yaml`, in particular in regards to the name of the observable.
> * I have removed the last bin: this is indeed the integral over the invariant mass range, but we are interested in the measurement differential in the various invariant mass bins.
> * I have implemented two variants of the uncertainties, one in which we only have the statistical and the systematic uncertainty, as per Hepdata, and one in which we have the breakdown of systematic uncertainties from Table 1 of the paper (under the assumption that it applies to all invariant mass bins). I have slightly modified the uncertainty treatment in this case - I agree with you that it is highly ambiguous to determine which uncertainty is correlated and which is not.

Dear @enocera Thank you! I have been on holiday for the past few weeks, and I was planning to dot the i's on this pull request in these days before the Morimondo meeting (indeed I had not finished with the metadata.yaml and the uncertainties). Thanks a lot for finishing it, I completely agree with the changes!

nnpdf_data/nnpdf_data/commondata/CMS_Z0_13TEV/metadata.yaml
```diff
@@ -0,0 +1,52 @@
...
+  data_uncertainties:
+  - uncertainties_AFB.yaml
+  variants:
+     breakdown:
+       data_uncertainties:
+       - uncertainties_breakdown_AFB.yaml
+  theory:
+    FK_tables:
+    - - CMS_Z0_13TEV_FWD
+      - CMS_Z0_13TEV_BWD
```
scarlehoff, line 49: We don't have these grids yet, right?

Do we want to merge it already or wait for the grids? (regardless of whether they are correct or not, just to keep track of the missing ones)

enocera:
> By eye it looks like the few points at the end are driving the difference. For a fit we can opt for option 2) and add a theory error for the uncertainty that we know we have which will deweight the more problematic points.

Why not a cut?

PR #2357, opened 2025-08-14
Adding tool to shift theoretical predictions due to correlated exp errors in data/theory comparison plots

enocera:
This PR adds the argument `with_shift` to the function `plot_fancy` in such a way that it is possible to display theoretical predictions shifted according to the correlated part of the experimental error in data/theory comparison plots. The new argument is of type `bool`, with default `True`, in which case the shift is applied and only the uncorrelated part of the experimental uncertainty is displayed. If `with_shift=False` the old behaviour is recovered. Documentation and tutorials have been updated accordingly. This PR supersedes #2102 .


validphys2/examples/data_theory_comparison.yaml
```diff
@@ -1,21 +1,30 @@
...
+    theoryid: 40_000_000
+    use_cuts: "internal"
+    with_shift: False
+    pdf: NNPDF40_nnlo_as_01180 
 
 dataset_inputs:
-      - { dataset: BCDMS_NC_NOTFIXED_P_EM-F2, variant: legacy}
       - { dataset: HERA_NC_318GEV_EAVG_BOTTOM-SIGMARED, variant: legacy}
+      - { dataset: ATLAS_1JET_8TEV_R06, variant: legacy}
+      - { dataset: BCDMS_NC_NOTFIXED_P_EM-F2, variant: legacy}
```
scarlehoff, line 21:
```suggestion
      - { dataset: HERA_NC_318GEV_EAVG_BOTTOM-SIGMARED}
      - { dataset: ATLAS_1JET_8TEV_R06_PTY, variant: "legacy_theory"}
      - { dataset: BCDMS_NC_NOTFIXED_P_EM-F2, variant: "legacy_dw"}
```

validphys2/src/validphys/config.py
```diff
@@ -213,6 +212,10 @@ def parse_use_cuts(self, use_cuts: (bool, str)):
 
         return res
 
+    def parse_with_shift(self, with_shift: bool):
+        """Shift to theory due to correlated systematics in data/theory plots"""
+        return with_shift
+
```
scarlehoff, line 218:
```suggestion
```

I believe this is not needed since it is just a `True/False` flag? Or am I missing something?

enocera: Done

PR #2359, opened 2025-08-21
Make the diagonal fit the default

jacoterh:
This PR addresses issue #2332

### DIFF
--- a/n3fit/runcards/example-nnpdf41.yml
+++ b/n3fit/runcards/example-nnpdf41.yml
@@ -6,94 +6,98 @@ description: Runcard for the NNPDF4.1 series of fits; WIP
 
 ######################################################################################
 dataset_inputs:
-- {dataset: NMC_NC_NOTFIXED_EM-F2, frac: 0.75, variant: legacy_dw}
-- {dataset: NMC_NC_NOTFIXED_P_EM-SIGMARED, frac: 0.75, variant: legacy}
-- {dataset: SLAC_NC_NOTFIXED_P_EM-F2, frac: 0.75, variant: legacy_dw}
-- {dataset: SLAC_NC_NOTFIXED_D_EM-F2, frac: 0.75, variant: legacy_dw}
-- {dataset: BCDMS_NC_NOTFIXED_P_EM-F2, frac: 0.75, variant: legacy_dw}
-- {dataset: BCDMS_NC_NOTFIXED_D_EM-F2, frac: 0.75, variant: legacy_dw}
-- {dataset: CHORUS_CC_NOTFIXED_PB_NU-SIGMARED, frac: 0.75, variant: legacy_dw}
-- {dataset: CHORUS_CC_NOTFIXED_PB_NB-SIGMARED, frac: 0.75, variant: legacy_dw}
-- {dataset: NUTEV_CC_NOTFIXED_FE_NU-SIGMARED, cfac: [MAS], frac: 0.75, variant: legacy_dw}
-- {dataset: NUTEV_CC_NOTFIXED_FE_NB-SIGMARED, cfac: [MAS], frac: 0.75, variant: legacy_dw}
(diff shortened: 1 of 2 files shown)


validphys2/src/validphys/n3fit_data.py
```diff
@@ -494,7 +500,7 @@ def fitting_data_dict(
     expdata = make_replica
     fittable_datasets = fittable_datasets_masked
 
-    inv_true, invcovmat_tr, invcovmat_vl, ndata_tr, ndata_vl, tr_mask, vl_mask, diag_rot = (
+    covmat, inv_true, invcovmat_tr, invcovmat_vl, ndata_tr, ndata_vl, tr_mask, vl_mask, diag_rot = (
```
jacoterh: @scarlehoff all covmat manipulation now happens as part of `_inv_covmat_prepared` which is cached so that it can be recycled across replicas.

jacoterh:
I have removed the commits related to the Cholesky decomposition from this PR. I have also rebased on top of main

validphys2/src/validphys/n3fit_data.py
jacoterh: Just to make this explicit, the hash id on masks is different for each of the replicas atm 
```
_inv_covmat_prepared called with masks id=6414820096, covmat hash=-1248862427858921237, diagonal_basis=True

_inv_covmat_prepared called with masks id=6332071232, covmat hash=-1248862427858921237, diagonal_basis=True

_inv_covmat_prepared called with masks id=6317245888, covmat hash=-1248862427858921237, diagonal_basis=True

_inv_covmat_prepared called with masks id=6416185200, covmat hash=-1248862427858921237, diagonal_basis=True
```

scarlehoff:
I tried running a MHOU fit (in order to try and get the baseline already with this) but it crashed at the vp-setupfit stage. I think this might be related with some of the errors in the tests.

kamillaurent:
Hi @scarlehoff, @jacoterh, here is the report of the default diagonal fit test on the latest settings (`251105-jcm-nnpdf41-mhou-001` runcard): https://vp.nnpdf.science/kSXuJUNETmCmPFoZM8oY2g==

jacoterh:
Hi @scarlehoff , there's something weird going on in the pseudodata test on master. Somehow, it doesn't take into account correlations across datasets and draws replicas only dataset by dataset, while this PR draws the pseudo data from the full covariance matrix. Therefore, the test `test_read_matches_recreate` which attempts to recreate the pseudodata that got saved during fit `pseudodata_test_fit_n3fit_250616` behaves differently on master compared to this branch. 

Specifically, https://github.com/NNPDF/nnpdf/blob/38e38bfb67635562e1878b18521b8dc8bf3b89c1/validphys2/src/validphys/pseudodata.py#L211

gets called for each dataset when running `pytest test_pseudodata.py` on master , while it only gets called once on this branch (assuming a single replica). I found this after looking more closely into *why* I had to regenerate pseudodata for the tests to pass again.

scarlehoff:
This makes sense, right? It should be taking it experiment by experiment and then eventually a dataframe is created separated by dataset (which I guess it is what the test is saving)

https://github.com/NNPDF/nnpdf/blob/38e38bfb67635562e1878b18521b8dc8bf3b89c1/validphys2/src/validphys/pseudodata.py#L218C1-L219C1

Since `pseudodata_test_fit_n3fit_250616` is old, you need to assume this was done with the old default, so you should test diagonal: off 
Actually, I think it is great that it failed, you want `test_read_matches_recreate` both for diagonal and not diagonal! So you can keep the non-diagonal one and add a diagonal one, to make sure that both are correct.

That said, you might want to also add a version check in `fit.as_input()` (like [here](https://github.com/NNPDF/nnpdf/blob/73572c1a4fa08a0f60e3b3e6073a2f97e577e88d/validphys2/src/validphys/overfit_metric.py#L76) so that when the version is older than 4.1.2, the diagonal matrix is set to `false` (we will tag as soon as this is merged)

PR #2360, opened 2025-08-24
CMS_Z0J_13TEV

enocera:
This PR incorporates the commondata implementation for the data set in the title.


nnpdf_data/nnpdf_data/commondata/CMS_Z0J_13TEV_PT/kinematics.yaml
```diff
@@ -0,0 +1,1141 @@
...
+    min: null
+    mid: 13000.0
+    max: null
+- pT:
+    min: 8.0
+    mid: 9.0
+    max: 10.0
+  m_ll2:
+    min: null
+    mid: 8317.44
```
scarlehoff, line 692: I'm afraid something went wrong here since all data is set as mll2 == 8317.44, mll ~ 91 but this dataset should be separated by ranges of mll

nnpdf_data/nnpdf_data/commondata/CMS_Z0J_13TEV_PT/filter.py
```diff
@@ -0,0 +1,120 @@
...
+            data_central.append(data_value["value"])
+        # Kinematic bins
+        kin_values = input["independent_variables"][0]["values"]
+        for kin_value in kin_values:
+            kin = {
+                'pT': {'min': kin_value['low'],
+                       'mid': 0.5 * (kin_value['low'] + kin_value['high']),
+                       'max': kin_value['high']},
+                'm_ll2': {'min': None, 'mid': 8317.44, 'max': None},
+                'sqrts': {'min': None, 'mid': 13000.0, 'max': None},}
```
scarlehoff, line 61:
```suggestion
```

scarlehoff (changes requested):
I think there's something wrong with the kinematics.

This might be a good opportunity to start using `m_ll` instead of the square...

RE data Vs theory something is wrong with the first bin (this is theory / data), it is systematically 30% off (probably some normalization or I did wrong a cut in my theory predictions).

The second bin and third bins seem ok.

Third and fourth a bit worse, I guess the statistics are all around the place but so is the data:

```
ratio = array([0.63142557, 0.69191467, 0.72222455, 0.72407745, 0.76986966,
       0.82255564, 1.04158693, 1.03487063, 1.01877841, 1.01176393,
       1.01778132, 1.00585585, 1.00522577, 0.99522485, 0.99550279,
       1.00101172, 1.00696618, 1.01148373, 1.02241088, 1.0466769 ,
       1.06189633, 1.07604775, 1.06241513, 1.1156114 , 1.03361906,
       0.98315902, 0.97514231, 0.97673345, 0.93802319, 1.00462591,
       0.98311333, 0.97416855, 0.88055417, 1.07630434, 1.00228661,
       0.8514168 ])
```

And the plots for the 2nd, 3rd and 5th bins (disregard the value of the mass which is wrong and filled just to get different plots)

Screenshot 2025-09-18 at 14 22 37
Screenshot 2025-09-18 at 14 23 25
Screenshot 2025-09-18 at 14 24 43

nnpdf_data/nnpdf_data/commondata/CMS_Z0J_13TEV_PT/kinematics.yaml
```diff
@@ -0,0 +1,1141 @@
+bins:
+- pT:
+    min: 0.0
+    mid: 1.0
+    max: 2.0
+  m_ll2:
+    min: null
+    mid: 8317.44
+    max: null
+  sqrts:
```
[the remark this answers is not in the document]
enocera: I have removed sqrts.

nnpdf_data/nnpdf_data/commondata/CMS_Z0J_13TEV/metadata.yaml
```diff
@@ -0,0 +1,63 @@
...
+  kinematics:
+    variables:
+      pT:
+         description: "lepton pair transverse momentum"
+         label: '$p_T$'
+         units: "GeV"
+      m_ll:
+         description: "Z boson mass squared"
+         label: '$m^2_{\ell \ell}$'
+         units: "GeV^2"
```
scarlehoff:
```suggestion
         label: '$m_{\ell \ell}$'
         units: "GeV"
```

Not sure whether I broke this during the rebase.

nnpdf_data/nnpdf_data/commondata/CMS_Z0J_13TEV/metadata.yaml
```diff
@@ -0,0 +1,63 @@
...
+    file: kinematics.yaml
+
+  data_central: data.yaml
+  data_uncertainties:
+  - uncertainties.yaml
+
+  theory:
+    conversion_factor: 1000.0
```
scarlehoff:
```suggestion
    conversion_factor: 0.001
```
Data is in pb and theory in fb

scarlehoff (commented):
I'm a bit worried about this dataset still because the agreement of the second bin is quite good (which is also special since it is the one that contains the Z mass).

But, here's the problem, I'm not normalising by the bin width, and according to the hepdata info, it is normalized. So I'd like to have a second look.

fwiw, the differences I see (the sames as I put above) are clearly not a factor of the bin width either.

enocera:
This PR has become very messy, as it incorporates changes well beyond those in the title. This PR is going to be split in three sub-PRs:
- [x]  one for the CMS_Z0J_13TEV data set #2468 
- [x] one for the fixed ATLAS_Z0J_13TEV data set - #2469 
- [ ] one for variants of both data sets (not needed for NNDPF4.1) - to appear

enocera:
@scarlehoff Concerning the second point above, as you may remember, we discovered that the commondata implementation was bugged because only a fraction of the pT bins were implemented and the uncertainty was one order of magnitude too large. So we really need a brand new data set implementation (this is not a variant, this is not a new observable of the existing data set, this is really a dataset implementation version 2,  the bugged version 1 having been used in the pheno paper). In this case, I have to confess that I don't remember what the workflow is. Is it sufficient to erase the previous implementation and have a new one (with version: 2 in the metadata) or should I create a new folder called, say `ATLAS_Z0J_13TEV_V1` or what? Thanks.

scarlehoff:
Version 2 in the metadata, comment: fix to bug in version 1. If you want to keep the bug as a variant you can do that.

enocera:
> Version 2 in the metadata, comment: fix to bug in version 1. If you want to keep the bug as a variant you can do that.

Sorry @scarlehoff : are you sure that I can keep the variant? The variant will also differ for the kinematics (the number of pT bins has changed).

scarlehoff:
Ah, no, in that case no. And  then also fktables will differ.

I would just replsce it. The pheno paper version lives in the tags of the repo.

enocera:
Very good. That's what I wanted to hear.

PR #2361, opened 2025-08-24
CMS_2JET_13TEV

enocera:
This PR incorporates the commondata implementation for the data set in the title.


enocera:
@scarlehoff As far as I understand, Alex Huss has computed grids for this data set, see https://arxiv.org/pdf/2412.21165, in particular Table 1. Incidentally, the computation incorporates sub-leading colour contributions for all the di-jet data. Do you know the whereabouts of these grids (they are not on Plougshare), by chance? Should I write to Alex to request these?

achiefa (commented):
All looks good to me. I'll add a data-theory comparison with the new full color grids/fktable that @andrpie produced.

Just a general question: Is there a reason why you chose the 3d distribution differential in m_jj rather than the one differential in p_T,avg? I see that they provide both variants in HepData. On the other hand, the CMS 8 TeV 3d distribution that @andrpie implemented is differential in p_T,avg, so the process option for DIJET_3D should also accept the p_T,avg (but this is something that @andrpie can do once THIS PR is merged).

nnpdf_data/nnpdf_data/commondata/CMS_2JET_13TEV/filter.py
```diff
@@ -0,0 +1,159 @@
...
+                # Symmetrise asymmetric uncertainties
+                unc[error["label"]] = Deltabar * (1. + delta*delta / (Deltabar * Deltabar))
+
+            elif "symerror" in error.keys():
+                unc[error["label"]] = error["symerror"]
+
+            uncertainty.update(unc)
+        
+        uncertainties.append(uncertainty)
+        shifts.append(shift)
```
achiefa: Is there a reson for not using the function `symmetries_errors` from `nnpdf_data.filter_utils.utils`? Otherwise I suggest to use it as follows

```suggestion
        for error in errors:
            unc = {}
            if "asymerror" in error.keys():
                delta_plus = error["asymerror"]["plus"]
                delta_minus = error["asymerror"]["minus"]
                if delta_plus > 0 and delta_minus > 0:
                    delta_minus = 0
                elif delta_plus < 0 and delta_minus < 0:
                    delta_plus = 0
                se_delta, se_sigma = symmetrize_errors(delta_plus=delta_plus, delta_minus=delta_minus)
                # Symmetrise asymmetric uncertainties
                unc[error["label"]] = se_sigma

            elif "symerror" in error.keys():
                unc[error["label"]] = error["symerror"]

            uncertainty.update(unc)
        
        uncertainties.append(uncertainty)
        shifts.append(se_delta)
```

achiefa:
I'm trying to understand the origin of the high chi2 for the 3D distribution. I'm focusing on the first bin in yb. Here is the data-theory comparison for the shifted and unshifted predictions.

<table>
<tr>
<td>theory_data_comparison_bin_0</td>
<td>theory_data_comparison_bin_0_unshifted</td>
</tr>
</table>

I've also computed the chi2 for this particular bin in yb:

| Column 1 | Column 2 |
|----------|----------|
| Exp. chi2     | 1.403e+03 |
| Exp.+th. chi2       | 19.6   | 
| Exp.+th.+pdf chi2 | 19.17 |

I hope this helps us find the problem with this dataset.

nnpdf_data/nnpdf_data/commondata/CMS_2JET_13TEV/filter.py
achiefa: So I have recomputed the experimental chi2 using the symmetrised version. With Emanuele's version, I get 2.66, whereas with the utility function implemented in the commond data I get 2.77. The (tiny) discrepancy makes sense -- Emanuele's is the first oder Taylor approximation of the one in the commondata. I think after this we can close this chapter.

PR #2362, opened 2025-08-26
ATLAS ZpT 8 TeV uncertainty variant

enocera:
This PR implements a variant of the ATLAS ZpT data in which the 100 systematic uncertainties are replaced by two systematic uncertainties, an uncorrelated and a correlated one. Statistical, luminosity and (optional) Monte Carlo uncertainties are as in the default variants. The implementation follows the tables available on Hepdata (see metadata).


enocera:
Dear @scarlehoff thanks for checking. I would nevertheless have this variant implemented.. It'd be interesting to see what happens when we change the cut in pT and add RES K-factors.

PR #2366, opened 2025-09-12
Add N3LO pch theories

giacomomagni:


giacomomagni:
> You added them as `.tar` instead of `.tgz`. Are they really tar files or are they "gz" compressed and you just used the wrong extensions? (if it is the second, just changing the name would be enough, but to be certain)

sorry yes they were `tar.gz` but I did a wrong rename.

Issue #2375, opened 2025-09-26
Wrong thcovmat in vp-comparefits if `user_covmat` is provided


achiefa:
When running `vp-setuptfit` with the following theory covmat setting:
```
theorycovmatconfig:
  point_prescriptions: [ad ihou, dis ihou, 7 point]
  user_covmat_path: <vp_link>/<user_covmat_name>.csv 
  pdf: <fitname>
  use_thcovmat_in_fitting: true
  use_thcovmat_in_sampling: true
```
The following tables are generated
```
- datacuts_theory_theorycovmatconfig_point_prescriptions0_theory_covmat_custom_per_prescription.csv
- datacuts_theory_theorycovmatconfig_point_prescriptions1_theory_covmat_custom_per_prescription.csv
- datacuts_theory_theorycovmatconfig_point_prescriptions2_theory_covmat_custom_per_prescription.csv
- datacuts_theory_theorycovmatconfig_theory_covmat_custom.csv
- datacuts_theory_theorycovmatconfig_total_theory_covmat.csv
- datacuts_theory_theorycovmatconfig_user_covmat.csv
```
`theory_covmat_custom.csv` is the sum of the different `point_prescriptions`. `user_covmat.csv` is the covmat for which we provide the path. `total_theory_covmat.csv` is the sum of these two.

When the fit is run, `n3fit` loads the __correct__ theory covmat using [produce_loaded_theory_covmat](https://github.com/NNPDF/nnpdf/blob/b30f00fd14fe1904a80d7f88af6b5be932946fd2/validphys2/src/validphys/config.py#L851) (see that it provides the proper name for the theory covmat based on the settings).

# The issue

However, when analysing the fit using vp (_e.g._ I found this problem using `vp-comparefits`), [produce_fitthcovmat](https://github.com/NNPDF/nnpdf/blob/b30f00fd14fe1904a80d7f88af6b5be932946fd2/validphys2/src/validphys/config.py#L1280) might be called. However, as you see below, only one name is provided.

https://github.com/NNPDF/nnpdf/blob/b30f00fd14fe1904a80d7f88af6b5be932946fd2/validphys2/src/validphys/config.py#L1309

 In the case of the `theirycovmatconfig` above, we only include the `point_prescriptions`, but the `user_covmat` overlooked. Morevoer, if we were to set the `theirycovmatconfig` as follows
```
theorycovmatconfig:
  user_covmat_path: <vp_link>/<user_covmat_name>.csv 
[...]

PR #2379, opened 2025-09-26
Solving issue 2375

achiefa:
Issue #2375


achiefa:
Now I think is ready to go

PR #2381, opened 2025-10-04
correct treatment of replica number for the photon evaluation

RoyStegeman:
I got an error when trying to do a QED fit with replica number larger than the Nrep of the luxpdf. I thought I fixed this before, and have used it before, so I'm not sure why it raised the error now (couldn't find a relevant commit that would explain it either).

The PDFs being evaluated by `self.luxpdfset.xfxQ(x, self.q_in, photonreplica, pid)` are used together with the computed photon PDF to do the evolution from the photon scale to the fitting scale.  For consistency, we want the non-photon PDFs to be the ones used to compute the structure functions entering the photon computation, which is done here.


scarlehoff:
i think this last commit https://github.com/NNPDF/nnpdf/commit/6f993b1cd2b87a7c555067e4fba314b838a7a721#diff-836eaf9c89330958e0d8c9c99d10b4f573dbf38a8b5fd56e81ede192d195202bR108-R118 refactored/reverted part of it?

I guess you've already used (and tested) this PR?

RoyStegeman:
Yes used it already

PR #2383, opened 2025-10-07
Implement TTB total cross section from arXiv:{2303.15340, 1911.13204}

t7phy:


jacoterh (approved):
Looks good! Just for reference, arXiv 2303.15340 (this PR) presents the total cross section in the full phase from the dilepton channel and therefore gives smaller uncertainties than the lepton + jet channel from https://arxiv.org/abs/2006.13076 (systematics are more under control)

nnpdf_data/nnpdf_data/commondata/ATLAS_TTBAR_13TEV_2L_140FB-1_TOT/kinematics.yaml
```diff
@@ -0,0 +1,13 @@
+bins:
+- zero:
+    min: null
+    mid: 0
+    max: null
```
scarlehoff, line 5: There's no need to add a fake kinematic. Please remove the zero.

Fwiw, there's also no need to sdd sqrts or min/max.

PR #2384, opened 2025-10-08
Add the possibility of adding extra data folders with the `data_path` variable.

scarlehoff:
This should facilitate adding new data, e.g., you can have a `nnprofile.yaml` in `$HOME/.config/NNPDF` with:

```yaml
data_path:
  - /tmp/my_own_data/
```

and then you can put your data there (with the typical `<dataset>/**` structure).

In order to do this I've moved the profile reading to the `nnpdf_data` package


ecole41 (approved):
I've tested this by moving a dataset to a new path and the code works correctly. I'm happy to merge

Issue #2251, opened 2024-12-20
sum rules for hessian PDFs


RoyStegeman:
When computing the uncertainty on the sum rules for hessian PDFs the standard deviation is taken over the moments computed per eigenvector

https://github.com/NNPDF/nnpdf/blob/050d723fb25e1dae8cf47b0a024c156a730d14bb/validphys2/src/validphys/sumrules.py#L214

comane:
Hi @RoyStegeman, can you perhaps give more details on why this is a bug.

Should a different formula be used for the computation of how much the sum rules for each eigenvector of a hessian set vary?

RoyStegeman:
For symmetric eigenvectors it's the average of the central-eigenvector shifts. For assymmetric the idea is the same but you take the max of the "plus" and "minus" eigenvectors.

PR #2386, opened 2025-10-23
Using Stats for sumrule statistics

achiefa:
Addresses issue #2251.


validphys2/src/validphys/sumrules.py
```diff
@@ -17,7 +17,7 @@
 from reportengine.checks import check_positive
 from reportengine.floatformatting import format_error_value_columns
 from reportengine.table import table
-from validphys.core import PDF
+from validphys.core import PDF, MCStats, SymmHessianStats
```
scarlehoff:
```suggestion
from validphys.core import PDF
```

validphys2/src/validphys/sumrules.py
```diff
@@ -206,14 +206,21 @@ def unknown_sum_rules(pdf: PDF, Q: numbers.Real):
...
-        d["std"] = np.std(arr)
-        d["min"] = np.min(arr)
-        d["max"] = np.max(arr)
+    stats_class = pdf.stats_class
+    for k, arr in data.items():
+        res[k] = stats_dict = {}
+        # Each entry in `data` is expected to be a vector with shape
+        # (central_member + error_members,), hence we reshape to (-1, 1) before
+        # passing to the stats class.
+        stats = stats_class(arr.reshape(-1,1))
```
scarlehoff: This works for sure, but doesn't `stats_class` do this automatically?

achiefa: I think this is what the following code snippet is meant to do
https://github.com/NNPDF/nnpdf/blob/7c3c42b240fef7c4a26088fe3327f08af4ea4201/validphys2/src/validphys/core.py#L841-L843
The problem is that it reshape `(size,)` into `(1,size)`, while we actually need `(size,1)`. This probably needs to be changed in the `Stats` class (?)

achiefa: Do you want me to make this change?

Edit: Actually we could simply transpose `self.data` in the initialiser.

andrpie (approved):

PR #2387, opened 2025-10-23
Excluded dataset page for vp-comparefits

jekoorn:
PR for the excluded (positivity) datasets in a vp-comparefits report.
This page is added if and only if the theory id, datacuts, and theory covmat (if present) are identical between fits.

### DIFF
--- a/validphys2/src/validphys/config.py
+++ b/validphys2/src/validphys/config.py
@@ -36,7 +36,7 @@ from validphys.filters import (
     default_filter_rules_input,
     default_filter_settings_input,
 )
-from validphys.fitdata import fitted_replica_indexes, num_fitted_replicas, match_datasets_by_name
+from validphys.fitdata import fitted_replica_indexes, match_datasets_by_name, num_fitted_replicas
 from validphys.gridvalues import LUMI_CHANNELS
 from validphys.loader import (
     DataNotFoundError,
@@ -1006,18 +1006,37 @@ class CoreConfig(configparser.Config):
         """
         Like produce_matched_datasets_from_dataspecs, but for mismatched datasets from a fit comparison.
         Returns the mismatched datasets, each tagged with more_info from the dataspecs they came from. Set up to work with plot_fancy.
+
+        Datasets are considered a mismatch if the name is different and if the variant is different.
         """
 
         self._check_dataspecs_type(dataspecs)
 
-        mismatched_dinputs = []
+        # Parse the data for the comparison so that only variant and dataset are actually tested
+        parsed_data = []
         for spec in dataspecs:
-            for dinput in spec["dataset_inputs"]:
-                # check whether the dataset exists in any of the other dataspecs
-                if any( dinput not in s["dataset_inputs"] for s in dataspecs ):
-                    # append (proc, ds) -> list of (dinput, spec)
-                    mismatched_dinputs.append( (dinput, spec) )
-        
+            tmp = [(i.name, i.variant) for i in spec["dataset_inputs"]]
+            parsed_data.append((spec, tmp))
+
+        # TODO:
+        # This is a convoluted way of checking whether there are mismatches
+        # between the lists of dataset inputs of a list of specs.
+        # This is not going to win any codegolf tournaments
+        already_mismatched = []
+        mismatched_dinputs = []
+        for spec, parsed_dinputs in parsed_data:
+            for spec_to_check, parsed_dinputs_to_check in parsed_data:
+                if spec == spec_to_check:
+                    continue
+                for i, parsed_dinput in enumerate(parsed_dinputs):
+                    # Use a list of already mismatched data to avoid duplicates
+                    if parsed_dinput in already_mismatched:
+                        continue
+                    if parsed_dinput not in parsed_dinputs_to_check:
+                        dinput = spec["dataset_inputs"][i]
+                        mismatched_dinputs.append((dinput, spec))
+                        already_mismatched.append(parsed_dinput)
+
         res = []
         # prepare output for plot_fancy
         for dsin, spec in mismatched_dinputs:
@@ -1029,7 +1048,7 @@ class CoreConfig(configparser.Config):
                     "pdfs": [i["pdf"] for i in dataspecs],
                     "fit": spec["fit"],
                 }
-                    ) 
+            )
         return res
 
     def produce_matched_positivity_from_dataspecs(self, dataspecs):
--- a/n3fit/runcards/examples/developing.yml
+++ b/n3fit/runcards/examples/developing.yml
@@ -81,7 +81,7 @@ parameters: # This defines the parameter dictionary that is passed to the Model 
   stopping_patience: 0.1 # percentage of the number of epochs
   layer_type: 'dense'
   dropout: 0.01
-  interpolation_points: 7
+  feature_scaling_points: 7
 
 fitting:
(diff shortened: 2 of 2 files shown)


validphys2/src/validphys/scripts/vp_comparefits.py
```diff
@@ -240,7 +243,29 @@ def get_config(self):
...
         c.update(self.complete_mapping())
         return self.config_class(c, environment=self.environment)
+    
+    def check_identical_theory_cuts_covmat(self):
+        args = self.args
+        l = self.environment.loader
+        resultspath = l.resultspath
+        
+        current_runcard_path = str(resultspath) + "/" + str(args['current_fit']) + "/filter.yml" 
+        reference_runcard_path = str(resultspath) + "/" + str(args['reference_fit']) + "/filter.yml"
```
scarlehoff: Since at this point you already have the loader, you can use `l.check_fit(fitname)` to get a reference to the fit (and with `fitname.as_input()`... I think, you get already most of the information you need).

The relevance of using `.check_fit` instead of looking directly at the path is that at this point the fit might not have been downloaded yet. `.check_fit` will ensure that they are downloaded and available (and at that point you will get a reference to them, so no need to generate yourself the reference or the load).


for future reference, when you are getting a path you can do directly:
```suggestion
        reference_runcard_path = resultspath / str(args['reference_fit']) / "filter.yml"
```
and so is python who will build the path instead of you. It's not that important for us but this line wouldn't work in systems for which the path is `\` .

validphys2/src/validphys/scripts/vp_comparefits.py
```diff
@@ -184,6 +183,11 @@ def get_commandline_arguments(self, cmdline=None):
...
         else:
             args['config_yml'] = comparefittemplates.template_path
+
+        #if args.get('current_fit') and args.get('reference_fit'):
+        #    if self.check_identical_theory_cuts_covmat(args):
+        log.info("Using excluded comparecard: identical theory cuts/covmat detected.")
+        args['config_yml'] = comparefittemplates.template_with_excluded_path
         return args
 
     def complete_mapping(self):
```
scarlehoff: This was almost done!

You just need to put the conditional here (well, at the end of this function)

```
are_the_same = self.check_identical_theory_cuts_covmat()
if are_the_same:
    log.info("Using excluded comparecard: identical theory cuts/covmat detected.")
    autosettings["template"] = "report_with_excluded.md"
```

These `autosettings` of `complete_mapping` is what is used to complete the template afterwards.
In this particular case (in which you are just changing the `template` but would be using the exact same runcard for everything else) I would not add a new `comparecard_excluded.yaml` but just the `report_with_excluded.md`

scarlehoff (commented):
Thanks. Seems to work fine (see img below)

Screenshot 2025-11-13 at 22 06 23

The two comments are 1) the necessary change to vp_comparefits to make it work, and 2) a comment on what we meant by excluded. Other than that this is good ^^

(btw, I understand the frustration when dealing with reportengine, I was there many moons ago, but you are navegating it quite well so don't get discouraged! After a while you learn to appreciate the guardrails and tools it offers, I promise!)

jekoorn:
Hi Juan, I still don't think I understand how to go from "returning a list of (pos)dataset names to actually generating plots in the report. 
With a rule like:

```
Positivity excluded from fit
--------------------------
{@with matched_excluded_positivity_from_dataspecs@}
{@posdataset_name@}
{@endwith@}
```
I can generate a list of dataset names to be included, as before. But I just don't understand why replacing `posdataset_name` with `plot_positivity` doesn't give me a plot. What exactly is required for this plot_positivity? Or should I use another function to plot? I tried also `plot_dataspecs_positivity`, but since an excluded dataset is only in one of both datasets, copying the methodology for the "included" plots will not work.
I wonder what you did to get an actual plot in your screenshot above^^

validphys2/src/validphys/config.py
```diff
@@ -995,13 +993,87 @@ def produce_matched_datasets_from_dataspecs(self, dataspecs):
...
+                for dsin in data_input:
+                    cd = self.produce_commondata(dataset_input=dsin)
+                    proc = get_info(cd).nnpdf31_process
+                    ds = dsin.name
+                    key = (proc, ds)
+
+                    # add (dsin, spec) to the set containing (proc, ds)
+                    if key not in all_sets:
+                        all_sets[key] = []
+                    all_sets[key].append((dsin, spec))
```
scarlehoff: What are you exactly doing here? If you are just checking the `dataset_input` from the `dataspecs` by name you should not need a lot of this.

Note that you are trying to reproduce `produce_matched_datasets_from_dataspecs` but what you _really_ want is `produce_matched_excluded_dataset_inputs_by_name`. And what you are really doing, this can be simplified quite a bit.

validphys2/src/validphys/fitdata.py
```diff
@@ -282,6 +282,35 @@ def t0_chi2_info_table(pdf, dataset_inputs_abs_chi2_data, t0pdfset, use_t0):
...
+#                names[(proc, ds)] = dsin
+#            all_used.append(names)
+#
+#    used_union = {ds for d in all_used for (_, ds) in d.keys()}
+#    excluded_set = implemented - used_union
+
+#    datasetcomp = match_datasets_by_name()
+
+
+#    return excluded_set
```
scarlehoff: Please remove this.

validphys2/src/validphys/scripts/vp_comparefits.py
```diff
@@ -223,8 +223,13 @@ def complete_mapping(self):
                     'unpolarized_bc': {'from_': 'positivity_bound'},
                 }
             )
+        are_the_same = self.check_identical_theory_cuts_covmat()
+        if are_the_same:
+            log.info("Adding mismatched datasets page: identical theory cuts/covmat detected")
+            autosettings["template"] = "report_mismatched.md"
```
scarlehoff: Instead of changing the whole template (and then needing a whole new report) you can change directly the mismatched_report part.

e.g., if `are_the_same` is `True` then the `template` part under `mismatched_report` stays, otherwise you drop it and change it to `template_text: "Difference settings, mismatched page cannot be added".

jekoorn:
Hi, I am now going through the failed checks, that I managed to reproduce locally. In particular, a plot_fancy plot fails because it doesn't look similar enough. See here the [baseline](https://www.nikhef.nl/~jkoorn/baseline.png), the [result](https://www.nikhef.nl/~jkoorn/result.png) and the [failed difference](https://www.nikhef.nl/~jkoorn/result-failed-diff.png). The only difference is a small lack of space below the y=0 axis.
I think that the reason why this happens is because of the fact that I needed to rebuild the limits of plot_fancy  https://github.com/NNPDF/nnpdf/pull/2387/files#diff-2257204401799e2ee4b8c681ae348465ff3c7c1f968116e33c3cee332ecefa6b to ensure that some current/reference predictions aren't out of the plot range. See e.g. [this plot](https://www.nikhef.nl/~jkoorn/mismatched_information_mismatched_datasets_by_name7_mismatched_report_plot_fancy_0.png), that is produced when I do not rebuild the limits. This is obviously not good, since the datapoints do not show up at all. 
I'm not sure how to proceed. Should I add a small tolerance around the rebuilt limits that I added in plot_fancy? Or keep as-is, since this is not a change that affects anything imo. Thanks.

PR #2388, opened 2025-10-27
Z plus jet (c/b) commondata

kamillaurent:
Commondata implementation for Z + b-jet and Z + c-jet production. The implemented observables are cross sections differential in P_T of Z, and differential in leading b(c)-jet. The implementation is divided in two folders: `ATLAS_Z1C_13TEV` and `ATLAS_Z1B_13TEV`


enocera:
Dear @kamillaurent for now you should not do anything with the correction factors, just leave the Hepdata tables in the rawdata folder. When we will compare the data to the theory we will  wonder whether it may make sense to correct the theory by means of supplementary K-factors. Thanks.

PR #2391, opened 2025-10-28
Penalty term in hyperotp figure of merit

tgiani:
@scarlehoff @Radonirinaunimi cold you please heck if the idea here is correct?


scarlehoff (approved):
This has had already O(1000) trials of testing, I think we can merge.

PR #2392, opened 2025-10-30
new theory cards for the production of QED FK tables for nlo, nnlo

vschutze-alt:


nnpdf_data/nnpdf_data/theory_cards/41041001.yaml
```diff
@@ -0,0 +1,42 @@
+ID: 41_041_001
+Comments: NNPDF4.1 NNLO QCD-only
```
ecole41:
```suggestion
Comments: NNPDF4.1 NNLO QCD QED
```

nnpdf_data/nnpdf_data/theory_cards/41041002.yaml
```diff
@@ -0,0 +1,42 @@
+ID: 41_041_002
+Comments: NNPDF4.1 NLO QCD QED
```
felixhekhorn:
```suggestion
Comments: NNPDF4.1 NLO QCD, NLO QED
```

nnpdf_data/nnpdf_data/theory_cards/41041002.yaml
```diff
@@ -0,0 +1,42 @@
...
+PTO: 1
+FNS: FONLL-C
```
felixhekhorn:
```suggestion
FNS: FONLL-B
```
eeehm, okay, I'll take the challenge to explain FONLL in brief :see_no_evil: 
1. the long explanation: read [our new FONLL paper](https://inspirehep.net/literature/2818030) and then take a look to Sec. 3.3 of [the original FONLL paper](https://arxiv.org/pdf/1001.2312)
2. the short explanation: FONLL combines fixed order ("FO") and resummed ("NLL") computations. The perturbative accuracy of the FO part is a parameter, as well as the logarithmic accuracy of the resummed part (so even though it is called "NLL" = next-to-leading log, we can in fact do it at any logarithmic order). The letter X in "FONLL-X" specifies those two accuracies.
   - for NLO PDF fits we could use "FONLL-A", but in practice we use "FONLL-B"
   - for NNLO PDF fits we always use "FONLL-C"
   - for N3LO PDF fits we use "FONLL-E" - remember #2385

felixhekhorn (changes requested):

nnpdf_data/nnpdf_data/theory_cards/41041002.yaml
vschutze-alt: Thank you so much!

Issue #2368, opened 2025-09-17
Implementation of future test data


enocera:
The future test data sets, listed on the [wiki](https://www.wiki.ed.ac.uk/spaces/nnpdfwiki/pages/432525241/List+of+new+data+sets+for+NNPDF4.1), are assigned as follows:

|dataset|c.m.e. (GeV)|L [fb-1]|arXiv|person in charge|
|:----------|:----------------:|:--------:|:------:|:-----------------------:|
|W-boson production differential cross sections|13|140|2502.21088|EC + JB|
|W+jet production differential cross sections|13|140|2412.11644|KL + AP |	
|Z+charm production differential cross sections|13|140|2403.15093|TG + KL|	
|W-boson production differential cross sections (low pile-up)|5.02/13|0.255/0.338|2502.09403|JB + AP|	

As suggested by @juanrojochacon two new PhD students (+ Ella) are assigned to each data set. The idea is that they independently implement the data set, and then they converge on a common interpretation/implementation. This should be useful for them to be acquainted with the NNPDF commondata framework. @vschutze-alt : I have not assigned you any specific data set because you no longer qualifies as "new" PhD student, but you're more than welcome to help implement any of the data sets, if you'd like to do so. @ecole41 I have assigned only one data set to you. If you want more, just ask. The last data set in the table does not have yet a Hepdata entry.

juanrojochacon:
Thanks @enocera this is most useful. Sounds like a great plan to me. We should also agree on a timescale to avoid that these things take too long - maybe we can say that by **1 Oct** or so there is a first version of the common data implementation that can be cross-checked among the two assigned people?

For Z+charm, maybe @vschutze-alt can also help to speed up things?

Once commondata is implemented, the next step will be to generate NNLO grids, and also here I would adopt the principle that two people do an independent production and then they cross-check each other, in the long term this will save a ton of work.

I understand that progress in future test datasets will be discussed at the code meetings?

juanrojochacon:
@jekoorn @kamillaurent let me know if anything is unclear, happy to discuss with you about the implementation process in the next days

juanrojochacon:
Concerning the ATLAS 5 TeV and 13 TeV low-PU cross-sections: despite what Eram said in morimondo, indeed it is still not in in HepData. I will write to him how to see if he can speed up things.

enocera:
[...]

PR #2398, opened 2025-11-06
Implementation of LHCB WPWM 5.02TeV

andrpie:
Implementation of the [2509.18817](https://arxiv.org/abs/2509.18817) future test dataset as per #2368.

The implementation and methodology was discussed with @kamillaurent on a call.

Points worth mentioning:
- implemented two observables: differential cross functions in muon transverse momentum, one for W+, another for W-. Not sure whether "DIFWP", "DIFWM" are good choices for observable names
- HEPData files are rather unconventionally named - instead of table numbers, there are names that only differ by sign: "+", "-". Thus the table fields in metadata.yaml contain these signs. 
- just to double check: is "DY_CC_PT" the right process type for both of the observables?
- I've noticed that some other implementations translate pb to fb - is it necessary?
- The luminosity uncertainty of 2% is implicitly mentioned in the paper and added to the central values. It's classified as MULT, LHCBLUMI5P02TEV (correlated across the experiment)


enocera:
> Points worth mentioning:
> 
>     * implemented two observables: differential cross functions in muon transverse momentum, one for W+, another for W-. Not sure whether "DIFWP", "DIFWM" are good choices for observable names

That's fine with me. Perhaps slightly better `DIF_WP` and `DIF_WM`, though I have no strong preferences.

>     * just to double check: is "DY_CC_PT" the right process type for both of the observables?

Yes.
> 
>     * I've noticed that some other implementations translate pb to fb - is it necessary?

In principle it is not. Because we want to match data to FK table predictions, the unity of measure is dictated by the default unity with which grids (and FK tables) were generated. So for the time being I would not convert pb to fb. We can reassess this when we will have theoretical predictions.

>     * The luminosity uncertainty of 2% is implicitly mentioned in the paper and added to the central values. It's classified as MULT, LHCBLUMI5P02TEV (correlated across the experiment)

Very good. We don't have any other LHCb measurements at 6.02 TeV, so you have total freedom in the choice of the label for the correlated luminosity uncertainty.

PR #2400, opened 2025-11-10
Add theory cards for mc variation

andrpie:


nnpdf_data/nnpdf_data/theory_cards/40022000.yaml
```diff
@@ -0,0 +1,41 @@
+ID: 40_022_000
+Comments: Same as 40_022_003, but alphas=0.115
+PTO: 2
+FNS: FONLL-C
+DAMP: 0
+IC: 1
+Q0: 1.65
+ModEv: TRN
```
jacoterh: We want EXA evolution everywhere rather than truncated

PR #2401, opened 2025-11-11
Make the usage of feature scaling + large x a bit clearer

### LINKED ISSUE #2377: Feature scaling in the run card
Hi @RoyStegeman as we discussed at the phone conf, we would like to find a way to select feature scaling in the run card which also enables to go back to the previous parametrisation, to check the effects of this variation everything else fixed. 

In NNPDF4.0 we parametrise PDFs as

f(x) = A x^alpha * (1-x)^beta * NN(x, log(x))

and in the run card we specify the values of the ranges for the small-x and large-x exponents alphas and beta.

Now in NNPDF4.1, we want to use feature scaling, with basically tells us that the PDFs are being parametrised as

f(x) = A x^alpha ( NN(x) - NN(x=1) )

where x in the NN input is distributed as explained in the FS paper. In this case, one does not need to indicate the large-x exponents in the run card.

So looking at the current run card and with the motivation to maintain backwards compatibility, it would be nice that

- [ ] have an option in the run card that more transparently allows the user to change between one option and the other (the current option is not very transparent), better to have some more clear selection

- [ ] If one uses FS, the code stops if the run card also lists large-x exponents

Does it make sense? Thanks in advance!

scarlehoff:
Closes #2377 

This PR deals with the clarity issues pointed out by #2377

1.  Allows the user to drop completely large-x from the runcard
2. If feature scaling is enabled, the iterated runcard will have nothing for large x (instead of having something that is then ignored completely by the fit)
3. Changes `interpolation_points` to `feature_scaling_points` so that it is clear when it is set to `False` or not set at all.

@kamillaurent @juanrojochacon let me know whether this is what you had in mind. This changes nothing on how the fit is run, right now the fit will just drop silently the `large-x` if feature scaling is set, now instead if you have large-x set it will refuse to run until you remove it.

### DIFF
--- a/validphys2/src/validphys/eff_exponents.py
+++ b/validphys2/src/validphys/eff_exponents.py
@@ -331,8 +331,11 @@ def previous_effective_exponents_table(fit: FitSpec):
     )
     basis = checked["basis"]
     flavours = checked["flavours"]
-    prev_a_bounds = [runcard_fl['smallx'] for runcard_fl in fitting["basis"]]
-    prev_b_bounds = [runcard_fl['largex'] for runcard_fl in fitting["basis"]]
+    prev_a_bounds = []
+    prev_b_bounds = []
+    for runcard_fl in fitting["basis"]:
+        prev_a_bounds.append(runcard_fl.get("smallx", (0.0, 0.0)))
+        prev_b_bounds.append(runcard_fl.get("largex", (0.0, 0.0)))
     # make single list alternating alpha and beta bounds
     data = [vals for pair in zip(prev_a_bounds, prev_b_bounds) for vals in pair]
     flavours_label = [f"${basis.elementlabel(fl)}$" for fl in flavours]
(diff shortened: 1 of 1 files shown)


kamillaurent:
@scarlehoff I tried to run the runcard in the branch `make_feature_scaling_clearer`. I then want to compare the fit to our baseline, `251124-jcm-nnpdf41-001`, but I obtain this error by validphys:

```
  File "/data/theorie/klaurent/nnpdf/validphys2/src/validphys/eff_exponents.py", line 335, in previous_effective_exponents_table
    prev_b_bounds = [runcard_fl['largex'] for runcard_fl in fitting["basis"]]
                     ~~~~~~~~~~^^^^^^^^^^
KeyError: 'largex'

scarlehoff:
> the same error message appears also if I try to compare the fit done with feature scaling and the same fit iterated,

That should not have happened instead!

> Let me know if you have other checks to suggest

The important one is actually that you iterated the fit (so `largex` was dropped) and were able to run (so it was, indeed, dropped).

But this PR was also about making the feature scaling part clearer so that's the important part. If you think it is indeed clearer, then good, otherwise (if the previous version seem clearer) better to roll it back.

PR #2403, opened 2025-11-12
fix positivity plot x label

jekoorn:
Implement the correct kinematic variable to be plotted (x-axis) on the positivity plots, so long as the option `pos_use_kin: True` is selected in the report.

See an example [report](https://vp.nnpdf.science/wh_aquiuRKmspRt2zLFU7w==/ )

### DIFF
--- a/validphys2/src/validphys/dataplots.py
+++ b/validphys2/src/validphys/dataplots.py
@@ -1252,7 +1252,7 @@ def plot_obscorrs(corrpair_datasets, obs_obs_correlations, pdf):
 
 
 @figure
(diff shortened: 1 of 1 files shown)


scarlehoff:
Thanks.

But... why don't we use this opportunity to remove this `pos_use_kin` and just plot always the kinematic variable? I've done a bit of archeology out of curiosity and turns out there's no reason why we have this idat thing anymore really ^^U https://github.com/NNPDF/nnpdf/pull/122

juanrojochacon (approved):

PR #2404, opened 2025-11-14
Disable tensorfloat32 type

scarlehoff:
Apparently the "new math" is just being sloppy as hardware instructions: https://blogs.nvidia.com/blog/tensorfloat-32-precision-format/

Note that the above is for GPU (or TPUs) but newer CPUs might be affected by the same issue in case we see it again:
https://www.intel.com/content/dam/develop/external/us/en/documents/lower-numerical-precision-deep-learning-jan2018-754765.pdf

For context here are the results, `251105-jcm-nnpdf41-mhou-001` is the new baseline and these are a few comparisons with a computer with this problem:

a) Default options, tensorfloat 32 is used for some operations with the consequence of a loss of precision
IMAGE 2025-11-14 14:14:30

b) `double_precision: true` in `n3fit`. By setting the default type to float64, only _some_ (not clear to me which) operations are downgraded to these tensorfloat types, but everything looks ok

IMAGE 2025-11-14 14:15:40

c) And finally comparison to a fit in which the tensorfloat is disabled and the entire fit is done in single precision, results are ok
IMAGE 2025-11-14 14:09:56

I'll benchmark cases b) and c) to check what is faster in cineca before merging this branch. Or if you prefer we can merge it asap and if during the benchmark we find that option b) is better (run by default in dp but keep these special features on) we can change the default.

NB: I'm only disabling it in TensorFlow, although in principle this is a hardware thing. However, apparently pytorch is more conservative in which operations have it on by default (perhaps that's why it is much slower...) and the results are actually ok. I haven't tested jax.

### DIFF
--- a/.github/actions/prepare_environment/action.yml
+++ b/.github/actions/prepare_environment/action.yml
@@ -19,7 +19,7 @@ inputs:
   python-version:
     required: true
     description: "Python version to use"
-    default: "3.12"
+    default: "3.13"
 
 runs:
   using: "composite"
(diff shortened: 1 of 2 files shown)


Radonirinaunimi (approved):
Tested and it works fine for me as well! Everything looks good.

tgiani:
@scarlehoff do you mean that b) makes runs 40% slower?

PR #2405, opened 2025-11-14
Add positivity data along W^2=10 GeV^2 line

kamillaurent:
This PR is to update positivity dataset for gluon, I added new datapoints along the W^2=10 GeV^2 line in `kinematic_XGL.yaml` and updated `metadata.yaml`. The baseline fit runcard does not need to be changed, since the dataset name is the same. @scarlehoff let me know if you have any question


kamillaurent:
Ok, I rebased on master and removed the extra runcard, let me know if any other action is required

Issue #2390, opened 2025-10-28
Implementation of data sets excluded from NNDPF4.0


enocera:
This is a list of experimental data that was excluded from the NNPDF4.0 baseline fit according to our selection criteria. In view of NNPDF4.1, we would like to revisit whether these data sets should still be excluded or not from the fit. The NNPDF4.1 methodology and the default theoretical settings (e.g. MHOUs, higher-twist corrections, etc.) are going to change w.r.t. NNPDF4.0, hence the selection of data may change.

Some of these data sets are already implemented in commondata, whereas others are not. I suggest to implement them ASAP. Data sets are listed in decreasing order of interest. I suppose that we need a legacy version and a default version for each of them.

|Dataset|Reference|In commondata|Person in charge|PR|
|----------- |:--------------:|:-----------------------:|:-------------------------:|:----:|
|CMS_2JET_8TEV_3D|[arXiv:1705.02628](http://arxiv.org/abs/1705.02628)|no|||
|ATLAS_1JET_7TEV_R06|[arXiv:1410.8857](http://arxiv.org/abs/1410.8857)|no|||
|CMS_1JET_7TEV_R07|[arXiv:1406.0324](http://arxiv.org/abs/1406.0324)|no|||
|ATLAS_WPWM_8TEV|[arXiv:1904.05631](http://arxiv.org/abs/1904.05631)|no|||
|D0_WPWM_1P96TEV_EM|[arXiv:1412.2862](http://arxiv.org/abs/1412.2862)|no|||
|LHCB_WPWM_8TEV_ELECTRON|[arXiv:1608.01484](http://arxiv.org/abs/1608.01484)|no|||
|ATLAS_PH_8TEV|[arXiv:1605.03495](http://arxiv.org/abs/1605.03495)|yes|n/a|n/a|
|ATLAS_TTBAR_8TEV_LJ_DIF (pTt and mtt)|[arXiv:1511.04716](http://arxiv.org/abs/1511.04716)|yes|n/a|n/a|
  
cc @scarlehoff

PR #2407, opened 2025-11-27
New commondata format for `ATLAS_1JET_7_TEV_R06`

achiefa:
This PR addresses #2390.

**EDIT** Since the version with R=0.4 is also available on Hepdata, I'm going to implement it.

Details
=====
Implementation of the double-differential distribution of the single-inclusive jet cross sections in proton-proton collisions measured with the ALTAS detector at $\sqrt{s}=7$ GeV. The double-differential cross sections are given as functions of th e the jet rapidity $y$ and the jet transverse momentum $p_T$. The energy interval covered by the jet momentum goes from $100$ GeV up to $2$.
- N. of bins: 140
- Distribution: $d\sigma / dy dp_T$
- C.o.m. energy: 7 TeV
- Energy range: from 100 GeV to 2 TeV

Comments
----------
This dataset was excluded from the NNPDF4.0 baseline dataset according to the selection criteria employed at the time.

Variants
-------
Other than the nominal version of the JES systematic uncertainties, which is used ad default variant in the present implementation, the dataset comes with two additional versions of the JES components. They are constructed by taking *stronger* and *weaker* correlations between the JES components with respect to the nominal uncertainties. Both versions are implemented in the present implementation as variants with names `stronger` and `weaker` respectively.

### DIFF
--- a/nnpdf_data/nnpdf_data/commondata/ATLAS_1JET_7TEV/filter.py
+++ b/nnpdf_data/nnpdf_data/commondata/ATLAS_1JET_7TEV/filter.py
@@ -48,9 +48,6 @@ def filter_ATLAS_1JET_7TEV_data_kinematics(observable: str = 'PTY-R06'):
         rap['min'], rap['max'] = rapidity_interval[0], rapidity_interval[1]
         rap['mid'] = 0.5 * (rap['min'] + rap['max'])
 
-        # center of mass energy
-        sqrts = float(input['dependent_variables'][0]['qualifiers'][4]['value'])
-
         # transverse momentum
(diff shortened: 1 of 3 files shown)


achiefa:
Just a comment:

In addition to the nominal JES uncertainties, the HEPDATA tables provide two other variants that differ in the way the experimental components of JES uncertainties are correlated. These have "stronger" and "weaker" correlation assumptions, respectively, with respect to the nominal configuration. I can generate variants for both, in case they are needed. @enocera

enocera:
Yes please. These were implemented in the old commondata implementation.

PR #2411, opened 2025-12-16
Implementation of ATLAS WPWM 8TeV - excluded from 4.0

vschutze-alt:


scarlehoff (commented):
For this one the grids are already available: https://github.com/NNPDF/theories_slim/pull/64
so you can add the theory to the commondata file (and even create the fktable and test the implementation!)

juanrojochacon:
Well we should be able to know exactly what goes on into the plots, else we are in trouble.

As far as I understand, the shifted data vs theory comparison plots include only PDF uncertainties. 

The value Exp chi2: 132.5 is consistent with the plot, but not those of the chi2 once MHOU and PDF errors are accounted for. Considering only central PDF errors brings the chi2 down to 1, as @enocera indicates.

I would be very surprised if accounting for PDF corrections makes the chi2 increase so dramatically.

So there is something that we need to understand here.

enocera:
I have recomputed the chi2 by regularising the experimental covariance matrix. The plots are unchanged, as it should, and I get the following values of the chi2:
Exp chi2: 1.834
Exp+th chi2: 1.504
Exp+th+pdf chi2: 1.27
In my opinion, these values are in line with what we observe at the level of the shifted plots.
For comparison, I attach the unshifted plot, where you can clearly see that there is a normalisation shift between data and theory.
plot_Wp
plot_Wm
So, all in all, I think that this falls in the usual category of data sets for which we are overcorrelating systematic uncertainties. Some work may be needed to understand which ones we are actually overcorrelating. Interestingly, on Hepdata there is a another version of the data set in which some of the uncertainties are combined. It may be useful to implement this version as a variant and see if we can get a better chi2 without regularisation.

enocera:
And just to complete the picture, here are the data-theory comparison plots including theory uncertainties (the blue bars are the sum in quadrature of the PDF uncertainty and of the diagonal elements of the MHOU covariance matrix)
plot_Wp
plot_Wm
Of course, the dominant MHOU component to the blue bar is strongly correlated across data points, hence the chi2 drops only by a moderate, and very reasonable, amount when including MHOU uncertainties (1.8->1.5)

enocera:
If one makes the luminosity uncertainty uncorrelated, something that is of course unphysical, one gets
Exp chi2: 1.935
Exp+th chi2: 0.1147
Exp+th+pdf chi2: 0.1117
This is peculiar, and very similar to what was observed in the companion Z measurement, see https://github.com/NNPDF/nnpdf/pull/2270. If I had to make a bet, I would say that the experimentalists got the normalisation of the central value of their data sets wrong.

enocera:
So, all in all, I don't see anything wrong in the implementation of this data set. The unshifted plots reveal a ~2.5\% normalisation shift, which is very close to the luminosity uncertainty, so there's no surprise that when one decorrelates an uncertainty of the same order of the shift the chi2 drops to about one. My suggestion is to merge this PR. The data set will then be excluded at the level of the data set selection.
cc @scarlehoff

scarlehoff (approved):
Ok, let's merge then so that we can exclude it happily later.

PR #2413, opened 2026-01-08
solve bug in evolven3fit

evagroenendijk:


felixhekhorn:
Can we please fix this deeper?
- remove `q_fin` from `construct_eko_photon_cards` - it is not used there: see [here](https://github.com/NNPDF/nnpdf/blob/6e071a8cb7fdc8fb92e1b03d79d1d49f72323563/n3fit/src/evolven3fit/eko_utils.py#L134)
- drop `q_fin` and `q_points` from the CLI

Can we please add a unit test? I'm surprised this has not come up earlier, because I suspect @jacoterh found the bug by just running the real life thing

evagroenendijk:
> Can we please fix this deeper?
> 
> * remove `q_fin` from `construct_eko_photon_cards` - it is not used there: see [here](https://github.com/NNPDF/nnpdf/blob/6e071a8cb7fdc8fb92e1b03d79d1d49f72323563/n3fit/src/evolven3fit/eko_utils.py#L134)
> * drop `q_fin` and `q_points` from the CLI
> 
> Can we please add a unit test? I'm surprised this has not come up earlier, because I suspect @jacoterh found the bug by just running the real life thing

@felixhekhorn thanks for the comment, I completely agree! I was indeed looking at your first point, the second I’ll try later. 

About the unit test, I tried to add something like it before, but did not add it eventually because it takes a long time even with few x points at LO (it has to produce a whole eko). Do we want to run that every time, or are you thinking of something else?

n3fit/src/n3fit/tests/regressions/quickcard_qed_3.json
```diff
@@ -51,22 +51,22 @@
...
+    1.047891586326594
   ],
   "integrability": [
-    0.03370336024090587,
-    0.00010960175177321396,
-    0.00019358846930050966,
-    0.06445395993068814,
-    0.03255362482741475
+    0.03370335418731024,
+    0.0337033541873098,
```
evagroenendijk: @scarlehoff Can I ask something? I regenerated these quickcards because they were not passing the regression tests anymore. But to be completely honest I don't understand 100% what these values are. The value that gave the regression error was the second of the integrability ones.

Now there is no problem anymore, but I wanted to check if I did not do it wrong

Thanks!

evagroenendijk: Ok, thanks a lot! Yes, my computer is a mac. Each time it was exactly the same value that differed though, is that not a bit suspicious?

And I am just testing the eko production on the cluster now, once it's done I'll rebase etc!

felixhekhorn (approved):

PR #2414, opened 2026-01-09
Fix for ruamel update

scarlehoff:
`ruamel.yaml` has updated to version 0.19, changes a few things internally, but most importantly there's a change of library that makes it incompatible with some LHAPDF info files.

I don't know whether the reason is they are badly formatted or a bug in ruamel, before investigating that I've just modified the dependences to install the old parser which should be ok for now. 

This is the reason for the recent errors in the CI e.g.,. https://github.com/NNPDF/nnpdf/actions/runs/20849541813/job/59901073339?pr=2393#step:4:443

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,7 +63,7 @@ matplotlib = "^3.9"
 pineappl = "^1.0.0"
 pandas = "*"
 numpy = "*"
-"ruamel.yaml" = "*"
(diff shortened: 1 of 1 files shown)


achiefa (approved):
Hi @scarlehoff, thanks for this. It looks good to me for now. Maybe we can open an issue just to remind ourselves that we need to make the NNPDF codebase compatible with the new ruemel version.

PR #2420, opened 2026-01-20
dataset excluded from NNPDF4.0: CMS_1JET_7TEV_R07

andrpie:
This is the initial implementation.
- the observable chosen is inclusive jet cross section with jet radius parameter R=0.7. There is another observable available, the same thing but with R=0.5. Should I implement both?
- for each bin, there is a nonperturbative correction factor that ensures a data-theory agreement. Should these also be implemented?

### DIFF
--- a/nnpdf_data/nnpdf_data/commondata/CMS_1JET_7TEV/metadata.yaml
+++ b/nnpdf_data/nnpdf_data/commondata/CMS_1JET_7TEV/metadata.yaml
@@ -35,9 +35,15 @@ implemented_observables:
     - uncertainties_R07.yaml
   theory:
     FK_tables:
-      - - CMS_1JET_7TEV_R07
+      - - CMS_1JET_7TEV_R07_BIN1
+        - CMS_1JET_7TEV_R07_BIN2
+        - CMS_1JET_7TEV_R07_BIN3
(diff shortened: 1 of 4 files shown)


nnpdf_data/nnpdf_data/commondata/CMS_1JET_7TEV/metadata.yaml
```diff
@@ -0,0 +1,43 @@
...
+    description: Double differential inclusive jet cross-section as a function of jet transverse momentum and absolute jet rapidity with jet radius parameter R=0.7
+    label: r"$d^2\sigma / dp_Tdy$"
+    units: '[pb/GeV]'
+  kinematics:
+    file: kinematics_R07.yaml
+    variables:
+      pT: {description: jet transverse momentum, label: r"$p_T$" , units: GeV}
+      y: {description: jet rapidity, label: r"$|y|$" , units: ''}
+      sqrts: {description: centre of mass energy, label: r"$\sqrt(s)$", units: GeV}
+  kinematic_coverage: [pT, y, sqrts]
```
achiefa: Could please change this order?

```suggestion
  kinematic_coverage: [y, pT, sqrts]
```

PR #2431, opened 2026-02-18
Make replicas Colibri compatible

ecole41:
This PR edits the `make_replica` function such that a central value array and covmat can be inputted and pseudodata can then be produced from this.

The selection of the random seed has been simplified such that the same seed for each replica is used. This function is currently incompatible with grouped data as the same random normal distribution will be used for different groups. To fix this, I have suggested to access the group name and edit the random seed using this. This means that the ordering of datasets or removing/adding datasets from the fit will not alter the random seed. Currently, the group name is not accessible in the `make_replica` function.

### DIFF
--- a/validphys2/src/validphys/tests/conftest.py
+++ b/validphys2/src/validphys/tests/conftest.py
@@ -107,15 +107,16 @@ def data_internal_cuts_config(data_config):
 
 
 @pytest.fixture(scope='module')
-def data_internal_cuts_new_theory_config(data_internal_cuts_config):
+def data_internal_cuts_closure_config(data_internal_cuts_config):
+    # Filterseed is not added so that it is changed by the tests
     config = dict(data_internal_cuts_config)
-    config["theoryid"] = THEORYID
+    config["fakepdf"] = PDF
     return config
 
 
 @pytest.fixture(scope='module')
-def data_fromfit_cuts_config(data_internal_cuts_new_theory_config):
-    config = dict(data_internal_cuts_new_theory_config)
+def data_fromfit_cuts_config(data_internal_cuts_config):
+    config = dict(data_internal_cuts_config)
     config.update(use_cuts="fromfit")
     return config
 
--- a/validphys2/src/validphys/tests/test_datafiles.py
+++ b/validphys2/src/validphys/tests/test_datafiles.py
@@ -1,8 +1,8 @@
 """
-    Test all datafiles
+Test all datafiles
 
-    The checks in ``test_all_datasets`` are run for each dataset independently so that one gets one
-    failure per dataset in case of problems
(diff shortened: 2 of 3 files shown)


ecole41:
@scarlehoff I have now edited this such that `make_replica` takes in a `cantral_values_array` and a `group_multiplicative_errors` dictionary. If this dictionary is `None` then the generation of the multiplicative shifts is skipped and only the covmat is used to produce shifts.

Also a new function `group_replica_mcseed` is added to produce the seed. Please let me know if this can be improved.

validphys2/src/validphys/n3fit_data.py
```diff
@@ -110,6 +110,20 @@ def replica_luxseed(replica, luxseed):
     return replica_nnseed(replica, luxseed)
 
 
+def group_replica_mcseed(replica_mcseed, groups_dataset_inputs_loaded_cd_with_cuts):
+    names_for_salt = []
```
scarlehoff:
```suggestion
    names_for_salt = []
    # Try to use the new dataset name, but make older runs reproducible by keeping the old names.
    # WARNING: don't rely on this behaviour, this might be removed in future releases
```
I'd say we remove it for the next tag but better to do it just before the tag.

scarlehoff (commented):
Thanks! I think it looks good. I've left a few small comments.

The only thing I'm not sure about is the mask part. I think it has nothing to do with `sep_mult`. Following the spirit of the rest of the changes you can create a function `group_positivity_masks` which creates what is called now `full_mask` and that returns an array of 0s or `None` when `resample_negative_pseudodata` is False (so that no positivity is checked).

That way you can remove the condition from `make_replica` as well.

ecole41:
> Thanks! I think it looks good. I've left a few small comments.
> 
> The only thing I'm not sure about is the mask part. I think it has nothing to do with `sep_mult`. Following the spirit of the rest of the changes you can create a function `group_positivity_masks` which creates what is called now `full_mask` and that returns an array of 0s or `None` when `resample_negative_pseudodata` is False (so that no positivity is checked).
> 
> That way you can remove the condition from `make_replica` as well.

I'm not too sure how I would remove the `resample_negative_pseudodata` condition from `make_replica` as this will be need to create the mask inside of `make_replica` when no commondata object is available. Whereas the `group_positivity_mask` requires a CommonData object. If you can think of a way to simplify this, please go for it.

I have now added some edits after your comments. I decided to remove `groups_dataset_inputs_loaded_cd_with_cuts` as an input into `make_replica` however this means removing the group name from the warning at the end of the function. Let me know if this should be added back somehow. Thanks

scarlehoff:
The bug here is actually coming from https://github.com/keras-team/keras/pull/22376

PR #2436, opened 2026-03-02
Update the process type of the 3D CMS 13 TeV

achiefa:
Just a small change in the process options. This is needed for the power correction analysis since I need to tell apart DIJET and DIJET_3D.


achiefa:
@scarlehoff before merging this I just want to check whether this has some side effects in validphys. I can't see any but I wnat to double check.

achiefa:
Yes, exactly.

scarlehoff (approved):

PR #2438, opened 2026-03-16
New baseline 26/03/13

kamillaurent:
I updated the `example-nnpdf41.yml` runcard with the new baseline (26-03-13).

Here is a comparison with the old baseline (`260202-jk-nnpdf41-mhou`):
https://vp.nnpdf.science/l2adpDSrRt2lGJK78Ncc1g==


scarlehoff:
I have the feeling we might need to go all the way up to 1000 replicas :'(

PR #2443, opened 2026-03-25
Load weights from a fit

ecole41:
This PR adds the `load_weights_from_fit` option which allows for the weights from a previous fit to be used as the initial weights of another fit, if the weight files are available for this previous fit.


validphys2/src/validphys/config.py
```diff
@@ -165,6 +165,20 @@ def parse_pdf(self, name, unpolarized_bc=None):
 
         return pdf
 
+    def produce_load_weights_dict(self, fit, load_weights_from_fit=None):
```
scarlehoff: Why do you need `fit` here as part of the input?

scarlehoff (commented):
Thanks! Seems fine to me. I want to try to do the version for multireplica before merging, I think it should work fine.

ecole41:
@scarlehoff Do you understand why the commondata test is failing here?

scarlehoff:
The filters for the jet data need to be rewritten to make sure their results are more robust and don't depend upon the versions of the tools, as they have many systematics. 
i.e, not related to this PR but something that would be nice to fix for sure.

validphys2/src/validphys/config.py
```diff
@@ -165,6 +165,27 @@ def parse_pdf(self, name, unpolarized_bc=None):
...
+            if weights_name.endswith(".h5"):
+                weights_name = weights_name[:-3]
+            weights_dict = {}
+            for p in fit_folder.glob("replica_*/{weights_name}.weights.h5"):
+                replica_folder = p.parent.name
+                replica_index = int(replica_folder.split("_")[1])
+                weights_dict[replica_index] = p
+            return weights_dict
+        except LoadFailedError as e:
+            raise ConfigError(str(e), load_weights_from_fit, self.loader.available_fits)
```
scarlehoff:
```suggestion
            raise ConfigError(str(e), load_weights_from_fit, self.loader.available_fits) from e
```

(little by little we should correct all the raise errors that lose the original exception, since they might be useful)

scarlehoff (approved):
I added a few changes (plus precommit, which changed quite a bit of some files because it hadn't been run in a while...) to make it work in the case of multireplica.
For the next tag if I remember I'll change the fitbot to use this, I think it'd be useful.

Would you mind adding a mention to `load_weights_from_fit: <somefit>` to the docs https://docs.nnpdf.science/n3fit/runcard_detailed.html#save-and-load-weights-of-the-model ? Other than that this is ready to be merged.

ecole41:
> I added a few changes (plus precommit, which changed quite a bit of some files because it hadn't been run in a while...) to make it work in the case of multireplica. For the next tag if I remember I'll change the fitbot to use this, I think it'd be useful.
> 
> Would you mind adding a mention to `load_weights_from_fit: <somefit>` to the docs https://docs.nnpdf.science/n3fit/runcard_detailed.html#save-and-load-weights-of-the-model ? Other than that this is ready to be merged.

It all looks good to me. I have just added a little section to the documentation, let me know if it needs editing at all.

PR #2453, opened 2026-04-14
Add prescriptions to use diagonal theory cov mat

enocera:
This PR implements two prescriptions to use either the envelope of the 7pt scale variations or the diagonal elements of the 7pt theory covariance matrix. These are meant as diagnosis tools to investigate the data/theory comparisons.


scarlehoff:
hi @ecole41 could you check that this is reasonable? As far as I'm aware it has already been used to compute the chi2 for https://github.com/NNPDF/nnpdf/pull/2411#issuecomment-4055174111 so it can be merged

(that said, @enocera could you also put the chi2 in that PR)

enocera:
This PR actually turns out to be immaterial for #2411, given that the problem has been identified with correlations of experimental uncertainties. Nevertheless, this PR may be of interest for other studies (e.g. ZpT at small pT), therefore I am in favour of keeping it. I may provide examples of cases in which this may be useful "soon".

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -141,6 +141,30 @@ def covmat_7pt(name1, name2, deltas1, deltas2):
...
+    return np.diag(np.diag(s))
+def covmat_7pt_sym_envelope(name1, name2, deltas1, deltas2):
+    """Returns the diagonal theory covariance sub-matrix constructed as the
+    symmetric envelope around the central scale of the 7pt variations"""
+    if name1 == name2:
+        deltas1_max = np.max(np.abs(deltas1), axis=0)
+        
+        s = np.outer(deltas1_max, deltas1_max)
+    else:
+        sys.exit("Inconsistent process name")
```
scarlehoff, line 181:
```suggestion
        raise ValueError("Inconsistent process name. This is a debugging prescription and should not be use for fits.")
```

Is this the intended behaviour?

validphys2/src/validphys/theorycovariance/construction.py
```diff
@@ -141,6 +141,30 @@ def covmat_7pt(name1, name2, deltas1, deltas2):
...
+        s = (1 / 3) * sum(np.outer(d, d) for d in deltas1)
+    else:
+        s = (1 / 6) * (
+            2 * (np.outer(deltas1[0], deltas2[0]) + np.outer(deltas1[1], deltas2[1]))
+            + (
+                np.outer((deltas1[2] + deltas1[3]), (deltas2[2] + deltas2[3]))
+                + np.outer((deltas1[4] + deltas1[5]), (deltas2[4] + deltas2[5]))
+            )
+        )
+    return np.diag(np.diag(s))
```
scarlehoff, line 172: I'm confused here because if name1 != name2 then there's nothing in the diagonal right? Should this also error out in that case. Or return all 0s if we only want the diagonal?

I think right now this one would be filling up random diagonals across the datasetX - datasetY offdiagonals

scarlehoff (changes requested):

Issue #43, opened 2018-02-08
Define and rationalize what  a "fit" is


scarrazza:
<a href="https://github.com/Zaharid"><img src="https://avatars2.githubusercontent.com/u/5675858?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [Zaharid](https://github.com/Zaharid)**
_Thursday Jun 29, 2017 at 16:07 GMT_
_Originally opened as https://github.com/NNPDF/nnpdfcpp/issues/19_

----

We should write and systematize somewhere what a:
 - Initialized fit is: I.e. the results of running `filter`.
 - A running fit is: Some replicas are being prepared.
 - A completed fit is. All replicas are finished and `postfit` ran successfully.

This makes it difficult for `postfit` to work correctly, and to write tools that operate on groups of fits like validphys.

More concretely, there are the following issues:
  - There is no one obvious way to distinguish a finished replica from a failed one from a running one. Replicas should  me moved to a folder with a different name after they are completed.
  - Postfit should take as input "completed" replicas and move them to "final" replicas, without destroying information in the process.
  - There are files outside the fit directory that are part of the "fit" (the file in `../config`).
  - There are files inside the fit directory that are not uploaded but are costly to recompute (the new preprocessing exponents).
  - `nnfit` writes files that start with the replica names and files that don't. None of the files should start with the replica name, since that makes processing more difficult for no reason (see https://github.com/NNPDF/validphys2/blob/07d7f297ab08f26e048583daac10c8e641f014fa/src/validphys/fitdata.py#L29). These prefixes are a disease from LHAPDF.
  - The files written by postfit should ideally be either yaml or csv (or csv with a yaml header). Both of these are easy to write (and to read using a library). No reason to write a parser for every file. Also some cannot be parsed easily as of now (e.g for `sumrules` one has to read the code to see which number is which sumrule.).  
  - There are many instances of the config file, that are only useful to make it confusing to decide the good one when they disagree (and to suck up disk space when uncompressed). Instead there should be only one file (the one currently corresponding to `filter.yml`).

PR #2455, opened 2026-04-14
Pseudodata in diagonal basis and saving rotation

### LINKED ISSUE #2445: Reproducibility of diagonal covmat fits (clarify or bug?)
In principle, when `savepseudodata` is set to `True` (the default), the fit should be saving a csv file with the training and validation data per replica. This `csv` includes the name of the datasets and the index of the data.

When using the diagonal basis, there is no one-to-one correspondence between the datasets and the data going into training and validation, however these `.csv` file still have the experiment name on them.

We have `pseudodata_test_fit_n3fit_251104` in the tests which is tested in `test_read_matches_recreate` so it seems we are able to recreate the pseudodata of a diagonal fit (separating training and validation). So, is this a bug? Or are we simply saving a dataset name that means nothing?

(or, rather, is the bug the fact that we are saving the name or are we saving something that is not really the data we are fitting?)

jacoterh:
WIP

- [x] make sure `nnfit_theory_covmat` is ordered consistently with the experimental covmat before converting both to a dataframe


validphys2/src/validphys/n3fit_data.py
```diff
@@ -586,11 +619,24 @@ def replica_pseudodata(experiment_indexed_make_replica, replica):
...
+        df.columns = [replica_column]
+        return df
+
+    df = pd.concat(experiment_diagonal_indexed_make_replica, ignore_index=True)
+    df.columns = [replica_column]
+    df.index = pd.Index([f"eigenmode {i}" for i in range(len(df))])
     return df
 
 
+experiment_diagonal_indexed_make_replica = collect(
```
achiefa: Can you please help me understand why we need to use collect here? Wouldn't `diagonal_indexed_make_replica` be enough alone? As far as I understand, the latter is already a dataframe with the correct index, comprising of group, dataset name, and data index. The collection just returns the same df wrapped in a array, with a single entry being the dataframe itself (so not very useful). This happens for the non-diagonal case too, see `experiment_indexed_make_replica`. What am I missing here?

achiefa (commented):
I'll have a closer look tomorrow morning. For the moment, this is the only thing that confuses me. But I'm surely missing something.

validphys2/src/validphys/n3fit_data.py
jacoterh: You're right, thanks! We can get rid of the collect because the diagonal rotation already acts on the whole space with all experiments combined

achiefa:
So just to make the point here and remind the myself of the future. @jacoterh has successfully implemented the possiblity of storing the eigenvectors and relative eigenvalues in vp-setupfit for reproducibility. This required us to default `use_t0` to True in Validphys, in a similar way to what is currently done in n3fit. If I'm understanding it correctly, an independent diagonalisation takes place also in n3fit. It would be great to use the stored values from vp-setupfit instead of recomputing them in n3fit (which, anyway, caches them if run on gpus), but I'd say this is not a priority for the moment.

There is one open problem with the current branch, namely that vp-setupfit sets `use_thcovmat_in_fitting` to False by default. Hence the diagonalisation in vp-setupfit doesn't take into account the theory covmat, unlike the n3fit case. So I was just wondering whether there's a reason why `use_thcovmat_in_fitting` is defaulted to False in first place. Do we break something if we set it to True? @scarlehoff 

Am I missing something @jacoterh ?

scarlehoff:
I guess by validphys you mean vp-setupfit. For `use_thcovmat_in_fitting`, since we are always using it we might as well set it true everywhere.

>  It would be great to use the stored values from vp-setupfit instead of recomputing them in n3fit (which, anyway, caches them if run on gpus),

Indeed, I think that, given that now vp-setupfit stores the diagonalization, it can be treated as the theory covmat and n3fit should read it instead of recalculating it.

achiefa:
> Indeed, I think that, given that now vp-setupfit stores the diagonalization, it can be treated as the theory covmat and n3fit should read it instead of recalculating it.

So do we want to implement this now, or shall we keep it for a new PR?

jacoterh:
Yes, let's do this here - I was already working on it on Friday but was encountering some issues. I continue with it this afternoon so we can hopefully merge this PR by tomorrow at the CM.

validphys2/src/validphys/covmats.py
```diff
@@ -484,6 +484,7 @@ def dataset_inputs_t0_total_covmat(dataset_inputs_t0_exp_covmat, loaded_theory_c
     """
     covmat = dataset_inputs_t0_exp_covmat
     covmat += loaded_theory_covmat
```
jacoterh: @scarlehoff Is there a way to have vp-setupfit write the theory covmat csv files to the table directory first before attempting to load them? We need the theory covmat already at the stage of the diagonalisation in vp-setupfit.

validphys2/src/validphys/config.py
jacoterh, line 937: This file doesn't exist yet during vp-setupfit so the diagonalisation is unable to read it and use it in the diagonalisation

validphys2/src/validphys/covmats.py
scarlehoff: I think it would be best to use whatever is in memory. But I remember we explicitly decide to forbid that a long time ago and I don't remember right now how easy / possible it is to revert that.

If you didn't already, try simply swapping the order of the theory covmat and the diagonalization in the vp_setupfit script. I don't remember when are they actually written down, but if it is upon creation that should be enough.

n3fit/src/n3fit/scripts/vp_setupfit.py
```diff
@@ -181,6 +181,8 @@ def from_yaml(cls, o, *args, **kwargs):
                 'datacuts::theory::theorycovmatconfig nnfit_theory_covmat'
             )
 
+        SETUPFIT_FIXED_CONFIG['actions_'] += [rotation_action]
```
jacoterh: @scarlehoff I changed the order of the actions in the hope the theory covmat csv would get written first, but again the same `FileNotFoundError` occurs. Maybe reportengine first resolves all dependencies regardless of their order? Not sure.

scarlehoff: Yes, it is very likely.

One option is to make it so the diagonalization can only be run from vp-setupfit and so it doesn't depend on the covmat that goes to the csv but on `nnfit_theory_covmat`, but then you need to also take into consideration the transformations that happen on the csv.

Another possibility is to change `nnfit_theory_covmat` so that, when it is coming from `vp-setupfit` it works normally, but when it is running from `n3fit` it just returns `None` (or perhaps `lambda : None` or whatever, not sure if it needs to be a function).
Then the `.csv` function can depend on `nnfit_theory_covmat` as well. When it is `None` it means that it is coming from `n3fit` (or from whatever that makes it return `None`) and needs to read the data from the `.csv`, otherwise use that `nnfit_theory_covmat` that just arrived.

As I said, we decided back in the day not to have the option to run without `vp_setupfit` first so I'm not sure this second option will work ootb.

Edit: can you stack two produce rules one of which is an `explicit_node`? I'm not sure :(
You can always mix and match these options. For instance, making sure that the order is fixed already when the covmat is written to the .csv file so that you don't need to think about that later.

jacoterh: Thanks Juan! I have gone for the for the first option you suggested, i.e. call `nnfit_theory_covmat` during vp-setupfit. Further checks are necessary but we have something working at the moment. For instance, I haven't paid close attention to the transformation you mentioned? Is this a reindexing of the pandas data frames when converting them to numpy arrays?

scarlehoff:
Before trying to solve the failing test, rebase on top of master to avoid doing the work twice (there are things added there like a more strict check on the documentation)

jacoterh:
@scarlehoff some worrying news perhaps. I found that the line 
https://github.com/NNPDF/nnpdf/blob/44a5ddf499d43775b1f0dae5b5b8f2e367e8563a/validphys2/src/validphys/covmats.py#L486
is called twice on master. As a result the theory covmat gets added twice, which then explains why I found different eigenvalues in the presence of a theory covmat. 

To reproduce this behaviour, use the runcard below and turn on the `pdb` debugger inside this function and you'll see that it gets called twice. Not sure why! I hope I'm wrong... maybe I'm going mad at this point

[260414-jth-diagonal_basis_test.yml](https://github.com/user-attachments/files/27176133/260414-jth-diagonal_basis_test.yml)

scarlehoff:
What do you mean by twice? It doesn't (shouldn't) matter how many times you call it, it is producing the sum of the experimental and theory covmat.

It would be inefficient (perhaps a leftover of having two covmats in the fit, the t0 and the normal) but it should be fine.

jacoterh:
Together with @achiefa we have checked that if you print dataset_inputs_t0_exp_covmat you will see that it changes between the two calls, and we suspect it already includes the theory covmat after the first time

scarlehoff:
That said, regardless how many times it gets called, this function should be just `return dataset_inputs_t0_exp_covmat +loaded_theory_covmat`

Change it to that and see whether results change. The `+=` is asking for trouble if pandas or numpy suddenly decide that means "in place" (if this is the issue then it is horrendous but luckily we have been compatible from pandas 3 only since january I think)

jacoterh:
Yes this is the issue indeed

scarlehoff:
Then, if you think this PR is close to finish, change it here so that it gets merged.

Otherwise, please open a new PR with this change so that we merge it to master asap (and perhaps even release 4.1.4)

achiefa:
So does this mean that all the fits performed since January are basically wrong?

jacoterh:
I'm starting to fear yes, the point is that `covmat += loaded_theory_covmat` modifies `covmat` in place (what else could it mean?), which links by reference to `dataset_inputs_t0_exp_covmat` so that this also changes...

Let's address this in a separate PR to fix this asap and check meanwhile which fits got affected by this

scarlehoff:
> So does this mean that all the fits performed since January are basically wrong?

I hope it is only since January.

Otherwise since we started using t0 for both sampling and fit.*

That function has been like that since forever and as a @jacoterh said it does say explicitly "modify covmat in place", my hope is that older versions of pandas/numpy were just creating a new object (so the original covmat is safe).

*unless going through it twice only happened recently, or only for the diagonal case. This is also a possibility.

achiefa:
> The += is asking for trouble if pandas or numpy suddenly decide that means "in place" (if this is the issue then it is horrendous but luckily we have been compatible from pandas 3 only since january I think)

BTW, I don't think `+=` is causing the problem here. It does what it's supposed to do, namely mutating the object in-place. The real problem (and for that we should blame both ReportEngine and Python) is the first `=`, which is silently aliasing, that is, acting by reference. Since `dataset_inputs_t0_exp_covmat` is cached by reportengine, any time we use `=` we create an alias to the global object, rather than a new variable in the local scope. This is very bad and counterintuitive, as there is no explicit way in Python to see whether a variable is passed by reference or by value. And ReportEngine doesn't help either.

I'm wondering whether similar patterns are used in other parts of the code: this in principle may affect other objects, not only the covmat.

scarlehoff:
I can reproduce this since 4.1.0 but not 4.0.10, so the alpha_s paper should be ok. Would be good to double check 

> and for that we should blame both ReportEngine and Python

You would have the same in Fortran or C when you pass a pointer around (and we would have for such a big array, we would've also been caching it). The code is wrong, it might have been ok in the past if a new object was made (both numpy and pandas used to create new objets for some operations) but it is wrong nonetheless.

The problem surfaces here because it is getting called twice and it seems that started in 4.1.0.
Looking at the changes, most likely the diagonal covmat make us go twice through there for some reason (regardless of whether it is on or off)

> I'm wondering whether similar patterns are used in other parts of the code: this in principle may affect other objects, not only the covmat.

Indeed.

achiefa:
> You would have the same in Fortran or C when you pass a pointer around (and we would have for such a big array, we would've also been caching it). The code is wrong, it might have been ok in the past if a new object was made (both numpy and pandas used to create new objets for some operations) but it is wrong nonetheless

Well, in C you have the type in the function's signature, so you'd know whether the parameter is a pointer, a reference, or a value. In python this not explicit. Here was a bit misleading and unfortunate.

Anyway, this is partially good news. The problem is that all the studies carried out in the diagonal covmat are affected. Is hyperopt also affected?

scarlehoff:
Also studies not in the diagonal covmat though. Anything using new versions of the code.

Hyperopt is not because we were not using the theory covmat (yey I guess) and Kamil's fits had to be redone anyway because of the fktables.

What worries me is the claim that we could reproduce the alpha_s fits 😅 hopefully it just means that this is not a big effect (which could be, it is a factor of 2 only for the theory covmat that goes into the chi2)

(btw, in python everything must be considered a pointer, although I agree that the fact that some types get copied can lead to confusion, but I don't think this was the case here :__ just a lack of attention 🙃)

jacoterh:
I reproduce the same bug on master when I run with `diagonal_basis: False`. The mtop paper used the version of master before the diagonal basis got enabled by default, yet all the functionality is there. (but it doesn't use it). This would mean that the mtop paper might be affected too... I'll check this today more carefully

achiefa:
> (btw, in python everything must be considered a pointer, although I agree that the fact that some types get copied can lead to confusion, but I don't think this was the case here :__ just a lack of attention 🙃)

Yeah, I think I learned a lesson here. Are you checking if the same bug is replicated somewhere else in the code? I think this might be a good task for an agent.

jacoterh:
> Yeah, I think I learned a lesson here. Are you checking if the same bug is replicated somewhere else in the code? I think this might be a good task for an agent.

Indeed, I will give this task to my agent, let's see

scarlehoff:
> Are you checking if the same bug is replicated somewhere else in the code?

No, I'm out the entire day for teaching duties. I won't be able to get back to this until the weekend so if you two can have a go it would be very helpfuk

jacoterh:
PR https://github.com/NNPDF/nnpdf/pull/2463 addresses the issues we discussed here :)

validphys2/src/validphys/n3fit_data.py
```diff
@@ -586,8 +672,16 @@ def replica_pseudodata(experiment_indexed_make_replica, replica):
...
-    df = pd.concat(experiment_indexed_make_replica)
-    df.columns = [f"replica {replica}"]
+    replica_column = f"replica {replica}"
+
+    if not diagonal_basis:
+        df = pd.concat(experiment_indexed_make_replica)
+        df.columns = [replica_column]
+        return df
+
+    df = diagonal_indexed_make_replica.copy()
```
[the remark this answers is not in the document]
jacoterh: This is fine as is given that concatenation is not necessary in the diagonal basis because there is only one group anyway

achiefa:
Hi @jacoterh, what's the status of this? Can I review it?

jacoterh:
Yes, sorry, this is ready for review. Please pay extra attention to the first point raised by copilot above. I don't think it breaks anything, but better safe than sorry

validphys2/src/validphys/pseudodata.py
```diff
@@ -123,7 +123,7 @@ def read_replica_pseudodata(fit, context_index, replica):
 def make_replica(
     central_values_array,
     group_replica_mcseed,
-    dataset_inputs_sampling_covmat,
```
achiefa, line 126: Why are we removing the function `dataset_inputs_sampling_covmat` in favour of `dataset_inputs_covmat_t0_considered`? The problem I see with this is partially exposed in Copilot's comment. As far as I can tell, `dataset_inputs_covmat_t0_considered` can only include or not t0, but does not consider any theory covmat. If we then use `dataset_inputs_covmat_t0_considered` in `make_replica`, we're then sampling without theory covmat, even if `use_thcovmat_in_sampling` is set to True in the runcard (at least this is what I gather from the function). Furthermore, I'm not entirely sure that `use_t0_sampling` is effective as of now: it was used by  `dataset_inputs_sampling_covmat`, while `dataset_inputs_covmat_t0_considered`'s signature is `use_t0`. I'm pretty sure that `use_t0` and `use_t0_sampling` must be decoupled.

So the code won't break if we use `use_t0_sampling` and `use_thcovmat_in_sampling`, but it doesn't lead to the expected behavour. Am I missing something?

validphys2/src/validphys/config.py
```diff
@@ -823,55 +824,6 @@ def produce_sep_mult(self, separate_multiplicative=False):
             return False
         return True
 
-    @configparser.explicit_node
-    def produce_dataset_inputs_sampling_covmat(
```
achiefa, line 827: See the other comment/

validphys2/src/validphys/covmats.py
```diff
@@ -406,14 +407,14 @@ def dataset_inputs_t0_covmat_from_systematics(
 
 
 def dataset_inputs_t0_total_covmat_separate(
-    dataset_inputs_t0_exp_covmat_separate, loaded_theory_covmat
```
achiefa, line 409: I'm quite sure that this was discussed and I forgot. But why aren't we loading the theory covmat anymore? `nnfit_theory_covmat` computes the theory covmat from scratch. Is this what we want? And why?

validphys2/src/validphys/covmats.py
```diff
@@ -502,16 +505,17 @@ def dataset_inputs_t0_exp_covmat(
         dataset_inputs_t0_predictions,
         False,
     )
-    return covmat
+    # TODO: how to know for sure if the index matches the covmat value ordering?
+    return pd.DataFrame(covmat, index=procs_index_matched, columns=procs_index_matched)
```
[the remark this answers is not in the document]
achiefa: Why do you need a dataframe? Is this because `nnfit_theory_covmat` is a dataframe too?

RE the returned type: my understanding of this function is that it serves at an intermediate level in the computation graph. In fact, there are other functions in vp that return the covmat as an indexed dataframe. So I'm wondering why we need to return a dataframe at this stage.

@scarlehoff, what do you think?

validphys2/src/validphys/n3fit_data.py
```diff
@@ -313,24 +313,102 @@ def fittable_datasets_masked(data):
...
+    """Prepare the covariance matrix and its inverse, optionally transforming to diagonal basis.
 
-    Since the masks and number of datapoints need to be treated for 1-point datasets
-    it also returns the right ndata and masks for training and validation:
+    Parameters
+    ----------
+    _hashed_dataset_inputs_fitting_covmat : Hashrray
+        Wrapped covariance matrix for caching purposes.
+    output_path : Path
+        Path to output directory containing diagonal basis data if needed.
```
achiefa, line 336:
```suggestion
    output_path : Path
        Path to output directory containing diagonal basis data if needed.
    use_thcovmat_in_fitting: bool, optional
        If True, load the total covariance matrix, which includes the theory covariance matrix. If False, load only the experimental covariance matrix. Default is False.
```

achiefa: Is `False` the desired default option for `use_thcovmat_in_fitting`?

validphys2/src/validphys/n3fit_data.py
```diff
@@ -313,24 +313,102 @@ def fittable_datasets_masked(data):
...
+        inv_total = (
+            np.diag(diag_inv_sqrt_total)
+            @ np.linalg.inv(cormat_total)
+            @ np.diag(diag_inv_sqrt_total)
+        )
+
+    return covmat, inv_total, diag_rot, eig_vals
+
+
+def _fiting_covmat(dataset_inputs_fitting_covmat, data_input, diagonal_basis=True):
```
achiefa:
```suggestion
def _fiting_covmat(dataset_inputs_fitting_covmat, diagonal_basis=True):
```

achiefa (commented):
Thank you @jacoterh for this. Overall I'd say that the PR solves the issue. However, there are some aspects that I don't completely understand and I'd like to clarify before we merge.

validphys2/src/validphys/covmats.py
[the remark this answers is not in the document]
jacoterh: Yes, exactly, I'd argue we need a dataframe here because it's later combined with `nnfit_theory_covmat` that is already a dataframe and I wanted to avoid any possible misalignment between the two, see `dataset_inputs_t0_total_covmat`.

validphys2/src/validphys/covmats.py
jacoterh: Because this function is now called already during vp-setupfit and there is no stored covmat available to load at that point yet!

validphys2/src/validphys/pseudodata.py
jacoterh: Thanks for spotting this, you're right. The issue is the following. We are now storing the covmat as part of vp-setupfit and we want n3fit to load these back in to avoid doing things twice. However, `make_replica` generates pseudodata in the data basis regardless of whether the user requests the diagonal basis or not, which means that this function needs access to the full covmat (exp + possibly th) in the data basis. But this we don't store when the diagonal basis is requested - we store the eigenvectors and eigenvalues. We can reconstruct it from here of course, but then we're going around in circles... The whole design with storing/loading needs some more thought!

validphys2/src/validphys/covmats.py
achiefa: Right, my bad!

validphys2/src/validphys/n3fit_data.py
jacoterh: Yes, it will be set to True if specified in the runcard

achiefa:
I'm trying the fix the errors, but I don't understand why some of the regression tests fail. For example, I don't understand why `no_positivty` gives a different best epoch compared to the one stored. I believe it shouldn't, right?
@scarlehoff @jacoterh

scarlehoff:
re the regression tests: https://github.com/NNPDF/nnpdf/issues/2464
a few tries may be needed

But there are many other tests failing right now, I would address the regression last.

achiefa:
Apparently `run_pytorch` keeps failing for an import error in `triton`, which is used by `torch`. Now, `triton` should not be used on a CPU-runner like ours, and the import logic should take this into account. Apparently `triton==3.7.0`, which is required by the just-released `torch==2.12.0`, has broken import logic and it segfaults when loading the C library (`triton._C.libtriton`).

A possible workaround could be pinning `torch<2.12` in the runner, which should transitively pin `triton<3.7`. What do you think?

.github/workflows/all_tests_nnpdf.yml
```diff
@@ -103,7 +103,11 @@ jobs:
         sudo rm -rf /opt/hostedtoolcache/Python/{3.9*}
         sudo rm -rf /opt/hostedtoolcache/CodeQL
         sudo rm -rf /usr/local/lib/android/sdk
-        pip install .[nolha,torch]
+        # torch 2.12.0 pulls triton 3.7.0, whose libtriton C extension segfaults
+        # at import on CPU-only Linux + CPython 3.13. Pinning torch<2.12 also
+        # pins triton to 3.6.x. Remove this pin once the upstream issue is
+        # resolved.
```
scarlehoff: Seems reasonable to me. If this is being discussed in the pytorch repo link the issue in the comments so that we can track it more easily.

achiefa: It's not actually. The most similar issue is [triton-#43](https://github.com/nascheme/triton/issues/43), but quite different. I'd mention it anyway.

achiefa:
Question for both @scarlehoff and @jacoterh. 

As of now, I'm able to set up and run a fit in the diagonal basis. The problem is that `read_replica_pseudodata` breaks in the diagonal basis. Specifically, the bug happens here:

https://github.com/NNPDF/nnpdf/blob/1ac88fba4b3cf5b55c37a9b340306b420ff02a4d/validphys2/src/validphys/pseudodata.py#L91-L93

In the diagonal basis, there is only one column-index (eigenmodes) and it should be `index_col=[0]`; contrary, in the non-diagonal fit there are three column-indices (group, dataset, id). Therefore we need to detect whether the fit was performed in the diagonal or non-diagonal basis. 

Everything that comes about after the CSV file loading

https://github.com/NNPDF/nnpdf/blob/1ac88fba4b3cf5b55c37a9b340306b420ff02a4d/validphys2/src/validphys/pseudodata.py#L103-L110

needs to be changed as well, beacuse in the diagonal basis we're not interested in any data label but just eigenmodes. Do you know an elegant and smart way to detect whether the fit was performed in the diagonal basis. In principle it would be sufficient something like
```
fit_input.get("diagonal_basis", True)
```
since it is defaulted to true. But if this changes in the future, folks have to remember to update also this one.

**Worth noting:**
The tests (for example `test_read_fit_pseudodata`) cannot catch the problem with diagonal pseudodata because (I think) they use fits without diagonal basis, for which the problem doesn't occur.

scarlehoff:
I think it is fine having two separate paths one for diagonal, one for non-diagonal. 
If parts of the code need the extra indices, that's fine I guess, we just have one BIGEXP dataset/group that contains everything.

> since it is defaulted to true. But if this changes in the future, folks have to remember to update also this one.

Ideally we should have lockfiles for the fit, but they need to be implemented so I guess memory is the only way. However, given that in one case you have 3 column and in the other you have 1, a hard crash if the input doesn't match the expectation is fine.

achiefa:
Why is `diagonal_basis` never used inside this function?

https://github.com/NNPDF/nnpdf/blob/1ac88fba4b3cf5b55c37a9b340306b420ff02a4d/validphys2/src/validphys/results.py#L245-L276

achiefa:
My latest commit fixes the last issue (I hope) with the diagonal basis. @scarlehoff and @jacoterh can you please have a look?

# TD;LR
Find a workaround for loading the covmat outside fit time for diagonal-basis fits. Needed to reconstruct pseudodata after the fit. Added two helper functions in `pseudodata.py`.

# Details
For a fit in diagonal basis, `read_fit_pseudodata` returns eigenmode-index (rotated) data, which is what's on disk. However, `recreate_fit_pseudodata` returned original-basis data, because it used directly `indexed_make_replica` but never applied the rotation. Hence the disagreement in the test `test_read_matches_recreate`.

In principle, one could use `diagonal_indexed_make_replica` instead of `indexed_make_replica` for diagonal fits. However this requires `fitting_data_dict`, which in turn is produced by `_inv_covmat_prepared` which reads the eigensystem from the CSV file in the fit folder. However, `output_path` is the fit directory at fit time. At vp time I'm not sure `output_path` points to the correct directory. Hence trying to route the recreation through `diagonal_indexed_make_replica` would yield FileNotFoundError. Is there a way to make `output_path` aware of the correct path?

Meanwhile, I solved this problem as follows:
1. Added `fit_diagonal_basis_rotation` which is the recreation-time counterpart of `_inv_covmat_prepared`. It reads the same eigensystem and the rotation is identical. It returns none for non-diagonal fits.
2. Added `diagonal_indexed_recreate_pseudodata`, which is similar to `diagonal_indexed_make_replica`, but sources the rotation from `fit_diagonal_basis_rotation`. For non-diagonal fits, it returns `indexed_make_replica` and thus falls back to the non-diagonal case  correctly handled even before.
3. The other functions such as `recreate_fit_pseudodata`, `recreate_pdf_pseudodata`, and `recreate_pdf_pseudodata_no_table`(thus the API) keep their existing signatures. The routing is handled by the two helper functions that I added.

There are two things that leave me a bit unsettled:
1. Different functions that do similar things, because they have to cope with reportengine's logic. This is the case of the two helper functions. But I don't think we can do any better.
2. Fits prior to the diagonal-basis are not compatible, unless we add `diagonal_basis: False` in the runcard. In fact, I had to update the pseudodata-test fit.

scarlehoff:
> Different functions that do similar things, because they have to cope with reportengine's logic. This is the case of the two helper functions. But I don't think we can do any better.

You mean the functions that do the sum? I guess you could have them be a wrapper for some function _combine_covmats but for just a sum not sure it makes sense. Other than that, they are distinct functions as they take different inputs.

I'm most worried about all the places in which the file to read the covmat from is defined `datacuts_theory_theorycovmatconfig_fitting_covmat_table.csv`, that should be limited to `config.py` (and even there it would better to reduce it, but at least it is concentrated).

> Fits prior to the diagonal-basis are not compatible, unless we add diagonal_basis: False in the runcard. In fact, I had to update the pseudodata-test fit.

This is ok. 4.1 is not compatible 100% with 4.0, the defaults are allowed to change.

achiefa:
I mean that the new function `diagonal_indexed_recreate_pseudodata` and the pre-existing `diagonal_indexed_make_replica` do similar things. However, the latter is meant to be used at n3fit-time while the former can be called through the API. My complaint was more me putting myself in the shoes of a new member browsing the codebase for the first time and getting confused seeing different functions with different scopes doing similar things.

achiefa:
@jacoterh what's the status of the two elements of your to-do list at the very top of this PR?

achiefa:
Another question: why do we save the fitting covmat **without** theory uncs under the name `datacuts_theory_fitting_covmat_table`? Can I remove the attribute `theory`?

https://github.com/NNPDF/nnpdf/blob/0dbb65a96975e7e9cead0469eba1defc41f88802/validphys2/src/validphys/n3fit_data.py#L361-L367

jacoterh:
For reference: https://vp.nnpdf.science/9PHAa9esR5qks0ukpImKqQ==/

The uncertainties are currently overestimated with respect to master - more work is needed before we can merge this

scarlehoff:
> Another question: why do we save the fitting covmat without theory uncs under the name datacuts_theory_fitting_covmat_table? Can I remove the attribute theory?


We might not have a theory covmat


> The uncertainties are currently overestimated with respect to master - more work is needed before we can merge this

Are we sure we are not considering the experimental covmat twice here? Is the written down covmat only the theory part or does it contain the experimental covmat already?

jacoterh:
> Another question: why do we save the fitting covmat **without** theory uncs under the name `datacuts_theory_fitting_covmat_table`? Can I remove the attribute `theory`?

We need the theory namespace to resolve `theoryID` that is needed to construct the t0 covmat

achiefa:
I'm still browsing the diagonalisation part looking for an explanation, but I found what I suspect is a bug. Take the following function:

https://github.com/NNPDF/nnpdf/blob/a5b2d93055cf27d6678ac79ddcad8ff4ee11a0a2/validphys2/src/validphys/n3fit_data.py#L387-L427

In L424, shouldn't we apply the inverse of `diag_inv_sqrt`. In other words, shouldn't we reverse the rotation to the correlation matrix? Here it seems to me we're applying this rotation twice.

achiefa:
Another comment on the same function. In `setupfit_fitting_covmat` we returns `covmat` from `dataset_inputs_fitting_covmat`. But this latter is not diagonalised. Shouldn't we return the diagonal matrix of eigvals, which is indeed the covariance matrix in the diagonal basis?

jacoterh:
> Shouldn't we return the diagonal matrix of eigvals, which is indeed the covariance matrix in the diagonal basis?

Have a look here.  We don't use `covmat` if `diagonal_basis` is `True`. The diagonal matrix of eigenvalues is instead stored in `diagonal_rotation`.
https://github.com/NNPDF/nnpdf/blob/a5b2d93055cf27d6678ac79ddcad8ff4ee11a0a2/validphys2/src/validphys/n3fit_data.py#L622

jacoterh:
> In L424, shouldn't we apply the inverse of `diag_inv_sqrt`. In other words, shouldn't we reverse the rotation to the correlation matrix? Here it seems to me we're applying this rotation twice.

What we do here is consistent with what we documented on https://docs.nnpdf.science/n3fit/methodology.html#diagonal-basis, so not sure why you think we apply the rotation twice?

achiefa:
> > In L424, shouldn't we apply the inverse of `diag_inv_sqrt`. In other words, shouldn't we reverse the rotation to the correlation matrix? Here it seems to me we're applying this rotation twice.
> 
> What we do here is consistent with what we documented on https://docs.nnpdf.science/n3fit/methodology.html#diagonal-basis, so not sure why you think we apply the rotation twice?

Yes, my bad. You're right.

achiefa:
I found the bug. The problem was that `nnfit_theory_covmat` used a different index compared to the experimental covmat. This is of no harm in master, because `nnfit_theory_covmat` in setupfit is just for storing the theory covmat, but never added to the experimental. `produce_loaded_theory_covmat` is then in charge of loading the covmat and changing the index (ordering using runcard order). In this PR, `nnfit_theory_covmat` is consumed in vp-setupfit and added to the experimental covmat, and then the eigendecomposition was computed. That's why comparing the diagonal rotation between this PR and master gave different matrices. I'll fix this.

**Comment**:
Can I slightly change the documentation for the diagonal basis, [link](https://docs.nnpdf.science/n3fit/methodology.html#diagonal-basis). I find the notation a bit confusing.

scarlehoff:
The ordering of the covariance matrices has been our main enemy for years now 😢 

Thank you for finding this! And yes, please, any changes towards making the documentation clearer are very much appreciated.

achiefa:
Let's wait for the tests to finish. Then, I kindly ask you to check thoroughly the changes: the covmat-related code can be tricky to deal with.

**Worth flagging a latent ordering fragility**
The exp+theory covmat sum relies on both piecese being in the same order. The theory is computed at setup-time, and loaded at n3fit-time. However, the computed and loaded theory covmats (which I checked to be the same) are reindexed in different ways:

- `loaded_theory_covmat` (n3fit) -> `data_input` (runcard order)
- `nnfit_theory_covmat` (vp-setupfit) -> `procs_index_matched` (follows `groups_index`)

*As of now*, these happen to agree, but only becuase diagonal-basis/thcovmat fits force `data_grouping = "ALL" (see config.py:1825, 1850‑1853), which keeps `groups_index` in runcard order. So it's more of a coincidence rather than a guarantee.

I suspect that if someone runs with a different `metadata_group` or a runcard that interleaves processes, the two diverge. This means that the map data <-> eigenmode would become misaligned.

Maybe this can be easily fixed by reindexing `nnfit_theory_covat` to `data_input` too.

jacoterh:
> _As of now_, these happen to agree, but only becuase diagonal-basis/thcovmat fits force `data_grouping = "ALL" (see config.py:1825, 1850‑1853), which keeps `groups_index` in runcard order. So it's more of a coincidence rather than a guarantee.

But that covers all cases then, doesn't it? If one does not run with the theory covmat there is nothing to combine in the first place.

achiefa:
> > _As of now_, these happen to agree, but only becuase diagonal-basis/thcovmat fits force `data_grouping = "ALL" (see config.py:1825, 1850‑1853), which keeps `groups_index` in runcard order. So it's more of a coincidence rather than a guarantee.
> 
> 
> 
> But that covers all cases then, doesn't it? If one does not run with the theory covmat there is nothing to combine in the first place.

This is the part I'm not entirely sure of. I did local tests using dis-only runcards. Might be different with a complete set of data.

scarlehoff:
> Maybe this can be easily fixed by reindexing nnfit_theory_covat to data_input too.

Let's do that.

As @jacoterh said 

> But that covers all cases then, doesn't it? If one does not run with the theory covmat there is nothing to combine in the first place.

Which means reindexing won't change anything (because it is already in that order although for a different reason) but it will ensure that if someone adds some other grouping in the future we are protected.

jacoterh:
> Let's do that.

Perfect, I can take care of that @achiefa

jacoterh:
**Problem**: `nnfit_theory_covmat` (written by vp-setupfit) was reindexed to `procs_index_matched` order (which follows `groups_index`, i.e., dataset-grouping order). But `loaded_theory_covmat `(read back by n3fit) was reindexed to `data_input` order (runcard order). If these differ, the exp+theory covmat sum silently misaligns rows/columns.

  **Fix** (by Claude) in construction.py:
  - Added `data_input_matched_procs_index(data_input, procs_index)` — a new provider that orders procs_index according to data_input (runcard order), mirroring the [str(i) for i in data_input] reindexing done in produce_loaded_theory_covmat.
  - Changed all three `*_fitting functions (theory_covmat_custom_fitting, total_theory_covmat_fitting, user_covmat_fitting)` to take `data_input_matched_procs_index` as their dependency instead of `procs_index_matched`.

achiefa:
I did a quick check using a reduced version of the 4.0 dataset, including different experiments other than DIS. I used the same runcard for master and this PR, and here is the report:
https://vp.nnpdf.science/sUKTC7aISBiSay2pleHG5Q==/

I think this is good news? I still want to see Jaco's report, just to double check.

scarlehoff:
This one does look very good :)

achiefa:
Yep, it does. Let's wait for @jacoterh 's report. Then we can review it and (hopefully) merge it.

jacoterh:
We're all good guys, the full report comparing this PR to the alphas baseline fit from Jelle: https://vp.nnpdf.science/Ok0eP8xbQ1ykL4NBE8CPMg==/

validphys2/src/validphys/n3fit_data.py
```diff
@@ -360,7 +381,56 @@ def _inv_covmat_prepared(_hashed_dataset_inputs_fitting_covmat, diagonal_basis=T
...
+        sigma_inv = 1 / sigma
+        cormat = np.einsum("i, ij, j -> ij", sigma_inv, covmat, sigma_inv)
+
+        # diagonalise the correlation matrix
+        eig_vals, v = np.linalg.eigh(cormat)  # cormat = V @ diag(eig_vals) @ V.T
+        u = np.einsum("i, ij -> ij", sigma, v)
+        rec_covmat = u @ np.diag(eig_vals) @ u.T
+        assert np.allclose(
+            covmat, rec_covmat
+        ), "Diagonalisation failed to reproduce original covmat"
```
scarlehoff: I'm trying to do `vp-setupfit` with `nnpdf40-like.yml` but that's failing here.

validphys2/src/validphys/pseudodata.py
```diff
@@ -70,10 +110,8 @@ def read_replica_pseudodata(fit, context_index, replica):
                             5    3.117819
                             6    0.771079
     """
-    # List of length 1 due to the collect
-    context_index = context_index[0]
-    # The [0] is because of how pandas handles sorting a MultiIndex
-    sorted_index = context_index.sortlevel(level=range(1, 3))[0]
+    # Detect whether fit performed in diagonal basis
+    diagonal_basis = fit.as_input().get("diagonal_basis", True)
```
scarlehoff, line 115:
```suggestion
    # TODO: change the fit object to return diagonal basis True or False depening on the NNPDF version
    diagonal_basis = fit.as_input().get("diagonal_basis", True)
```

I (hopefully) will add this so that older fits are automatically understood as diagonal_basis: False.

validphys2/src/validphys/n3fit_data.py
scarlehoff:
```suggestion
     np.testing.assert_allclose(covmat, rec_covmat, rtol=1e-4, atol=1e-5, err_msg="Diagonalisation failed to reproduce original covmat")
```

These are the tolerances that I need to write for this test to pass. This is an experimental-only matrix which I guess it is the reason why.

What should we do in this scenario?
Crash and say "this covariance matrix is not very well defined, please set diagonal_basis: False" or accept the lower tolerance?

jacoterh: Interesting, probably you are right and the off-diagonal blocks in the case of a theory covmat help regularise the covmat. Maybe we should do the eigendecomposition block by block in the case of an experimental covmat only?

scarlehoff: Perhaps that would solve the issue, ys.

jacoterh: For now I would accept the lower tolerance and take care of my suggestion above in a separate smaller PR

scarlehoff: Ok to separate both issues but I think it is something that should be dealt with more or less urgently after merging this.

scarlehoff: I see you merged directly the suggestion, it was just a proposal I didn't checked that it matched the indentation. Also, it is a very long line, it would need a `pre-commit` run to comply with the style of the rest of the code.

scarlehoff:
I also see a lot of 

```
nnpdf/validphys2/src/validphys/covmats_utils.py:88: RuntimeWarning: overflow encountered in matmul
  return np.diag(diagonal) + corr_sys_mat @ corr_sys_mat.T
```

shouldn't we worry a bit? If not, why not?

achiefa:
> I also see a lot of
> 
> ```
> nnpdf/validphys2/src/validphys/covmats_utils.py:88: RuntimeWarning: overflow encountered in matmul
>   return np.diag(diagonal) + corr_sys_mat @ corr_sys_mat.T
> ```
> 
> shouldn't we worry a bit? If not, why not?

I've never seen this error before, and I think it should be addressed. Which runcard are you using?

scarlehoff:
The 4.0 example with the diagonal set to true

validphys2/src/validphys/n3fit_data.py
achiefa: > This is an experimental-only matrix which I guess it is the reason why.

I'm pretty sure this is the sum of both experimental and theory covmats. In fact, the matrix comes from `dataset_inputs_t0_total_covmat`

https://github.com/NNPDF/nnpdf/blob/b31822d2ae122a1a00e431b2a5874587c65a8424/validphys2/src/validphys/covmats.py#L498-L506


> Interesting, probably you are right and the off-diagonal blocks in the case of a theory covmat help regularise the covmat. Maybe we should do the eigendecomposition block by block in the case of an experimental covmat only?

I'm not sure I get this.

scarlehoff: in the 4.0 case there are no theory errors so the diagonalization happen to the experimental part only

achiefa:
Do you know the reason why `test_parallel_restart` is failing?

scarlehoff:
```
AssertionError: assert 5 == 6
```

Mmmm, 5 is not 6 indeed. Not sure, let's rerun the test.

achiefa:
Ok, is there anything left to do?

validphys2/src/validphys/n3fit_data.py
```diff
@@ -319,39 +319,60 @@ def _hashed_dataset_inputs_fitting_covmat(dataset_inputs_fitting_covmat) -> Hash
...
-        eig_vals, uT = np.linalg.eigh(cormat)
-        uT = np.einsum("i, ik -> ik", diag_inv_sqrt, uT)
-        diagonal_rotation = uT.T
-
-        ndata = len(eig_vals)
-
+        if use_thcovmat_in_fitting:
+            diagonal_basis_saved = "datacuts_theory_theorycovmatconfig_fitting_covmat_table.csv"
+        else:
+            diagonal_basis_saved = "datacuts_theory_fitting_covmat_table.csv"
```
scarlehoff: The `False` path is broken at the moment because even when `use_thcovmat_in_fitting: false`, the file saved is `datacuts_theory_theorycovmatconfig_fitting_covmat_table.csv`.

But also, this should not be set up here, this should be getting `fitting_covmat_name`.

n3fit/src/n3fit/scripts/vp_setupfit.py
```diff
@@ -171,8 +176,14 @@ def from_yaml(cls, o, *args, **kwargs):
             filter_action = 'datacuts::theory::fitting filter'
             check_n3fit_action = 'datacuts::theory::fitting n3fit_checks_action'
 
+        # Add rotation action for the total covariance matrix
+        if file_content.get('theorycovmatconfig') is not None:
+            rotation_action = 'datacuts::theory::theorycovmatconfig fitting_covmat_table'
```
scarlehoff, line 185: I think the culprit might be this set of lines. You should check that it is not None _and_ that use theory covmat in fitting blablabla is True

validphys2/src/validphys/n3fit_data.py
```diff
@@ -548,6 +622,27 @@ def fitting_data_dict(
 exps_fitting_data_dict = collect("fitting_data_dict", ("group_dataset_inputs_by_metadata",))
 
 
+@table
+def fitting_covmat_table(output_path, setupfit_fitting_covmat, data_index, diagonal_basis=True):
```
scarlehoff, line 622: Note for the future, because this had me trying to find the blabla_in_sampling bug for a longer time than I wanted, we should either not use reportengine's `@table` decorator at all for the covmat or, if we do, then regenerate the names automagically (i.e., not set them up in config or anywhere).
The current mix (which was already there before this PR fwiw) is a nightmare.

scarlehoff (commented):
The theory covmat modules are such a minefield that one day we will be tried in The Hague.

fwiw, the opposite combination to the one that failed works ok.

So:

crashes fort he fit
```
use_thcovmat_in_fitting: false
use_thcovmat_in_sampling: true
```

works:
```
use_thcovmat_in_fitting: true
use_thcovmat_in_sampling: false
```

n3fit/src/n3fit/scripts/vp_setupfit.py
achiefa: Just to double check: does this fail at the level of vpsetupfit or n3fit?

scarlehoff: `vp_setupfit` saves a file, but it is `datacuts_theory_theorycovmatconfig_fitting_covmat_table.csv`, `n3fit` then does not find the right one (which is the one without theorycovmat).
So it crashes at the level of `n3fit` but the issue is at the level of `vpsetupfit`.

achiefa: ah got it!

achiefa:
Hi @scarlehoff, now the bug you reported is fixed. Can you please check it again? If everything looks ok, we can merge it.

Edit: I dind't have time to look into the table business, but I guess that it's something we should keep in mind for the future.

scarlehoff:
Thanks! Seems to work. Let's wait for the tests then merge.

scarlehoff (approved):
*gesture of relief *

achiefa:
Btw, I've also tested this PR in my NTK stuff, and I was able to reconstruct the loss function epoch-by-epoch using the stored pseudodata in the diagonal basis.

Issue #2461, opened 2026-04-27
Fail if the runcard doesn't match the md5 as generated by vp-setupfit


scarlehoff:
Because otherwise you can generate a diagonal covmat, then change the order of the dataset (or even remove some) and all hell would break loose.

I just realized https://github.com/NNPDF/nnpdf/pull/2455#discussion_r3147250997 this is a possibility. Since we have the guardrails for that let's use them.

jacoterh:
Yes this should have high priority. I would even say this needs to be addressed first before merging #2455

PR #2462, opened 2026-04-28
Check md5 integrity against vp-setupfit in n3fit

achiefa:
This PR should address #2461. Please have a look.

Also, I was wondering whether we want to hash also the files in `filter` directory. This would be an easy fix.

### DIFF
--- a/n3fit/src/n3fit/scripts/n3fit_exec.py
+++ b/n3fit/src/n3fit/scripts/n3fit_exec.py
@@ -110,14 +110,14 @@ class N3FitEnvironment(Environment):
         self.input_folder.mkdir(exist_ok=True)
 
     def _verify_setupfit_md5(self):
-        if getattr(self, "skip_md5_check", False):
+        if sekf.skip_md5_check:
             log.warning("Skipping md5 check against vp-setupfit (--skip-md5-check is set).")
             return
 
         md5_path = self.output_path / MD5_FILENAME
         if not md5_path.exists():
             raise N3FitError(
-                f"No {MD5_FILENAME} file not found at {md5_path}. "
+                f"{MD5_FILENAME} file not found at {md5_path}. "
                 "Run vp-setupfit on this runcard before n3fit."
             )
         stored = md5_path.read_text().strip()
@@ -345,6 +345,7 @@ class N3FitApp(App):
             "--skip-md5-check",
             help="Skip the integrity check against the md5 written by vp-setupfit.",
             action="store_true",
+            default=False,
         )
         return parser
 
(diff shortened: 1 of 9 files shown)


achiefa:
I think I'm done with this. Can you please check so that we can merge it into Jaco's PR?

achiefa:
I tested this and it works fine. Should I remove the hashing of the theorycovamat?

scarlehoff:
> Should I remove the hashing of the theorycovamat?

Yes, since we are not going to recompute the theorycovmat when running n3fit.

n3fit/src/n3fit/scripts/n3fit_exec.py
```diff
@@ -103,6 +109,29 @@ def init_output(self):
         self.input_folder = self.replica_path / INPUT_FOLDER
         self.input_folder.mkdir(exist_ok=True)
 
+    def _verify_setupfit_md5(self):
+        if getattr(self, "skip_md5_check", False):
```
scarlehoff: Wouldn't it be enough with a `self.skip_md5_check`? why `getattr`?

n3fit/src/n3fit/scripts/n3fit_exec.py
```diff
@@ -103,6 +109,29 @@ def init_output(self):
...
+            log.warning("Skipping md5 check against vp-setupfit (--skip-md5-check is set).")
+            return
+
+        md5_path = self.output_path / MD5_FILENAME
+        if not md5_path.exists():
+            raise N3FitError(
+                f"No {MD5_FILENAME} file not found at {md5_path}. "
```
scarlehoff:
```suggestion
                f"{MD5_FILENAME} file not found at {md5_path}. "
```

n3fit/src/n3fit/scripts/n3fit_exec.py
achiefa: Idk, just habit. `self.skip_md5_check` will do.

achiefa:
Hi @scarlehoff, I've implemented your suggestions. Let me know.

scarlehoff:
I think you can rebase only the md5 part on top of master (instead of pointing to `diag_covmat_reproducibility`) and we merge this by itself.
It's mostly orthogonal and it can be useful already I would say.

achiefa:
Yes, I agree. I've reversed the changes inherited from `diag_covmat_reproducibility`. Now this points to master. Once the tests finish, we can merge this.

Edit: let me rebase on top of master before

achiefa:
Rebased on top of master

achiefa:
I've updated the tests, and the documentation as well. The tests are now running. Can you please check if this is what you had in mind?

scarlehoff (approved):
Thanks! And thanks for updating the docs as well.

PR #2463, opened 2026-04-29
covmat-by-reference-bug

jacoterh:
This PR addresses a bug in the construction of the total covariance matrix entering the fit:

https://github.com/NNPDF/nnpdf/blob/44a5ddf499d43775b1f0dae5b5b8f2e367e8563a/validphys2/src/validphys/covmats.py#L479-L487

The line 

https://github.com/NNPDF/nnpdf/blob/44a5ddf499d43775b1f0dae5b5b8f2e367e8563a/validphys2/src/validphys/covmats.py#L485

links `covmat` to `dataset_inputs_t0_exp_covmat ` by reference, i.e. creates an alias, such that any subsequent changes to `covmat ` also mutate the original object `dataset_inputs_t0_exp_covmat `. 

- We have checked that n3fit calls/resolves the provider `dataset_inputs_t0_total_covmat` twice before starting the fit, which means that `dataset_inputs_t0_exp_covmat ` already includes the theory covmat after the first call such that the returned object that enters the fit after the second call is the experimental covmat plus **twice** the theory covmat.
- We have run an agent and it turns out this behaviour shows up in `covmats.py` in 3 other places. This PR fixes this too. It also added a dedicated test to check whether things are safe now.
- The filter utils function has also been modified for safety, because also here we modify the argument directly. I don't think this matters too much because the filter utils are not part of report engine, but better safe than sorry.

Report comparing the impact of the bug before and after the fix: https://vp.nnpdf.science/Gx02F1SqSaKSvTtOLhXFZg==/

### DIFF
--- a/.github/workflows/all_tests_nnpdf.yml
+++ b/.github/workflows/all_tests_nnpdf.yml
@@ -78,7 +78,7 @@ jobs:
       - name: Build recipe
         shell: bash -l {0}
         run: |
-          conda -n base install conda-build --yes
+          conda install -n base conda-build --yes
           conda build -q conda-recipe --package-format=1
       - name: Keep conda package as artifact
         if: ${{ !cancelled() }}
--- a/.github/workflows/all_tests_nnpdf.yml
+++ b/.github/workflows/all_tests_nnpdf.yml
@@ -78,7 +78,7 @@ jobs:
       - name: Build recipe
         shell: bash -l {0}
         run: |
-          conda install conda-build --yes
+          conda -n base install conda-build --yes
           conda build -q conda-recipe --package-format=1
       - name: Keep conda package as artifact
         if: ${{ !cancelled() }}
--- a/nnpdf_data/nnpdf_data/filter_utils/utils.py
+++ b/nnpdf_data/nnpdf_data/filter_utils/utils.py
@@ -284,13 +284,12 @@ def concat_matrices(rows, columns, list_of_matrices):
         the final, fully concatenated matrix row by row.
 
     """
-    for i in range(len(list_of_matrices)):
-        list_of_matrices[i] = np.array(list_of_matrices[i])
+    matrices = [np.array(matrix) for matrix in list_of_matrices]
     col_list = []
     for i in range(rows):
         row_list = []
         for j in range(columns):
-            row_list.append(list_of_matrices[j + i * columns])
+            row_list.append(matrices[j + i * columns])
         col_list.append(np.concatenate(tuple(row_list), axis=1))
     final_mat = np.concatenate(tuple(col_list), axis=0)
     final_mat_list = []
@@ -438,6 +437,7 @@ def check_xq2_degenearcy(Q2, x):
             unique kinematics are: {unique_pairs.shape[1]}, original size: {size}"""
         ) from exc
 
+
 def uncert_skip_variant(source_file, skip_file, uncert_file, uncert_name, remove_source=True):
     r"""
     Create two new uncertainty files, one where the specified uncertainty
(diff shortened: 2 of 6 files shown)


achiefa:
Hi @jacoterh, thanks for opening this PR. I'll have a look asap!

nnpdf_data/nnpdf_data/filter_utils/utils.py
```diff
@@ -284,13 +284,12 @@ def concat_matrices(rows, columns, list_of_matrices):
         the final, fully concatenated matrix row by row.
 
     """
-    for i in range(len(list_of_matrices)):
-        list_of_matrices[i] = np.array(list_of_matrices[i])
+    matrices = [np.array(matrix) for matrix in list_of_matrices]
```
achiefa: I'm OK with this. I just want to check if this affects any of the datasets. `regenerate_commondata` should tell us if there's any difference.

validphys2/src/validphys/tests/test_covmats.py
```diff
@@ -153,3 +161,24 @@ def test_single_datapoint(single_data_single_point_internal_cuts_config):
     # Ensure the dataset is only a single datapoint
     assert ld.ndata == 1
     ld.systematic_errors(t0_predictions)
+
+
```
achiefa, line 165: The test is fine per se, but I'd like to hear @scarlehoff's opinion. In my view, the test prevents us from repeating the same mistake, although now that we are aware of the issue it might seem a bit superfluous. Still, I think we should keep it as a reminder for the future and as a signal for new members.

achiefa (commented):
It looks good to me. I just want to hear whether @scarlehoff has something to say on this matter.

jacoterh:
Why do we get a failing test because of some eko incompatibility?

validphys2/src/validphys/tests/test_covmats.py
scarlehoff: Many of our tests are built on the blood of our previous mistakes. I'm happy with superflous tests.

scarlehoff:
It looks like the macos workers in github have changed (and have python 3.14 now? we should not be testing that)

achiefa:
Does this mean that we can merge this? Is it possible to change the Python version un the MacOS runners?

scarlehoff:
No, it means that we need to look over the github docs to understand what changed (or what are we doing wrong in the workflow for mac os)

.github/workflows/all_tests_nnpdf.yml
```diff
@@ -78,7 +78,7 @@ jobs:
       - name: Build recipe
         shell: bash -l {0}
         run: |
-          conda -n base install conda-build --yes
+          conda install -n base conda-build --yes
```
scarlehoff, line 81: If I was both clever and handsome I would be too powerful for this world.
Anyway, this can be merged I guess.

(I've done this change because reading https://github.com/conda-incubator/setup-miniconda#important I thought macos, which has its own flavour of bash due to licence issues, might be messing with the environment activation, but I don't fully know _why_ this fixes it other than "apple bad")

nnpdf_data/nnpdf_data/filter_utils/utils.py
scarlehoff: Since we are concatenating, maybe we don't actually need to do a copy for these ones though 😅

scarlehoff (approved):
Thanks for this. Please make sure to put here also the report you did comparing the fit with and without the bugfix.

nnpdf_data/nnpdf_data/filter_utils/utils.py
jacoterh: Yeah it's not part of reportengine so modifying directly the argument should be safe upon repeated use - I've restored the version from master

scarlehoff: tbf I think also the version from master is an unnecessary copy, in any case, please merge when you want

achiefa (approved):
Also green light from me. Thanks @jacoterh !

jacoterh:
Somehow I'm not authorised to merge? I see "Merging is blocked. You're not authorised to push to this branch". Weird.

scarlehoff:
I wanted to put strong protections to master. Perhaps I exaggerated :__  I'll double check

PR #2467, opened 2026-05-11
Add theory cards for hadronic N3LO QCD, theory ID 41000100

kamillaurent:
This PR is intended to add the `theory_cards` for generating N3LO QCD FK Tables of hadronic processes. The theory ID of the central scale is `41000100`, so the added files are `410001*.yaml`.


felixhekhorn (changes requested):
- I think the agreement is - and that seems to be the case on master - to NOT commit the "sub-theory" cards, i.e. the cards for the individual FONLL runs, so you should remove those; e.g. 41_000_000_00
- I see this is also the case for 41_000_000 , but why is there a gap in numbers to the SV cards? i.e. central = 00, SV = 10, 11, 12, etc. so why not 01, 02, etc.? there might be a legitimate reason

scarlehoff:
Indeed. Please do not commit the nfonll runcards.

kamillaurent:
I removed the nfonll runcards, @felixhekhorn I assume the gap in numbers is to avoid confusion with the FONLL runcards, which are {01..06}, I used {10..15} following the convention used for the scale variation of 41000000.

felixhekhorn:
> I removed the nfonll runcards

dito for 41_041_000_00 and friends

nnpdf_data/nnpdf_data/theory_cards/.gitignore
```diff
@@ -0,0 +1,6 @@
+410000*01.yaml
```
scarlehoff: Please remove the .gitignore. 

If you want to have a local gitignore in a particular folder you can always add `.gitignore` and it will ignore itself.

(ex. this one right now would ignore `41000002.yaml` which is perfectly legit.

scarlehoff:
Thanks. I've rebased and will merge once tests pass.

felixhekhorn (approved):

kamillaurent:
@scarlehoff can this branch be merged or some other change is needed?

PR #2468, opened 2026-05-12
Implementation of CMS_Z0J_13TeV - general cleanup

enocera:
This PR replaces and supersedes part of #2360 . This PR contains the cleaned-up implementation of the CMS ZpT 13 TeV data set. Comments in #2360 relevant to this data set are taken into account For the time being, standard cuts due to lack of resummation and of EW corrections are implemented.

The values of the chi2 are as follows (with `NNPDF40_nnlo_as_01180` as input PDF set)
Exp chi2: 1.354
Exp+th chi2: 0.7604
Exp+th+pdf chi2: 0.7573

Here are the data/theory comparison plots. For once, all seems good to me.
plot_bin4
plot_bin3
plot_bin2
plot_bin1


scarlehoff:
I was about to ask you about https://github.com/NNPDF/nnpdf/pull/2360 you read my mind!

validphys2/src/validphys/cuts/filters.yaml
```diff
@@ -279,6 +279,18 @@
...
+- dataset: CMS_Z0J_13TEV_PT-M
+  reason: Avoid the region where resummation effects become important.
+  rule: "pT >= 30"
+
+- dataset: CMS_Z0J_13TEV_PT-M
+  reason: Avoid the region where EWK corrections are important.
+  rule: "pT <= 170"
+
+- dataset: CMS_Z0J_13TEV_PT-M
+  reason: Remove first invariant mass bin which is completely off.
```
scarlehoff: Did we have an idea why?

enocera: No I don't. And I don't see any reason why the NNLOjet runcard should be wrong. At this point life is too short, and I would just impose a cut, as I did. If you want I can add "which is completely off for unknown reasons".

scarlehoff: Yes please.

enocera: Done.

nnpdf_data/nnpdf_data/commondata/CMS_Z0J_13TEV/metadata.yaml
```diff
@@ -0,0 +1,63 @@
...
+    variables:
+      pT:
+         description: "lepton pair transverse momentum"
+         label: '$p_T$'
+         units: "GeV"
+      m_ll:
+         description: "Z boson mass squared"
+         label: '$m^2_{\ell \ell}$'
+         units: "GeV^2"
+    file: kinematics.yaml
```
scarlehoff, line 49: I don't think the m_ll variable is actually squared.

scarlehoff (approved):
Please check hte m_ll comment. Looking at kinematics.yaml they look like GeVs

nnpdf_data/nnpdf_data/commondata/CMS_Z0J_13TEV/metadata.yaml
enocera: You're totally right. Fixed. I am regenerating the data-theory comparison plots.

enocera:
> Please check hte m_ll comment. Looking at kinematics.yaml they look like GeVs

You're right. These are GeV. Just look at the Z-mass peak in the data-theory comparison plots. The square does not make sense at all. Apologies.

nnpdf_data/nnpdf_data/commondata/CMS_Z0J_13TEV/metadata.yaml
enocera: Here are the plots with the correct label.
plot_bin1
plot_bin2
plot_bin3
plot_bin4

scarlehoff:
fwiw, I would actually prefer to keep m_ll and use that in process options (so it needs to be changed to accept it)

It's easier to understand the numbers/plots/cuts that way (but no strong opinion).

enocera:
> fwiw, I would actually prefer to keep m_ll and use that in process options (so it needs to be changed to accept it)
> 
> It's easier to understand the numbers/plots/cuts that way (but no strong opinion).

OK, so let me do that.

enocera:
@scarlehoff If you're happy with the changes, I'll proceed to merge this PR, and then rebase #2469 .

scarlehoff (approved):

PR #2469, opened 2026-05-12
ATLAS_Z0J_13TEV - bugfix

enocera:
This PR is a re-implementation of the ATLAS_Z0J_13TEV data set. In comparison to the previous implementation, there are two improvements.
1) A bug fix. We realised that the original implementation was bugged, because. In the previous version two HepData tables were used, separately for the central value and for the uncertainties (Tables 26 and 5, respectively). However, it was not realised that the two differ for the bin normalisation, which is, on average, of 2 GeV. Uncertainties were therefore, on average, a factor of 2 smaller than what they should be. 
2) An extension. In the previous version only a subset of pT bins were implemented. The full pT spectrum is now implemented. 

The bugged version, used for the pheno paper, is removed, but can be retrieved from the github history.

The chi2 is as follows
Exp chi2: 57.13
Exp+th chi2: 8.509
Exp+th+pdf chi2: 8.424
and here is a data-theory comparison (with shift). For the time being, there are cuts to remove the small pT (resummation) and large pT (EWK corrections) regions. 
plot_ATLAS_ZpT_13TEV

My opinion is that the unimpressive chi2 is a consequence of the very small experimental uncertainties. I do no tthink that there is anything wrong with this implementation. Results are pretty much in line with the experimental paper, see in particular Fig. 6 of  https://arxiv.org/pdf/1912.02844.

This PR superseded and replaces the second part of #2360 .


scarlehoff (approved):
@enocera given the issues with the jets and the statistical uncertainty being a bit underestimated, I wonder whether it is the same case here and the sub-% I got from NNLOJET is also underestimated. In Fig. 6 of the experimental paper there are less ups and downs.

enocera:
> @enocera given the issues with the jets and the statistical uncertainty being a bit underestimated, I wonder whether it is the same case here and the sub-% I got from NNLOJET is also underestimated. In Fig. 6 of the experimental paper there are less ups and downs.

I can implement a variant, as we have for the 8 TeV data, in which I add a 1% uncorrelated MC uncertainty for the time being.

scarlehoff:
If we do so, we should do them for all. Let's leave it like this for now. If/when I have time I'll check the chi2 including the MC uncertainty as reported by NNLOJET (which will be around 0.5% I think) for the V+J sets.

PR #2470, opened 2026-05-13
Higher Mellin moments in sumrules.py

achiefa:
I thought this could be useful, for instance in comparing with lattice results.


validphys2/src/validphys/sumrules.py
```diff
@@ -24,6 +24,19 @@
 LIMS = [(1e-9, 1e-5), (1e-5, 1e-3), (1e-3, 1)]
 POL_LIMS = ((1e-4, 1e-3), (1e-3, 1))
 
+# For high Mellin moments x**n * f(x) the integrand is dominated by the
+# large-x region; slice (1e-3, 1) further so quad does not have to span
+# many orders of magnitude in a single sub-interval.
```
scarlehoff: For me this comment and the code below are at odds.
We don't need to span many orders of magnitude so we break it down even further into each slice of order of magnitude.

achiefa: I'm not sure I understand this. The intervals (1e-3, 1e-2) and (1e-5, 1e-4) span two different order of magnitudes. So if we included these two sub-intervals into one integration, than the quadrature would range between two different orders of magnitude and the resolution of the integration would be affected. Am I wrong?

scarlehoff: You are not, that's why we separate in limits for the others. But I don't understand why the separation is even finer here if the lower subintervals are less important.

achiefa: Ah right, now I got your point. If the small-x sub-intervals are really subdominant, why are we slicing them even further. Honestly I didn't spend too much time thinking about it: I just took the original intervals and added more sub-intervals at large-x. We can collpse the small-x ones into one sub-interval.

scarlehoff: Ok, so, next question. Do the intervals at large x make a difference at all?
At low-x it does because of the different orders of magnitude as you said. For large-x instead, if you are sampling linearly there's no much of a difference is there? Or does it help `quad` a lot?

edit: trying to understand the logic here. I just found it confusing.

achiefa: Here is a concrete example:

Screenshot 2026-05-15 at 11 19 58

So the small-x region is negligible, ok. The reson for the different sub-invervals at large-x is more phenomenological than algorithmic. For example, I was interested in seeing which region at large-x contributes the most. I think this is relevant for lattice determinations, that's why I split the large-x interval. Note that the user can always specify a custom list of sub-intervals. So I'm fine with removing the global `MELLIN_LIMS` and using `LIMS` as default. Regarding the impact on `quad` this I haven't checked.

scarlehoff: > The reson for the different sub-invervals at large-x is more phenomenological than algorithmic

Okok. This is enough for me tbh. I just wanted to understand why the choice. If you could add that to the comment that'd be great.
I'm fine with keeping the subinterval at both small-x and large-x, it doesn't really make a big difference. I was just very confused in general because you kept the small one while saying that they were not relevant and added a few more in a region in a way in which (numerically-integrating-wise) wouldn't make a difference.

But "I'm interested on this particular partial split" is good enough for me.

validphys2/src/validphys/sumrules.py
```diff
@@ -90,6 +103,24 @@ def f(x, lpdf, Q):
     return f
 
 
+def _make_mellin_moment_integrand(fldict, n: int):
+    """Make an integrand for the n-th Mellin moment ``∫ x**n · f(x) dx`` of the linear
+    combination of defined by ``fldict``.
+
+    Note that LHAPDF returns ``x*f(x)``, so we multiply by ``x**(n-1)``. Reduces to
+    :py:func:`_make_pdf_integrand` for n=0 and to :py:func:`_make_momentum_fraction_integrand`
+    for n=1.
```
scarlehoff: I would propose merging these three functions and just apply the right value of `n` in each case.

scarlehoff (approved):
lgtm, but if you were to merge these three functions with

```
_make_pdf_integrand(fldict) = make_mellin_moment_integrand(fldict, n=0)
_make_momentum_fraction_integrand(fldict) = make_mellin_moment_integrand(fldict, n=1)
```

I would be very happy.

achiefa:
Once tests pass, this can be merged.

PR #2478, opened 2026-06-01
Add `neopdf` PDF interpolation backend

Radonirinaunimi:
The following introduces a PDF backend module so that one can in principle support any PDF interpolation library. Currently, LHAPDF and [NeoPDF](https://github.com/QCDLab/neopdf) are supported and they can be used interchangeably given that they are numerically equivalent (bit-for-bit). (This eventually paves the way to a Conda-free dependency).


scarlehoff:
Thanks for this.

My main comment is that the validphys is already provider agnostic and the backend selecting code already exists: https://github.com/NNPDF/nnpdf/blob/master/validphys2/src/validphys/lhapdf_compatibility.py

I think what you did should be easy to modify so that it works similarly to pdfflow, without breaking it or adding a third way of changing the backend.

As to how to change the backend, I would put it in the profile. With pdfflow is just a "in case of failure finding lhapdf". Then we can eventually change the default.

(the conda dependence will be there 4ever for pandoc reasons I'm afraid :__)

validphys2/src/validphys/neopdfset.py
```diff
@@ -0,0 +1,107 @@
...
+        q2grid = np.asarray(qgrid) ** 2
+        xgrid = np.asarray(xgrid)
+        pids = list(flavors)
+        nx, nq, nfl = len(xgrid), len(qgrid), len(pids)
+
+        results = []
+        for member in self.members:
+            vals = np.array(member.xfxQ2s(pids, xgrid, q2grid))
+            results.append(vals.reshape(nfl, nx, nq))
+        return np.array(results)
```
scarlehoff: This is repeating actually a lot of codes (which I'm strongly against)

As far as I can see you can do the same I did for pdfflow but the wrapper would be just 4 lines or so instead of an entire class.

Then the make_pdf in pdf_compatibility is the only place in which a choice needs to be made.

Most importantly, `core.py` should have no changes.

Radonirinaunimi: Thanks @scarlehoff for having looked into this already. This is still very much in WIP (although things are already in place and working) and thus I am not entirely convinced yet about how to best implement the interface. However, I fully agree with you about the redundancy part and started addressing this in  [484110](https://github.com/NNPDF/nnpdf/pull/2478/commits/4484110e8fd4b259565eda5ce1592143c490183f). Now, the unnecessary modules/codes have been removed.

> As to how to change the backend, I would put it in the profile. With pdfflow is just a "in case of failure finding lhapdf". Then we can eventually change the default.

Yes, I agree with this.

> (the conda dependence will be there 4ever for pandoc reasons I'm afraid :__)

Yeah, which we may want to get rid of at some point... although the alternative is not so clear.

pyproject.toml
```diff
@@ -112,6 +114,7 @@ tests = ["pytest", "pytest-mpl", "hypothesis", "pytest-cov"]
 docs = ["sphinxcontrib-bibtex", "sphinx-rtd-theme", "sphinx", "tabulate"]
 qed = ["fiatlux"]
 nolha = ["pdfflow", "lhapdf-management"]
+neopdf = ["neopdf-hep"]
```
scarlehoff:
```suggestion
neopdf = ["neopdf-hep", "lhapdf-management"]
```

I haven't maintained `pdfflow` in a while and you are actively maintaining `neopdf`. Perhaps it makes sense to change `nolha` to `pdfflow` and make `neopdf` be `nolha`.

validphys2/src/validphys/lhapdf_compatibility.py
```diff
@@ -115,6 +167,22 @@ def make_pdf(pdf_name, member=None):
     --------
         list(pdf_sets)
     """
+    backend = os.environ.get(_BACKEND_ENV_VAR, "lhapdf").lower()
+
+    if backend not in _VALID_BACKENDS:
```
scarlehoff, line 193: fwiw, to get the information from the profile just `get_nnpdf_profile()` is enough (from `nnpdf_data::utils`)

I'm not sure right now whether one can add variables at will without changing that function (in any case it would be trivial). The important thing is, don't add a new default to `get_nnpdf_profile` for `pdf_backend` but rather add explicitly to `validphys/nnprofile_default.yaml` `pdf_backend: lhapdf`.

The goal I have in mind is that people can start using `neopdf` there if they like and in a few months we try to swap the default and see what happens.

validphys2/src/validphys/lhaindex.py
```diff
@@ -12,19 +12,28 @@
...
 
 
+def _paths():
+    """Return the data paths for the currently active PDF backend."""
+    if _active_backend() == "neopdf":
+        import neopdf
+
+        return neopdf.paths()
+    return lhapdf.paths()
+
```
scarlehoff, line 25: Since the function is still paths() I think it would be better if lhapdf_compatibility just did lhapdf = neopdf

(or better, within lhapdf_compatibility do a pdf_backend or whatever that points to lhapdf or neopdf)

Otherwise the imports to neopdf (and thus thing to fix and change) would be scattered across the code.

Radonirinaunimi: You mean along the lines of [c59b5bb](https://github.com/NNPDF/nnpdf/pull/2478/commits/c59b5bbe2fb4f5abb7a5c863fed333638da8706c)?

scarlehoff: Yes. More generally, neopdf (like lhapdf now, up to things that might not have been cleaned up yet) should be only imported in thst file and should be the only file that knows there is more than one backend.

(like n3fit with keras, where all torch/tensorflow conditions are within the backend module)

Radonirinaunimi: Yes, I get the idea. Will look into this more asap.

PS: I resolved the conversations above because I think they were addressed. If you think some of them have not been addressed, please re-open again.

Radonirinaunimi: So, actually, this has already been resolved and AFAIC this is now ready.

(As opposed to what I mentioned during the CM, this is should go before #2514 because having the following changes going into that PR makes life _infinitely_ easier on an HPC systems where a pure python env is enough...)

.github/actions/set_neopdf_datapath/action.yml
```diff
@@ -0,0 +1,12 @@
...
+# Point NeoPDF at the directory where lhapdf-management stores PDF sets,
+# so it reads local files instead of downloading from the LHAPDF servers.
+
+runs:
+  using: "composite"
+  steps:
+    - name: Point NeoPDF at the LHAPDF data directory
+      shell: bash -l {0}
+      run: |
+        echo "NEOPDF_DATA_PATH=$(python -c 'from lhapdf_management.configuration import environment; print(environment.datapath)')" >> $GITHUB_ENV
```
scarlehoff: I added https://github.com/scarlehoff/lhapdf_management/issues/8 which I guess is relevant here (god knows when I will do it)

validphys2/src/validphys/tests/test_neopdfset.py
```diff
@@ -0,0 +1,318 @@
...
+neopdf = pytest.importorskip("neopdf", reason="neopdf not installed")
+
+# In view of making LHAPDF Optional in the future.
+try:
+    import lhapdf as _lhapdf
+
+    _lhapdf.setVerbosity(0)
+    HAS_LHAPDF = True
+except ModuleNotFoundError:
+    HAS_LHAPDF = False
```
scarlehoff: This should not be necessary, it is already part of conftest.

validphys2/src/validphys/tests/test_neopdfset.py
```diff
@@ -0,0 +1,318 @@
...
+
+requires_lhapdf = pytest.mark.skipif(not HAS_LHAPDF, reason="lhapdf not installed")
```
scarlehoff: This should be moved to conftest.

validphys2/src/validphys/tests/test_neopdfset.py
```diff
@@ -0,0 +1,318 @@
...
+
+PDF_NAME = "NNPDF40_nnlo_as_01180"
```
scarlehoff: This should be imported from conftest.

validphys2/src/validphys/tests/test_neopdfset.py
```diff
@@ -0,0 +1,318 @@
...
+        for fl in [21, 1, 2, -1]:
+            if fl not in neo_pdfset_t0.flavors:
+                continue
+            neo_val = neo_pdfset_t0.xfxQ(X_VALUE, Q_VALUE, n=0, fl=fl)
+            lha_val = lha_pdfset_t0.xfxQ(X_VALUE, Q_VALUE, n=0, fl=fl)
+            np.testing.assert_allclose(neo_val, lha_val, **CROSS_BACKEND_TOL)
+
+
+class TestNeoPDFSetInterface:
+    """The following simpy checks that that ``NeoPDFSet`` exposes the same
```
scarlehoff:
```suggestion
    """The following simply checks that that ``NeoPDFSet`` exposes the same
```

I was wondering why did you need `simpy` to do this check 😆

validphys2/src/validphys/tests/test_neopdfset.py
```diff
@@ -0,0 +1,318 @@
...
+    def test_xfxQ_all_members(self, neo_pdfset, lha_pdfset, fl):
+        if fl not in neo_pdfset.flavors:
+            pytest.skip(f"pid {fl} not in set")
+        for n in range(neo_pdfset.n_members):
+            neo_val = neo_pdfset.xfxQ(X_VALUE, Q_VALUE, n=n, fl=fl)
+            lha_val = lha_pdfset.xfxQ(X_VALUE, Q_VALUE, n=n, fl=fl)
+            np.testing.assert_allclose(neo_val, lha_val, **CROSS_BACKEND_TOL)
+
+    @pytest.mark.parametrize("x", [1e-5, 1e-3, 0.1, 0.5, 0.9])
+    @pytest.mark.parametrize("Q", [1.7, 10.0, 100.0])
```
scarlehoff: And same for the rest of the class: given that you are using already constants at the top, use those for the parametrization. They are mostly the same, and having a few more tests of this type is cheap.

validphys2/src/validphys/tests/test_neopdfset.py
```diff
@@ -0,0 +1,318 @@
...
+        from validphys.lhapdf_compatibility import InvalidPDFBackend, make_pdf
+
+        with pytest.raises(InvalidPDFBackend, match="Unknown backend"):
+            make_pdf(PDF_NAME)
+
+    def test_neopdf_single_member_returns_neopdf_member(self, monkeypatch):
+        monkeypatch.setenv("NNPDF_PDF_BACKEND", "neopdf")
+        from validphys.lhapdf_compatibility import _NeoPDFPDF, make_pdf
+
+        members = make_pdf(PDF_NAME, member=0)
```
scarlehoff: I would test this with a member different from 0 tbh, 0 is already tested elsewhere. Or test 0 and another one.

validphys2/src/validphys/lhapdf_compatibility.py
```diff
@@ -25,17 +30,56 @@
...
+    """
+    backend = os.environ.get(_BACKEND_ENV_VAR)
+    if backend is None:
+        try:
+            from nnpdf_data.utils import get_nnpdf_profile
+
+            backend = get_nnpdf_profile().get("pdf_backend", "lhapdf")
+        except Exception:
+            backend = "lhapdf"
+    return backend.lower()
```
scarlehoff: The `_VALID_BACKENDS` check should be done here.

validphys2/src/validphys/lhapdf_compatibility.py
```diff
@@ -25,17 +30,56 @@
...
+    Resolution order (highest priority first):
+    1. ``NNPDF_PDF_BACKEND`` environment variable
+    2. ``pdf_backend`` key in the NNPDF profile (``nnprofile.yaml``)
+    """
+    backend = os.environ.get(_BACKEND_ENV_VAR)
+    if backend is None:
+        try:
+            from nnpdf_data.utils import get_nnpdf_profile
+
+            backend = get_nnpdf_profile().get("pdf_backend", "lhapdf")
```
scarlehoff: Please, modify the `get_nnpdf_profile` so that `pdf_backend` defaults to `lhapdf` and remove the try-except here (also because we don't want a try-except with exception)

validphys2/src/validphys/lhapdf_compatibility.py
```diff
@@ -25,17 +30,56 @@
...
+def active_pdf_module():
+    """Return the module providing the LHAPDF-like set-management API
+    (``paths()``, ``pathsPrepend()``, ...) for the currently active PDF
+    backend.
+    """
+    if _active_backend() == "neopdf":
+        import neopdf
+
+        return neopdf
+    return lhapdf
```
scarlehoff: In my view, this should not be a function, but just a module-level constant defined at import time (if you want, do it lazily).

Here, if `lhapdf` is not installed, you are returning `lhapdf-management` because `lhapdf` has ben made equal to `lhapdf_management` at this point. All usages of `lhapdf` through here should then be management stuff.

Anyway, my point is that this header needs to be modified so it doesn't the same thing a few different times. You should decide when using `neopdf` who does the management? 
    a) If it is `lhapdf-management` then you can skip this `active_backend` function and leave the if-else at the top (since neopdf is lhapdf-compatible you'll be fine even if both are there)
    b) If it is `neopdf` then do a top-level check for the profile and on first import one of the three (`lhapdf-management`, `neopdf` or `lhapdf`) should be `active_pdf_backend` regardless of the rest.

validphys2/src/validphys/lhapdf_compatibility.py
```diff
@@ -115,11 +200,27 @@ def make_pdf(pdf_name, member=None):
     --------
         list(pdf_sets)
     """
-    if USING_LHAPDF:
+    backend = _active_backend()
+
+    if backend not in _VALID_BACKENDS:
```
scarlehoff: The check should've already been done at this point.

validphys2/src/validphys/lhapdf_compatibility.py
```diff
@@ -115,11 +200,27 @@ def make_pdf(pdf_name, member=None):
...
+        from neopdf.pdf import PDF as _NeoPDF
+
+        members = _NeoPDF.mkPDFs(pdf_name)
+        if member is None:
+            return [_NeoPDFPDF(m) for m in members]
+        return [_NeoPDFPDF(members[member])]
+
     pdf_meta = lhapdf.load_pdf_meta(pdf_name)
     if member is None:
         return [_PDFFlowPDF(pdf_meta, m) for m in range(len(pdf_meta))]
```
scarlehoff: Please, make this function into if-else case-like statements. Right now it is a weird mix.

Now there is a backend key it should return whatever the user asked for.

pyproject.toml
```diff
@@ -110,7 +113,9 @@ pymongo = {version = "<4", optional = true}
 tests = ["pytest", "pytest-mpl", "hypothesis", "pytest-cov"]
 docs = ["sphinxcontrib-bibtex", "sphinx-rtd-theme", "sphinx", "tabulate"]
 qed = ["fiatlux"]
-nolha = ["pdfflow", "lhapdf-management"]
+pdfflow = ["pdfflow", "lhapdf-management"]
+neopdf = ["neopdf-hep", "neopdf-cli", "lhapdf-management"]
+nolha = ["neopdf-hep", "lhapdf-management"]
```
scarlehoff: Remove the `nolha`.

Also, why do we need the `neopdf-cli` in one case and not the other. Do we need it at all?

scarlehoff (commented):
The main change requested is that `lhapdf_compatibility` should honor the backend selected by the user. Before the pdfflow-lhapdf_management part was just a fallback.
Now if the user selects lhapdf, and there is no lhapdf, bad luck. Similarly if the user selects pdfflow/neopdf.


One thing to check is how much faster it is to use neopdf than lhapdf when doing a report with a lot of PDF calls (like luminosity plots).

I see you left a few python loops within the `_NeoPDFPDF` and perhaps those can be lifted to the rust side?

Radonirinaunimi:
Hi @scarlehoff, I believe I addressed all of your comments (as far as I understood them).

One of the main changes, perhaps, is commit [e661e14](https://github.com/NNPDF/nnpdf/pull/2478/commits/e661e146fd3b19d1e8ae5d9028daf9ee16494010) where I defined how the hierarchy of backend should be set. Before, if LHAPDF is not installed and no alternative backend is specified, then it falls silently to `pdfflow` although it is not installed at all. So now, if that's the case, then it raise an error. The consequence of this is that before one was still able to run `postfit` even if none of the backends are installed, but now that's no longer possible. And hence why the `run_pytorch` fail (easily fixable though).

Regarding the `neopdf` loop in `_neoPDFPDF._xfxQ_all_pid`, there is actually already a Rust interface for that, which I already implemented. For a quick benchmark with $\mathcal{O} (10^{16})$ points, NeoPDF is ~6 times faster during PDF loading and ~8 times faster during evaluation than LHAPDF. The situation about the luminosity computation is, however, completely different because the integration is done per point. I actually had a quick look at this and it seems that one can improve this by at least 10 times by vectorizing (but I will do that in a separate PR).

One more thing, I also fixed a minor bug in `N3LHAPDFSet.grid_values` which arises in the case `len(xgrid) != len(qgrid)`.

scarlehoff:
> One more thing, I also fixed a minor bug in N3LHAPDFSet.grid_values which arises in the case len(xgrid) != len(qgrid).

We never have that situation right? (just to understand whether the bug had an effect at some point)

validphys2/src/validphys/lhapdf_compatibility.py
```diff
@@ -25,17 +30,61 @@
...
+    """Return the active PDF backend.
+
+    Resolution order (highest priority first):
+    1. ``NNPDF_PDF_BACKEND`` environment variable
+    2. ``pdf_backend`` key in the NNPDF profile (``nnprofile.yaml``), which
+       defaults to ``lhapdf`` (see ``nnprofile_default.yaml``)
+    """
+    backend = os.environ.get(_BACKEND_ENV_VAR)
+    if backend is None:
+        from nnpdf_data.utils import get_nnpdf_profile
```
scarlehoff, line 54: Why is this imported here and not at the top?

validphys2/src/validphys/lhapdf_compatibility.py
```diff
@@ -1,13 +1,18 @@
 """
-    Module for LHAPDF compatibility backends
+Module for LHAPDF compatibility backends
 
-    If LHAPDF is installed, the module will transparently hand over everything to LHAPDF
-    if LHAPDF is not available, it will try to use a combination of the packages
-        `lhapdf-management` and `pdfflow`
-    which cover all the features of LHAPDF used during the fit (and likely most of validphys)
+If LHAPDF is installed, the module will transparently hand over everything to LHAPDF.
+If LHAPDF is not available, it will try to use a combination of the packages
```
scarlehoff: This is no longer true. If LHAPDF is not installed, and NNPDF_PDF_BACKEND (or profile::pdf_backend) is not set, then vp won't work.

validphys2/src/validphys/lhapdf_compatibility.py
```diff
@@ -115,12 +210,29 @@ def make_pdf(pdf_name, member=None):
...
+        members = _NeoPDF.mkPDFs(pdf_name)
+        if member is None:
+            return [_NeoPDFPDF(m) for m in members]
+        return [_NeoPDFPDF(members[member])]
+    elif backend == "pdfflow":
+        pdf_meta = lhapdf.load_pdf_meta(pdf_name)
+        if member is None:
+            return [_PDFFlowPDF(pdf_meta, m) for m in range(len(pdf_meta))]
+        return [_PDFFlowPDF(pdf_meta, member)]
+    else:
```
scarlehoff: Since they all return you can remove the else and save an indentation below.

scarlehoff (commented):
Thanks. A few small comments. I think this is good to go but I want to test it before. I'll do it asap.

validphys2/src/validphys/lhapdf_compatibility.py
Radonirinaunimi: I believe we still need this fallback in case no backend is specified (either in the profile or in the env var). It already happened to me a few times and it is much better to panic here.

scarlehoff: Yes, we need it! I meant the else in the conditional. Since all the ifs are returning you can put the raise out of the else (but leaving the exception)

Radonirinaunimi: Sorry for the late reply, quite busy lately...

Ah, right! I didn't check exactly when I wrote the reply 🙈 Now addressed. I also resolved the outdated comments above.

validphys2/src/validphys/lhapdf_compatibility.py
Radonirinaunimi: There was no specific reason TBH... As far as that function (`_active_backend`) is actually concerned, it can be removed... Personally, I don't have a preference.

Radonirinaunimi:
@jacoterh, @kamillaurent, this is now (close to) final so if you could test/play with that'd be great.

validphys2/src/validphys/lhapdf_compatibility.py
scarlehoff: Better at the top, everything more organized and easier to understand (that said, if it can be removed, also ok)

PR #2480, opened 2026-06-03
Regression stability + new regression tests

### LINKED ISSUE #2464: Improve reproducibility of regression tests (and add one with a theory covmat)
We have two issues with the regression tests,

the first one, that they are very sensitive to very small changes. There are at least two families (not sure what the differences are) of workers in github, and there are two possible results for some of these tests.
As far as I can see this is the result of numerical differences that grow (and a fit means many numbers being updated for many epochs) and testing it is pretty much a nightmare (because it depends on hitting one of the "bad" workers during the test).
I don't have a good idea for this right now I'm afraid.

The second is, adding a theory covmat. We haven't done this because it is extremely expensive on the worker and cannot afford it (we hit the 6GB limit, if you look at the workflows I'm already removing files inside the docker image to free space...).
We will need to 1) probably remove even more files 2) upload the result of `vp-setupfit` for every regression test to the server so that it can be downloaded directly from the nnpdf server without running it in the worker.
This would've caught bugs like the one fixed by #2463 and now that the theory covmat is our default, this is important.
And would be made much better once #2461-#2462 are done since that ties the vp-setupfit with the n3fit run through the md5 file.

I've put this at the top of my to-do list but I cannot commit any time to my to-do list until about Thursday next week so if someone takes it before that, I would be extremely happy.

scarlehoff:
- [x] Add a test for vp-setupfit
- [x] Improve the stability of the regression tests
- [x] Include a theory covmat test
- [x] Generate the data in a dedicated worker
- [x] Set the tolerance with respect to said worker

Fixes #2464


scarlehoff:
Ok, so the good news is that after many label-unlabel I have now the two samples that produce different results and make the test fail.
The failure seems to be coming from workers in different Azure regions. And there is no way to choose among them. Also, no guarantee that within the same region all workers are the same.

There are two options:
1. Run the regression in our own hardware.
2. Use those two samples to set the tolerances

I think the solution I'll go for is:
- Regenerate the regression in our own hardware (if our github plan allows for it...)
- Use the two samples to select a threshold compatible with the drift with respect to our own hardware.

This should be the most robust option.

Sorry for the spam whoever is still subscribed to this issue/PR but making the CI do it for me in a loop was the lowest-effort way to fish for the errors 😅

Radonirinaunimi:
> Regenerate the regression in our own hardware (if our github plan allows for it...)

For this, I think we can host our own runner in the Nikhef cluster (or somewhere else) as we did with `pineko` for the regression test.

scarlehoff:
Ah, fantastic. We should be able to use the same one I think. I was worried the custom runner was a paid feature, I didn't realize we were already using it.

Radonirinaunimi:
Btw, I will be waiting for this before resuming #2478.

scarlehoff:
Ok, the custom runner is working. I've created it in Sevilla. I don't know how stable it will be but that's what I can do.

Let me now prepare the tolerances with respect to the worker's results.

scarlehoff:
I'm afraid self-hosting is not a possibility, the difference is too big and I don't think we have the infrastructure to run the regression tests on the self-hosted runner for every commit :(

Let's see whether I can find a tolerance value for which is acceptable and for which the regressions don't need to be redone that often.

Radonirinaunimi:
> I'm afraid self-hosting is not a possibility, the difference is too big and I don't think we have the infrastructure to run the regression tests on the self-hosted runner for every commit :(
> 
> 
> 
> Let's see whether I can find a tolerance value for which is acceptable and for which the regressions don't need to be redone that often.

Why don't we then use a container? This would allow use to have infinite reproducibility?

scarlehoff:
We are effectively running in a container. I haven't tested, but if the differences are on the underlying hardware, the container won't fix it.

Radonirinaunimi:
But is it really at the level of the hardware though? Not at the level of system packages, compilers, etc.?

scarlehoff:
Well, unless the `ubuntu-latest` means different things, yes. If you get two logs from two workers that generate the differences the only difference is the AWS region in which they are running (I don't have enough statistics to know whether it would happen in the same region or whether it would work in different regions though, but it's what I have... so the only way around is a tolerance...)

Radonirinaunimi:
Selecting `ubuntu-latest` for example only ensures OS upgrades but on top of this, GitHub continuously updates the runner images themselves (system packages, compiler, tools, etc.). And I am pretty sure these are the causes of the non-reproducibility. And this is really the main reasons (on top of LHAPDF) we run regressions in container in PineAPPL and NeoPDF for example. I could try to give a look whenever I have some time.

scarlehoff:
The non-reproducibility here happens non deterministically though. But indeed, it could be updates that made one azure region be different from the other

scarlehoff:
Ok, for the time being generating the reference in a custom worker and defining the tolerances with respect to it seems to work. I can try to make it run a few more times to get more statistics.

@Radonirinaunimi if you have time to hace a quick look at this it would be very much appreciated. I'm not very happy with the tolerances and if that can be fixed with a docker image please do so. But at the moment I'm more worried about having regressions that we can rely on (so that people don't learn to simply ignore the X).

Radonirinaunimi (approved):
Should this be merged? I believe the documentation that we discussed over the CM could be done in another PR.

Issue #2399, opened 2025-11-07
Issue with shifts in plotting utilities


achiefa:
When producing a validphys report through `vp-comparefits`, with fits that include the photon PDF, I get this problem with the plotting pipeline used in the report
```
Traceback (most recent call last):
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nnpdf/bin/vp-comparefits", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/s2569857/Codes/NNPDF4.0/nnpdf/validphys2/src/validphys/scripts/vp_comparefits.py", line 247, in main
    a.main()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nnpdf/lib/python3.12/site-packages/reportengine/app.py", line 421, in main
    self.run()
  File "/Users/s2569857/Codes/NNPDF4.0/nnpdf/validphys2/src/validphys/app.py", line 150, in run
    super().run()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nnpdf/lib/python3.12/site-packages/reportengine/app.py", line 406, in run
    rb.execute_sequential()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nnpdf/lib/python3.12/site-packages/reportengine/resourcebuilder.py", line 210, in execute_sequential
    result = self.get_result(callspec.function,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nnpdf/lib/python3.12/site-packages/reportengine/resourcebuilder.py", line 380, in get_result
    fres = function(**kwdict)
           ^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniconda/base/envs/nnpdf/lib/python3.12/site-packages/reportengine/figure.py", line 143, in wrapper
    return list(func(*args, **kwargs))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/s2569857/Codes/NNPDF4.0/nnpdf/validphys2/src/validphys/dataplots.py", line 567, in plot_fancy_dataspecs
    yield from _plot_fancy_impl(
  File "/Users/s2569857/Codes/NNPDF4.0/nnpdf/validphys2/src/validphys/dataplots.py", line 272, in _plot_fancy_impl
    shifts, alpha = shifts_from_systematics(lcd_wc,theory_predictions)
[...]

PR #2484, opened 2026-06-14
Fix shifts in plot_fancy_dataspecs

### LINKED ISSUE #2481: Shift in data/theory comparison plots in vp reports
I suspect that the shift due to correlated systematic uncertainties in data/theory comparison plots in vp reports is not applied to both current and reference fits.

this is the unshifted data/theory comparison for the ZpT data (as representative data sets)

https://vp.nnpdf.science/Wq-EhBBDSk2mVsf_o67yPw==

and here is the same shifted data/theory comparison.

https://vp.nnpdf.science/GwytXkkQSrCZl95NEHgUvA==

You clearly see that the shift is applied only to the first prediction, not to the subsequent(s).

The issue applies to all data sets, not just ZpT.

enocera:
This PR ensures that the shift due to correlated systematic uncertainties is applied to all theoretical predictions when more than one prediction is compared to the data, for example when using plot_fancy_dataspecs. This PR fixes https://github.com/NNPDF/nnpdf/issues/2481.

Here are the same plots as in https://github.com/NNPDF/nnpdf/issues/2481, but with the fix applied.

https://vp.nnpdf.science/TCmkXajzSeq8TGsHVLE4Dw==


validphys2/src/validphys/dataplots.py
```diff
@@ -281,36 +277,49 @@ def _plot_fancy_impl(
...
+
+        # Compute shifts due to the correlated part on the experiemntal ucnertainty
+        if with_shift:
+            lcd_wc = loaded_commondata_with_cuts(commondata, cuts)
+            theory_predictions = result.central_value
+
+            # For unknown reasons, `shifts_from_systematics` may randomly fail.
+            # If a LinAlgError is raised, shifts are not included in the final plot. 
+            try:
+                shifts, alpha = shifts_from_systematics(lcd_wc, theory_predictions)
```
scarlehoff: I think `alpha` should not be an output of this function. The complicated/costly part of this function (inverting a matrix, etc) doesn't need to be done for the data part (which is the only reason you might want to enter this loop with the data), right? I think it would be worth it to compute alpha, given solely `lcd_wc` as a separate function (`shifts_from_systematics` can also call that same function).

Btw, one thing that might reduce the report time is to make the `result` hashable (depends only on commondata and on the theory number -if a theory prediction-) such that the amount of shifts that need to be computed for a report are not duplicated... writing but probably an issue to tackle later.

validphys2/src/validphys/dataplots.py
```diff
@@ -281,36 +277,49 @@ def _plot_fancy_impl(
...
-        # Shift the theory when with_shift option is True
-        if i == 1 and with_shift:
-            cv[mask] = result.central_value - shifts
+
+        shifts = None
+        alpha = None
+        do_shift = with_shift
+
+        # Compute shifts due to the correlated part on the experiemntal ucnertainty
+        if with_shift:
```
scarlehoff, line 295: I think it would be cleaner if this were instead:

```suggestion        
        if do_shift:
```

where 
```
do_shift = with_shift and not isintance(result, DataResult)
```
(you might need to import DataResults from results.py)

So that you don't enter the if condition with the Data, so you don't have to check if i>=1, which might not always be reliable if things change in validphys. You still need the alpha of course, so make sure that you add:

```
if with_shift and isintance(result, DataResult):
    err[mask] = compute_alpha(lcd_wc)
```

validphys2/src/validphys/dataplots.py
```diff
@@ -281,36 +277,49 @@ def _plot_fancy_impl(
...
+            if i >= 1 and do_shift:
+                cv[mask] = result.central_value - shifts
+            else:
+                cv[mask] = result.central_value
+
+            # Retain only the uncorrelated part of the data error if shifting the data
+            if i == 0 and do_shift:
+                err[mask] = alpha
+            else:
+                err[mask] = result.std_error
```
scarlehoff:
```suggestion
                shifts = 0.0

            cv[mask] = result.central_value - shifts
            err[mask] = result.std_error                
```

Since when you are dealing with the data, you are not entering in the if condition, this is safe.

validphys2/src/validphys/dataplots.py
```diff
@@ -263,10 +263,6 @@ def _plot_fancy_impl(
     nkinlabels = len(table.columns)
     ndata = len(table)
 
-    # Compute shifts due to the correlated part of the exp cov matrix
-    lcd_wc = loaded_commondata_with_cuts(commondata, cutlist[0])
```
scarlehoff, line 267: I would leave this here. Perhaps with `if with_shift` to avoid unnecessary calculations. But it makes sense to do it before the loop.

scarlehoff (commented):
Added a few comments. The general idea is that `alpha` is only needed for the data so it doesn't need to be part of the output.
And the data doesn't need the shift so it shouldn't go through the costly part of the equation.

If you check for data beforehand, then you avoid also having to guard against using data/theory relying on the order in which enters the plot, because it is not a given that tomorrow we won't change the order for whatever reason (and the bug would completely escape me as it did when I reviewed these changes, because it didn't occur to me to test with more than one prediction...)

validphys2/src/validphys/dataplots.py
enocera: I don't understand this suggestion. If I do this, then I'm re-instating the offending bit, which is to call the function with cutlist[0].

scarlehoff: This, I realize now, is the other point you mentioned during the CM. Indeed you are right, I didn't consider the cuts could be different for each plot.

validphys2/src/validphys/dataplots.py
enocera: I have separated the computation of the shifts and the extraction of the uncorrelated and correlated parts part of the uncertainty.

validphys2/src/validphys/dataplots.py
enocera: In my opinion, we have to enter the condition with the Data, also to get the uncorrelated part of the uncertainty, because these are Data with cuts, and cuts may differ.

validphys2/src/validphys/dataplots.py
enocera: Done.

enocera:
@scarlehoff I hope to have correctly addressed your ocmments. perhaps, let me generate a vp comparefit reports to check that everything is in order.

validphys2/src/validphys/dataplots.py
```diff
@@ -281,36 +277,40 @@ def _plot_fancy_impl(
...
+                shifts = shifts_from_systematics(lcd_wc, theory_predictions)
+            except np.linalg.LinAlgError:
+                log.warning(
+                    "Error occurred in computing systematic shifts for "
+                    f"{info.ds_metadata.name}. These will be disregarded in the plots."
+                )
+                shifts = 0.
+
             cv[mask] = result.central_value - shifts
+            err[mask] = unco_unc(lcd_wc)
```
scarlehoff: Here the check for data/theory is missing with `DataResult`, right? Or the same needs to be done in all cases?

(but maybe then `alpha` will need to be define in both path, in the `try` and the `except`)

validphys2/src/validphys/dataplots.py
```diff
@@ -281,36 +277,40 @@ def _plot_fancy_impl(
...
         err = np.full(ndata, np.nan)
-        # Shift the theory when with_shift option is True
-        if i == 1 and with_shift:
+
+        shifts = None
+        alpha = None
+        do_shift = with_shift
+        
+        # Compute shifts due to the correlated part on the experiemntal ucnertainty
+        if do_shift:
```
scarlehoff:
```suggestion        
        # Compute shifts due to the correlated part on the experimental uncertainty
        if with_shift:
```

enocera: Done.

validphys2/src/validphys/dataplots.py
enocera: After some thoughts, I think that here, if try fails, we just don't want to apply the shifts, and we want  to fall into the unshifted case by default (which is not the case now). I am thinking that perhaps we should fall in this case also when alpha is zero (that is when there's no uncorrelated part of the uncertainty because e.g. the experimentalists provide us only with a covariance matrix). Right now we just set the uncertainty displayed in data/theory plots to zero and do not shift the data (and say in the title that the data is shifted).

scarlehoff: I think the way of making sure there is no shift applied is to do
```
try:
    shifts = shift_calculation()
    alpha = alpha_calculation()
except:
     shifts = 0.0
     alpha = result.std_err
```
so that it is equivalent to not applying the shift, no?

enocera: Not exactly, because I fear that the label printed on the plot will still be "shifted" because try is in the with_shift if. I'm checking right now.

enocera: Simple fix, turn with_shift to False.

enocera:
@scarlehoff I have rethought how we apply shifts, based on your comments, that I didn't completely understand in the first place, but that I hope to have understood now. What I hopefully do now is to separate the uncertainty on the data and the shift to the theory and to more clearly identify four cases: with shift; no shift (default); no shift, because even if the user requires a shift, the shift cannot be computed because uncorrelated uncertainties are zero; no shift, because, even if the user requires a shift, the computation of the shift fails.

scarlehoff:
My comments were ill-informed in that I didn't appreciate the fact that two different predictions might have different cuts.

I'll look at the new version of the changes with that in mind. I like the 4 cases though.

validphys2/src/validphys/dataplots.py
```diff
@@ -281,39 +277,89 @@ def _plot_fancy_impl(
...
+                    log.warning(
+                        "Error occurred in computing systematic shifts for "
+                        f"{info.ds_metadata.name}. These will be disregarded in the plots."
+                    )
+                    fail = True
+                    
+            # Determine data uncertainty   
+            else:
+                alpha = unco_unc(lcd_wc)
+                if alpha.all() == 0. or fail:
```
scarlehoff: queston, if fails, it should fail for everyone, right?

enocera: What do you mean?

scarlehoff: In a plot we have data, theory1, theory2. Any failure should remove the shifts for everyone, right?

enocera: Ah! In that sense. So yes, if it fails for one theory, but not for another, then the plot must collapse to the unshifted version for everything. I hadn't thought about this case which, I think, is not encompassed by the current structure.

scarlehoff: because right now, if I understand it correctly, if the data comes first the fail will never activate?

enocera: I think that now theory comes first. The way I understand your comment is the following: suppose you are comparing theory 1 and theory 2. Shifts are set to 0 initially. Then suppose that theory 1 fails because of the error in computing shifts. Then everything falls back to the unshifted version of the plot. Then suppose that theory 2 also fails for the same reason. No problem, everything falls back to the unshifted version of the plot. I understand that you're saying that there is an inconsistency when only one of the two theories fails, irrespective of the order. In the current implementation, the plot never falls back to the unshifted version, because some theories will be shifted and some will not, and the error will be the total error or only the uncorrelated part of the error depending on whether the failing theory is the last (or not). Is this what you are thinking of?
If so, I guess that this can be fixed by initialising fail to False outside the loop on i and by putting try except in an if fail==False.

enocera: BTW, do you remember for which data set the computation of the shifts was failing?

scarlehoff: >  I understand that you're saying that there is an inconsistency when only one of the two theories fails, irrespective of the order. In the current implementation, the plot never falls back to the unshifted version, because some theories will be shifted and some will not, and the error will be the total error or only the uncorrelated part of the error depending on whether the failing theory is the last (or not). Is this what you are thinking of?

But more generally, the fact that it depends on the order and that is not guaranteed.

> If so, I guess that this can be fixed by initialising fail to False outside the loop on i and by putting try except in an if fail==False.

Yes, probably

> BTW, do you remember for which data set the computation of the shifts was failing?

No ^^U
But you can force the failure.

enocera:
Dear @scarlehoff I decided to investigate a little more the reasons for the failure reported in https://github.com/NNPDF/nnpdf/issues/2399 (something that I should have done long ago). The reason for the failure, i.e. the reason why the computation of the shifts fails, is that the legacy variant of the offending data sets (ATLAS_SINGLETOP_7TEV-T-Y-NORM and ATLAS_SINGLETOP_7TEV-TBAR-Y-NORM), used by @achiefa in his QED fit, is bugged. There is a statistical uncertainty set to 1e-10, which code-wise is nonzero, but which is effectively triggering a singularity. That uncertainty must be removed. The recent fit run by @vschutze-alt are not affected, because she's using correctly the default (not legacy) variant. 

So I'll first open another PR in which I fix the offending commondata, then rebase this PR on that PR, and in this PR I will get rid of the case in which the computation of the shifts fails. As a future reminder to myself and to everybody, any future failure of the shifts is a symptom of a bug somewhere else.

enocera:
P.S. Incidentally, the implementation of the uncertainties of the default (non-legacy) variant is also bugged (uncertainties are orders of magnitude larger than what they should be).

scarlehoff:
Thanks @enocera 

> I will get rid of the case in which the computation of the shifts fails. As a future reminder to myself and to everybody, any future failure of the shifts is a symptom of a bug somewhere else.

And link this as a comment in the code and we will frame it somewhere.

enocera:
Bug on ATLAS single top 7 TeV differential distributions fixed in https://github.com/NNPDF/nnpdf/pull/2489

validphys2/src/validphys/dataplots.py
```diff
@@ -281,36 +277,40 @@ def _plot_fancy_impl(
...
+                cv[mask] = result.central_value
+                alpha = unco_unc(lcd_wc)
+                if alpha.all() == 0.:
+                    err[mask] = result.std_error
+                    with_shift = False
+                else:
+                    err[mask] = alpha
+            # Determine shift           
+            else:
+                if with_shift:
```
scarlehoff: This second if will always be true, because if the previous sets it to False then the next run of the loop won't get here (since the first if will not pass). So I think you can remove the if/else.
What I don't know if perhaps you want to simply add (in the function that computes the shift) that when alpha=0 then return 0.

validphys2/src/validphys/covmats.py
```diff
@@ -292,9 +285,57 @@ def shifts_from_systematics(lcd_wc, theory_predictions):
         # Compute the shifts
         shifts = -np.matmul(beta * alpha[:, np.newaxis], r)
 
-    return shifts, alpha
+    return shifts
```
scarlehoff, line 288: Ah, never mind, you are already doing that.

Then yes, simply remove the second if (the `else` path can never be reached) and we are good to go.

enocera:
@scarlehoff For future reference, let me summarise some facts here.
- We are interested in the following cases: w/ shift (default) and w/o shift. If w/ shift, then there may be two sub-cases: the uncorrelated part of the uncertainty is zero or not. In the former case, we should just fall back to the w/o shift case.
- The failure reported by @achiefa in #2399 was triggered by a bug in the commondata implementation, which has been fixed in #2489 and #2490. Here is the output if I run his script (`with_shift=True`): https://vp.nnpdf.science/YDPfRZ4UR7eJaPJd9Fyaww==. As you can see, because the uncorrelated part of the uncertainty is zero, the behaviour falls back to that of the unshifted plots. The output is the same as that obtained with `with_shift=False`, as it should be https://vp.nnpdf.science/S0uYJU9BTDahmeYZbgJWxA==
- Here are a couple of reports in which I compare two theory predictions to the 4.0-like data set, once `with_shift=True` https://vp.nnpdf.science/0kAOnJ1hT1248VUwRosohA== and once `with_shift=False` https://vp.nnpdf.science/zqH_GpF-TBefFR4G4y-Y_A==. It seems to me that everything is as it should be.

validphys2/src/validphys/covmats.py
enocera: Done, thanks!

scarlehoff (approved):

PR #2485, opened 2026-06-15
Add the possibility of evolving the fit with hoppet

scarlehoff:
This PR implements the evolution of the PDF using HOPPET with the `--hoppet` flag.

The generation of the xgrid, qgrid, etc, is the same as with the vanilla evolution and the results are, at NNLO, equivalent for Q > mc.

For Q < mc the way HOPPET and EKO invert the threshold is different, see:

EKO: https://eko.readthedocs.io/en/latest/theory/Matching.html#backward-evolution
HOPPET: https://github.com/hoppet-code/hoppet/blob/31a2ff92aa82f227505ee0ce111db8acfce340a0/src/evolution.f90#L329

We would need to set `inversion_method = expanded` and the mass of the charm to 0 to get the exact same results with both.

N3LO to be tested.

### DIFF
--- a/n3fit/src/evolven3fit/evolve.py
+++ b/n3fit/src/evolven3fit/evolve.py
@@ -82,6 +82,11 @@ class ExportGrid:
         if self.labels is None:
             self.labels = [PIDS_DICT[i] for i in self.pids]
 
+        # Setting the charm to 0 gets matching results with EKO's expanded inversion
+
+    #         self.pdfgrid[:,self.pids.index(-4)] = 0.0
+    #         self.pdfgrid[:,self.pids.index(4)] = 0.0
+
     @property
     def pdfvalues(self):
         """Return the PDF, i.e., pdfgrid / xgrid,
--- a/n3fit/src/evolven3fit/evolve.py
+++ b/n3fit/src/evolven3fit/evolve.py
@@ -24,6 +24,8 @@ from eko.io import manipulate
 from validphys.pdfbases import PIDS_DICT
 from validphys.utils import yaml_safe
 
+from . import hoppet_evolve
+
 _logger = logging.getLogger(__name__)
 
 LOG_FILE = "evolven3fit.log"
@@ -95,58 +97,13 @@ class ExportGrid:
         return self.pdfgrid.T / self.xgrid
 
 
-def evolve_exportgrid(eko_path: pathlib.Path, exportgrids: list[ExportGrid]):
+def _evolve_exportgrids_with_eko(eko_op: pathlib.Path, exportgrids: list[ExportGrid]):
     """
-    Takes the path to an EKO and a list of exportgrids,
-    returns a tuple with an info file and the
-    evolved exportgrid as a dictionary of the form:
-
-    .. code-block:: python
-
-        {
-            (Q_1^2, nf1): (replica, flavours, x),
-            (Q_2^2, nf1): (replica, flavours, x),
-            ...
-            (Q_3^2, nf2): (replica, flavours, x),
-        }
-
-
-    with the output grouped by nf and sorted in ascending order by Q2.
-
-    Parameters
-    ----------
-    eko_path: pathlib.Path
-        Path to the evolution eko
-    exportgrids: list[ExportGrid]
-        List of ExportGrid objects to be evolved
-
-    Returns
-    -------
-    info_file: eko_box.info_file
-        Dict-like object with the info file information.
-    evolved_replicas: dict
-        a dictionary containing all evolved PDFs.
-        The format of the output is
-        { (q2, flavour number): np.ndarray(replica, flavours, x) }
+    Evolves the set of exportgrids using EKO.
     """
-    # Check that all exportgrid objects have been evaluated for 1) The same value of Q, the same value of x
     ref = exportgrids[0]
-    hessian_fit = ref.hessian
-
-    for egrid in exportgrids:
-        assert egrid.q20 == ref.q20, "Different values of q0 found among the exportgrids"
-        np.testing.assert_allclose(
-            ref.xgrid, egrid.xgrid, err_msg="ExportGrids are not all evaluate at the same x nodes"
-        )
-        assert (
-            hessian_fit == egrid.hessian
-        ), "Trying to evolve hessian and non-hessian fit at the same time"
-
-    # Read the EKO and the operator and theory cards
-    eko_op = eko.EKO.read(eko_path)
     theory = eko_op.theory_card
     op = eko_op.operator_card
-
     assert ref.q20 == op.mu20, f"The EKO can only evolve from {op.mu20}, PDF asked for {ref.q20}"
 
     _logger.debug(f"Theory card: {json.dumps(theory.raw)}")
@@ -160,7 +117,6 @@ def evolve_exportgrid(eko_path: pathlib.Path, exportgrids: list[ExportGrid]):
     if XGrid(x_grid) != eko_original_xgrid:
         new_xgrid = XGrid(x_grid)
         new_metadata = dataclasses.replace(eko_op.metadata, xgrid=new_xgrid)
-
         new_operators = {}
         for target_key in eko_op.operators:
             elem = eko_op[target_key.ep]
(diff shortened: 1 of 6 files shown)


n3fit/src/evolven3fit/evolve.py
```diff
@@ -194,17 +155,84 @@ def evolve_exportgrid(eko_path: pathlib.Path, exportgrids: list[ExportGrid]):
...
+        assert (
+            hessian_fit == egrid.hessian
+        ), "Trying to evolve hessian and non-hessian fit at the same time"
+
+    if hoppet:
+        info, sorted_evolved = hoppet_evolve.evolve_exportgrids_with_hoppet(
+            eko_path, exportgrids, theory_id
+        )
+    else:
+        # We read the EKO to a temporary directory that will vanish upon exiting
```
felixhekhorn, line 224:
```suggestion
        # We extract the EKO to a temporary directory that will vanish upon exiting
```
why is the native `EKO.read` behaviour not good enough? like `with eko.EKO.read`

n3fit/src/evolven3fit/evolve.py
```diff
@@ -306,13 +345,16 @@ def evolve_fit(fit_folder, force, eko_path, hessian_fit=False):
...
         force: bool
             whether to force the evolution to be done again
-        eko_path: str or pathlib.Path
-            path where the eko is stored (if None the eko will be
-            recomputed)
         hessian_fit: bool
             wether the fit is hessian
+        eko_path: str or pathlib.Path
+            path where the eko is stored (if None the eko will be recomputed)
+        hoppet: bol
```
felixhekhorn, line 356:
```suggestion
        hoppet: bool
```

n3fit/src/evolven3fit/hoppet_evolve.py
```diff
@@ -0,0 +1,251 @@
...
+# we could
+def _make_hoppet_pdf_callback(exportgrid, interpolation_degree=4):
+    """Creates an interpolation pdf(x) given an exportgrid.
+    NOTE: the function includes Q but this is dropped anyway
+    """
+    xgrid = exportgrid.xgrid
+    pid_columns = [exportgrid.pids.index(pid) for pid in HOPPET_QCD_PIDS]
+    pdfgrid = exportgrid.pdfgrid[:, pid_columns]
+
+    dispatcher = InterpolatorDispatcher(xgrid, interpolation_degree, mode_N=False)
```
felixhekhorn, line 68: just remember that you rely on `log=True` [here](https://github.com/NNPDF/eko/blob/d41da2f5ed4b24fed7bc3d6de89b32f2e63851eb/src/eko/interpolation.py#L415), which is what you want

n3fit/src/evolven3fit/hoppet_evolve.py
```diff
@@ -0,0 +1,251 @@
...
+
+    nloop = nnpdf_theory.PTO + 1
+    fns = nnpdf_theory.FNS.split("-")[0]
+
+    return HoppetTheory(
+        alphas=nnpdf_theory.alphas,
+        qref=nnpdf_theory.Qref,
+        q0=nnpdf_theory.Q0,
+        nloop=nloop,
+        mur_over_q=nnpdf_theory.XIR,
```
felixhekhorn, line 105: This looks wrong. Maybe you mean
```suggestion
        mur_over_q=nnpdf_theory.XIF,
```
? Recall that in NNPDF nomenclature the renormalization scale (and hence `XIR`) refers to the hard scattering, the partonic matrix elements and not to evolution. However, in SV scheme A you shift the evaluation of the strong coupling which multiplies the anomalous dimensions by a factor (see [here](https://eko.readthedocs.io/en/latest/theory/MHOU.html#shifting-anomalous-dimensions) or the mentioned papers there) and most likely this is what hoppet calls `mur_over_q` (I haven't check hoppet explicitly). This is because I suspect very much that hoppet can only do scheme A (as with Pegasus for example). Thus at the minimum you need to check the SV scheme here in addition and also match that.

n3fit/src/evolven3fit/hoppet_evolve.py
```diff
@@ -0,0 +1,251 @@
...
+def _nnpdf_theory_to_hoppet(nnpdf_theory, polarized=False):
+    """Translation layer between the NNPDF theory card and hoppet's parameters."""
+    if polarized:
+        raise NotImplementedError("Polarized not implemented for hoppet")
+
+    if nnpdf_theory.QED != 0:
+        raise NotImplementedError("QED evolution not yet supported for hoppet")
+
+    if nnpdf_theory.PTO > 2:
+        raise NotImplementedError("Only up to NNLO for now with hoppet")
```
felixhekhorn, line 92: actually, why? hoppet should be able to do N3LO by now, no?

n3fit/src/evolven3fit/hoppet_evolve.py
```diff
@@ -0,0 +1,251 @@
...
+
+    masses = (nnpdf_theory.mc, nnpdf_theory.mb, nnpdf_theory.mt)
+    # Note: for hoppet this is both the masses and the thresholds
```
felixhekhorn, line 95: so we should check they are the same for us, i.e. k=1

n3fit/src/evolven3fit/hoppet_evolve.py
```diff
@@ -0,0 +1,251 @@
...
+
+    nloop = nnpdf_theory.PTO + 1
+    fns = nnpdf_theory.FNS.split("-")[0]
```
felixhekhorn, line 98: Is this correct? it looks too simple to me (but I haven't checked hoppet). I suspect however, you need either (some variation of) "VFNS" or "FFNS". However, in many cases our FNS should be "FONLL" - which should be mapped to "VFNS".

n3fit/src/evolven3fit/hoppet_evolve.py
```diff
@@ -0,0 +1,251 @@
...
+    ymax = -np.log(xmin)
+
+    qmin = min(np.min(q_values), theory.q0)
+    qmax = max(np.max(q_values), theory.q0)
+    hp.StartExtended(
+        ymax,
+        dy,
+        qmin,
+        qmax,
+        dy / 4.0,  # hoppet's suggestion
```
felixhekhorn, line 146: if they say so ... but shouldn't we just change the default value?

n3fit/src/evolven3fit/hoppet_evolve.py
```diff
@@ -0,0 +1,251 @@
...
+    return theory_card, op_card_eko
+
+
+def _configure_hoppet(hp, theory: HoppetTheory, x_grid, q_values, dy=0.025):
+    """
+    Calling a few hoppet function that need to be called before starting.
+    We use hoppet's extended start to make sure we have the same
+    range as the lhapdf grid that we use in NNPDF.
+    """
+    hp.SetExactDGLAP(True, True)
```
felixhekhorn, line 134: I'm guessing in the wild: does "Exact" here refer to the solution method? should we check that we actually want that? Is there another option?

n3fit/src/evolven3fit/evolve.py
scarlehoff: https://github.com/NNPDF/eko/issues/533

n3fit/src/evolven3fit/hoppet_evolve.py
scarlehoff: I should check, probably I read mur and translated it to xir (I only tested central fits so wouldn't see the error in my test)

n3fit/src/evolven3fit/hoppet_evolve.py
scarlehoff: Is to be tested! I was waiting for our own 4.1 n3lo fits for that

n3fit/src/evolven3fit/evolve.py
felixhekhorn: mh? how is this related to my question here? this is rather the proper issue to this line https://github.com/NNPDF/nnpdf/blob/9bf30b17d6ac2e5686dfa105a35485fafee3bbb1/n3fit/src/evolven3fit/hoppet_evolve.py#L121

n3fit/src/evolven3fit/hoppet_evolve.py
scarlehoff: Nono, their parametrization to make it faster. We want exact (to avoid numerical noise). But as I said in the OP to get the exact results all across Q we need to also change our parameters (no way around that but for the purposes of this PR I don't care dor Q < mc)

n3fit/src/evolven3fit/hoppet_evolve.py
scarlehoff: I played around with these parameters and I didn't really see any relevant change tbh

n3fit/src/evolven3fit/evolve.py
scarlehoff: Ah, sorry, I'm looking at the comments with my phone, that's the only problem with eko's default behaviour I can remember right now.
I'd need to have a closer look to remember ^^U

PR #2486, opened 2026-06-19
Shapley analysis

scarrazza:
Includes Shapley analysis tools to the NNPDF framework.


pyproject.toml
```diff
@@ -55,12 +55,15 @@ vp-list = "validphys.scripts.vp_list:main"
 vp-nextfitruncard = "validphys.scripts.vp_nextfitruncard:main"
 vp-hyperoptplot = "validphys.scripts.vp_hyperoptplot:main"
 vp-deltachi2 = "validphys.scripts.vp_deltachi2:main"
+vp-shapley = "validphys.shapley.scripts.vp_shapley:main"
 
 [tool.poetry.dependencies]
 # Generic dependencies (i.e., validphys)
 python = ">=3.9"
+shapley-values = {git = "https://github.com/rbonnetguerrini/shapley-values.git", branch = "main"}
```
scarrazza, line 63: I would suggest @rbonnetguerrini to make a release of his code and upload to PyPI, so we avoid installing from git.

scarrazza (commented):
@rbonnetguerrini after discussing with @scarlehoff, we think we should add a section in the sphinx documentation. For example, you could move the execution tutorial from the scripts readme directly there (see doc folder) and include the example in the jupyter notebook.

scarlehoff:
Hi @rbonnetguerrini I think if you rebase on top of the current master (rebase would be much preferred and since this is a separate feature it should be painless) most of the test will pass.

pyproject.toml
scarlehoff: This will be necessary before merging to master yes, but we can wait.

validphys2/src/validphys/shapley/slurm/run_vp_shapley.slurm
```diff
@@ -0,0 +1,292 @@
+#!/bin/bash
```
scarlehoff, line 1: It worries me that such a long slurm script is necessary to run this?

In any case, I would prefer just documentation, I don't think uploading slurm scripts to the repo is necessary (since they would only work for the system you were running this on, most clusters are configured differently).

validphys2/src/validphys/shapley/runcards/DIS/DIS-FT.YAML
```diff
@@ -0,0 +1,95 @@
+#
```
scarlehoff, line 1: I think the runcards should go to the paper repository, which is where we have most of the "reproduce X paper" runcards.

In this repository we should have just one example runcard at most (in a folder called example or something like that). Or even none, if the whole example can be written in the documentation.

validphys2/src/validphys/shapley/setup.py
```diff
@@ -0,0 +1,609 @@
...
+"""
+
+import functools
+from typing import List, Optional
+
+import numpy as np
+import pandas as pd
+from scipy.linalg import cholesky, solve_triangular
+
+from validphys.loader import Loader
```
scarlehoff, line 21: Instead of Loader, if you use FallbackLoader, when you load something that is not available locally (but it is available in the NNPDF server) it will be automatically downloaded.

validphys2/src/validphys/shapley/setup.py
```diff
@@ -0,0 +1,609 @@
...
+    block_diags = []
+    special_corrs = []
+    for stat, sys in zip(all_stat_errors, all_sys_errors):
+        is_intra = sys.columns.isin(INTRA_DATASET_SYS_NAME)
+        block_diags.append(construct_covmat(stat, sys.loc[:, is_intra]))
+        special_corrs.append(sys.loc[:, ~is_intra])
+
+    special_sys = pd.concat(special_corrs, axis=0, sort=False).fillna(0)
+    diag = la.block_diag(*block_diags)
+    return diag + special_sys.to_numpy() @ special_sys.to_numpy().T
```
scarlehoff, line 522: Why do you need functions like this one?

validphys2/src/validphys/shapley/setup.py
```diff
@@ -0,0 +1,609 @@
...
+        cfac = tuple(entry.get("cfac", []))
+        return name, variant, cfac
+    else:
+        raise TypeError(f"Dataset entry must be str or dict, got {type(entry)}")
+
+
+def setup_observables(
+    pdf_name: str,
+    datasets: list,
+    theory_id: int = 708,
```
scarlehoff, line 314:
```suggestion
    theory_id: int = 40000000,
```

scarlehoff (commented):
Just a few first comments.

From a quick look I think many functions should be using internal valiphys function instead of mirrors of it to ensure things are taken in exactly the same way. I am mainly worried about the covariance matrices but it really applies to everything.

But I think most of the issue is `setup_observables` which should be modified to be as small as possible, reusing as much internal validphys stuff as possible, I would do the following:

1) Use the API instead of Loader, and ask the API for the different pieces that you need (covmat, fktables, etc). Then from those outputs you construct your own output (instead of having to compute the covmat yourself it would be `API.covmat_inverse_whatever_something_else(**config)`
2) With that you will know which input do you actually need
3) Remove all functions that will become unused  at that point

Then, later instead of calling it in the script with `main()` it can be called as a subclass of ValidphysApp.run() and the API part can also be removed.

PR #2487, opened 2026-06-24
Update fitbot

scarlehoff:
Just an update of the fitbot to use the latest versions of the software (updated to python 3.13, etc) to have a pre-tag reference.

Updated also in this branch the commondata, since the eko update brought with it a numpy update which changes a bit the numbers.


scarlehoff:
https://github.com/NNPDF/nnpdf/actions/runs/28166943200/attempts/1

image

PR #2490, opened 2026-07-01
Update docstrings in validphys with new datasets names

kamillaurent:
I am still not sure when to add `"variant": "legacy"` and `"frac": 0.75` when switching names.


scarlehoff:
In general, if you are not doing a fit, then the fraction is not needed (since it is the training/validation split)

RE legacy, use the variant that is used in the 4.1 runcard. If, for whataver reason, it cannot be used (e.g. it is an example that only runs for older theories and data for some reason) then use legacy.

kamillaurent:
Regarding datasets that are not in the `dataset_names.yml` (i.e. `NTVNUDMNFe` from `validphys2/src/validphys/n3fit_data.py`), should I remove them or should I change them with the closest name (in this case `NTVNUDMNFe_dw_ite`)?

scarlehoff:
> should I remove them or should I change them with the closest name (in this case NTVNUDMNFe_dw_ite)?

I'd say the closest one.

But again, if it is an exmaple, you can choose a different dataset and make sure that the example is showing the same.

scarlehoff:
Hi @kamillaurent is this complete?

kamillaurent:
I don't spot any old name in the docstrings, so I believe this is complete @scarlehoff

kamillaurent:
I don't spot any old name in the docstrings, so I believe this is complete @scarlehoff

validphys2/src/validphys/theorycovariance/higher_twist_functions.py
```diff
@@ -90,7 +90,7 @@ def get_pc_type(
     Returns
     -------
     str or tuple of (str, str)
-        The PC type key(s). For the NMC ratio dataset
+        The PC type key(s). For the ratio dataset
```
scarlehoff, line 93: here I think NMC was acting more as a noun referring to the NMC collaboration itself. But it's ok, it reads good either way

validphys2/src/validphys/pseudodata.py
```diff
@@ -104,7 +104,7 @@ def read_replica_pseudodata(fit, context_index, replica):
     >>> rep_info.pseudodata.loc[rep_info.tr_idx].head()
                                 replica 1
     group dataset           id
-    ATLAS ATLASZPT8TEVMDIST 1   30.665835
+    ATLAS ATLAS_Z0J_8TEV_PT-M 1   30.665835
```
scarlehoff: Here, instead of changing this manually you should run the code above and update the info (you will see that you cannot use `pseudodata_test_fit_n3fit` and have to use an updated version such as `pseudodata_test_fit_n3fit_260518`), which has 15 replicas instead of 10 and which has `ATLAS_Z0_7TEV_46FB_CC-Y` as first dataset.

validphys2/src/validphys/utils.py
```diff
@@ -88,10 +88,10 @@ def experiments_to_dataset_inputs(experiments_list):
     >>> fit = API.fit(fit='NNPDF31_nnlo_as_0118_1000')
```
scarlehoff: Same here. Instead of updating the results of the code, you should update the fit instead.

A recent one you can use is: `nnpdf40-like_t0_sampling`, and below instead of `experiments` which doesn't exist anymore, you need to use `dataset_inputs = fit.as_input()['dataset_inputs']`

scarlehoff (changes requested):

kamillaurent:
@scarlehoff now it should be ok. Let me know if this PR can be merged or needs some other change.

scarlehoff:
Thanks, let's wait for https://github.com/NNPDF/nnpdf/pull/2499 to merge since this affects doctests so better to rebase on top of that one so this is merged tested.

scarlehoff (approved):
thanks!

PR #2494, opened 2026-07-06
4.1 N3LO theory cards

jekoorn:
Write the N3LO theory cards for:
- [x] central theory
- [x] anomalous dimensions variations
- [x] scale variations
- [x] DIS coeff functions fact variations -> cf variations still required
- [x] IHOU + MHOU missing prescription
- [x] alpha_s variations

And:
- [x] write the scalevariationtheoryids.yaml accordingly


nnpdf_data/nnpdf_data/theory_cards/41300040.yaml
```diff
@@ -0,0 +1,46 @@
...
+GF: 1.1663788e-5
+SIN2TW: 0.22348
+TMC: 1
+MP: 0.938
+Qref: 91.1880
+alphas: 0.118
+alphaqed: 0.0075700
+nfref: 5
+nf0: 4
+n3lo_ad_variation: [3, 0, 0, 0, 0, 0, 0]
```
felixhekhorn: these are the wrong ad IHOU variation settings: `3` would refer to the old (Giacomo) way - in the new (Moch et al.) way there are only 3 options: `0` (central), `1`, and `2` - but no `3`. Compare to theory 1000+

nnpdf_data/nnpdf_data/theory_cards/41300038.yaml
```diff
@@ -0,0 +1,46 @@
+ID: 41_300_038
+Comments: NNPDF4.1 N3LO QCD QED
```
felixhekhorn: If you gave the alpha_s variation a dedicated title, do so here! For example `NNPDF4.1 N3LO QCD QED Pgg up variation` - see theory 40005001+ (Please double check that the first entry would be Pgg, e.g. by just checking that theory)

nnpdf_data/nnpdf_data/theory_cards/41300000.yaml
```diff
@@ -0,0 +1,46 @@
...
+FNS: FONLL-E
+DAMP: 0
+IC: 1
+Q0: 1.65
+ModEv: EXA
+IterEv: 60
+XIR: 1.0
+XIF: 1.0
+NfFF: 5
+QED: 2
```
felixhekhorn: wait a second are we really supposed to do QED here? I don't think so ... and if that is true, of course, remove "QED" also from the comment everywhere

nnpdf_data/nnpdf_data/theory_cards/41300000.yaml
```diff
@@ -0,0 +1,46 @@
+ID: 41_300_000
```
felixhekhorn: Actually, I'm not sure `41_300_*` is the best range ... in particular because we already have `41_041_*`; I think the two should be close, e.g. this could be `41_042_*` referring to QCD only while the other is "the main theory", i.e. including QED

nnpdf_data/nnpdf_data/theory_cards/41300010.yaml
```diff
@@ -0,0 +1,46 @@
+ID: 41_300_010
```
felixhekhorn: I think I have asked this before, but why would we leave a gap of 9 number here?

felixhekhorn (changes requested):

nnpdf_data/nnpdf_data/theory_cards/41300000.yaml
Radonirinaunimi: If we ever want to make incremental comparisons, which I believe we do, then we definitely do not want QED for these theories. That'll come after.

nnpdf_data/nnpdf_data/theory_cards/41300010.yaml
jekoorn: No particular reason other than that the 41000000 theories also skip the first 9... And at some point I had already named too many files to change all.

nnpdf_data/nnpdf_data/theory_cards/41300000.yaml
jekoorn: Fair enough. So we don't want QED by default in this N3LO series?

nnpdf_data/nnpdf_data/theory_cards/41300038.yaml
jekoorn: gotcha!

nnpdf_data/nnpdf_data/theory_cards/41300040.yaml
jekoorn: Ok perfect, so the only variations would be these?

` - theoryid: 1000
    variations:
      (0, 0, 0, 0, 0, 0, 0): 1000 # central th
      (1, 0, 0, 0, 0, 0, 0): 1001 # anomalous dimensions variations
      (2, 0, 0, 0, 0, 0, 0): 1002
      (0, 1, 0, 0, 0, 0, 0): 1003
      (0, 2, 0, 0, 0, 0, 0): 1004
      (0, 0, 1, 0, 0, 0, 0): 1005
      (0, 0, 2, 0, 0, 0, 0): 1006
      (0, 0, 0, 1, 0, 0, 0): 1007
      (0, 0, 0, 2, 0, 0, 0): 1008
      (0, 0, 0, 0, 1, 0, 0): 1009
      (0, 0, 0, 0, 2, 0, 0): 1010
      (0, 0, 0, 0, 0, 1, 0): 1011
      (0, 0, 0, 0, 0, 2, 0): 1012
      (0, 0, 0, 0, 0, 0, 1): 1013
      (0, 0, 0, 0, 0, 0, 2): 1014
      (0.5, 1): 1015 # scale variations`
      
in fact (and what appears from how I created these cards), I have no idea what these numbers represent. Can I find out somewhere about the numbers, what do they mean?

validphys2/src/validphys/scalevariations/scalevariationtheoryids.yaml
```diff
@@ -423,6 +423,105 @@ n3lo: &n3lo
   (0.125): 40_007_012
   (0.130): 40_007_013
 
+n3lo41: &n3lo41
+  (0, 0, 0, 0): 41_300_000 # central th
+  (1, 0, 0, 0): 41_300_038 # anomalous dimensions variations
```
Radonirinaunimi: As mentioned by @felixhekhorn above, for the baseline N3LO we want the FHMRUVV parametrization and hence use the following variations:

https://github.com/NNPDF/nnpdf/blob/a90cf665ac65a0f07fc57968297b2e9b84754c2e/validphys2/src/validphys/scalevariations/scalevariationtheoryids.yaml#L700-L714

Radonirinaunimi (commented):
Pretty much overlaps with what @felixhekhorn already mentioned. The main point being the use of FHMRUVV parametrization as baseline.

nnpdf_data/nnpdf_data/theory_cards/41300040.yaml
felixhekhorn: > in fact (and what appears from how I created these cards), I have no idea what these numbers represent. Can I find out somewhere about the numbers, what do they mean?

https://eko.readthedocs.io/en/latest/theory/N3LO_ad.html

nnpdf_data/nnpdf_data/theory_cards/41300000.yaml
felixhekhorn: no, no QED

felixhekhorn:
Let's remember to make a pinefarm release with the new yadism version, @jekoorn please take care of that.

cc @scarlehoff

felixhekhorn:
Please resolve comments, which you believe to have addressed so we know what to worry about

felixhekhorn:
> Let's remember to make a pinefarm release with the new yadism version, @jekoorn please take care of that.

https://github.com/NNPDF/pinefarm/pull/131

jekoorn:
> > Let's remember to make a pinefarm release with the new yadism version, @jekoorn please take care of that.
> 
> [NNPDF/pinefarm#131](https://github.com/NNPDF/pinefarm/pull/131)

[New release ](https://github.com/NNPDF/pinefarm/releases/tag/v0.4.2)
Should we create a new release from the tag? I see there are no existing proper releases yet

validphys2/src/validphys/scalevariations/scalevariationtheoryids.yaml
```diff
@@ -423,6 +423,54 @@ n3lo: &n3lo
   (0.125): 40_007_012
   (0.130): 40_007_013
 
+n3lo41: &n3lo41
+  (0, 0, 0, 0, 0, 0, 0): 41_042_000 # central th
+  (1, 0, 0, 0, 0, 0, 0): 41_042_001 # anomalous dimensions variations
```
Radonirinaunimi: Some variations seem to be missing, for instance `(2, 0, 0, 0, 0, 0, 0)` or `(0, 0, 0, 0, 0, 2, 0)`. Why aren't they included?

validphys2/src/validphys/scalevariations/scalevariationtheoryids.yaml
```diff
@@ -423,6 +423,54 @@ n3lo: &n3lo
...
+  (2, 2): 41_042_018
+  # DIS coeff functions fact variations
+  lower bound DIS: 41_042_019
+  central DIS: 41_042_000
+  upper bound DIS: 41_042_020
+  # IHOU + MHOU missing prescription
+  (1, 0.5 missing): 41_042_021  # As 787 but DIS and N3LO DY from 722
+  (1, 2 missing): 41_042_022 # As 790 but DIS and N3LO DY from 722
+  (1, 0.5 hadronic): 41_042_023  # As 787 but DIS from 722
+  (1, 2 hadronic): 41_042_024 # As 790 but DIS from 722
```
Radonirinaunimi: You can remove these comments as they are no longer applicable.

validphys2/src/validphys/scalevariations/scalevariationtheoryids.yaml
```diff
@@ -423,6 +423,54 @@ n3lo: &n3lo
...
+  (2, 1): 41_042_012
+  (0.5, 0.5): 41_042_013
+  (1, 0.5): 41_042_014
+  (2, 0.5): 41_042_015
+  (0.5, 2): 41_042_016
+  (1, 2): 41_042_017
+  (2, 2): 41_042_018
+  # DIS coeff functions fact variations
+  lower bound DIS: 41_042_019
+  central DIS: 41_042_000
```
Radonirinaunimi, line 455: Not related to this PR but perhaps something for the future. This is a redundant information (and hence a room for mistakes), we should only keep `(0, 0, 0, 0, 0, 0, 0)`.

Radonirinaunimi (commented):
Some minor comments/requests.

validphys2/src/validphys/scalevariations/scalevariationtheoryids.yaml
jekoorn: Because I forgot to. Now they're included!

nnpdf_data/nnpdf_data/theory_cards/41042014.yaml
```diff
@@ -0,0 +1,47 @@
...
+PTO: 3
+FNS: FONLL-E
+DAMP: 0
+IC: 1
+Q0: 1.65
+ModEv: EXA
+IterEv: 60
+XIR: 1.0
+XIF: 1.0
+ModSV: expanded
```
Radonirinaunimi:
```suggestion
XIF: 1.0
```
`ModSV` must be omitted when `XIF: 1.0` and included only when $\xi_F \neq 1.0$.

Radonirinaunimi (changes requested):
Sorry, I missed this before.

Radonirinaunimi (approved):
This looks good to me now.

nnpdf_data/nnpdf_data/theory_cards/41042011.yaml
```diff
@@ -0,0 +1,46 @@
+ID: 41_042_011
+Comments: NNPDF4.1 N3LO QCD-only (nsm = 2)
```
felixhekhorn:
```suggestion
Comments: NNPDF4.1 N3LO QCD-only (nsm = 1)
```

felixhekhorn (changes requested):
Image
there is something fishy

nnpdf_data/nnpdf_data/theory_cards/41042011.yaml
felixhekhorn: I guess this means you didn't automatize everything :upside_down_face: let's hope there are not more typos around :see_no_evil:

jekoorn:
> Image there is something fishy

what do you mean?

felixhekhorn:
> > Image there is something fishy
> 
> what do you mean?

sorry - too much German I guess :see_no_evil: the string`(xir, xif =` was found 12 times in the changed files - this is 4 times too much; it should be 3*3-1 = 8 times

nnpdf_data/nnpdf_data/theory_cards/41042011.yaml
jekoorn: it means the number of files was small enough that it was a task I could do by hand (resulting, as we found, in typos here and there)

jekoorn:
I added it also in the runcards for IHOU + MHOU missing prescription (that @Radonirinaunimi ) recommended be removed for future iterations.

nnpdf_data/nnpdf_data/theory_cards/41042011.yaml
felixhekhorn: > it means the number of files was small enough

44 :upside_down_face:

jekoorn: writing an automatized script would probably take as long

felixhekhorn:
> I added it also in the runcards for IHOU + MHOU missing prescription (that @Radonirinaunimi ) recommended be removed for future iterations.

I'm slightly confused ... what is the difference between 41_042_021 vs. 41_042_026 vs. 41_042_028? they are identical theory cards, but they differ by the FK table they contain? (@Radonirinaunimi I guess?)

(cc @scarlehoff this is again what we were discussing the other day I think)

scarlehoff:
If the theory cards are identical then the content should be identical (thus, they should not exist)

If there is any difference that cannot be reflected in the theory card (for instance that one uses yadism old version and the other yadism new version) that difference should be explicit in the description.

jekoorn:
> > I added it also in the runcards for IHOU + MHOU missing prescription (that @Radonirinaunimi ) recommended be removed for future iterations.
> 
> I'm slightly confused ... what is the difference between 41_042_021 vs. 41_042_026 vs. 41_042_028? they are identical theory cards, but they differ by the FK table they contain? (@Radonirinaunimi I guess?)
> 
> (cc @scarlehoff this is again what we were discussing the other day I think)

I did it like this because some other theories also did it like this; I assumed these would differ by FK-Table as you say, since potentially they would take DY/DIS tables from other theories (like is don ein some other theories like the 1000 series)

scarlehoff:
We are trying not to repeat the errors of the past :__)

Many of those were actually the same to the point that in the server they are linked among them.

felixhekhorn:
> I did it like this because some other theories also did it like this; I assumed these would differ by FK-Table as you say, since potentially they would take DY/DIS tables from other theories (like is don ein some other theories like the 1000 series)

that is what we, @scarlehoff and I, concluded the other day (if I understood correctly): they differ by FK tables - and, no, such an information has no direct counterpart in the theory card, but I agree with 

> that difference should be explicit in the description.

so instead of a yaml comment in this specific file `validphys2/src/validphys/scalevariations/scalevariationtheoryids.yaml`, an additional ~ "As 40_005_049 but DIS from 40_005_027" in the description would be good

scarlehoff:
> "As 40_005_049 but DIS from 40_005_027" in the description would be good

Rather I would explain what they are "Scale variations applied only to NNLO data (hadronic and DIS+jet) and not to N3LO DIS" (if that's the case). You can add also that hadronic data is like 049 and DIS like 027 as well, of course.

This is very important since here we are first creating the theory cards but then those theories will need to be filled with something.

Radonirinaunimi:
Yes, these are Frankenstein theories for which FK tables are taken from different theories and thus are intrinsically different. But, @jekoorn, I asked to remove deprecated information because we **cannot** take these FK from old theories at all because we now have completely different theory parameters.

As for the comment, I completely agree that we should add comments to the card. Specifically, I would now only explain the description what they are but also point to the IDs explicitly. Having the explicit IDs always avoid unnecessary guessing.

jekoorn:
Is this what's meant with the more clear comments? I'm not exactly sure from the previous messages if having these "Frankenstein" theories, i.e. the ones `41_042_025 - 41_042_028` is actually a good or useful thing. Do we want to keep them? In any case, I added more clear descriptions. And is the current list of theory cards ok?

felixhekhorn:
The additional comments definitely help me. Let's ask for confirmation also to @scarlehoff . As far as I understand there is no simple way to avoid the Frankenstein theories, as they are genuinely different from their relatives, but that has no corresponding setting in the theory card. (By the way: Frankenstein is a very appropriate name, since they are the fusion of two halves :zombie: :see_no_evil: )

(you may need to rebase again, so the unrelated files (hopefully) disappear)

jekoorn:
I rebased on top of master just this morning, but I can see if I can do another trick. I like Frankenstein too!

EDIT: the unrelated commits disappeared after some magic!

Issue #2288, opened 2025-02-28
Update theory documentation


felixhekhorn:
We should update the theory documentation:

We should replace the old documentation here https://docs.nnpdf.science/theory/FastInterface.html as it is outdated.
Instead the page should contain the updated way:
- to calculate FK tables we need two (main) programs: pinefarm + pineko
- the mathematical concept of FK tables (with updated notation) is here: https://pineko.readthedocs.io/en/latest/theory/fktables.html
- the actual computation is a two step process:
  1. generate a (process-dependent) grid with pinefarm
  2. transform the grid into an FK table
- for 1.
  - point to https://pinefarm.readthedocs.io/en/latest/index.html
  - point to https://github.com/NNPDF/pinecards
- for 2: point to https://pineko.readthedocs.io/en/latest/index.html

(as per our email exchange just now)

felixhekhorn:
Questions for @scarlehoff : 
1. I'd like to drop the explicit table in `theoryindex.rst`, fine?
2. is `vp-list -r theories` still working? is anybody using that?
3. is `vp-checktheory` still working? is anybody using that?
4. is `validphys.loader.Loader.check_theoryID` stil working? is anybody using that?

scarlehoff:
1. Mmm, that one is painless and autogenerated, I think it is fine.
2. It does work, not sure how many people use `vp-list` in general, but it does work.
3. Same. I don't use it, not sure others.
4. Yes. It is what vp uses internally to download theories.

2 and 3 will work as long as the theories work because there's no really much functionality in those scripts, it is just a handy way of reading a theory and printing it

RoyStegeman:
I use neither of these commands. 

`vp-list` seems quite useless since the list is too long to read "manually" and I can't really many people grepping to see if a theoryID exist. 

`vp-checktheory` doesn't even fit the full `Comments`, so again, not sure how useful that really is. As soon as you want to do something as trivial as comparing two theories you already need something else. 

I'd be in favour of removing them. If people actually rely on them we'll hear the complaint

felixhekhorn:
re 1: How useful is it to have a table with 52 columns and 643 rows? on my screen it takes ~5500 x 21000 pixels ... yes I have used the page before myself, but still I wonder ... I'm worried about the power set of PTO x MHOU x QED x mt x alphas x ...

scarlehoff:
[...]

PR #2495, opened 2026-07-07
Update hyperlinks in documentation

### LINKED ISSUE #572: Integrate github with the documentation
@siranipour It would be great if links such as

https://docs.nnpdf.science/_modules/validphys/calcutils.html#all_chi2_theory

pointed to the github repo rather than to some generated page, like it does for e.g. numpy when you click on *Source*

https://docs.scipy.org/doc/numpy-1.17.0/reference/generated/numpy.linalg.eig.html#numpy.linalg.eig

Also it would be nice if there were shortcuts to refer to a given issue and PR within the docs.

evagroenendijk:


evagroenendijk:
@felixhekhorn I decided to also fix the equation references in https://docs.nnpdf.science/theory/PTevol.html so in the end it will still take some time

felixhekhorn:
wait a moment: I suggest to drop that page as is and instead point to https://eko.readthedocs.io, where all of this is already explained in great detail

@scarlehoff do you agree?

felixhekhorn:
cc #2288

scarlehoff:
It's the information exactly the same?

If yes, ok from me.

evagroenendijk:
@felixhekhorn Ah ok, I had not seen that in #2288 this stuff is already being done. Shall I check if the info is the same & drop it if yes, or leave it to #2288?

felixhekhorn:
> It's the information exactly the same?

> @felixhekhorn Ah ok, I had not seen that in #2288 this stuff is already being done. Shall I check if the info is the same & drop it if yes, or leave it to #2288?

all the evolution information is covered by EKO - and actually this page here is severely outdated, e.g. it does not discuss QED, that we use exact these days, or SV scheme B, so better we drop it. And yes, @evagroenendijk please drop it here and then tick off the relevant box over there

Instead, TMC - which has nothing to do with evolution - should point to https://yadism.readthedocs.io/en/latest/implementation/TMC.html ... aaand then that page is empty, but then we should fix that there, where it belongs. The current content can not be simply transferred since it uses Mellin space (which yadism is not doing). Also the implementation in yadism is more complex as we allow, e.g., different levels (/orders) if TMC. Note that the code documentation does have some explanation ...

Any way, don't forget the main purpose of this PR :upside_down_face:

doc/sphinx/source/external-code/pdf-codes.rst
```diff
@@ -16,7 +16,7 @@ release.
 PDF evolution
 -------------
 
-`APFEL <https://apfel.hepforge.org/>`_ ('A PDF Evolution Library') is the PDF evolution code currently
+`APFEL <https://github.com/scarrazza/apfel>`_ ('A PDF Evolution Library') is the PDF evolution code currently
```
felixhekhorn: "currently used"? definitely not. Note that "Note that at the time of writing ..." has passed 🙃  -> replace with EKO. It can be a very short paragraph in my opinion, because EKO has it's detailed documentation. Drop APFEL as a whole. HOPPET can stay in the light of #2485 under "Other codes", though the content should also be cut down to the bare minimum such that it does not specify outdated information (e.g. by now HOPPET can also do N3LO)

doc/sphinx/source/theory/PTevol.rst
felixhekhorn, line 1: I would drop the file as a whole - and if you wish to keep the entry in the menu, which maybe is fine, point it directly to the EKO documentation.

TMC has nothing to do with evolution and in fact now that DIS is becoming less prominent, it only applies to this very specific set of observables

doc/sphinx/source/external-code/pdf-codes.rst
```diff
@@ -42,7 +42,7 @@ the subset that most truthfully reproduces the underlying probability distributi
 Other codes
```
felixhekhorn, line 36: the pineline should be featured prominently here (again with a (very) short paragraph pointing to the relevant documentation https://nnpdf.github.io/pineline )

felixhekhorn (commented):
As I already said: updating the documentation is nice and good - but was not the intended task for this PR

evagroenendijk:
> As I already said: updating the documentation is nice and good - but was not the intended task for this PR

Sure! That's why I did not look at the text, I just updated the links. Actually as far as the nnpdf documentation goes, I pushed all the changes I made already (I did not find any other deprecated links)

felixhekhorn:
> As I already said: updating the documentation is nice and good - but was not the intended task for this PR

ehm, actually, looking at [28ae19b](https://github.com/NNPDF/nnpdf/pull/2495/commits/28ae19bdede581336f5bc2b9c7c24bacd2dba62a) let me ask to be sure: is it clear what #572 wants to do? Specifically, we are not talking about any explicitly written URL in our documentation (like the ones you change in that commit) - we are talking about the automatic link generated by sphinx

> I did not find any other deprecated links

(maybe this indicates we are not talking about the same thing :see_no_evil: )

evagroenendijk:
> > As I already said: updating the documentation is nice and good - but was not the intended task for this PR
> 
> ehm, actually, looking at [28ae19b](https://github.com/NNPDF/nnpdf/pull/2495/commits/28ae19bdede581336f5bc2b9c7c24bacd2dba62a) let me ask to be sure: is it clear what #572 wants to do? Specifically, we are not talking about any explicitly written URL in our documentation (like the ones you change in that commit) - we are talking about the automatic link generated by sphinx
> 
> > I did not find any other deprecated links
> 
> (maybe this indicates we are not talking about the same thing 🙈 )

Ah okay, then indeed I misunderstood the issue...

evagroenendijk:
@felixhekhorn I think I can use this https://www.sphinx-doc.org/en/master/usage/extensions/linkcode.html then to solve the issue, would you agree? It can then point to the n3fit and validphys directories in the repository

felixhekhorn:
> @felixhekhorn I think I can use this https://www.sphinx-doc.org/en/master/usage/extensions/linkcode.html then to solve the issue, would you agree? It can then point to the n3fit and validphys directories in the repository

not sure - this is the right direction, but then you would need to provide the mapping between name of "the thing" (function, class) and the URL yourself. I hope there are better solutions out there, which are already targeted at GitHub and do all the magic for you

scarlehoff:
> Any way, don't forget the main purpose of this PR 🙃

Indeed.

I was thinking something like this: https://github.com/TDKorn/sphinx-github-style not sure whether it is the best option (we can check what numpy is doing exactly I guess) but something like this.

RE linking yadism or EKO. Whenever the information is there, I'm happy with just linking it. But if it is not, better to leave porting the content for later (ideally, open an issue so that it gets assigned later!)

felixhekhorn:
> I was thinking something like this: https://github.com/TDKorn/sphinx-github-style not sure whether it is the best option (we can check what numpy is doing exactly I guess) but something like this.

that's better, but maybe still not the best way - which I would agree is the numpy way. It might be worth checking their implementation explicitly
 
> RE linking yadism or EKO. Whenever the information is there, I'm happy with just linking it. But if it is not, better to leave porting the content for later (ideally, open an issue so that it gets assigned later!)

https://github.com/NNPDF/yadism/issues/386 - however, I would remove the current stuff since at best it is confusing (we're surely not doing it in Mellin space) and at worst it is just wrong

evagroenendijk:
@felixhekhorn @scarlehoff it looks like numpy is directly using sphinx.ext.linkcode (https://github.com/numpy/numpy/blob/main/doc/source/conf.py#L496C1-L598C1). I'll check what would be easiest for us

felixhekhorn:
> @felixhekhorn @scarlehoff it looks like numpy is directly using sphinx.ext.linkcode (https://github.com/numpy/numpy/blob/main/doc/source/conf.py#L496C1-L598C1). I'll check what would be easiest for us

if that works, that's just fine - but then we should take some heavy inspiration (a.k.a. copying) from what they do ... I was mostly worried that mapping would be complicated, but if we can just copy, that is fine

evagroenendijk:
> > @felixhekhorn @scarlehoff it looks like numpy is directly using sphinx.ext.linkcode (https://github.com/numpy/numpy/blob/main/doc/source/conf.py#L496C1-L598C1). I'll check what would be easiest for us
> 
> if that works, that's just fine - but then we should take some heavy inspiration (a.k.a. copying) from what they do ... I was mostly worried that mapping would be complicated, but if we can just copy, that is fine

I think roughly copying what they do should work. 

But I'm only wondering one thing, namely why do we care about doing all this if we have exclude_patterns = ['modules/**'] (https://github.com/NNPDF/nnpdf/blob/4ae1a8c0777d883dc40735c67325a9742a0710da/doc/sphinx/source/conf.py#L78)? Won't that just ignore all these modules when building the actual documentation? Or am I missing something?

scarlehoff:
We are currently not linking the docs into the documentation. We want to start linking the docs, so it might be necessary to change `conf.py` (I haven't looked at this exclude pattern in particular but yes, it might be that it needs to be changed)

evagroenendijk:
@felixhekhorn @scarlehoff I implemented the changes in conf.py. I basically copied the numpy method. The only things I changed in their implementation are:
- changed "numpy" -> n3fit, evolven3fit, validphys, nnpdf_data and changed the corresponding github repo
- removed lines where they account for the fact that some of their code is in c

Then I tested it by asking codex to run it and provide me with a table that shows the modules and corresponding github links, which were the correct ones. 

Note two things:
1) it's possible now to also provide the links for nnpdf_data, but it's not compiled in the Makefile at the moment
2) if we want to start linking the docs as Juan commented there are some extra changes to be made but the setup is there

Let me know if it looks ok to you!

felixhekhorn:
> Then I tested it by asking codex to run it

please test at least once yourself, i.e. build the documentation yourself and click on a link

And this maybe will help you also to fix the action https://github.com/NNPDF/nnpdf/actions/runs/29324443681/job/87057256138?pr=2495 (which you still need to do of course) (I had a brief glance there but I don't know what actually is failing ...)

evagroenendijk:
> > Then I tested it by asking codex to run it
> 
> please test at least once yourself, i.e. build the documentation yourself and click on a link
> 
> And this maybe will help you also to fix the action https://github.com/NNPDF/nnpdf/actions/runs/29324443681/job/87057256138?pr=2495 (which you still need to do of course) (I had a brief glance there but I don't know what actually is failing ...)

Ok, will do!

I think the failing action has something to do with the rust files that I adjusted, I'll fix it when I finish that!

doc/sphinx/source/theory/PTevol.rst
felixhekhorn, line 1: I'm happy with the removal - but someone else was pointing to this page and you need to fix that as well 🙃

evagroenendijk:
@felixhekhorn I'm finished making the changes. 

About testing the links, it's not possible to click a link in the generated documentation since they are currently not explicitly used inside the documentation (see comment by Juan). But I generated a test file containing one of the links, and this worked fine.

felixhekhorn:
> About testing the links, it's not possible to click a link in the generated documentation since they are currently not explicitly used inside the documentation (see comment by Juan).

I'm not sure I understand - why is it not working? which comment do you mean? are we not linking to our code? (in which case this whole issue and PR would be void, so I don't think that can be the case, can it?) or are the link targets not there? then we need to provide them of course ...

evagroenendijk:
> > About testing the links, it's not possible to click a link in the generated documentation since they are currently not explicitly used inside the documentation (see comment by Juan).
> 
> I'm not sure I understand - why is it not working? which comment do you mean? are we not linking to our code? (in which case this whole issue and PR would be void, so I don't think that can be the case, can it?) or are the link targets not there? then we need to provide them of course ...

That is also what I was wondering, if I understood @scarlehoff correctly we are now not linking in the docs, but at some point we want to and that's why we have to make this change. In any case I think
 https://github.com/NNPDF/nnpdf/blob/4ae1a8c0777d883dc40735c67325a9742a0710da/doc/sphinx/source/conf.py#L78

makes sure that at the moment these modules are ignored by the makefile. Did I understand you correctly @scarlehoff ?

scarlehoff:
At the moment what we have in the docs is this https://docs.nnpdf.science/_modules/index.html
and what we would like is to have instead the a
structure linked with github.

I haven't been able to follow the discussion so not sure what the issue is in the PR. But the issue to be solved is that one.

felixhekhorn:
> At the moment what we have in the docs is this https://docs.nnpdf.science/_modules/index.html and what we would like is to have instead the a structure linked with github.

do we have a link somewhere which actually points there @scarlehoff ?

evagroenendijk:
> https://docs.nnpdf.science/_modules/index.html

Ok, then I think we have to make some more changes as well. I.e. remove the modules from exclude_pattern and I think add some other things as well. Because as far as I understand, this sphinx viewcode makes this page, but linkcode (which I've implemented now) does not automatically

(I think we have to make a dedicated rst file for it then)

scarlehoff:
> do we have a link somewhere which actually points there @scarlehoff ?

I think we removed it? It used to be in the main page.
(Not sure whether some are left in there)

evagroenendijk:
@scarlehoff @felixhekhorn Alright, so what do we want to do? Do we want to have this page, or do we want to leave it like this? If we would ever want to start putting links in the docs we can easily manage already

scarlehoff:
We would like it (as an API reference page) but we want it to look more like this:

https://numpy.org/doc/stable/reference/generated/numpy.datetime_as_string.html#numpy.datetime_as_string

i.e., just the docstring (ideally the examples, if any, separated)


However, and maybe this was part of the confusion (and I didn't realize until now, sorry!) I just noticed that numpy no longer has the link to the source in the documentation! 

It used to be like this, with [source] which would take you too the github code.
https://numpy.org/doc/1.19/reference/generated/numpy.ma.MaskedArray.__int__.html#numpy.ma.MaskedArray.__int__

I wonder why it changed... I liked it a lot.

In any case, the point is that our API reference is just a dump of the code (like https://docs.nnpdf.science/_modules/n3fit/layers/DIS.html ) which is not very useful.
Given the numpy change, let's just ensure that we have an API reference that is the docstring and examples, like EKO
https://eko.readthedocs.io/en/latest/modules/eko/eko.html#module-eko.beta

and let's leave the source linking to github for later given that numpy dropped it (which means the original instruction of just following numpy could never work...)

evagroenendijk:
> We would like it (as an API reference page) but we want it to look more like this:
> 
> https://numpy.org/doc/stable/reference/generated/numpy.datetime_as_string.html#numpy.datetime_as_string
> 
> i.e., just the docstring (ideally the examples, if any, separated)
> 
> However, and maybe this was part of the confusion (and I didn't realize until now, sorry!) I just noticed that numpy no longer has the link to the source in the documentation!
> 
> It used to be like this, with [source] which would take you too the github code. https://numpy.org/doc/1.19/reference/generated/numpy.ma.MaskedArray.__int__.html#numpy.ma.MaskedArray.__int__
> 
> I wonder why it changed... I liked it a lot.
> 
> In any case, the point is that our API reference is just a dump of the code (like https://docs.nnpdf.science/_modules/n3fit/layers/DIS.html ) which is not very useful. Given the numpy change, let's just ensure that we have an API reference that is the docstring and examples, like EKO https://eko.readthedocs.io/en/latest/modules/eko/eko.html#module-eko.beta
> 
> and let's leave the source linking to github for later given that numpy dropped it (which means the original instruction of just following numpy could never work...)

I agree that it's not nice to have a dump of the code. But wouldn't it be nicer if I just make a page like this https://docs.nnpdf.science/_modules/index.html, only then linking the associated github pages to the modules? At least in my opinion the github linking is nice

scarlehoff:
Yes. But what I think would be nicer is to have the docstrings in the documentation (just like with EKO). Many of the functions are well documented now so it is a pity not to take advantage of it.

(and then linking to github even better of course)

scarlehoff:
The test that is not passing (`deploy_docs`) is due to the many warnings that are being produced by the documentation. I guess this is due to the new stuff being now "compiled" into the documentation.

I suggest, before fixing it, to rebase (rebase!) this on top of https://github.com/NNPDF/nnpdf/pull/2499 as Ella has fixed many docstring which will create many conflicts if you change things here as well.

Issue #2457, opened 2026-04-21
Check whether normalized distributions covmats have a 0 eigenvalue


scarlehoff:
If so, diagonalizing the covmat and removing 0s would be enough.

jacoterh:
[normalised_dist_eigvals.pdf](https://github.com/user-attachments/files/27497390/normalised_dist_eigvals.pdf)

Conclusion: contrary to what one naively expects, the experimental covariance matrix of normalised differential distributions does not have a single vanishing eigenvalue.

scarlehoff:
I guess then the answer is no. Tagging @enocera (but he might already be aware I guess) and closing.

PR #2496, opened 2026-07-07
Add missing cuts on top pair data

enocera:
This PR implements some additional cuts on new top-quark pair distributions to remove the potentially linearly dependent bin of each normalised distribution.


validphys2/src/validphys/cuts/filters.yaml
```diff
@@ -49,18 +43,6 @@
    linearly dependent on the others
   rule: "m_ttBar<1000"
 
-- dataset: CMS_TTBAR_8TEV_LJ_DIF_YT-NORM
-  reason: |
-   We remove the last bin of the LHCB_DY_8TEV_MUON_Ynormalised distribution because it is
```
scarlehoff, line 54: I wonder how did LHCB end up there 😆

scarlehoff (approved):
I'm fine with this but I wonder whether it wouldn't be better to use `idat` for all of them, so that we select one particular bin (and not based on a kinematic variable).

Also, the bin is completely arbitrary, right? Was it important in the end which bin was chosen? I don't remember whether we did test this in the end.

I remember @jacoterh checked the covmat didn't actually have a 0 though... https://github.com/NNPDF/nnpdf/issues/2457

validphys2/src/validphys/cuts/filters.yaml
enocera: There were a bunch of these, actually - obviously I removed all of them but one.

enocera:
> I'm fine with this but I wonder whether it wouldn't be better to use `idat` for all of them, so that we select one particular bin (and not based on a kinematic variable).
> 
> Also, the bin is completely arbitrary, right? Was it important in the end which bin was chosen? I don't remember whether we did test this in the end.
> 
> I remember @jacoterh checked the covmat didn't actually have a 0 though... #2457

The rationale right now is to always use the last bin (with the poorest statistics). I seems to remember that there was an executive agreement on that. The specific bin we remove is of course arbitrary, however consistently removing the last one seems to me the easiest convention and the more convenient (we are always on the tail of the distribution).

In all the top-quark pair data sets the removed bin is always the last. I can switch to the bin number consistently.

validphys2/src/validphys/cuts/filters.yaml
enocera: Fixed.

enocera:
@scarlehoff I have now moved to using the rule on idat for all data sets.

scarlehoff (approved):
i wonder whether `idat!=-1` would work...

but ok, not very important, thanks!

PR #2498, opened 2026-07-08
N3lo benchmark

evagroenendijk:
The theory cards used to produce DIS predictions for the N3LO benchmark. Related to nnpdf/pinecards#199

reference document: https://docs.google.com/document/d/1xGwpdlhm93cKxmXuLVRMK4WINwYxmmZJ/edit


nnpdf_data/nnpdf_data/theory_cards/41002200.yaml
```diff
@@ -0,0 +1,44 @@
+ID: 41_002_200
+Comments: N3LO benchmark, NNPDF4.1 LO QCD-only (GM-VFNS=FONLL)
+PTO: 0
+FNS: FONLL-C
```
evagroenendijk: @felixhekhorn for this order is FONLL-C fine? I now put C for all of them, except for N3LO where I have E

felixhekhorn:
```suggestion
FNS: ZM-VFNS
```
We never implemented "FONLL-0" (which is a term we invented ourselves), i.e. FONLL at LO, so we can actually not do it properly - however, since we will only deal with perturbative charm in this exercise here, there is no GM-VFNS in the first place and thus we can safely assume ZM here.

(I think pineko will not like putting "FONLL-C" here ...)

nnpdf_data/nnpdf_data/theory_cards/41002201.yaml
```diff
@@ -0,0 +1,44 @@
+ID: 41_002_201
+Comments: N3LO benchmark, NNPDF4.1 NLO QCD-only (GM-VFNS=FONLL)
+PTO: 1
+FNS: FONLL-C
```
felixhekhorn:
```suggestion
FNS: FONLL-A
```

nnpdf_data/nnpdf_data/theory_cards/41002000.yaml
```diff
@@ -0,0 +1,44 @@
+ID: 41_002_000
+Comments: N3LO benchmark, NNPDF4.1 LO QCD-only (ZM-VFNS)
+PTO: 0
+FNS: ZM-VFNS
+DAMP: 0
+IC: 1
+Q0: 1.65
```
felixhekhorn:
```suggestion
Q0: 1.0
```
Here and everywhere else

nnpdf_data/nnpdf_data/theory_cards/41002000.yaml
```diff
@@ -0,0 +1,44 @@
...
+MZ: 91.19
+MW: 80.3692
+GF: 1.1663788e-5
+SIN2TW: 0.22348
+TMC: 0
+MP: 0.938
+Qref: 91.19
+alphas: 0.118
+alphaqed: 0.0075700
+nf0: 4
```
felixhekhorn:
```suggestion
nf0: 3
```
Here and everywhere else

felixhekhorn (changes requested):
If you want to read more about FONLL schemes (i.e. FONLL-A vs. FONLL-B vs. FOLL-C etc) see https://inspirehep.net/literature/2818030 or better the original paper https://inspirehep.net/literature/842785

nnpdf_data/nnpdf_data/theory_cards/41002000.yaml
evagroenendijk: I don't think we want this for the generation of the observables right? For the F2's we want to use our own PDFs, while only for evolving the toy PDF we want Q0=1, correct?

nnpdf_data/nnpdf_data/theory_cards/41002200.yaml
felixhekhorn: see comment above - we need to change it

nnpdf_data/nnpdf_data/theory_cards/41002000.yaml
felixhekhorn: As we discussed (and hopefully will soon be officially confirmed): I understood we want toy for everything

nnpdf_data/nnpdf_data/theory_cards/41002202.yaml
```diff
@@ -3,8 +3,8 @@ Comments: N3LO benchmark, NNPDF4.1 NNLO QCD-only (GM-VFNS=FONLL)
 PTO: 2
 FNS: FONLL-C
 DAMP: 0
```
evagroenendijk, line 5: @felixhekhorn Do we want DAMP=0 for the FONLL predictions? Seemed like the most simple thing to me, but I don't know if you think differently

PR #2504, opened 2026-07-27
Crash when combining point prescriptions with a user covmat

achiefa:
I was tryin to add the user covmat generated for the PC uncertainties on top of the point prescription. `vp-setupfit` crashed during reportengine graph resolution:

```
File ".../reportengine/utils.py", line 56, in saturate
    kwargs = {name: d[name] for name, param in s.parameters.items()
              ~^^^^^^
KeyError: 'group_dataset_inputs_by_metadata'
```

This goes away if I remove the user covmat from the runcard.

# Possible explanation
When both covmats are provided, `total_theory_covmat` is the provider which carries a check: `@check_fit_dataset_order_matches_grouped` and the related `@make_argcheck`. Now, my understanding is that `group_dataset_inputs_by_metadata` is a production rule that is not resolved in the namespec at the time of the argcheck, which datacuts::theory::theorycovmatconfig (argcheck does not trigger p`roduce_ `rules). I think this chain used to be correct by accident before @jacoterh and I modified this part of the code for the bug with the diagonal basis. Before the *_fitting providers (including `total_theory_covmat_fitting`) used
```
procs_index_matched(groups_index, procs_index),
```
taking `groups_index` and forcing the production of `group_dataset_inputs_by_metadata`. Now `procs_index_matched` is replaced by `data_input_matched_procs_index(data_input, procs_index)`, which doesn't call `groups_index`. 

At the same time, this was obselete and wasn't captured by the tests. In `total_theory_covmat_fitting`, the covmat is already reindexed using `data_input_matched_procs_index`. Therefore, it can be removed.

### DIFF
--- a/validphys2/src/validphys/tests/test_theorycovariance.py
+++ b/validphys2/src/validphys/tests/test_theorycovariance.py
@@ -1,0 +1,101 @@
+"""
+test_theorycovariance.py
+
+Tests for the theory covariance matrices built in
+:py:mod:`validphys.theorycovariance.construction`, and in particular for the
+``nnfit_theory_covmat`` production rule, which is what ``vp-setupfit`` asks for
+whenever a fit runcard contains a ``theorycovmatconfig``.
+"""
+
+import numpy as np
+import pytest
+
+from reportengine.table import savetable
+from validphys.api import API
+
+# The user covmat is looked up with ``Loader.check_vp_output_file``, which
+# rejects absolute paths, so it has to be found relative to the cwd. The name is
+# made distinctive so that it cannot collide with anything in the vp cache.
+USER_COVMAT_FILENAME = "test_user_covmat.csv"
+POINT_PRESCRIPTION = "3 point"
+
+
+@pytest.fixture(scope="module")
+def user_covmat_on_disk(tmp_path_factory, thcovmat_config):
+    """Write out a "user" theory covmat and return ``(directory, matrix)``.
+
+    The matrix written out is the ``3 point`` scale variation covmat for the very
+    same data, so that once it has been read back by ``fromfile_covmat`` the
+    resulting ``user_covmat`` is numerically identical to ``theory_covmat_custom``.
+    This exercises the real cut and index handling of ``fromfile_covmat`` and
+    gives the tests below an expected value with no hardcoded numbers in it.
+
+    ``theory_covmat_custom`` is float32; it is promoted to float64 before being
+    written, because the shortest repr of a float32 only round trips to float32
+    and would introduce a ~1e-8 relative error.
+    """
+    covmat = API.theory_covmat_custom(
+        point_prescriptions=[POINT_PRESCRIPTION], **thcovmat_config
+    ).astype(np.float64)
+    path = tmp_path_factory.mktemp("user_covmat")
+    savetable(covmat, path / USER_COVMAT_FILENAME)
+    return path, covmat
+
+
+@pytest.mark.parametrize(
+    ("covmat_config", "expected_factor"),
+    [
+        # Scale variations *and* a user covmat. This is the branch that used to
+        # fail during graph resolution with
+        # ``KeyError: 'group_dataset_inputs_by_metadata'``.
+        (
+            {"point_prescriptions": [POINT_PRESCRIPTION], "user_covmat_path": USER_COVMAT_FILENAME},
+            2.0,
+        ),
+        # Scale variations only.
+        ({"point_prescriptions": [POINT_PRESCRIPTION]}, 1.0),
+        # User covmat only.
+        ({"user_covmat_path": USER_COVMAT_FILENAME}, 1.0),
+        # User covmat only, rescaled by ``mult_factor``.
+        ({"user_covmat_path": USER_COVMAT_FILENAME, "mult_factor": 2.5}, 2.5),
+    ],
+    ids=["scalevar_and_user", "scalevar_only", "user_only", "user_only_rescaled"],
+)
+def test_nnfit_theory_covmat(
+    monkeypatch, thcovmat_config, user_covmat_on_disk, covmat_config, expected_factor
+):
+    """Every branch of ``produce_nnfit_theory_covmat`` must resolve and return a
+    covmat indexed in runcard order on both axes.
+
+    The ``scalevar_and_user`` case is a regression test: combining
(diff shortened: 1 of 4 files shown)


validphys2/src/validphys/tests/test_theorycovariance.py
```diff
@@ -0,0 +1,101 @@
...
+
+import numpy as np
+import pytest
+
+from reportengine.table import savetable
+from validphys.api import API
+
+# The user covmat is looked up with ``Loader.check_vp_output_file``, which
+# rejects absolute paths, so it has to be found relative to the cwd. The name is
+# made distinctive so that it cannot collide with anything in the vp cache.
```
scarlehoff: why not a temporary name??

validphys2/src/validphys/theorycovariance/theorycovarianceutils.py
```diff
@@ -85,30 +85,6 @@ def check_correct_theory_combination_internal(
...
-        for dsinput in group["data_input"]:
-            grouped_ds = dsinput.name
-            input_ds = next(data_input_iter).name
-            check(
-                grouped_ds == input_ds,
-                "Dataset ordering is changed by grouping, this will cause "
-                "errors when running fits with theory covmat. Datasets should "
-                f"be ordered by {processed_metadata_group} in the runcard.",
-            )
-
```
scarlehoff, line 110: I see this check is from the time the fitting code was c++ so I'm ready to believe it is obsolete.
However, I also know that two different PhD students three year apart spent many hours on a misordered sum of covmats.
Now it would be about 3 years later from last time... are you absolutely sure this is safe to remove?

Btw, a way to check whether the hypothesis is correct from what you say would be to simply add groups_index to data_input_matched_procs_index as a dummy argument. That should be enough to trigger the production.

validphys2/src/validphys/tests/test_theorycovariance.py
```diff
@@ -0,0 +1,101 @@
...
+
+    runcard_index = API.data_index(**thcovmat_config).droplevel(0)
+    process_index = API.procs_index(**thcovmat_config).droplevel(0)
+
+    # Guard against this test quietly becoming vacuous. It only has teeth as long
+    # as grouping DATA_THCOVMAT by process actually reorders it: LHCB_Z0_8TEV_MUON_Y
+    # is DY NC and so joins the group opened by CMS_Z0J_8TEV_PT-Y, overtaking
+    # ATLAS_WJ_8TEV_WP-PT, which is DY CC. If DATA_THCOVMAT is ever changed into an
+    # order that the grouping leaves alone, this test must be given its own
+    # dataset_inputs rather than being left silently trivial.
```
scarlehoff: Please simplify this comment so it can be quickly parsed by eye:
```
# Check this test is not triviak by ensuring the input datasets are sorted incorrectly 
```
big words scare me.

Another option is (since the index functions are cheap) to reorder so this function always check inputs with a "misaligned" runcard. As you prefer.

validphys2/src/validphys/theorycovariance/theorycovarianceutils.py
achiefa: > Btw, a way to check whether the hypothesis is correct from what you say would be to simply add groups_index to data_input_matched_procs_index as a dummy argument. That should be enough to trigger the production.

Yes, I tried to add `groups_index` as a dummy argument and it worked. However, I'm not in favour of this solution for two reasons. First, it's very hard to understand the reason why the dummy variable has been included. However, a comment may help in this case. Second, as you said we'd have an important part of the code (checking ordering of covmats) that rely on obselete code. Moreover, if we were to keep this obselete code, we'd have two functions checking the same thing in two not-so-different branches of the code.

validphys2/src/validphys/tests/test_theorycovariance.py
achiefa: This comment is way too long and says nothing. I'll change it.

validphys2/src/validphys/theorycovariance/theorycovarianceutils.py
scarlehoff: No, I don't think it should be added as an argument of course.
But it is a way to check whether it was a possible issue or the actual issue.

achiefa: Good. I reach these conclusions:
- `check_fit_dataset_order_matches_grouped` checks if `group_dataset_inputs_by_metadata` follows or not runcard order.
- However, `group_dataset_inputs_by_metadata` is never used to construct an index in the theory covmat pipeline.
- The method itself is called in https://github.com/NNPDF/nnpdf/blob/abee4f09bb67bdc797d972a9d640fda00c027106/validphys2/src/validphys/config.py#L1928-L1929
but it's a different thing.
- Anyway, this checked was a guard and didn't fix the order of the code: it just refused to run. Now the code fixes the ordering by reindexing. On the writing side,  `theory_covmat_custom_fitting` / `total_theory_covmat_fitting` / `user_covmat_fitting` push the covmat through `_reindex_covmat_to_fitting_order` with `data_input_matched_procs_index`. This latter matches with the order of the runcard, which is the same as the exp. covmat. On the reading side, `produce_loaded_theory_covmat` always reindexes.
- The critical part is https://github.com/NNPDF/nnpdf/blob/abee4f09bb67bdc797d972a9d640fda00c027106/validphys2/src/validphys/theorycovariance/construction.py#L544-L549
where both matrices are added. However, `fromfile_covmat` uses the same index as in `theory_covmat_custom_per_prescription`, namely `procs_index`. If their index mismatches, then the sum would complain rasing an error.

Given these points, I think it's safe to remove this function.

validphys2/src/validphys/tests/test_theorycovariance.py
achiefa: How would you then restructure the parametrization. Note that `tmp_path_factory` is already a tmp directory, and the test already goes into it. Would you build the config dict inside the test?

scarlehoff: I would not restructure anything. The comment says the name is to avoid clashes. An unique temporary name (with uuid for instance) would for sure avoid clashes.

vschutze-alt (approved):
I have had a look and ran some vp reports / vp-setupfit and all seems to be working

Issue #2086, opened 2024-05-18
Highlight variant differences in comparefits reports


scarlehoff:
At the moment only different datasets and cuts are mentioned, but datasets using a different variant should also be mentioned.

scarlehoff:
Related to https://github.com/NNPDF/nnpdf/issues/2397, if the hash is included even when the theory is the same we can know when the fktables have changed.

PR #2505, opened 2026-07-28
Highlight variant differences

andrpie:
addresses #2086

I still need to test whether this works as intended!!

It doesn't feel like the cleanest way to do it, especially with legacy dataset names. `match_datasets_by_name` already finds the intersection of datasets used in the fits, but it returns `DataSetSpec` rather than `DataSetInput` (the former has no variant information). Perhaps it would be better to parse variant name into `DataSetSpec`?

### DIFF
--- a/validphys2/src/validphys/comparefittemplates/report.md
+++ b/validphys2/src/validphys/comparefittemplates/report.md
@@ -120,6 +120,7 @@ Dataset differences and cuts
 ----------------------------
 {@print_dataset_differences@}
 {@print_different_cuts@}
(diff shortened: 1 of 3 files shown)


andrpie:
Dear @scarlehoff, thank you for your patience on this one.
This is a sample output of `vp-comparefits` run on this branch: https://vp.nnpdf.science/iu35Q7S7Qk27_IwAfodB3w==
At the very bottom, there is a table summarising variant differences. I believe this can look better - please let me know what you think.

Issue #2264, opened 2025-01-22
Fix ugly kinematics


scarlehoff:
In the new implementation, the kinematics have been left as they were in the old implementation. Without the magic plotoptions within validphys, some of these are ugly-ish in plot, like e.g., https://vp.nnpdf.science/EK2J-BJyRV2_zKT2-Zeoyg==/ where m^2 is used when it would be nicer to have m.

These doesn't have any effect anywhere (if the cuts are updated accordingly), so please, everyone, feel free to update the kinematic selection of any dataset you find produces ugly plots :)

scarlehoff:
To work on this the ideal workflow I'd say is:

1. Prepare a validphys runcard with `plot_fancy` (+ use_cuts internal and a kinematic coverage plot) with the dataset to be changed.
2. Create the plots, check whether they look ugly. For instance a logarithmic scale is used when a linear one would be more convenient or viceversa. Or pt^2 is used when pt is a more natural variable.
3. Change the `filter.py` or the kinematics file for the dataset (as well as the metadata) to use a different variable or to use linear/log plot, etc.
4. Redo the plot and check how they look.

PR #2507, opened 2026-08-24
Fix kinematics

ecole41:
This PR aims to fix the issue raised in[ Issue #2264](https://github.com/NNPDF/nnpdf/issues/2264#event-28267925655).

- [x] Use mass instead of masses square for DY (both FT and collider)
- [x] Remove sqrts from any dataset for which it is included in the name.


ecole41:
There are a few datasets which use `sqrts` for the plotting (e.g `plot_x: sqrts`). In this case, `sqrts` is needed in the kinematics files even if it is available in the dataset name.

scarlehoff:
> There are a few datasets which use `sqrts` for the plotting (e.g `plot_x: sqrts`). In this case, `sqrts` is needed in the kinematics files even if it is available in the dataset name. 

Which datasets do that? That's a very strange (and probably wrong) choice, because all points would be on top of each other.

ecole41:
> > There are a few datasets which use `sqrts` for the plotting (e.g `plot_x: sqrts`). In this case, `sqrts` is needed in the kinematics files even if it is available in the dataset name.
> 
> Which datasets do that? That's a very strange (and probably wrong) choice, because all points would be on top of each other.

It looks like they are in total ttbar cross-section measurements which only include one point, such as `ATLAS_TTBAR_13TEV_TOT`

scarlehoff:
Ah, I see, one quick solution would be to just change plot_x to the mass of the top which is often (always?) one of the variables. Just to make all datasets consistent.

scarlehoff (approved):
Thanks.

I prepared a report with all the figures https://vp.nnpdf.science/LV2MZ80IRAm6Lb0c1fHfOw==

but trying to go through them proved more than what my old eyes can do :__)

The only two points which might be worth updating for completeness are the absolute value in these datasets:

https://vp.nnpdf.science/LV2MZ80IRAm6Lb0c1fHfOw==/#ATLAS_DY_7TEV_36PB_ETA_plot_fancy_0

and

https://vp.nnpdf.science/LV2MZ80IRAm6Lb0c1fHfOw==/#ATLAS_Z0_7TEV_46FB_CF-Y_plot_fancy_0


And, a logarithmic scale for

https://vp.nnpdf.science/LV2MZ80IRAm6Lb0c1fHfOw==/#ATLAS_Z0J_8TEV_PT-M_plot_fancy_2

If you could update these I think this can be merged immediately.

ecole41:
@scarlehoff I have just edited the metadata files to address your comments

PR #2511, opened 2026-08-29
Fix-ish hyperopt leak

scarlehoff:
I haven't been able to actually fix the memory leak, so I took a different route. Instead of fixing the leak, now each hyperopt trial is run as forked process that dies once the trial finishes.

With this we can run hyperopt for as many trials we want without hitting OOM. I'm bypassing it when running hyperopt in parallel because in that case one can just leave the database running and run many small jobs and then it would be better not to have the overhead of these changes.

Not very beautiful but functional*. I've left also my previous attempts (in `internal_state.py`) at cleaning up tensorflow which were not enough but they did clean up things that were dangling so I think it might be worth having them.

(tested by submitting many jobs at BSC and none of them failed due to an OOM, which was the case before! I'll add an actual benchmark)

### DIFF
--- a/n3fit/src/n3fit/backends/keras_backend/internal_state.py
+++ b/n3fit/src/n3fit/backends/keras_backend/internal_state.py
@@ -18,7 +18,15 @@ import keras
 from keras import backend as K
 import numpy as np
 
+from validphys.convolution import central_predictions, predictions
+
 log = logging.getLogger(__name__)
+
+
+def _internal_backend_clear():
+    """Keras might need extra help to clear some backends."""
+    pass
+
 
 # Prepare Keras-backend dependent functions
 if (kback := K.backend()) == "torch":
@@ -36,6 +44,7 @@ if (kback := K.backend()) == "torch":
 
 elif K.backend() == "tensorflow":
     import tensorflow as tf
+    from tensorflow.python.framework import ops
 
     def set_eager(flag=True):
         """Set eager mode on or off
@@ -61,6 +70,14 @@ elif K.backend() == "tensorflow":
             log.warning(
                 "Could not set tensorflow parallelism settings from n3fit, maybe tensorflow is already initialized by a third program"
             )
+
+    def _internal_backend_clear():
+        """TensorFlow saves the gradient of the custom models we create as custom gradients.
+        These are not followed by keras and are thus not cleared during the clear_backend call."""
+        registry = ops.gradient_registry._registry
+        for name in tuple(registry):  # so python don't complain about the waning dictionary
+            if name.startswith("CustomGradient-"):
+                registry.pop(name)
 
 elif K.backend() == "jax":
 
@@ -120,6 +137,12 @@ def clear_backend_state():
     and unused memory.
     """
     log.info("Clearing session")
+    # The cache of validphys' predictions might be keeping a reference to the N3PDF model, remove
+    central_predictions.cache_clear()
+    predictions.cache_clear()
+    # Then clear the backend
+    _internal_backend_clear()
+    # and finally Keras
     K.clear_session()
 
 
--- a/n3fit/src/n3fit/backends/keras_backend/internal_state.py
+++ b/n3fit/src/n3fit/backends/keras_backend/internal_state.py
@@ -18,8 +18,6 @@ import keras
 from keras import backend as K
 import numpy as np
 
-from validphys.convolution import central_predictions, predictions
-
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


Radonirinaunimi:
Thanks @scarlehoff for this. I will have a look ASAP before the end of this week.

scarlehoff:
Thanks! Turns out this is going to be important because we need to redo the hyperopt with MHOU (which certainly requires be very light on memory :__ )

scarlehoff:
The promised benchmark, running in `montblanc` in Milan, with the memory peaks:

A first run small run (so take the times with a grain of salt, since due to the computer and the size of the runcard, the overhead is irrelevant, the memory was certainly the cause of the difference)
```
master: 22 GB, 2.5 hours

this branch:  16 GB, 30 minutes
```

A second run, with a greater number of datasets and replicas

```
master: nothing, OOM, i got a peak of 28 GB before the OOM which happened after 1 hours

this branch: 24 GB, it finished after 6 hours
```

n3fit/src/n3fit/backends/keras_backend/internal_state.py
```diff
@@ -18,8 +18,16 @@
 from keras import backend as K
 import numpy as np
 
+from validphys.convolution import central_predictions, predictions
```
Radonirinaunimi: It seems to me that this now violates the entire purpose of `internal_state.py`. It is supposed to be a generic backend but now it imports `validphys.convolution` and nukes its process-wide cache on every trial.

For the sake of maintainability, I think it would be best to pass `central_predictions.cache_clear` and `predictions.cache_clear` as arguments to `clear_backend_state` and hence avoid the module import.

n3fit/src/n3fit/backends/keras_backend/internal_state.py
```diff
@@ -62,6 +71,14 @@ def set_threading(threads, cores, double_precision=False):
                 "Could not set tensorflow parallelism settings from n3fit, maybe tensorflow is already initialized by a third program"
             )
 
+    def _internal_backend_clear():
+        """TensorFlow saves the gradient of the custom models we create as custom gradients.
+        These are not followed by keras and are thus not cleared during the clear_backend call."""
+        registry = ops.gradient_registry._registry
```
Radonirinaunimi: I am always hesitant about the usage of these private API without safeguards. Can we at least enclose it in a try/except + warnings? That'd be better than a hard failure when this changes in newer versions.

n3fit/src/n3fit/hyper_optimization/hyper_scan.py
```diff
@@ -45,9 +48,55 @@
...
+        finally:
+            sender.close()
+
+    process = context.Process(target=_run_objective)
+    process.start()
+    # The parent only receives, close its sender
+    sender.close()
+    try:
+        # Wait for the child to finish running
+        success, ret = receiver.recv()
```
Radonirinaunimi, line 83: In principle, for some god knows what reasons, a child process could deadlock, causing the job to hang (with no way of telling if it's stuck or slow) (?). Should we perhaps add a timeout to avoid this?

Radonirinaunimi (commented):
This looks generally fine to me. I only have very minor comments/suggestions below and they are mainly pedantic aspects. What I think are really missing though are unit tests, especially for `_run_trial_in_subprocess`.

We could for example have simple tests that:
- the actual forks really have different PID than the parent
- in case of no forks/macos it falls back directly to the current process
- the timeout is enforced
- ...

n3fit/src/n3fit/backends/keras_backend/internal_state.py
scarlehoff: I'll do the `try-except` (although fwiw it could even be removed, its effect is tiny in comparison with the separate process)

n3fit/src/n3fit/hyper_optimization/hyper_scan.py
scarlehoff: You are right, but at the same time we don't know beforehand we don't know how long a particular architecture should take.
I _think_ this is a problem that will be solved by the cluster admin's email saying "your job has been blocking a GPU at 0% usage for the last 24 hours" :P

(more seriously, I wouldn't know how to solve this problem effectively)

n3fit/src/n3fit/backends/keras_backend/internal_state.py
scarlehoff: Done (but since it is only these two, I've opted for keeping themin the model trainer hyperoptimization)

scarlehoff: Done (but since it is only these two, I've opted for keeping themin the model trainer hyperoptimization)

n3fit/src/n3fit/hyper_optimization/hyper_scan.py
Radonirinaunimi: That's true. In most cases, it'd be even by the scheduler when reaching the maximum time. As I mentioned, this is really a very minor thing so we could disregard.

My thinking was that setting a "reasonable" threshold would be a slightly better safeguard in that it will free up resources for other trials/processes to run (instead of just blocking it). And by reasonable I simply mean a number that we can tolerate (ie do we even expect or tolerate a run that exceeds 20 hrs?).

scarlehoff: We might want that if we are allowing for an extremely long number of epochs.

scarlehoff: We might want that if we are allowing for an extremely long number of epochs.

Radonirinaunimi (approved):

PR #2512, opened 2026-09-07
Added LLM guidelines to the documentation

scarlehoff:
As discussed in Morimondo. Let me know if you'd like to modify anything.

### DIFF
--- a/doc/sphinx/source/contributing/rules.rst
+++ b/doc/sphinx/source/contributing/rules.rst
@@ -55,9 +55,9 @@ Our policy for LLM generated contributions is similar to that of the `Debian pro
 
 * The committer is solely responsible for the content of the commit. This includes not only the quality of the code, legibility and future maintainability but also that attribution is given when attribution is due (be careful with licenses!)
 
-* The committer should be the **human**, not the model. This is important, we are an academic project where developers come and go as they finish their projects, so it is important to know **who** actually introduced the changes.
+* The committer must be the **human**, not the model. This is important, we are an academic project where developers come and go as they finish their projects, so it is important to know **who** actually introduced the changes.
 
-* AI usage should be documented in the PR. Please include also the harness and models you used. This helps the reviewer understand the scope of certain changes (e.g., the tests were done automatically, while the documentation was written by hand) and allow us to keep a record, and also understand the evolution of the technology.
+* AI usage must be documented in the PR. Include for example the harness and models you used. This helps the reviewer understand the scope of certain changes (e.g., the tests were done automatically, while the documentation was written by hand) and allow us to keep a record, and also understand the evolution of the technology.
 
 * We are an academic collaboration, the goal of the documentation and code comments should be to teach and train current students and future researchers, and should be written with that goal in mind, not LLMs. More generally, anything that could be useful for LLMs' ``skills.md`` or ``agents.md`` file, would be better suited as an edit to our online documentation so that it is also useful for humans reading the code!
 
--- a/doc/sphinx/source/contributing/rules.rst
+++ b/doc/sphinx/source/contributing/rules.rst
@@ -46,6 +46,20 @@ implements new convolution methods:
 It demonstrates how to add a new feature, together with relevant tests and
 documentation, and refine it based on the discussion.
 
+.. _aiusage:
+
+LLM Usage Guideliens
+--------------------
+
+Our policy for LLM generated contributions is similar to that of the `Debian project <https://lwn.net/Articles/1091231/>`_ and can be summarized as:
+
+* The committer is solely responsible for the content of the commit. This includes not only the quality of the code, legibility and future maintainability but also that attribution is given when attribution is due (be careful with licenses!)
+
+* The committer should be the **human**, not the model. This is important, we are an academic project where developers come and go as they finish their projects, so it is important to know **who** actually introduced the changes.
+
+* AI usage should be documented in the PR. Please include also the harness and models you used. This helps the reviewer understand the scope of certain changes (e.g., the tests were done automatically, while the documentation was written by hand) and allow us to keep a record, and also understand the evolution of the technology.
+
+* We are an academic collaboration, the goal of the documentation and code comments should be to teach and train current students and future researchers, and should be written with that goal in mind, not LLMs. More generally, anything that could be useful for LLMs' ``skills.md`` or ``agents.md`` file, would be better suited as an edit to our online documentation so that it is also useful for humans reading the code!
 
 .. _reviews:
 
--- a/doc/sphinx/source/contributing/rules.rst
+++ b/doc/sphinx/source/contributing/rules.rst
@@ -59,7 +59,9 @@ Our policy for LLM generated contributions is similar to that of the `Debian pro
 
 * AI usage must be documented in the PR. Include for example the harness and models you used. This helps the reviewer understand the scope of certain changes (e.g., the tests were done automatically, while the documentation was written by hand) and allow us to keep a record, and also understand the evolution of the technology.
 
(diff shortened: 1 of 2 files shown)


doc/sphinx/source/contributing/rules.rst
```diff
@@ -46,6 +46,20 @@ implements new convolution methods:
...
 documentation, and refine it based on the discussion.
 
+.. _aiusage:
+
+LLM Usage Guideliens
+--------------------
+
+Our policy for LLM generated contributions is similar to that of the `Debian project <https://lwn.net/Articles/1091231/>`_ and can be summarized as:
+
+* The committer is solely responsible for the content of the commit. This includes not only the quality of the code, legibility and future maintainability but also that attribution is given when attribution is due (be careful with licenses!)
```
felixhekhorn: Mention references? speaking about academia ...

doc/sphinx/source/contributing/rules.rst
```diff
@@ -46,6 +46,20 @@ implements new convolution methods:
...
+
+* The committer should be the **human**, not the model. This is important, we are an academic project where developers come and go as they finish their projects, so it is important to know **who** actually introduced the changes.
```
felixhekhorn:
```suggestion
* The committer must be the **human**, not the model. This is important, we are an academic project where developers come and go as they finish their projects, so it is important to know **who** actually introduced the changes.
```
? Otherwise the above statement ("The committer is solely responsible") does not make sense, right?

doc/sphinx/source/contributing/rules.rst
```diff
@@ -46,6 +46,20 @@ implements new convolution methods:
...
+
+* AI usage should be documented in the PR. Please include also the harness and models you used. This helps the reviewer understand the scope of certain changes (e.g., the tests were done automatically, while the documentation was written by hand) and allow us to keep a record, and also understand the evolution of the technology.
```
felixhekhorn:
```suggestion
* AI usage must be documented in the PR. Include for example the harness and models you used. This helps the reviewer understand the scope of certain changes (e.g., the tests were done automatically, while the documentation was written by hand) and allow us to keep a record, and also understand the evolution of the technology.
```
?

felixhekhorn (commented):
- There is the case of dependabot - which is clearly no LLM, but also not a human. I'm still quite in favor of allowing that since it fixes known security problems with a simple (typically few line) change in a lock file. Should we mention it explicitly someway?
- There is the case of pre-commit: we already removed the `autoupdate` from all repos (I think). Should we explicitly discourage `autofix` (and `autoupdate`)?

doc/sphinx/source/contributing/rules.rst
scarlehoff: You mean adding something like this?

```suggestion
* The committer is solely responsible for the content of the commit. This includes not only the quality of the code, legibility and future maintainability but also that attribution is given when attribution is due. Be careful with licenses and remember to add references and citations when needed!
```

scarlehoff:
They changed the interface ever so slightly and I am now many years old, so I manage to write my reply to you as the commit msg instead of here, lol
I'll rebase later to amend that because I feel very silly.

> There is the case of dependabot - which is clearly no LLM, but also not a human. I'm still quite in favor of allowing that since it fixes known security problems with a simple (typically few line) change in a lock file. Should we mention it explicitly someway?

You know my opinion on dependabot is quite different :_)

But that said, it is not LLM-generated and it is not prompted by a human anyway so I think it doesn't belong here. If there were a case of "we need to bump this library because of this specific feature" I would prefer that to be in the commit rather than simply "bump to version x".

> There is the case of pre-commit: we already removed the autoupdate from all repos (I think). Should we explicitly discourage autofix (and autoupdate)?

What do you mean? In general the changes should come "pre-committed" in the PRs (but again, a separate problem)
In /this/ repository in particular we have the problem that we still have pre-pre-commit code so sometimes the files get huge unwarranted changes but that should not happen in the others (and it is happening less and less here)

felixhekhorn:
> What do you mean? In general the changes should come "pre-committed" in the PRs (but again, a separate problem) 

yes, but for lazy people you can also let the CI (= a bot) do that job for you. We could add a sentence á là "We discourage the use of `autofix` and `autoupdate` with pre-commit.ci" or similar if we consider that necessary. (I'm just looking to the contributors in our repos.)

scarlehoff:
For me they are a different category as they don't add features nor they solve bugs. They increase the noise, as one has to go one commit down in the blame to find who introduced some particular change, but they don't hide attribution.

I'm generally against anything that touches the code without a human at the helm, but not enough to write it in stone I think.

doc/sphinx/source/contributing/rules.rst
```diff
@@ -46,6 +46,20 @@ implements new convolution methods:
...
+
+* The committer must be the **human**, not the model. This is important, we are an academic project where developers come and go as they finish their projects, so it is important to know **who** actually introduced the changes.
+
+* AI usage must be documented in the PR. Include for example the harness and models you used. This helps the reviewer understand the scope of certain changes (e.g., the tests were done automatically, while the documentation was written by hand) and allow us to keep a record, and also understand the evolution of the technology.
+
+* We are an academic collaboration, the goal of the documentation and code comments should be to teach and train current students and future researchers, and should be written with that goal in mind, not LLMs. More generally, anything that could be useful for LLMs' ``skills.md`` or ``agents.md`` file, would be better suited as an edit to our online documentation so that it is also useful for humans reading the code!
```
Radonirinaunimi: I have a hard. time understanding the last part of this paragraph. What do you mean by this?

Generally, is it planned to add a `skills.md` to specify the basic required information/workflows?

scarlehoff: in morimondo at some point someone mentioned we should add skills.md to run fits with mhou, with dataset variations, etc.

My opinion was (is) that we **don't** want to do that. Whatever people would like to add as a skills.md for an llm should be instead written as part of the documentations for humans to read. This would help LLMs just the same but will keep researchers/students as the main focus.

Your comment wad very useful, it means that without this context the text I wrote is hars to understand ^^ I'll try to reword it better.

scarlehoff: Let me know if my last modification keeps the intent more clear.

Radonirinaunimi: Understood! The text is now clearer. Thanks.

scarlehoff:
Can i merge this then?

Radonirinaunimi (approved):
Fine by me.

felixhekhorn (approved):
I would be even more happy if @enocera could also add his point of view though

felixhekhorn:
why are the changes not visible online yet https://docs.nnpdf.science/contributing/rules.html ? (So I can point to them in eko) The workflow seemed to have worked fine https://github.com/NNPDF/nnpdf/actions/runs/34844760646/job/103977900086 ...

scarlehoff:
Perhaps a race condition with the previous PR that I merged at the same time.

doc/sphinx/source/contributing/rules.rst
```diff
@@ -46,6 +46,22 @@ implements new convolution methods:
 It demonstrates how to add a new feature, together with relevant tests and
 documentation, and refine it based on the discussion.
 
+.. _aiusage:
+
+LLM Usage Guideliens
```
felixhekhorn, line 51:
```suggestion
LLM Usage Guidelines
```

@scarlehoff oh no, this will change the URIs 🙈

scarlehoff: I should've asked a LLM to proofread my text...

I'll update it directly in master.

felixhekhorn: https://github.com/NNPDF/nnpdf/commit/eda7cf0dbb6e2ec46c170ec4be003b7e3d3a084a

PR #2514, opened 2026-09-09
Extend HyperOpt backend to support the `itwinai` interface

Radonirinaunimi:
This extends the hyperopt backend in view of interfacing with [itwinai](https://itwinai.readthedocs.io/stable/index.html#). Such an interface will provide several advantages as this will allow us to:

- run distributed HPO on a Ray cluster without mongoDB
- use some of the Ray Tune search algorithms
- use advanced scheduler such as SHA (Successive Halving Algorithm) for example to stop unpromising trials early
- have a genuine memory isolation that'd have fully prevented https://github.com/NNPDF/nnpdf/pull/2511
- a true automatic resume of crashed/interrupted trials
- have live-monitoring through which one can see trials-status and resource usages
- ...

These minor changes are sufficient for the [plugin](https://itwinai.readthedocs.io/stable/getting-started/plugins.html) (to be hosted in separate repo) while still allowing the old hyperopt to run exactly as before.

### DIFF
--- a/n3fit/src/n3fit/backends/keras_backend/MetaModel.py
+++ b/n3fit/src/n3fit/backends/keras_backend/MetaModel.py
@@ -350,7 +350,10 @@ class MetaModel(Model):
 
     def get_layer_re(self, regex):
         """Get all layers matching the given regular expression"""
-        check = lambda x: re.match(regex, x.name)
+
+        def check(x):
+            return re.match(regex, x.name)
+
         return list(filter(check, self.layers))
 
     def get_replica_weights(self, i_replica):
@@ -534,7 +537,7 @@ def set_layer_replica_weights(layer, weights, i_replica: int):
         layer.get_layer(f"{NN_PREFIX}_{i_replica}").set_weights(weights)
         return
 
-    full_weights = [w.numpy() for w in layer.weights]
+    full_weights = [ops.variable_to_numpy(w) for w in layer.weights]
     for w_old, w_new in zip(full_weights, weights):
         w_old[i_replica : i_replica + 1] = w_new
 
--- a/n3fit/src/n3fit/backends/keras_backend/internal_state.py
+++ b/n3fit/src/n3fit/backends/keras_backend/internal_state.py
@@ -40,6 +40,13 @@ if (kback := K.backend()) == "torch":
         log.info("Setting max number of threads to: %d", threads)
         torch.set_num_threads(threads)
 
+    def enable_op_determinism():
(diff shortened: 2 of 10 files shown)


n3fit/src/n3fit/backends/keras_backend/MetaModel.py
```diff
@@ -350,7 +350,10 @@ def apply_as_layer(self, x):
 
     def get_layer_re(self, regex):
         """Get all layers matching the given regular expression"""
-        check = lambda x: re.match(regex, x.name)
+
+        def check(x):
+            return re.match(regex, x.name)
+
```
scarlehoff, line 356: Why changes like this one? (not that I'm against, I'd just like to understand why, to me most of the changes look generally sensible but nothing to do with hyperopt?)

Radonirinaunimi: Yes, only a few parts of the changes directly have to do with hyperopt while the remaining are minor (bug) fixes when using `KERAS_BACKEND=torch` and (in this particular case) minor style changes to adhere more with the [PEP 8](https://peps.python.org/pep-0008/#programming-recommendations) recommendations.

scarlehoff: ah, okok

Are you using torch for some particular reason? I started a few changes in Morimondo (that I need to test well still) that seem to make torch faster so if it s needed I can try to finish it.

Radonirinaunimi: Yep, I needed to test end-to-end the itwinai plugin with both torch and tensorflow.

Radonirinaunimi:
For some reasons, there are still a couple of data that fail...

scarlehoff:
It's onle dijets now, pobably need to crank the precision up a bit. Although they are so bad that I'm not sure there's enough digits in the known universe 😬

I'll open a branch and just keep testing in a loop until I'm sated.

scarlehoff:
Ah, no, scratch that. Good news, I missed two covmat_to_artfuncs!

scarlehoff:
Rebase on master again. I've rerun the test now 3 times and hasn't failed yet so hopefully it's fixed now.

ps: there's one of the star - 2jet that I'm not sure about and if it fails again I will just "disconnect" that `filter.py` because I don't think it is worth more time.

Radonirinaunimi:
Perfect! Thanks.

PR #2515, opened 2026-09-10
Use mpmath for reproducibility of experimental uncertainties

scarlehoff:
All the issues on the generation of the data involve either `covmat_to_artunc` (most) or `decompose_covmat` (a few). I'm a bit tired of that so this PR should hopefully fix it 4ever.

I'm using `mpmath` to arbitrary precision for _just_ the eigenvalue decomposition, hopefully this is enough. It doesn't seem to make it slower in a noticeable. If they still fail, we increase the precision until they don't fail anymore.

The only two datasets where the changes happen all the way to the most significant digits are the two JET datasets because the covmats are all around the place.


nnpdf_data/nnpdf_data/filter_utils/utils.py
```diff
@@ -17,10 +17,37 @@
...
+    """
+    if full_precision:
+        with mp.workdps(ARBITRARY_PRECISION):
+            covariance = mp.matrix([[mp.mpf(str(value)) for value in row] for row in covmat])
+            # We use str(value) to truncate the floats, in exchange we end up
+            # with something non-symmetrical that we need to correct for
+            covariance = (covariance + covariance.T) / 2
+            eigval, eigvec = mp.eigsy(covariance)
+            # And back to numpy
+            eigval = np.array(eigval.tolist(), dtype=float).reshape(-1)
```
andrpie, line 44: Should we use the function `sort_eigenvalues` here to avoid the issue of ordering of uncertainties?

scarlehoff: I added it below so that both mpm and numpy are ordered by `sort_eigenvalues`

andrpie (commented):
Thank you so much for doing this @scarlehoff, I believe this could solve the issue with experimental uncertainties. Just left a brief comment on ordering of eigenvalues (which was one of the problems we addressed previously).

nnpdf_data/nnpdf_data/filter_utils/utils.py
andrpie: Ah I see, all good then!

andrpie (approved):

PR #2516, opened 2026-09-16
Conservative ds

enocera:
This PR contains a proposal for a runcard with all the conservative data sets and an updated proposal of K-folds. Double-counted data sets are deweighted, and put always in the same fold.


enocera:
cc @evagroenendijk at some point to be assessed in view of the Shapley values.

evagroenendijk:
> cc @evagroenendijk at some point to be assessed in view of the Shapley values.

@enocera Is this already the fold partitioning that I should do the Shapley analysis on? I can produce a plot such as Fig. 13 in https://arxiv.org/pdf/2607.02647

enocera:
> > cc @evagroenendijk at some point to be assessed in view of the Shapley values.
> 
> @enocera Is this already the fold partitioning that I should do the Shapley analysis on? I can produce a plot such as Fig. 13 in https://arxiv.org/pdf/2607.02647

@evagroenendijk I understand that this is indeed the case.

PR #2518, opened 2026-09-16
Theory cards for nnpdf4.1 with fitting scale Q0 = 1 GeV

kamillaurent:
Theory cards for nnpdf4.1 NNLO QCD with fitting scale $Q_0 = 1 GeV$. The analogous 4.0 theory had id `40001000`, but `41001000` is already taken. For this reason I used `41002000`.


nnpdf_data/nnpdf_data/theory_cards/41002000.yaml
```diff
@@ -0,0 +1,42 @@
+ID: 41_002_000
+Comments: NNPDF4.1 perturbative charm, NNLO QCD-only
```
scarlehoff, line 2: Add here Q = 1 GeV

Not perturbative charm! That's a fit choice!

nnpdf_data/nnpdf_data/theory_cards/41002000.yaml
```diff
@@ -0,0 +1,42 @@
...
+PTO: 2
+FNS: FONLL-C
+DAMP: 0
+IC: 1
```
felixhekhorn, line 6: as discussed: this is subject to https://github.com/NNPDF/pineko/issues/254

felixhekhorn (commented):
else looks good.

nnpdf_data/nnpdf_data/theory_cards/41002000.yaml
felixhekhorn: just to make the discussion from the other side explicit: using `IC: 1` is safe for the moment. It can be then removed here or also later.

